PR #2, opened 2024-09-20
add an example form that has ALL controls

nathanstitt:
page is at: http://localhost:3000/study/edit/new


nathanstitt:
closing, code was moved into #9

PR #5, opened 2024-10-04
Add a few members pages

nathanstitt:
Adds a simple page at: `/member/openstax` and an admin page at: `/admin/members`

Both are completely insecure and will need clerk added.  The start page doesn't actually create a study yet either 😭


nathanstitt:
Test for GH!

PR #4, opened 2024-10-03
WIP: Add clerk

therealmarv:
Add a running Clerk implementation

see https://openstax.atlassian.net/browse/OTTER-37

PR #6, opened 2024-10-08
Add clerk v2

therealmarv:
this version has the newest main (2024-10-08) integrated. Old PR is [add-clerk](https://github.com/safeinsights/management-app/pull/4) is closed.

see https://openstax.atlassian.net/browse/OTTER-37


src/components/providers.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const Providers: React.FC<Props> = ({ children }) => {
+    return (
+        <MantineProvider theme={theme}>
+            <ModalsProvider>
+                <ClerkProvider>{children}</ClerkProvider>
+            </ModalsProvider>
+        </MantineProvider>
+    )
+}
```
nathanstitt, line 24: I moved this file as part of #5 to src/app/providers.tsx since it was only used on layout.  it has Clerk loaded on it already though so you maybe can just remove this file.

nathanstitt (approved):
Looks good @therealmarv but see my comment about removing providers.tsx file

package.json
```diff
@@ -32,6 +32,7 @@
         "@vitejs/plugin-react": "^4.3.2",
         "@vitest/coverage-v8": "^2.1.1",
         "dotenv": "*",
+        "esbuild": "^0.24.0",
```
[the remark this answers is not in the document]
therealmarv: We got problems when it was not there. It's a sub requirement of vitejs above unfortunately and it broke our build out if we don't install it in the dev requirements.

src/components/providers.tsx
therealmarv: deleted it

PR #8, opened 2024-10-09
move pre-propsal screen to study-request location

nathanstitt:
Also add a link to it from homepage for people who are lost

PR #9, opened 2024-10-10
revised study proposal form

jbwilson8:


nathanstitt:
I'm going to merge this because it'll conflict with #8.  after merge, I'll pull main into #8 and fix it up that way.  Thanks so much @jbwilson8 !

PR #13, opened 2024-10-15
Add authentication to routes

nathanstitt:
Also:
* moves start page to `/researcher/study/request/openstax`
* changes default port for development to 4000 - this was needed to avoid conflicts with other apps in enclave
* adds a `/researcher/studies` page where researchers can view status and launch additional runs
* adds `/member/openstax/studies/review` as a bare-bones entry to review screen for members


src/middleware.ts
```diff
@@ -1,12 +1,35 @@
-import { clerkMiddleware } from '@clerk/nextjs/server'
+'use server'
+import 'server-only'
+import { clerkMiddleware, createRouteMatcher } from '@clerk/nextjs/server'
+const isMemberRoute = createRouteMatcher(['/fix-me/member(.*)'])
+const isResearcherRoute = createRouteMatcher(['/fix-me/researcher(.*)'])
```
nathanstitt, line 5: @therealmarv this is where we can check roles, I added the `fix-me` so it didn't match and stop everything from working

therealmarv (approved):
I've wrapped my head around it and tried to find any big errors. No big errors I could find. So I give it a ✅

PR #19, opened 2024-10-23
adding remaining study proposal form fields and style

jbwilson8:
[OTTER-46](https://openstax.atlassian.net/browse/OTTER-46)

**TODO:**  make button functionality better

[OTTER-46]: https://openstax.atlassian.net/browse/OTTER-46?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/researcher/study/[encodedStudyId]/edit/form.tsx
+++ b/src/app/researcher/study/[encodedStudyId]/edit/form.tsx
@@ -11,15 +11,9 @@ import { FormValues, schema, zodResolver } from './schema'
 import Link from 'next/link'
 
 export const Form: React.FC<{ studyId: string; study: FormValues }> = ({ studyId, study }) => {
-    const {
-        control,
-        getValues,
-    } = useForm<FormValues>({
+    const { control, getValues } = useForm<FormValues>({
         resolver: zodResolver(schema),
-        defaultValues: { ...study,
-            highlights: true,
-            eventCapture: true,
-        },
+        defaultValues: { ...study, highlights: true, eventCapture: true },
         mode: 'onChange',
     })
     const [isDisabled, setIsDisabled] = useState(true)
@@ -30,7 +24,7 @@ export const Form: React.FC<{ studyId: string; study: FormValues }> = ({ studyId
     const checkFormValid = () => {
         const values = getValues()
         const hasChecked = values.highlights || values.eventCapture
-        setIsFormValid(hasChecked)
+        setIsFormValid(hasChecked ?? false)
     }
 
     return (
@@ -90,13 +84,13 @@ export const Form: React.FC<{ studyId: string; study: FormValues }> = ({ studyId
                 <Flex p={2} gap="lg">
                     <Text className={labelStyle}>Datasets of Interest</Text>
                     <Stack>
-                        <Checkbox 
+                        <Checkbox
                             name="highlights"
                             control={control}
                             value="highlights"
                             label="Highhlights and Notes"
                         ></Checkbox>
-                        <Checkbox 
+                        <Checkbox
                             name="eventCapture"
                             control={control}
                             value="event-capture"
--- a/src/app/researcher/study/[encodedStudyId]/edit/form.tsx
+++ b/src/app/researcher/study/[encodedStudyId]/edit/form.tsx
@@ -1,29 +1,44 @@
 'use client'
 
-import React from 'react'
+import React, { useState } from 'react'
 
 import { Form as HookForm, useForm } from 'react-hook-form'
-import { Checkbox, Select, Textarea, TextInput } from 'react-hook-form-mantine'
-
+import { Checkbox, Textarea, TextInput } from 'react-hook-form-mantine'
 import { inputStyle, labelStyle } from './style.css'
-import { Button, Flex, Stack, Text } from '@mantine/core'
+import { Button, Flex, Group, Stack, Text } from '@mantine/core'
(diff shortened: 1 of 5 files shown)


src/app/researcher/study/[encodedStudyId]/edit/form.tsx
```diff
@@ -1,84 +1,130 @@
...
+import Link from 'next/link'
 
 export const Form: React.FC<{ studyId: string; study: FormValues }> = ({ studyId, study }) => {
-    const { control } = useForm<FormValues>({
+    const {
+        control,
+        // formState: { isValid },
+        getValues,
+    } = useForm<FormValues>({
         resolver: zodResolver(schema),
```
Comment (chrisbendel) [inline src/app/researcher/study/[encodedStudyId]/edit/form.tsx:19]: This is a nitpick but i prefer having the schema defined  in the same file (its just easier for me to read) but no changes here necessary if your team prefers otherwise

chrisbendel (commented):
I was going down a bit of a rabbit hole with the form stuff. A couple small comments, but maybe a larger discussion - is there a reason why @safeinsights/otter-devs decided to bring in `react-hook-form`? It looks like y'all are using some [adapter library](https://github.com/aranlucas/react-hook-form-mantine) which their last update was ~1 year ago, im not sure if it's still maintained. 

If you want to use react-hook-form with mantine components you might have to wrap all of the mantine input components with a `<Controller>` which gets super cumbersome, or it leads you to abstract those out with a library like the one you're using which is more maintenance. 

TLDR is there any reason why we can't just use the mantine `useForm` hook? i understand it might not provide every feature that react-hook-form does but its one less dependency and it also allows integration with zod/yup/etc

Hope im not derailing this PR with this, but thought it might be important to bring up for cross team learning and alignment!

src/app/researcher/study/[encodedStudyId]/edit/schema.ts
```diff
@@ -2,14 +2,21 @@ import { z } from 'zod'
...
+        piName: z.string().min(3).max(100),
+        highlights: z.preprocess((value) => value === 'on', z.boolean()).nullish(),
+        eventCapture: z.preprocess((value) => value === 'on', z.boolean()).nullish(),
+        outputMimeType: z.string().nullish(),
+        irbDocument: z.string().nullish(),
+        containerLocation: z.string().min(3).max(250),
+    })
+    .refine((data) => data.highlights || data.eventCapture, {
+        message: 'At least one checkbox must be selected',
+        path: ['highlights', 'evntCapture'],
```
Comment (chrisbendel) [inline src/app/researcher/study/[encodedStudyId]/edit/schema.ts:18]: Spelling error on 'evntCapture' which might have been leading you to manually check the valid state?

src/app/researcher/study/[encodedStudyId]/edit/form.tsx
```diff
@@ -1,84 +1,130 @@
...
         defaultValues: study,
+        mode: 'onChange',
     })
+    const [isDisabled, setIsDisabled] = useState(true)
+    const [isFormValid, setIsFormValid] = useState(true)
+    const enableButton = () => {
+        setIsDisabled(false)
+    }
+    const checkFormValid = () => {
```
Comment (chrisbendel) [inline src/app/researcher/study/[encodedStudyId]/edit/form.tsx:28]: I started digging into "why" you might be needing this - could be related to a spelling error in `schema.ts` in the `refine` callback

src/app/researcher/study/[encodedStudyId]/edit/schema.ts
Comment (jbwilson8) [reply src/app/researcher/study/[encodedStudyId]/edit/schema.ts:18]: Thanks for catching that

nathanstitt:
Hey @chrisbendel thanks for taking a look at the form stuff ❤️     

We choose react-hook-form over mantine forms because it's more efficient, matine forms re-renders the entire form on every event while react-hook-form only re-renders the field that's being interacted with.  [comment to that effect](https://github.com/orgs/mantinedev/discussions/5263) but I also verified it. That's not a huge deal but react-hook-form is also very popular and I suspected docs would be more available.

Use of the adapter did make me a little nervous but each component is just a tiny wrapper so it seems like it should be bug-free https://github.com/aranlucas/react-hook-form-mantine/blob/master/src/Input/Input.tsx for an example.

src/app/researcher/study/[encodedStudyId]/edit/form.tsx
Comment (nathanstitt) [reply src/app/researcher/study/[encodedStudyId]/edit/form.tsx:19]: putting the schema in a separate file lets us re-use it for server-side validation in the `actions.ts` file to validate the incoming values before using them with the DB.    It also can generate the typescript types for the form and that can be imported separately as well

chrisbendel:
> Hey @chrisbendel thanks for taking a look at the form stuff ❤️
> 
> We choose react-hook-form over mantine forms because it's more efficient, matine forms re-renders the entire form on every event while react-hook-form only re-renders the field that's being interacted with. [comment to that effect](https://github.com/orgs/mantinedev/discussions/5263) but I also verified it. That's not a huge deal but react-hook-form is also very popular and I suspected docs would be more available.
> 
> Use of the adapter did make me a little nervous but each component is just a tiny wrapper so it seems like it should be bug-free https://github.com/aranlucas/react-hook-form-mantine/blob/master/src/Input/Input.tsx for an example.

That makes sense! Just for y'alls knowledge it looks like a more recent version of the mantine/forms library has an [uncontrolled mode](https://mantine.dev/form/uncontrolled/) which [does not re-render all the components](https://github.com/orgs/mantinedev/discussions/1193#discussioncomment-9883362)

PR #21, opened 2024-11-04
Website login, signup and pw reset

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-38

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
         "pg": "^8.13.0",
         "react-hook-form": "^7.53.0",
         "react-hook-form-mantine": "^3.1.3",
+        "react-router-dom": "^6.27.0",
         "remeda": "^2.14.0",
         "server-only": "^0.0.1",
         "uuid": "^10.0.0",
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
         "pg": "^8.13.0",
         "react-hook-form": "^7.53.0",
         "react-hook-form-mantine": "^3.1.3",
-        "react-router-dom": "^6.27.0",
         "remeda": "^2.14.0",
         "server-only": "^0.0.1",
         "uuid": "^10.0.0",
--- a/src/components/signin.tsx
+++ b/src/components/signin.tsx
@@ -1,8 +1,8 @@
 import { useState } from 'react'
 import { reportError } from './errors'
-import { Anchor, Button, Group, Loader, PasswordInput, Stack, Text, TextInput, Title } from '@mantine/core'
+import { Anchor, Button, Group, Loader, PasswordInput, Stack, Text, TextInput, Title, Paper, CloseButton, Checkbox } from '@mantine/core'
 import { isEmail, isNotEmpty, useForm } from '@mantine/form'
-import { useRouter } from 'next/navigation'
+import { useRouter, Link } from 'next/navigation'
 import { useSignIn } from '@clerk/nextjs'
 import { SignInResource } from '@clerk/types'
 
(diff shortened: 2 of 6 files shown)


jbwilson8 (approved):

src/components/reset-password.tsx
```diff
@@ -0,0 +1,159 @@
...
+                    />
+                    <Stack align="center" mt={15}>
+                        <Button type="submit">Send Reset Code</Button>
+                        <Anchor onClick={() => router.push('/')}>Back to Login</Anchor>
+                    </Stack>
+                </Paper>
+            </form>
+        </Stack>
+    )
+}
```
nathanstitt, line 159: it feels to me like this component is really two separate components,.

Maybe we could split most of it into other components and then this could be:

```ts
export function ResetPassword() {
     const { isLoaded, signIn, setActive } = useSignIn()
     const [pendingReset, setPendingReset] = useState<Boolean>(false)

     if (pendingReset) return <PendingReset />

     return <ResetForm onComplete={() => setPendingReset(true)} />
}
```

package.json
```diff
@@ -31,6 +31,8 @@
         "@testing-library/user-event": "^14.5.2",
         "@types/pg": "^8.11.10",
         "@types/react": "18.3",
+        "@typescript-eslint/eslint-plugin": "^8.13.0",
+        "@typescript-eslint/parser": "^8.13.0",
```
nathanstitt, line 35: curious why we needed to add these, was lint not working correctly for you?

src/components/signin.tsx
```diff
@@ -93,29 +95,37 @@ export function SignIn() {
 
     return (
         <Stack>
-            <Title order={3}>Login</Title>
+            {/* <Title order={3}>Login</Title> */}
```
nathanstitt, line 98: was this intentional?

src/components/signin.tsx
```diff
@@ -93,29 +95,37 @@ export function SignIn() {
...
+                        withAsterisk
+                        key={form.key('password')}
+                        {...form.getInputProps('password')}
+                        mt={10}
+                        placeholder="Password"
+                        aria-label="Password"
+                    />
+                    <Stack align="center" mt={15}>
+                        <Button type="submit">Login</Button>
+                        {/* <Checkbox mt={5} label="Remember me" aria-label="Remember me" /> */}
```
nathanstitt, line 124: same q, I think this wasn't used, but we could just remove it from code altogether.

nathanstitt (changes requested):
Looks good overall, but think we should cleanup the large reset component

package.json
nathanstitt: I see the commit message now, that makes sense to silence the unknown TS version warning

therealmarv: I was getting this error which would not go away with a simple `npm install`:
```WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
You may find that it works just fine, or you may not.
SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.6.0
YOUR TYPESCRIPT VERSION: 5.6.2
```

It's 99% a my-side problem which maybe does not need to get into code. I'm totally okay with reverting it because the CI does not have this errors it seems with a fresh install.

therealmarv:
It's not done yet (end of wip and I want to remove some more lines) but thanks for reviews!

nathanstitt:
Oh derp, sorry, I missed that you hadn't requested reviews yet 😭

therealmarv:
I think this is the *final* state for the PR now 🤞

nathanstitt (approved):
looks good, thanks Marvin!

PR #29, opened 2024-11-07
download results from runs

nathanstitt:
Also adds synthetic results generation in dev mode and prod until enclave is deployed

image


src/server/config.ts
```diff
@@ -4,6 +4,13 @@ export const TEST_ENV = !!(process.env.CI || process.env.NODE_ENV === 'test')
 
 export const PROD_ENV = process.env.NODE_ENV === 'production'
 
-export const getUploadTmpDirectory = () => {
-    return process.env.UPLOAD_TMP_DIRECTORY || '/tmp'
-}
+export const getUploadTmpDirectory = () => process.env.UPLOAD_TMP_DIRECTORY || '/tmp'
+
+export const ALWAYS_CREATE_ECR = process.env.ALWAYS_CREATE_ECR === 't'
```
chrisbendel, line 9: just out of curiosity why are the env variables stored as `t` instead of true/false? something outside of our control / we cant use booleans?

nathanstitt: yes they come from the process.env and all values are strings.  I could have just tested for their existence, but this way they can also be turned off if needed by setting to anything other than `'t'`

src/app/researcher/studies/page.tsx
```diff
@@ -9,7 +9,7 @@ export default async function ListStudiesPage() {
 
     const studies = await db
         .selectFrom('study')
-        .select(['id', 'title', 'containerLocation', 'description'])
+        .select(['study.id', 'title', 'containerLocation', 'description'])
```
chrisbendel, line 12: for my own knowledge are these effectively different?

PR #32, opened 2024-11-13
Add role auth

therealmarv:
https://openstax.atlassian.net/browse/OTTER-36

It's in regards to the 403 pages a MVP. But the focus is more that members cannot be researchers and vice versa (protect routes). Also non logged in users should not access any member or researcher pages (which before this PR was possible). The 403 page could definitely look "nice" or "styled". If that's needed on top I would prefer to built on top of this PR.


therealmarv:
It's ready for review.

Things which should be improved for future (after pilot?):
* Tests for middleware but excluding/mocking Clerk (I'm currently trying on how this can be achieved).

Things which could be improved:
* Nice looking 403.

nathanstitt (approved):
looks great, thanks @therealmarv !

PR #34, opened 2024-11-14
add IaM policy for enclave access when creating ECR

nathanstitt:
Needed to allow os enclave to pull from repos.  In the future we'll need to figure out how to pass the account numbers in programically instead of hardcoding

### DIFF
--- a/src/server/aws-ecr-policy.ts
+++ b/src/server/aws-ecr-policy.ts
@@ -1,0 +1,28 @@
+import { type AwsIAMPolicy, AwsEcrActions as Action } from 'aws-iam-policy-types'
+import { ENCLAVE_AWS_ACCOUNT_NUMBERS } from './config'
+
+export const EcrPolicy: AwsIAMPolicy = {
+    Version: '2012-10-17',
+    Statement: [
+        {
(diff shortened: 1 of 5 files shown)


src/server/aws-ecr-policy.ts
```diff
@@ -0,0 +1,28 @@
+import { type AwsIAMPolicy, AwsEcrActions as Action } from 'aws-iam-policy-types'
+import { ENCLAVE_AWS_ACCOUNT_NUMBERS } from './config'
+
+export const EcrPolicy: AwsIAMPolicy = {
+    Version: '2012-10-17',
```
chrisbendel, line 5: I'm sure this has no side effects but my curiosity is wondering about the significance of this date 😆

chrisbendel (approved):

src/server/aws-ecr-policy.ts
nathanstitt: yah, it's AWS specific and refers to the schema for the IaM statement.  Your comment prompted me to look it up and it appears there's only two valid values: '2012-10-17' | '2008-10-17'

PR #36, opened 2024-11-19
adds action to container registry policy

philschatz:
This is necessary for ECS to pull the image


nathanstitt (approved):
also probably should add to users perms below this, but I'll add it as a followup. Thanks!

PR #39, opened 2024-11-21
Misc fixes for demo

nathanstitt:
fixes a few minor issues noted noted before/during demo:

* Researcher's study review urls used unencoded uuids 
* push instructions did not open after adding a new run
* AWS resources were all being tagged with "Sandbox" environment and it failed with UTF chars in study title
* Study title length was too limited 
* `a` tags were nested on on researcher studies listing, causing hydration errors


nathanstitt:
Thanks @jbwilson8 !

PR #44, opened 2024-12-11
file upload tests

jbwilson8:
[OTTER-54](https://openstax.atlassian.net/browse/OTTER-54)

### DIFF
--- a/tests/researcher-create-study.spec.tsx
+++ b/tests/researcher-create-study.spec.tsx
@@ -30,4 +30,43 @@ test.describe('app', () => {
 
         await expect(page.getByText(testTitle).first()).toBeVisible()
     })
+
+    test('validate file upload', async ({ page }) => {
+        await visitClerkProtectedPage({ page, url: '/', role: 'researcher' })
+
+        // Navigate to the upload section
+        await page.getByRole('button', { name: /propose/i }).click()
+        await page.getByRole('button', { name: /submit/i }).click()
+        await page.getByRole('button', { name: /next/i }).click()
+
+        //Verify that study proposal has a highlight checked
+        await page.getByRole('checkbox', { name: /highlights/i }).check()
+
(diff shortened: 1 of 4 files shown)


tests/fixtures/temp/large-file
nathanstitt, line 1: hm, I don't think we should checkin a large file like this, because it will bloat the repo for everyone who pulls it.  Unfortunately, to remove it you'll need to do a git filter-branch vs just `git delete`   https://www.deployhq.com/git/faqs/removing-large-files-from-git-history
Once @iccgoncalves specs this feature a little bit more we can come up with a way to generate a large file on the fly in the spec that just barely exceeds the limit.,

tests/researcher-create-study.spec.tsx
```diff
@@ -30,4 +30,43 @@ test.describe('app', () => {
...
+        // const invalidFileType = 'tests/fixtures/invalid.txt'
+        // await page.setInputFiles('input[type="file"]', invalidFileType)
+        // await expect(page.getByText('File type must be .zip')).toBeVisible()
+
+        // Test file size limit
+        const largeFile = 'tests/fixtures/temp/large-file.zip'
+        await page.setInputFiles('input[type="file"]', largeFile)
+        await expect(page.getByText('File size cannot exceed')).toBeVisible()
+    })
 })
```
nathanstitt, line 72: love this, tests it very throughly

tests/fixtures/temp/large-file
iccgoncalves: Thank you @nathanstitt for sharing this with me. I agree, this needs further specification. 
A couple of aspects I have in mind as first pass:
- File Size
- File Type
- Language Supported (I have raised this to leadership for further conceptualization)

@nathanstitt @jbwilson8 feel free to add more thoughts in case there's additional limitations we might have missed

PR #46, opened 2024-12-11
add a button that shows loading state while nav

nathanstitt:


chrisbendel:
@nathanstitt i dug up the [relevant response](https://openstax.slack.com/archives/C07H80W7656/p1733960170905959?thread_ts=1733959961.383229&cid=C07H80W7656) from @iccgoncalves in slack about this PR 

i wonder if we should close this for now? just in the spirit of cleaning up the PR list

nathanstitt:
Closing because it hasn't been an issue that we need to worry about

PR #50, opened 2025-01-10
Add combined code coverage

nathanstitt:


tests/unit.helpers.ts
```diff
@@ -1,30 +1,13 @@
...
-            password: process.env.E2E_CLERK_RESEARCHER_PASSWORD!,
-        },
-    })
-}
-
-export const readTestSupportFile = (file: string) => {
-    return fs.promises.readFile(path.join(__dirname, 'support', file), 'utf8')
-}
+export * from './common.helpers'
+export { fs, path }
```
chrisbendel, line 10: is there any benefit to this vs just importing path/fs where needed in the consuming files?

chrisbendel (approved):

therealmarv (approved):

tests/unit.helpers.ts
nathanstitt: not really, just a tiny benefit it just makes it easier for consumers to import from one spot.

PR #57, opened 2025-02-06
Add new member management page with form

chrisbendel:
Additionally organize some common actions and forms, clean up some redundant definitions, etc.

General note on kysely-codegen, this PR starts utilizing and tries to organize type definitions for a model based on the columns [using the `Selectable` type from kysely](https://github.com/RobinBlomberg/kysely-codegen?tab=readme-ov-file#using-the-type-definitions), rather than defining ad hoc interfaces in each component. We can also expand this to use the updatable and insertable types as well for CRUD things


tests/e2e.helpers.ts
```diff
@@ -103,54 +103,3 @@ export const visitClerkProtectedPage = async ({ page, url }: VisitClerkProtected
         },
     })
 }
-
-export const insertTestStudyData = async (opts: { memberId: string }) => {
```
chrisbendel, line 107: Unused

src/schema/member.ts
```diff
@@ -4,22 +4,21 @@ import { Member as DefinedMember } from '@/database/types'
...
-export type ValidatedMember = z.infer<typeof schema>
+export type ValidatedMember = z.infer<typeof memberSchema>
 
 export const NEW_MEMBER: NewMember = {
     identifier: '',
     name: '',
     email: '',
     publicKey: '',
 }
-Object.freeze(NEW_MEMBER)
```
chrisbendel, line 22: This was causing an issue on the add new member form, this shouldn't be necessary 🤔

src/app/admin/members/table.tsx
```diff
@@ -65,24 +53,51 @@ export function MembersAdminTable() {
...
+        mutationFn: deleteMemberAction,
+        onSettled: async () => {
+            return await queryClient.invalidateQueries({ queryKey: ['members'] })
+        },
+    })
+
+    return (
+        <Group gap={4} justify="center" wrap="nowrap">
+            <Modal opened={opened} onClose={close} title={`Edit ${member.name}`} closeOnClickOutside={false}>
+                <EditMemberForm member={member} onCompleteAction={close} />
```
chrisbendel, line 93: I laid the groundwork here to potentially hoist the need for passing down an `onCompleteAction`, previously it was being passed down through 3-4 different components just to close the modal when the form was submitted. Each form should have it's own modal now which eliminates another layer, i still think this can be improved on though 🤔 can come back to this down the road

src/schema/member.ts
nathanstitt: I think this should stay as a safeguard.  If the NewMember object gets mutated then the next user of the form will see the previous users data 😱 

If the new member form complains, you can use `initialValues: { ...NEW_MEMBER }` to clone it

src/components/surety-guard.tsx
```diff
@@ -1,6 +1,7 @@
 import { Button, Popover, ActionIcon, Text, Flex } from '@mantine/core'
 import { IconTrash } from '@tabler/icons-react'
 
+// TODO Mantine does provide a confirm modal https://mantine.dev/x/modals/#confirm-modal
 export const SuretyGuard: React.FC<{ children?: React.ReactNode; message?: string; onConfirmed: () => void }> = ({
```
nathanstitt, line 5: hehe, yes but this one pops up right beside the thing you clicked :)

tests/unit.helpers.tsx
```diff
@@ -5,9 +5,35 @@ import os from 'os'
...
+    const testQueryClient = createTestQueryClient()
+
+    return render(
+        <QueryClientProvider client={testQueryClient}>
+            <MantineProvider theme={theme}>
+                <ModalsProvider>{ui}</ModalsProvider>
+            </MantineProvider>
+        </QueryClientProvider>,
+    )
+}
```
nathanstitt, line 34: instead of this, we can use the 2nd context argument of `render` along with the `TestingProvidersWrapper`   Example: https://github.com/safeinsights/management-app/blob/main/src/app/researcher/study/request/%5BmemberIdentifier%5D/form.test.tsx#L16-L17

I do see this being a handy abstraction though so maybe we still export it.  up to you I guess

nathanstitt (approved):
looks great Chris, ty.  I'll approve for faster merging but i do think we should still freeze the NEW_MEMBER object

src/schema/member.ts
chrisbendel: I see what you're worried about, but having tested it without the freeze, filling out the form, submitting and cancelling the form, and then reopening the add member form it will be cleared out entirely. I don't know if there's any edge case im missing here in testing though. However I thought of maybe a simpler solution, we can just have a method that returns a "new member" with the blank object, then pass that into the component, rather than having the internal form spread the values of any member it gets
```
export const getNewMember = (): NewMember => {
    return {
        identifier: '',
        name: '',
        email: '',
        publicKey: '',
    }
}
```

let me know how that sounds!

tests/unit.helpers.tsx
chrisbendel: yeah i suppose this is totally a stylistic thing right?

These would be functionally equivalent 🤔 
```
        const {container} = render(
            <ManageMemberPage params={props.params} />,
            TestingProvidersWrapper
        )

        const { container } = renderWithProviders(await ManageMemberPage(props))
```

chrisbendel: im happy to go either way on this one! if we want to leave it in and see how it goes organically that works for me too

src/schema/member.ts
nathanstitt: yah that sounds great!  basically this entire thing was an attempt to get a new blank member for forms which that accomplishes.  Really it's probably an over-optimization and the InitialValues could just be specified in the useForm vs imported from here

PR #53, opened 2025-02-03
initial setup for member validation test

jbwilson8:

PR #59, opened 2025-02-11
enforce MFA for all accounts

nathanstitt:
Redirects to /account/mfa whenever a user without mfa enabled uses SI

image

image


package.json
```diff
@@ -107,5 +107,6 @@
         "vite-tsconfig-paths": "^5.1.4",
         "vitest": "*",
         "vitest-monocart-coverage": "^3.0.0"
-    }
+    },
+    "overrides": {}
```
chrisbendel, line 111: do we need the `overrides`?

types/globals.d.ts
chrisbendel, line 1: is this something we need or should the `auth()` hook have the typing for this? these global type files i usually try to avoid when possible 😆

src/app/account/mfa/sms/page.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+// Display phone numbers that are not reserved for MFA
+const ManageAvailablePhoneNumbers = () => {
+    const { user } = useUser()
+
+    if (!user) return null
+
+    // Check if any phone numbers aren't reserved for MFA
+    const availableForMfaPhones = user.phoneNumbers
+        .filter((ph) => ph.verification.status === 'verified')
```
chrisbendel, line 73: general nitpick of preferring to be verbose with the callback arguments ie `ph` -> `phoneNumber` or even `phone` just for legibility 🤔 but this is entirely subjective

chrisbendel (changes requested):

package.json
nathanstitt: lol, nope.  I thought I might for the qrcode lib but then the warnings went away

types/globals.d.ts
nathanstitt: no this is the flag that we added to the clerk session (per their docs), this types it.

src/app/account/mfa/sms/page.tsx
nathanstitt: yah that's kinda weird anyway, why are there two filters?  this code is mostly from the clerk docs btw. I rewrote and moved it inline to the component

chrisbendel (approved):
good stuff!! nice work

chrisbendel:
@nathanstitt i handled the merge conflict from merging #53 upstream in [this commit](https://github.com/safeinsights/management-app/pull/59/commits/97c7b836280545d1ecbf3b37e890f03c951a17dd) let me know if this looks good to you

PR #60, opened 2025-02-13
Refactor site root layout, routing based on user roles, and Introduce member dashboard page

chrisbendel:


jbwilson8 (approved):

package.json
```diff
@@ -45,6 +45,7 @@
         "@mantine/modals": "^7.16",
         "@mantine/notifications": "^7.16",
         "@mantinex/dev-icons": "^1.1.0",
+        "@phosphor-icons/react": "^2.1.7",
```
nathanstitt, line 48: why pull in additional icon package for a `gear`?  We're also using https://tabler.io/icons which has that.  I'd suggest that if we do want to use phosphor, we remove @mantinex/dev-icons and Tabler in favor of it

src/app/providers.tsx
```diff
@@ -9,14 +9,18 @@ import { ClerkProvider } from '@clerk/nextjs'
...
 
 function makeQueryClient() {
     return new QueryClient({
         defaultOptions: {
             queries: {
                 // With SSR, we usually want to set some default staleTime
-                // above 0 to avoid refetching immediately on the client
+                // above 0 to avoid re-fetching immediately on the client
                 staleTime: 60 * 1000,
+                refetchOnWindowFocus: false,
```
nathanstitt, line 21: it seems like `refetchOnWindowFocus` might be beneficial?  as I understand react-query, it pauses refetch while unfocused, so the refetch on focus is good for cases where the page is in the background for extended period of time. perhaps that isn't correct though?

nathanstitt (approved):
looks good, just a few things to debate :D

package.json
chrisbendel: oh yeah, i meant to leave context on this, devika has decided that's the icon package we're going to use going forward (i didn't have a say in this) - we will have to replace and remove all the tabler icons and use phosphor

src/app/providers.tsx
chrisbendel: yeah this is totally up for debate, there's a good article here https://tanstack.com/query/latest/docs/framework/react/guides/important-defaults I was just playing with how this would interact with page focusing and stuff - happy to remove this here and we can decide on what defaults we want (think it makes sense choosing defaults per app since different apps have different use cases)

PR #62, opened 2025-02-16
allow totp codes starting with zero

therealmarv:
totp codes starting with zero gave an error message (bug)

image


nathanstitt (approved):
good catch @therealmarv , thank you!

PR #63, opened 2025-02-19
Add database table for member user public keys

rnathuji:


src/database/migrations/1739918380436_member_user_public_key.ts
```diff
@@ -0,0 +1,19 @@
+import { type Kysely, sql } from 'kysely'
```
rnathuji, line 1: I used `_` in the filename here instead of `-` as well as the singular noun, but I wasn't sure what the convention is in the code base as there seems to already be a bit of inconsistency with `uuid_fn` and `study-runs` . Happy to change as needed :man_shrugging: .

src/database/migrations/1739918380436_member_user_public_key.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema
+        .createTable('member_user_public_key')
```
rnathuji, line 5: I opted for a more verbose table name so it's clear we are storing just the public key (e.g. instead of `member_user_key` or something), but I am open to suggestions.

src/database/migrations/1739918380436_member_user_public_key.ts
```diff
@@ -0,0 +1,19 @@
...
+        .addColumn('id', 'uuid', (col) => col.defaultTo(sql`v7uuid()`).primaryKey())
```
rnathuji, line 6: IDK if we need this (e.g. if we plan to return an ID value to users via API), but included it as the PK for now.

src/database/migrations/1739918380436_member_user_public_key.ts
```diff
@@ -0,0 +1,19 @@
...
+        .addColumn('user_id', 'uuid', (col) => col.notNull())
```
rnathuji, line 7: This is a stab at a column name for referencing a user, but I'm not sure if it aligns with the eventual vision :sweat_smile: . (e.g. could this be a FK to a future table? :man_shrugging: )

src/database/migrations/1739918380436_member_user_public_key.ts
```diff
@@ -0,0 +1,19 @@
...
+        .addColumn('value', 'text', (col) => col.notNull())
+        .addColumn('fingerprint', 'text', (col) => col.notNull())
```
rnathuji, line 9: I believe the naming here is consistent with other tables where it's `{table noun} - {attribute}` like `study-run-status`. So here we have `member-user-public-key-value` and `member-user-public-key-fingerprint`. This is in contrast to, say, naming the `value` column `key` as that would be `member-user-public-key-key` which seems inconsistent to me :man_shrugging: .

src/database/migrations/1739918380436_member_user_public_key.ts
```diff
@@ -0,0 +1,19 @@
...
+        .addColumn('created_at', 'timestamp', (col) => col.defaultTo(sql`now()`).notNull())
+        .addColumn('updated_at', 'timestamp', (col) => col.defaultTo(sql`now()`).notNull())
```
rnathuji, line 11: I added these audit columns for consistency, but I'm curious how the `updated_at` column actually works in this codebase. Is that some Kysely magic or a trigger we configure?

src/database/migrations/1739918380436_member_user_public_key.ts
```diff
@@ -0,0 +1,23 @@
...
+        .execute()
+
+    db.schema
+        .createIndex('member_user_public_key_fingerpring_indx')
```
rnathuji, line 15: I added this index with the expectation we're going to query by fingerprint. Perhaps that's premature optimization and / or other indexes might be useful?

rnathuji:
In the interest of getting rapid feedback, here is a quick, initial stab at the migration for the new table. I have liberally populated the PR with commentary / questions for reviewers to help with getting feedback :sweat_smile: .

nathanstitt (approved):
thanks for the PR @rnathuji!

Unfortunately we're currently lacking a user table to map this to.    Clerk doesn't give us a uuid for users, their user id is a string column that looks like `cl_jsfdjafas`. I'm not sure we'll want to use that everywhere, I'd rather make our own user table with a `clerk_id` column and map this.

This does set us up for that so I think we can merge and then add a foreign key constraint to this user_id column later

rnathuji:
@nathanstitt -- So, what had happened was :sweat_smile: : I commented my PR with questions / thoughts, but I kind of forgot to click them through :upside_down_face: . I saw you approved, but if you can, take a quick look at those just to be sure you don't want any changes based on them.

src/database/migrations/1739918380436_member_user_public_key.ts
rnathuji: Yes, I used the CLI, but it was something like:

```bash
$ npx kysely migrate:make member_user_public_key
```

So, the "convention" is user provided vs kysely itself :man_shrugging: .

chrisbendel:
hey all im just catching up after being out for the count with the flu over the past few days, i think i missed a lot of the conversational context around this stuff

im just trying to understand it in my head in terms of how this might affect the UI / what might need to change down the line in response to these changes, what would the new difference be between a member setting a public key at the member level (ie `member.public_key`) vs a user setting their own public key? would the `user.public_key` be the source of truth that would bubble up to member(s)? 

this question might not make any sense cause my brain is a little foggy 😆

rnathuji:
@chrisbendel - I believe that the current UI screens are for the public key used by the member to authenticate with BMA API (e.g. it's for "services") while these are for data encryption for "users". So they are orthogonal in terms of use cases / purpose. I believe the mockups during last week's platform sync may have highlighted the potential new UI screens. I can't find file links, but you can check the recording linked in the [agenda doc](https://docs.google.com/document/d/1ETTY1q4pnHWs9xY8bakwbn91s3L2FVur3D_ci1rle9c/edit?tab=t.0) to take a look.

PR #80, opened 2025-03-12
Remove B64 <-> UUID Conversions

chrisbendel:
Things were getting complicated constantly converting UUID <-> B64 from the ui <-> db. Per a quick chat with Nathan, I believe we can just simplify for now to use the uuids that are in the database


nathanstitt:
### BUILD STATUS ✅

[Preview Link](https://pr80.dev.safeinsights.org/)

nathanstitt (approved):
Looks good @chrisbendel   I suspect we may have some lingering bugs from this but we can address them as they pop up.

PR #86, opened 2025-03-17
changes & misc fixes from post-vaca 🅿️ 

nathanstitt:
* Fixes specs
* excludes /api from middleware to restore previous functionality of using JWT auth
* adds study db fields for IRB and Description document upload path


nathanstitt:
### BUILD STATUS ❌



src/server/actions/study-actions.ts
```diff
@@ -70,6 +68,8 @@ export const getStudyAction = async (studyId: string) => {
         .executeTakeFirst()
 }
 
+export type SelectedStudy = NonNullable<Awaited<ReturnType<typeof getStudyAction>>>
```
chrisbendel, line 71: We may not need this for the `<ReviewControls>` prop typing, instead we can just add to the getStudyAction select clause and the typing should be happy
```
            'study.descriptionDocPath',
            'study.irbDocPath',
```

i do see the idea behind this though - since the server actions can just select whatever they want it might not always match up with the types directly from the database schema - if we're picking and choosing other columns from joined tables, we will hit this same issue. 

Just thinking out loud about the long term approach that makes sense as the codebase grows, do we have a bunch of types like this for individual queries? One alternative I could see is using react-query in client components that need the data and relying on the types implicitly from the return type of the action calling the db.execute(), vs having to pass them down via props. whats cool about react query is it will de-dupe the query so we could call the same action 7 times in 7 components and it should just fetch once and re-use the value across those components 🤔 

anyway no change here needed i think, but this brought up some thoughts 🤣

src/server/aws.ts
```diff
@@ -107,22 +113,33 @@ const calculateChecksum = async (body: ReadableStream) => {
...
             ChecksumSHA256: hash,
-            Key: pathForStudyJobResults(info),
+            Key,
             Body: upStream,
         },
     })
     await uploader.done()
 }
 
+// export const storeResultsFile = async (info: MinimalJobResultsInfo, body: ReadableStream) => {
```
chrisbendel, line 139: can remove? or keeping around for later?

src/server/study.ts
chrisbendel, line 1: empty file here 🤔

src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx
```diff
@@ -1,18 +1,19 @@
...
-import { StudyProposalFormValues, studyProposalSchema, zodResolver } from './studyProposalSchema'
+import { StudyProposalFormValues, studyProposalSchema, zodResolver } from './schema'
 
 export const StudyProposalForm: React.FC<{ memberId: string; memberIdentifier: string }> = ({ memberId }) => {
     const router = useRouter()
 
-    const handleFileChange = (key: string, file: File | null) => {
-        studyProposalForm.setFieldValue(key as keyof StudyProposalFormValues, file)
-    }
+    // const handleFileChange = (key: string, file: File | null) => {
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx:14]: still needed?

src/app/researcher/study/request/[memberIdentifier]/actions.ts
```diff
@@ -2,16 +2,17 @@
 
 import { USING_CONTAINER_REGISTRY } from '@/server/config'
 import { createAnalysisRepository, generateRepositoryPath, getAWSInfo } from '@/server/aws'
-import { StudyProposalFormValues, studyProposalSchema } from './studyProposalSchema'
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberIdentifier]/actions.ts:5]: i lean towards favoring explicit naming on these schema files, because eventually we're going to have a bunch of different `schema.ts` files and global searching will start to get a little messy (its one of the few things that bothers me with nextjs having multiple `page.tsx` or `route.ts` files). Is there any reason to not keep the explicit naming?

src/middleware.ts
```diff
@@ -106,10 +106,14 @@ export default clerkMiddleware(async (auth, req) => {
...
     matcher: [
-        // Skip Next.js internals and all static files, unless found in search params
-        '/((?!_next|[^?]*\\.(?:html?|css|js(?!on)|jpe?g|webp|png|gif|svg|ttf|woff2?|ico|csv|docx?|xlsx?|zip|webmanifest)).*)',
+        // Skip:
+        //   Next.js internals
+        //   api requests: the api access wrapper accesses DB, but nextjs middleware doesn't support a full node env
+        //   and all static files, unless found in search params
+        '/((?!_next|api|[^?]*\\.(?:html?|css|js(?!on)|jpe?g|webp|png|gif|svg|ttf|woff2?|ico|csv|docx?|xlsx?|zip|webmanifest)).*)',
         // Always run for routes below
         '/(dl|member|researcher)(.*)',
```
chrisbendel, line 118: do we want to include `/admin` in here as well?

chrisbendel (commented):
awesome!! thanks for hitting so much in here, it looks good just a couple small things - as well as a few bigger picture ideas that don't really need to be addressed here but wanted to highlight/talk about in the future 😃

tests/playwright.setup.ts
```diff
@@ -12,4 +18,21 @@ setup('global setup', async ({}) => {
...
         }
     }
+
+    const pubKey = await readTestSupportFile('public_key.pem')
+    const fingerprint = await readTestSupportFile('public_key.sig')
+
+    for (const clerkId of CLERK_MEMBER_TEST_IDS) {
+        const userId = await findOrCreateSiUserId(clerkId, 'Test Researcher User')
+
+        const pkey = await db.selectFrom('memberUserPublicKey').where('userId', '=', userId).executeTakeFirst()
```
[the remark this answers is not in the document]
nathanstitt: yah but I just thought of an issue.  I bet devs who run docker will have issue bc they've ran `npm install` inside docker and have linux libs, will probably get errors running this 😭   I think I'll rework to move this to the `bin/migrate-dev-db` script

tests/study-flow.spec.ts
```diff
@@ -4,9 +4,12 @@ import { visitClerkProtectedPage, test, expect } from './e2e.helpers'
 // this ensures they run in order and will share the study title
 test.describe.configure({ mode: 'serial' })
 
-test.describe('Studies', () => {
-    // const codeLine = 'print("Hello, Tester")'
+test.beforeEach(async ({}, testInfo) => {
+    // Extend timeout for all tests running this hook by 30 seconds.
+    testInfo.setTimeout(testInfo.timeout + 30_000)
```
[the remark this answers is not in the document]
nathanstitt: no really, but my guess is that GH actions sometimes runs much slower than normal, probably because you are unlucky and your job gets assigned an overloaded machine

src/server/actions/study-actions.ts
nathanstitt: yah I thought about that but I don't think we should be selecting fields we don't need just to make our lives easier with TS typing.   Instead I'd prefer to use a type alias kinda like what I used here.

I don't think the doc paths won't be useful at all on the client side because all they're going to hold is the hashed filename.  We'll need to make an endpoint to download the docs which redirect to s3 for the actual download.

src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx
Comment (nathanstitt) [reply src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx:14]: heh, I'm not sure what this was intended do do, I commented it out to see if anything broke.  since it didn't we should remove it.  ty!

src/middleware.ts
nathanstitt: those aren't actually needed since the other regex should match everything EXCEPT the named urls.  They are a good safeguard though case the hat other regex is somehow messed up.  I think I'll comment them to that effect and move them to the front of the array.  I think that'll be a tiny bit more efficient

src/server/actions/study-actions.ts
chrisbendel: totally makes sense - thanks!

src/server/config.ts
```diff
@@ -80,3 +80,9 @@ export type SSOCookieConfig = {
     private_key: string
     public_key: string
 }
+
+export const CLERK_MEMBER_TEST_IDS: Set<string> = new Set(PROD_ENV ? [] : ['user_2srdGHaPWEGccVS6hzftdroHADi'])
+
+export const CLERK_RESEARCHER_TEST_IDS: Set<string> = new Set(
+    PROD_ENV ? [] : ['user_2nGGaoA3H84uqeBOHCz8Ou9iAvZ', 'user_2oiQ37cyMUZuHnEwxjLmFJJY5kR'],
```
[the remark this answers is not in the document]
nathanstitt: yah, I thought about storing them there but then we have to keep GH actions vars env synced up as well as all devs will need to once again update their `.env`  Like you said, I don't see the harm in publicly disclosing the test ids, esp since they're not valid in prod envs.

src/middleware.ts
chrisbendel: gotcha - sounds good!

src/app/researcher/study/request/[memberIdentifier]/actions.ts
```diff
@@ -2,16 +2,17 @@
...
+import { StudyProposalFormValues, studyProposalSchema } from './study-prooposal-schema'
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberIdentifier]/actions.ts:5]: just a typo on the filename `prooposal`

chrisbendel (approved):
looks good just a small typo and we can merge 😄

PR #88, opened 2025-03-18
Admin invite users

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-61

Known issue:
- "invite user with email" is not implemented because no email provider available at this moment. Needs to be implemented once we can send emails ourselves.


nathanstitt:
### BUILD STATUS ❌



src/components/clerkcustom/invite-form.tsx
chrisbendel, line 1: For file organization, I keep going back and forth on this issue personally, but I'm generally leaning towards:

- If the components aren't being reused anywhere and are simply just being defined as client components to be used in a server component, they can live with that server component (ie in this case inside `app/admin/invite` alongside `page.tsx`. I think this might help keep the components folder from becoming the wild west
- If the component is going to be re-used, then we bring it into the shared component folder


I am open to discussion on this for sure, I just have noticed in my own work the components folder has been growing a lot (because I was doing the same thing as you were initially)

src/components/clerkcustom/invite-form.tsx
```diff
@@ -0,0 +1,207 @@
...
+        } catch (error: unknown) {
+            const errorMessage = error instanceof Error ? error.message : 'User creation failed'
+            setError(errorMessage)
+        } finally {
+            setLoading(false)
+        }
+    }
+
+    return (
+        <form onSubmit={handleSubmit}>
```
chrisbendel, line 102: This form can probably be re-written and simplified a ton with [mantine's use-form](https://mantine.dev/form/use-form/) hook and zod for validation, we have a few examples in the codebase using `use-form`. Taking this approach would simplify your `handleSubmit` method considerably as well

i'm happy to jump in and implement the form if you haven't used it before, but lmk! 👍

src/app/admin/organization/table.tsx
```diff
@@ -35,11 +37,16 @@ export function MembersAdminTable() {
 
     return (
         <Flex direction={'column'}>
+            {globalError && (
+                <Alert icon={<Warning size={16} />} color="red" mb="md">
+                    {globalError}
```
chrisbendel, line 42: I think i see what you're trying to do here - i think we can avoid all this prop drilling (ie passing down `onError` to all the different forms, and instead just use [mantine's notifications](https://mantine.dev/x/notifications/). you can do something like:
```notifications.show({ message: 'Your error message', color: 'red' })```

wherever the error is actually coming out, ie [here](https://github.com/safeinsights/management-app/pull/88/files#diff-4135e16cee7d3541903c61727bec9d1aadc891f7879b0dcf46f8f51d92a07015R25)

File: src/server/actions/user-actions.ts [hunk 3]
```diff
@@ -3,7 +3,7 @@
 import { db } from '@/database'
 
 export const getUserIdByClerkId = async (clerkId: string) => {
```
chrisbendel, line 5: We can remove this method entirely, as we have `siUser()` now

src/server/actions/user-actions.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        transaction: vi.fn(),
+    },
+}))
+
+describe('User Actions', () => {
+    afterEach(() => {
+        vi.clearAllMocks() // Reset all mocks between tests
+    })
+
+    describe('getUserIdByClerkId', () => {
```
chrisbendel, line 19: we can also remove all the related `getUserIdByClerkId` tests now

src/server/actions/user-actions.test.ts
```diff
@@ -0,0 +1,309 @@
...
+    db: {
+        selectFrom: vi.fn(),
+        insertInto: vi.fn(),
+        transaction: vi.fn(),
+    },
+}))
+
+describe('User Actions', () => {
+    afterEach(() => {
+        vi.clearAllMocks() // Reset all mocks between tests
```
chrisbendel, line 16: I wonder if we want to add this to the global vitest.setup.ts file in `afterEach`? Is this behavior we'd want for every test? cc @nathanstitt

src/server/actions/user-actions.ts
```diff
@@ -40,3 +40,55 @@ export const findOrCreateSiUserId = async (clerkId: string, name: string) => {
...
+
+export const createUserAction = async (
+    clerkId: string,
+    name: string,
+    isResearcher: boolean,
+    memberId: string,
+    isReviewer: boolean,
+) => {
+    // Run all related inserts in a transaction for atomicity
+    return await db.transaction().execute(async (trx) => {
```
chrisbendel, line 52: There seems to be some overlap now with this method and the `findOrCreateSiUserId` method above. Maybe worth looking at to see if we need both? I can't tell 🤔

src/server/actions/user-actions.ts
```diff
@@ -40,3 +40,55 @@ export const findOrCreateSiUserId = async (clerkId: string, name: string) => {
...
+            })
+            .returningAll()
+            .execute()
+
+        if (!memberUserResults.length) {
+            throw new Error('Failed to create member_user record')
+        }
+
+        // Return the created user record.
+        return results[0]
```
chrisbendel, line 92: Rather than getting an array back from the query (using `.execute()`) you can use `.executeTakeFirst()` or `.executeTakeFirstOrThrow()` if you want it to throw on error

chrisbendel (changes requested):

src/server/actions/user-actions.test.ts
nathanstitt: we shouldn't need the `clearAllMocks` because we have `mockReset: true,` turned on in the vitest.config.ts which does this automatically

src/components/clerkcustom/invite-form.tsx
```diff
@@ -0,0 +1,207 @@
...
+                password,
+            })
+            // save user in DB
+            await createUserAction(
+                result.clerkId,
+                `${firstName} ${lastName}`,
+                formValues.roles.researcher,
+                selectedOrganization, // Use the selected organization ID directly
+                formValues.roles.reviewer,
+            )
```
nathanstitt, line 77: something to keep in mind is that all actions happen over the network the same as if you were firing off `fetch` calls.  This means that you're relying on the user having a stable network connection and not closing the page, navigating away etc so that they both occur.  it'd be much better to create a single action that performs these steps inside a transaction if possible

src/components/clerkcustom/invite-form.tsx
```diff
@@ -0,0 +1,207 @@
...
+                </Alert>
+            )}
+            <div
+                style={{
+                    display: 'flex',
+                    width: '100%',
+                    gap: '1rem',
+                    alignItems: 'flex-end',
+                    marginBottom: '1rem',
+                }}
```
nathanstitt, line 121: two suggestions: 
* Mantine has a lot of components to do this sort of thing, such as Group, Stack, Flex, each of which have props for setting styles such as `mb` (for marginBottom) etc.
* I'd prefer to use raw style calls only as a last resort, instead we can use the style from pandacss if we really do need to apply custom styling

src/server/actions/user-actions.test.ts
```diff
@@ -0,0 +1,309 @@
...
+import { getUserIdByClerkId, getMemberUserPublicKey, findOrCreateSiUserId, createUserAction } from './user-actions'
+
+// Mock the database
+vi.mock('@/database', () => ({
+    db: {
+        selectFrom: vi.fn(),
+        insertInto: vi.fn(),
+        transaction: vi.fn(),
+    },
+}))
```
nathanstitt, line 12: I don't think we should mock the DB unless we really need to.  we're using `pg-transactional-tests` library which runs everything inside a transaction that's rolled back at the end of each spec.

bin/migrate-dev-db
```diff
@@ -5,10 +5,11 @@ set -e
 TYPES=src/database/types.ts
 
 npx kysely migrate:latest
-npx kysely seed:run
 
 npx kysely-codegen --camel-case --dialect postgres --out-file $TYPES
 
+npx kysely seed:run
```
chrisbendel, line 11: just out of curiosity wondering why the order of operations changed

src/server/actions/user-actions.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+describe('User Actions', () => {
+    afterEach(() => {
+        vi.clearAllMocks() // Reset all mocks between tests
```
chrisbendel, line 16: Should be able to remove this afterEach now

src/server/actions/member-actions.ts
```diff
@@ -23,6 +30,10 @@ export const upsertMemberAction = async (member: Member | NewMember) => {
     return results[0]
```
chrisbendel, line 30: do we want to `executeTakeFirstOrThrow` for this call on line 15 instead of returning[0] and throwing an error manually? Might be simpler

src/server/actions/study-actions.ts
```diff
@@ -26,7 +26,7 @@ export const fetchStudiesForMember = async (memberIdentifier: string) => {
             'study.researcherId',
             'study.status',
             'study.title',
-            'user.name as researcherName',
+            'user.fullName as researcherName',
```
chrisbendel, line 29: in general should we just prefer using fullName over firstName everywhere? for my own knowledge going forward 😄

src/lib/string.ts
```diff
@@ -21,3 +21,13 @@ export function slugify(str: string) {
 
     return str.slice(0, 50)
 }
+
+export function randomString(length: number) {
```
chrisbendel, line 25: We're using this to mock fake data, but we're already pulling in faker so we can prob use that in place of this, probably `faker.string.alpha(8) // 'fEcAaCVbaR'`

src/app/admin/invite/organization-select.tsx
```diff
@@ -0,0 +1,40 @@
+import { Text, Select, type SelectProps } from '@mantine/core'
+import { useQuery } from '@tanstack/react-query'
+import { fetchMembersForSelectAction } from '@/server/actions/member-actions'
+
+export const OrganizationSelect: React.FC<SelectProps> = (props) => {
+    const {
+        data: members = [],
```
chrisbendel, line 7: just in general, we're doing a lot of renaming from member -> organization, do we want to follow this general rule in code? ie here we have `OrganizationSelect` which internally uses `members` as a variable, i think this can get confusing down the line

dont think we have to do this now for time sake, but i don't want to lose track of it

also... i think this is a US vs EU thing but we're doing both `organization` and `organisation` in here 😆

src/app/admin/invite/invite-form.tsx
```diff
@@ -0,0 +1,99 @@
...
+        validateInputOnBlur: true,
+        initialValues: {
+            firstName: '',
+            lastName: '',
+            email: '',
+            organizationId: '',
+            password: randomString(8),
+            isReviewer: false,
+            isResearcher: false,
+        } as InviteUserFormValues,
```
chrisbendel, line 27: we shouldn't need this `as InviteUserFormValues` right?

src/app/admin/invite/invite-form.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const studyProposalForm = useForm({
+        mode: 'uncontrolled',
+        validate: zodResolver(inviteUserSchema),
+        validateInputOnBlur: true,
+        initialValues: {
+            firstName: '',
+            lastName: '',
+            email: '',
+            organizationId: '',
+            password: randomString(8),
```
chrisbendel, line 24: commented in the `lib/string` about using faker to generate the string here if we want to

src/app/admin/invite/admin-users.schema.ts
```diff
@@ -0,0 +1,25 @@
...
+        email: z.string().nonempty().email('This is not a valid email.'),
+        password: z.string().nonempty('password is required'),
+        isReviewer: z.boolean().optional(),
+        isResearcher: z.boolean().optional(),
+        organizationId: z.string().nonempty('origination must be selected'),
+    })
+    .superRefine((data, ctx) => {
+        if (!data.isReviewer && !data.isResearcher) {
+            ctx.addIssue({
+                path: ['role'],
```
chrisbendel, line 18: i might be blind here but is `role` defined in the schema above? 🤔

chrisbendel (approved):
just some topical stuff / comments for my own learning, but can push some of the bigger stuff off in favor of getting this in sooner

bin/migrate-dev-db
nathanstitt: because the seed files had references to changed columns and it failed because the types file wasn't updated yet, I should add a comment to that effect in the script

src/server/actions/member-actions.ts
nathanstitt: great point, thanks!

src/server/actions/study-actions.ts
nathanstitt: yes, I think so, unless there's a table where we want to sort by lastName column or something like that

src/lib/string.ts
nathanstitt: hm, maybe but I like this because it skips easily confused letters like `0O` etc

src/app/admin/invite/organization-select.tsx
nathanstitt: yah, I think we should discuss if we should rename `member` references in code post-mvp

src/app/admin/invite/invite-form.tsx
nathanstitt: we kinda do, putting it here also types the useForm so that the it's typed everywhere.  However it'd be clearer to be explicit and use `useForm<InviteUserFormValues>`, I'll update to do that

src/app/admin/invite/admin-users.schema.ts
nathanstitt: no it's not,  adding it here is kinda adding some meta info but isn't really typed.  There might be a better way to do this, I'm not sure.

src/lib/string.ts
chrisbendel: there is an [exclude characters](https://fakerjs.dev/api/string#alpha) in the options for this too if thats a concern

src/app/admin/invite/invite-form.tsx
chrisbendel: yeah i generally try to avoid `as` because it's bitten me in the past pretending like it actually complies when it doesnt if you add or remove a prop later on 😭

src/app/admin/invite/admin-users.schema.ts
chrisbendel: yea im still super flaky on the zod stuff too 🤣 it seems theres a million ways to do the same thing

src/app/admin/invite/organization-select.tsx
therealmarv: Beginning post mvp sounds like a good time frame to tackle this (and we should stick to organiZation when singular according to US English).

PR #89, opened 2025-03-18
check for MFA in layout vs middleware

nathanstitt:
This way it'll run on only client rendered pages as well as server-rendered ones

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -1,4 +1,8 @@
+'use server'
+
 import type { Metadata } from 'next'
+import { redirect } from 'next/navigation'
+import { currentUser } from '@clerk/nextjs/server'
 
 import './globals.css'
 import '@mantine/core/styles.layer.css'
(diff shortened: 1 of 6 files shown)


nathanstitt:
### BUILD STATUS ✅

[Preview Link](https://pr89.dev.safeinsights.org/)

src/app/layout.tsx
```diff
@@ -7,21 +11,28 @@ import '@mantine/dropzone/styles.layer.css'
...
-import { ReactNode } from 'react'
+import { type ReactNode } from 'react'
 
-export const metadata: Metadata = {
-    title: 'SafeInsights Management Application',
-    description: 'Manages studies, members, and data',
-    icons: {
-        icon: '/icon.png',
-    },
+export async function generateMetadata(): Promise<Metadata> {
```
chrisbendel, line 16: any reason to switch this other than stylistic? since this is all static data?

https://nextjs.org/docs/app/api-reference/functions/generate-metadata

src/app/layout.tsx
```diff
@@ -7,21 +11,28 @@ import '@mantine/dropzone/styles.layer.css'
 
 import { Providers } from './providers'
 import { AppLayout } from '@/components/layout/app-layout'
-import { ReactNode } from 'react'
+import { type ReactNode } from 'react'
```
chrisbendel, line 14: curious why this had to become a type? i think we're doing just `ReactNode` in other components, just wondering if theres something i can learn here 😆

chrisbendel (approved):

src/app/layout.tsx
nathanstitt: the file switched to server-rendered, and nexts complained that "only functions may be exported from server-rendered components"

src/app/layout.tsx
nathanstitt: heh, TBH, I don't even recall why I switched that.  As far as I know exporting a type is a little better because it makes the code easier to treeshake, the compiler will now know that it's completely safe to remove that from the build.  Of course we'll always have React included in the build so this doesn't really help at all

src/app/layout.tsx
chrisbendel: weird!! good to know

tests/vitest.setup.ts
```diff
@@ -3,16 +3,44 @@ import { beforeAll, beforeEach, afterEach, afterAll, vi } from 'vitest'
 import { testTransaction } from 'pg-transactional-tests'
 import { createTempDir } from '@/tests/unit.helpers'
 import fs from 'fs'
+//import * as mockRouter from 'next-router-mock';
```
chrisbendel, line 6: comment here

PR #91, opened 2025-03-19
store publicKey as binary data

nathanstitt:
Is more efficient and believe it will parse easier then hex on the browser

### DIFF
--- a/src/app/account/keys/user-key-actions.ts
+++ b/src/app/account/keys/user-key-actions.ts
@@ -1,31 +1,27 @@
 'use server'
 
 import { db } from '@/database'
-import { getUserIdByClerkId } from '@/server/actions/user-actions'
+import { siUser } from '@/server/queries'
 
 export const getMemberUserPublicKey = async (clerkId: string) => {
     const result = await db
-        .selectFrom('memberUserPublicKey')
-        .innerJoin('user', 'memberUserPublicKey.userId', 'user.id')
-        .select(['memberUserPublicKey.value as memberUserPublicKey'])
+        .selectFrom('userPublicKey')
+        .innerJoin('user', 'userPublicKey.userId', 'user.id')
+        .select(['userPublicKey.publicKey as memberUserPublicKey'])
         .where('user.clerkId', '=', clerkId)
         .executeTakeFirst()
 
     return result?.memberUserPublicKey
 }
 
-export const setMemberUserPublicKey = async (clerkId: string, publicKey: string, fingerprint: string) => {
-    const userId = await getUserIdByClerkId(clerkId)
-
-    if (!userId) {
-        throw new Error(`User for clerk id ${clerkId} doesn't exist!`)
-    }
+export const setMemberUserPublicKey = async (publicKey: ArrayBuffer, fingerprint: string) => {
+    const user = await siUser()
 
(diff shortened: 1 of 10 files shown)


src/app/account/keys/user-key-actions.ts
```diff
@@ -1,31 +1,27 @@
 'use server'
 
 import { db } from '@/database'
-import { getUserIdByClerkId } from '@/server/actions/user-actions'
+import { siUser } from '@/server/queries'
 
 export const getMemberUserPublicKey = async (clerkId: string) => {
```
nathanstitt, line 6: this is dangerous, what would happen if user A called this endpoint with user B's clerkId?  Instead we should always lookup user info from Clerk directly, we have a few helpers for that, I used siUser since it'll throw if the user isn't found

src/database/migrations/1742320602314_rename_member_user_public_key_table.ts
```diff
@@ -1,12 +1,18 @@
...
-    await db.schema.alterTable('memberUserPublicKey').renameTo('userPublicKey').execute()
+    await db.schema.alterTable('member_user_public_key').renameTo('user_public_key').execute()
+
+    await db.schema.alterTable('user_public_key').renameColumn('value', 'public_key').execute()
+    await db.schema.alterTable('user_public_key').addUniqueConstraint('unique_user_id', ['user_id']).execute()
+
+    await db.schema
+        .alterTable('user_public_key')
+        .alterColumn('public_key', (col) => col.setDataType(sql`bytea USING public_key::bytea`))
+        .execute()
```
nathanstitt, line 13: migrations don't use camelcase, they have to match what the tables/fields are in the DB

nathanstitt:
@chrisbendel no we shouldn't have to.   when we select the publicKey from the db, kysley returns it as a `Buffer`.   then as you noticed in the test, JSON.stringify(Buffer) will encode it as `{ type: 'Buffer', data: <binary array> }`.      If we pass Buffer through an action call, nextjs handles all that for us though and it's a Buffer on the client as well.

chrisbendel:
@nathanstitt gotcha - so basically it will be a buffer everywhere? With a [route like this](https://github.com/safeinsights/management-app/blob/main/src/app/api/job/%5BjobId%5D/keys/route.ts) where the TOA is requesting public keys, would this impact them in any way either? Or are strings and buffers effectively the same thing 🤔

src/app/account/keys/user-key-actions.ts
chrisbendel: I forgot that all server actions are technically public endpoints 😭

nathanstitt:
Yeah was just considering that route and adding changes to the encryption pr to better support it.  

Strings & Buffers are similar datatypes but Buffers are for holding binary data and Strings typically hold things that can be viewed by humans.  

What we were doing previously was transmitting the public keys as hex encoded binary data in a string, which requires parsing and converting to a Buffer.     This switches to transmitting data in a format that is easier to convert because it allows us to skip the hex conversion step.

chrisbendel (approved):
per conversation in 🅿️ we should be good to go on this with all the other changes in TOA/encryption lib

PR #92, opened 2025-03-21
redesign of mfa pages

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-76


nathanstitt:
### BUILD STATUS ❌



src/app/account/mfa/sms/page.tsx
```diff
@@ -1,199 +1,186 @@
...
-}
-
-export default function ManageSMSMFA() {
-    const [showBackupCodes, setShowBackupCodes] = React.useState(false)
-    const { openUserProfile } = useClerk()
-    const { isLoaded, user } = useUser()
-
     if (!isLoaded) return null
 
     if (!user) {
```
chrisbendel, line 102: I think this can be brought up before all the other function definitions, right after your `useState` calls. we should probably be redirecting too since they can't access this page at all if there's no user?

Moving this up would also allow all your other `user?` references to be just `user` as there's no null possibility. I would also move the `if (!isLoaded) return null` up too

src/app/account/mfa/sms/page.tsx
```diff
@@ -1,199 +1,186 @@
...
+    const [verificationSuccess, setVerificationSuccess] = React.useState(false)
+    const [codeSent, setCodeSent] = React.useState(false)
+    const [isSendingCode, setIsSendingCode] = React.useState(false)
+    const [isVerifying, setIsVerifying] = React.useState(false)
+    const [phoneResourceToVerify, setPhoneResourceToVerify] = React.useState<PhoneNumberResource | null>(null)
+
+    // Determine if there's an existing phone record
+    const existingPhone = user?.phoneNumbers?.[0]
+    const prefilledPhone = existingPhone ? existingPhone.phoneNumber : phoneNumber
+    const isPhoneEditable = !existingPhone
```
chrisbendel, line 26: it seems that at this point, if they already have an existing phone - they cant do anything in the form. why go through all the trouble of rendering all the disabled state if they are effectively locked out of using it? We can probably bail out and remove a lot of the logic around this later on in the component if we don't need to handle this case? 🤔

src/app/account/mfa/sms/page.tsx
```diff
@@ -1,199 +1,186 @@
...
+export default function ManageSMSMFA() {
+    const { isLoaded, user } = useUser()
+    const [phoneNumber, setPhoneNumber] = React.useState('')
+    const [verificationCode, setVerificationCode] = React.useState('')
+    const [error, setError] = React.useState('')
+    const [verificationSuccess, setVerificationSuccess] = React.useState(false)
+    const [codeSent, setCodeSent] = React.useState(false)
+    const [isSendingCode, setIsSendingCode] = React.useState(false)
+    const [isVerifying, setIsVerifying] = React.useState(false)
+    const [phoneResourceToVerify, setPhoneResourceToVerify] = React.useState<PhoneNumberResource | null>(null)
```
chrisbendel, line 21: There is a lot of state in this single component which seems to get pretty complicated in the JSX, i wonder if we can extract some of these out into their own smaller components so its more easily testable? one immediate area i notice is the three different conditions in the jsx for `!verificationSuccess &&` and `verificationSuccess &&`. though maybe it doesn't make sense to over abstract this stuff since we don't really need to reuse it anywhere else..

chrisbendel (changes requested):

src/app/account/mfa/app/page.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+        mockUseUser.mockReturnValue({
+            isLoaded: true,
+            isSignedIn: true,
+            user: {
+                id: 'user_123',
+                createTOTP: mockCreateTOTP,
+                verifyTOTP: mockVerifyTOTP,
+                // Mock any other user properties accessed by the component
+            },
+        } as unknown as UseUserReturn)
```
nathanstitt, line 55: the aforementioned mockClerkSession will do this for you

src/app/account/mfa/sms/page.tsx
therealmarv: @nathanstitt Yes, that idea was rejected during the process. I wasn’t aware that editing the Clerk profile screen was the preferred approach; otherwise, I likely would have voiced my objections to that design/UX decision.

nathanstitt: eh, I don't know if it was _preferred_ , just didn't know what we'd decided 😁 ty!

nathanstitt (approved):

PR #95, opened 2025-03-25
Check user access before calling actions

nathanstitt:
Each NextJS action is a network endpoint that could be guessable and callable by an unauthenticated user.  

This adds a set of functions to wrap each action and check the proper level of access before calling it

It also adds a linter to check that actions conform to our standards:
* should be in a file ending in `actions.ts`
* the function name should end in `Action`
* it must be wrapped by one of the designated access checks: 'adminAction', 'researcherAction', 'memberAction', 'userAction', 'anonAction'


nathanstitt:
### BUILD STATUS ❌



therealmarv (approved):

PR #97, opened 2025-03-25
Various UI Improvements

chrisbendel:


src/database/migrations/1742904221263_add_reviewed_by_to_study.ts
```diff
@@ -3,10 +3,10 @@ import { Kysely } from 'kysely'
...
         .alterTable('study')
-        .addColumn('reviewed_by', 'uuid', (col) => col.references('user.id'))
+        .addColumn('reviewer_id', 'uuid', (col) => col.references('user.id'))
         .execute()
 }
 
 export async function down(db: Kysely<unknown>): Promise<void> {
-    await db.schema.alterTable('study').dropColumn('reviewed_by').execute()
+    await db.schema.alterTable('study').dropColumn('reviewer_id').execute()
 }
```
nathanstitt, line 12: since we've been deploying, you'll have to create a new migration to rename the column.  otherwise it'd would fail to run on environments where the migration previously ran

nathanstitt (changes requested):
you'll have to add a new migration

nathanstitt:
### BUILD STATUS ✅

[Preview Link](https://pr97.dev.safeinsights.org/)

src/database/migrations/1742904221263_add_reviewed_by_to_study.ts
chrisbendel: damn, really shot myself in the foot with this 😭 

in the future, if things don't get merged to dev (cause we had merged this earlier to get the big PR out of the way) - we wouldn't have to rewrite a migration until its merged and deployed correct?

PR #98, opened 2025-03-25
Trigger code build when study is approved

nathanstitt:
Previously image builds were triggered when a `approved` file is uploaded but that stopped working because the code didn't find a manifest.json file it needed.

Rather than dig through that IaC code, it is simpler to call the codebuild itself with the info it needed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
             "version": "0.0.1",
             "license": "AGPL-3.0-or-later",
             "dependencies": {
+                "@aws-sdk/client-codebuild": "^3.775.0",
                 "@aws-sdk/client-ecr": "^3.758.0",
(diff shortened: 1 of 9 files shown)


philschatz (approved):
I'm not super-familiar but the crux seems to be around the call to `triggerBuildImageForJob`. 

If it's left in the transaction then when `triggerBuildImageForJob` fails the study will not be marked `APPROVED` which seems a little unexpected.

So, there is an additional implied state of a study: `APPROVED_BUT_BUILD_FAILED` or something but that's a future problem

nathanstitt:
Good point, we should really have a BUILD_FAILED status.  Right now the image builds _should_ never fail though because they're literally just a Dockerfile `COPY . .`

in the future they may be a lot different though especially when we support multiple languages.

PR #99, opened 2025-03-26
BMA - Study Details

jbwilson8:
[OTTER-86](https://openstax.atlassian.net/browse/OTTER-86)

[OTTER-86]: https://openstax.atlassian.net/browse/OTTER-86?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nathanstitt:
### BUILD STATUS ❌



src/app/dl/study/[memberIdentifier]/[studyId]/docs/[...path]/route.ts
```diff
@@ -0,0 +1,85 @@
...
+            headers: {
+                'Content-Disposition': `attachment; filename="${filename}"`,
+                'Content-Type': 'application/octet-stream',
+            },
+        })
+    } catch (error) {
+        console.error('Document download error:', error)
+        return NextResponse.json({ error: 'Internal server error' }, { status: 500 })
+    }
+}
```
Comment (nathanstitt) [inline src/app/dl/study/[memberIdentifier]/[studyId]/docs/[...path]/route.ts:85]: Sorry @jbwilson8 this is outdated, we now have a catchall route at `src/app/dl/[...slug]/route.ts` to handle this.  that file has different handlers for each file type

src/app/member/[memberIdentifier]/study/[studyIdentifier]/review/study-job-files.tsx
```diff
@@ -13,21 +13,32 @@ export const StudyJobFiles: FC<{ job: StudyJob }> = ({ job }) => {
...
     }
 
+
     const fileNames = Object.keys(data?.manifest.files || {})
 
+
     const fileChips = fileNames.map((fileName) => {
+        const downloadUrl = data.jobInfo
+            ? `/analysis/${data.jobInfo.memberIdentifier}/${data.jobInfo.studyId}/${data.jobInfo.studyJobId}/code/${fileName}`
+            : '#'
```
Comment (nathanstitt) [inline src/app/member/[memberIdentifier]/study/[studyIdentifier]/review/study-job-files.tsx:35]: this should be updated in main to point to the new dl handler

src/app/researcher/study/request/[memberIdentifier]/study-document-files.tsx
```diff
@@ -0,0 +1,42 @@
...
+        path: string
+    }
+
+    const documentChips = data.documents.map((doc: Document) => (
+        <Badge
+            key={doc.path}
+            color="#D4D1F3"
+            c="black"
+            component="a"
+            href={`/dl/study/${data.studyInfo.memberIdentifier}/${studyId}/docs/${doc.path}`}
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberIdentifier]/study-document-files.tsx:31]: we need to always use the lib/path.ts helpers for determining the path to a file

src/components/study/study-code-details.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+    const member = use(getMemberIdFromIdentifierAction(memberIdentifier))
+    if (!member) {
+        return <AlertNotFound title="Member was not found" message="no such member exists" />
+    }
+
+    const study = use(getStudyAction(studyIdentifier))
+    if (!study) {
+        return <AlertNotFound title="Study was not found" message="no such study exists" />
+    }
```
nathanstitt, line 25: since StudyCodeDetails is called from other pages which should also be checking  for member & study existence, why do we need the checks and alerts?  Why don't we have the pages pass in the member & study for this to display?

src/components/study/study-proposal-details.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const study = use(getStudyAction(studyIdentifier))
+    if (!study) {
+        return <AlertNotFound title="Study was not found" message="no such study exists" />
+    }
+
+    const documents = use(dataForStudyDocumentsAction(studyIdentifier))
+    if (!documents) {
+        return <AlertNotFound title="Documents were not found" message="no such documents exists" />
+    }
+
```
nathanstitt, line 32: same comment as StudyCodeDetails, I think we can pass in member and study.  Maybe it would makes sense for this to check if the docs exists, but if we type the Study that's passed in to have those properties be non-optional then we wouldn't need this check as well

src/server/actions/study-job.actions.ts
```diff
@@ -65,7 +71,7 @@ export const rejectStudyJobResultsAction = memberAction(async (info) => {
     revalidatePath(`/member/[memberIdentifier]/study/${info.studyId}/review`)
 }, minimalJobInfoSchema)
 
-export const dataForJobAction = memberAction(async (studyJobIdentifier) => {
+export const dataForJobAction = userAction(async (studyJobIdentifier) => {
```
nathanstitt, line 74: While it's always hard to come up with good names, a good rule is that methods should never be called `data` 😁 instead it's better to use a more descriptive name.  Maybe this could be `loadStudyJobAction` ?

src/server/actions/study-job.actions.ts
```diff
@@ -104,27 +110,83 @@ export const dataForJobAction = memberAction(async (studyJobIdentifier) => {
     return { jobInfo, manifest }
 }, z.string())
 
-export const latestJobForStudyAction = memberAction(async (studyId) => {
+export const dataForStudyDocumentsAction = async (studyId: string) => {
```
nathanstitt, line 113: Maybe `loadDocumentsForStudyAction` ?

src/server/actions/study-job.actions.ts
```diff
@@ -104,27 +110,83 @@ export const dataForJobAction = memberAction(async (studyJobIdentifier) => {
...
+            path: studyInfo.irbDocPath,
+        })
+    }
+
+    if (studyInfo.agreementDocPath) {
+        documents.push({
+            name: studyInfo.agreementDocPath || 'Agreement Document',
+            path: studyInfo.agreementDocPath,
+        })
+    }
```
nathanstitt, line 155: It doesn't seem like these if statements and object assignments are valid to me.  You're checking if the path is set, and then if it is you're pushing an object with the path for both name and path properties.  The name has a `||` but that will always be true because of the surrounding `if` statement.

nathanstitt: instead of all the `if` statements, I think you could use an array with filter/Boolean:
```
return {
  studyInfo,
  documents: [studyInfo.agreementDocPath, studyInfo.descriptionDocPath,studyInfo.irbDocPath].filter(Boolean)
} 
```
The Boolean will cast each to a bool and filter will remove any that are not true

src/server/storage.ts
```diff
@@ -72,6 +73,71 @@ async function storeFile(filePath: string, info: MinimalStudyInfo, file: File) {
     }
 }
 
+async function retrieveFile(
```
nathanstitt, line 76: why is this needed vs the existing `fetchFile` function, or the `urlOrContentForFile` ?

src/server/actions/study-job.actions.ts
```diff
@@ -116,26 +111,29 @@ export const dataForJobAction = memberAction(async (studyJobIdentifier) => {
...
         .innerJoin('study', 'study.id', 'studyJob.studyId')
-        .innerJoin('member', (join) =>
-            join.on('member.identifier', '=', slug).onRef('member.id', '=', 'study.memberId'),
+        .$if(Boolean(ctx?.orgSlug), (qb) =>
+            qb.innerJoin('member', (join) =>
+                join.on('member.identifier', '=', ctx.orgSlug!).onRef('member.id', '=', 'study.memberId'),
+            ),
         )
-
+        .$if(Boolean(ctx?.userId && !ctx?.orgSlug), (qb) => qb.where('study.researcherId', '=', ctx?.userId || ''))
```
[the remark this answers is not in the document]
nathanstitt: we can leave it for now but I think we can change this to be simpler like what  I reworked here: https://github.com/safeinsights/management-app/blob/main/src/server/db/queries.ts#L38-L40

src/server/actions/study.actions.ts
```diff
@@ -78,17 +85,20 @@ export const fetchStudiesForCurrentMemberAction = memberAction(async () => {
...
 
         // security, check user has access to record
-        .innerJoin('member', (join) =>
-            join.on('member.identifier', '=', slug).onRef('member.id', '=', 'study.memberId'),
+        .$if(Boolean(ctx?.orgSlug), (qb) =>
+            qb.innerJoin('member', (join) =>
+                join.on('member.identifier', '=', ctx.orgSlug!).onRef('member.id', '=', 'study.memberId'),
+            ),
         )
+        .$if(Boolean(ctx?.userId && !ctx?.orgSlug), (qb) => qb.where('study.researcherId', '=', ctx?.userId || ''))
```
nathanstitt, line 102: same comment as above we can simplify to https://github.com/safeinsights/management-app/blob/main/src/server/db/queries.ts#L38-L40

src/server/db/queries.ts
```diff
@@ -136,10 +135,19 @@ export const getMemberUserPublicKeyByClerkId = async (clerkId: string) => {
...
         .selectAll('studyJob')
-
+        // security, check user has access to record
+        .innerJoin('study', 'study.id', 'studyJob.studyId')
+        .$if(Boolean(ctx?.orgSlug), (qb) =>
+            qb.innerJoin('member', (join) =>
+                join.on('member.identifier', '=', ctx.orgSlug!).onRef('member.id', '=', 'study.memberId'),
+            ),
+        )
+        .$if(Boolean(ctx?.userId && !ctx?.orgSlug), (qb) => qb.where('study.researcherId', '=', ctx?.userId || ''))
```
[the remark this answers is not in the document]
nathanstitt: same, use https://github.com/safeinsights/management-app/blob/main/src/server/db/queries.ts#L38-L40

nathanstitt (approved):
@chrisbendel and @jbwilson8 looks great! ty both very much.  I think it would be good to simplify the access checks as noted in the comments but I'm also fine if we do that as a follow up if ya'll would prefer that

src/server/actions/study.actions.ts
chrisbendel: much cleaner - thank you!

PR #100, opened 2025-03-26
Fix admin invite and admin org view

therealmarv:
Hot fix for pages https://openstax.atlassian.net/browse/OTTER-61

```
http://localhost:4000/admin/invite
http://localhost:4000/admin/organization
```

Errors:
* Next JS error on organisations list view
* no org visible
* org selector broken on invite

Now it works if user is in SafeInsights Clerk org (bottom):
image

### DIFF
--- a/src/server/actions/wrappers.ts
+++ b/src/server/actions/wrappers.ts
@@ -82,7 +82,7 @@ export function adminAction<S extends Schema, F extends WrappedFunc<S>>(func: F,
     const wrappedFunction = async (arg: z.infer<S>): Promise<any> => {
         const store = localStorageContext.getStore()
         if (store?.orgSlug !== 'safe-insights') {
-            logger.error("Current orgSlug in adminAction:", store?.orgSlug)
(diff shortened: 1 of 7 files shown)


src/server/actions/wrappers.ts
```diff
@@ -80,8 +81,8 @@ export function adminAction<S extends Schema, F extends WrappedFunc<S>>(func: F,
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const wrappedFunction = async (arg: z.infer<S>): Promise<any> => {
         const store = localStorageContext.getStore()
-
-        if (store?.orgSlug !== 'safeinsights') {
+        if (store?.orgSlug !== 'safe-insights') {
```
nathanstitt, line 84: lets add the slug as an exported variable somewhere, maybe in server/config ?  That way hopefully we don't typo it again like I did here 😭

therealmarv: will do! One moment...

PR #101, opened 2025-03-26
Fix containerLocation on studies

nathanstitt:
The containerizer process pushes all images into a single ECR, this fixes the path to point to it instead of the old one-ecr-per-study


scb6 (approved):

src/server/aws-ecr-policy.ts
```diff
@@ -1,47 +0,0 @@
-import { type AwsIAMPolicy, AwsEcrActions as Action } from 'aws-iam-policy-types'
```
rnathuji, line 1: Just a note: It makes sense that we're removing this since there is no longer a 1:1 between repo and study. However, a couple of thoughts:

1. I think we may need to twiddle the shared repo to allow pulls from the enclave accounts like this policy used to
2. Eventually, will it be okay that members can (at least technically if you ignore the "security" of discoverability) pull RC images from others? (unless I am missing something about the new model, I think that would be an implication at least :man_shrugging: )

rnathuji (approved):

src/server/aws-ecr-policy.ts
nathanstitt: yes for 1) I added that here: https://github.com/safeinsights/iac/blob/mgmntapp-deploy/management-app/containerizer-stack.ts#L54-L71 
2) hm, yes that's a great point, we will need to think about that.   Also, we should probably have a different repo for each base image in order to see the storage savings.    So maybe we'll end up with one-repo-per member+base image?  Something to think about when we move off ECR

PR #107, opened 2025-03-27
fix paths for encrypted/decrypted results

nathanstitt:
ideally decrypted results would be stored with the original filename and encrypted are encrypted.zip. we're not quite there yet but hopefully this fixes the paths for it


nathanstitt:
### BUILD STATUS ✅

[Preview Link](https://pr107.dev.safeinsights.org/)

nathanstitt:
tested on dev with pr env, merging

PR #108, opened 2025-03-27
HiFi Notes

chrisbendel:


nathanstitt:
### BUILD STATUS ❌



src/app/safeinsights-logo.png
nathanstitt, line 1: should we remove the existing logo?

src/theme.ts
```diff
@@ -1,5 +1,109 @@
-import { createTheme } from '@mantine/core'
+import { createTheme, DefaultMantineColor, MantineColorsTuple } from '@mantine/core'
+
```
nathanstitt, line 2: I don't really understand why Mantine theme's require so many colors, they must be using it for something though.  I'm assuming you generated those using their theme builder?

nathanstitt (approved):
LGTM, thanks!

src/theme.ts
chrisbendel: oh no, we have [our own custom color palettes](https://www.figma.com/design/IqiKTkT92Dq9YHn5FjlAKr/SI-UI-Component-Library?node-id=3104-692&t=sQzPU185uTiXxFGB-0) now which is what part of this PR is doing for the hi-fi designs

PR #109, opened 2025-03-28
Fix path issues with app

nathanstitt:
Basically unifies all path and storage handling to server/storage.ts so it's easier to debug and figure out the correct paths for everything


src/server/mailgun.ts
```diff
@@ -0,0 +1,52 @@
...
+            template: 'welcome email',
+            'h:X-Mailgun-Variables': JSON.stringify({
+                fullName,
+                resetPasswordLink: `https://${process.env.DOMAIN_NAME}/account/reset-password`,
+            }),
+        })
+    } catch (error) {
+        logger.error.log('Welcome email error: ', error)
+    }
+}
```
nathanstitt, line 52: @chrisbendel heads up this moves all the lib/mailgun to server/mailgun and reads the config from server/config.  doing it this way means the values can be read from secrets when deployed.

nathanstitt:
### BUILD STATUS ✅

[Preview Link](https://pr109.dev.safeinsights.org/)

PR #112, opened 2025-03-31
view results panel for researchers

nathanstitt:
Also makes changes to how the member review code approves the job.  It now sends the file contents as-is to the action so the name can also be saved

image


src/app/layout.tsx
```diff
@@ -5,7 +5,6 @@ import type { Metadata } from 'next'
 import './globals.css'
 import '@mantine/core/styles.layer.css'
 import 'mantine-datatable/styles.layer.css'
-import '@mantine/dropzone/styles.layer.css'
```
chrisbendel, line 8: removing this removes the styles for the dropzone stuff, so we still need it included

src/server/db/queries.ts
```diff
@@ -26,13 +26,64 @@ export const queryJobResult = async (jobId: string): Promise<MinimalJobResultsIn
     return { ...results, resultsType: results.resultsPath ? 'APPROVED' : 'ENCRYPTED' } as MinimalJobResultsInfo
 }
 
-export const checkMemberAllowedStudyReview = async (studyId?: string, identifier = getOrgSlugFromActionContext()) => {
+export const checkUserAllowedJobView = async (jobId?: string, memberIdentifier?: string) => {
```
chrisbendel, line 29: the `checkUserAllowedJobView` and `checkUserAllowedStudyView` methods are seemingly getting a bit complex with a lot of logic around both optional arguments, im just thinking out loud but reading through it i could see how this might be confusing at first glance. i wonder if there's a way to split out these methods to be more granular with required args 🤔

src/server/actions/study-job.actions.ts
```diff
@@ -159,20 +177,22 @@ export const onFetchStudyJobsAction = memberAction(async (studyId) => {
         .execute()
 }, z.string())
 
-export const fetchJobResultsCsvAction = memberAction(async (jobId: string): Promise<string> => {
+export const fetchJobResultsCsvAction = userAction(async (jobId): Promise<string> => {
+    await checkUserAllowedJobView(jobId)
```
chrisbendel, line 181: one thing im curious about re: implementation on the `xAction` wrappers, is if we followed the same pattern as this `checkUserAllowedNounVerb` for the authentication checks, ie rather than wrapping the method we would simply just call something like `await validateUserAction()` at the top of the method, very similar to these calls where we're validating. 

for me personally its a little confusing to keep track of both, i understand they're different checks - no need to address this on the PR but wanted to highlight it and maybe spark a conversation around the implementation of the server action auth checks, i think when we merged that work things were a bit rushed and didn't have time to fully process it 🤔

src/app/researcher/study/[studyId]/review/page.tsx
```diff
@@ -1,23 +1,26 @@
...
+    const job = await latestJobForStudy(studyId)
+
     if (!study) {
         return <AlertNotFound title="Study was not found" message="no such study exists" />
     }
 
     return (
         <Center>
-            <Paper w="70%" shadow="xs" p="sm" m="xs">
+            <Paper shadow="xs" p="xl" w="100%">
```
Comment (chrisbendel) [inline src/app/researcher/study/[studyId]/review/page.tsx:23]: we don't need the `<Center>` and the subsequent w="100%" on the `Paper` element, im sure there will be a round of hi-fi work to do on these components so its fine to leave for now too

chrisbendel (approved):
just a couple small tweaks and codepilot had a good suggestion on the missing `await`

mostly just some high level stuff to talk about in the future/for me to understand better, but nothing urgent on this PR

src/app/researcher/study/[studyId]/review/page.tsx
Comment (nathanstitt) [reply src/app/researcher/study/[studyId]/review/page.tsx:23]: hah, nope we do not, I'll remove Center

PR #115, opened 2025-04-02
Convert table to server component, add seed data

chrisbendel:


nathanstitt:
### BUILD STATUS ❌



src/app/member/[memberIdentifier]/dashboard/display-study-status.test.tsx
Comment (chrisbendel) [inline src/app/member/[memberIdentifier]/dashboard/display-study-status.test.tsx:1]: Just moved from the studies-table.test.tsx file

kysely.config.ts
```diff
@@ -1,11 +1,13 @@
 import { defineConfig } from 'kysely-ctl'
 import { dialect } from './src/database/dialect'
+import { CamelCasePlugin } from 'kysely'
 
 export default defineConfig({
     dialect,
     migrations: {
         migrationFolder: 'src/database/migrations',
     },
+    plugins: [new CamelCasePlugin()],
```
chrisbendel, line 10: this should let us do camelcase stuff in the migrations/seed files now

src/app/member/[memberIdentifier]/dashboard/display-study-status.tsx
Comment (chrisbendel) [inline src/app/member/[memberIdentifier]/dashboard/display-study-status.tsx:1]: some edge case happened with the `validate-actions` script, it turns out if we try to
- mark a file with `use server`
- export some async functions (ie just smaller server components) from the file

then validate-actions fails, saying it should be DisplayStudyStatusAction

however since the script is explicitly parsing `use server`, this is just technically working around that and doing the implicit `use server` by default. something to deal with later, maybe we can get away from the validate actions when we circle back on the access checks/wrapper stuff

src/app/member/[memberIdentifier]/dashboard/page.tsx
```diff
@@ -17,6 +35,31 @@ export default async function MemberDashboardPage(props: { params: Promise<{ mem
         return <AlertNotFound title="Member was not found" message="no such member exists" />
     }
 
+    const studies = await fetchStudiesForCurrentMemberAction()
+
+    const rows = studies.map((study) => (
+        <TableTr key={study.id}>
```
Comment (chrisbendel) [inline src/app/member/[memberIdentifier]/dashboard/page.tsx:41]: Took some inspiration from you @nathanstitt now this is also a server component 😄

src/database/seeds/1728493930771_add-openstax-member.ts
```diff
@@ -1,36 +1,83 @@
 import type { Kysely } from 'kysely'
+import { DB } from '@/database/types'
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export async function seed(db: Kysely<any>): Promise<void> {
-    await db
+export async function seed(db: Kysely<DB>): Promise<void> {
```
chrisbendel, line 4: @nathanstitt anything else of value you think we should add here in the seed while we're updating it?

nathanstitt: while I understand why we'd want to, I don't think we can do this here 😭 

the seed files run on all environments, including prod, where we do not want testing accounts with semi-public passwords created.

Instead we have a `tests/configure-test-env.ts` that does roughly this, but could probably use more ❤️ 

Also, we can't edit existing seed or db files because they will not re-run on existing db, instead we should create new files using `npx kysely seed` or `npx kysely migrate`

src/database/seeds/1743608138837_test_users.ts
```diff
@@ -0,0 +1,99 @@
+import type { Kysely, Selectable } from 'kysely'
+import { DB, MemberUser, User } from '@/database/types'
+import { db } from '@/database'
+import { v7 as uuidv7 } from 'uuid'
+
+export async function seed(db: Kysely<DB>): Promise<void> {
```
nathanstitt, line 6: we need to check an env var check so this doesn't run in prod, maybe something like:
```suggestion
export async function seed(db: Kysely<DB>): Promise<void> {
    if (process.env.NO_TESTING_DATA) return;
```

nathanstitt (approved):
looks good but do need an EVN var check added

PR #119, opened 2025-04-03
Reset every property after invitation except the selected organisation

therealmarv:
fixes part of this comment https://openstax.atlassian.net/browse/OTTER-61?focusedCommentId=15701 on user invitation.

### DIFF
--- a/src/app/admin/invite/invite-form.tsx
+++ b/src/app/admin/invite/invite-form.tsx
@@ -76,8 +76,14 @@ export function InviteForm() {
                 </Text>
                 <Flex direction="column" gap="sm">
                     <Flex gap="md">
-                        <Checkbox label="Reviewer" {...studyProposalForm.getInputProps('isReviewer', { type: 'checkbox' })} />
-                        <Checkbox label="Researcher" {...studyProposalForm.getInputProps('isResearcher', { type: 'checkbox' })} />
+                        <Checkbox
+                            label="Reviewer"
+                            {...studyProposalForm.getInputProps('isReviewer', { type: 'checkbox' })}
+                        />
+                        <Checkbox
+                            label="Researcher"
+                            {...studyProposalForm.getInputProps('isResearcher', { type: 'checkbox' })}
+                        />
                     </Flex>
                     <Input.Error>{studyProposalForm.errors.role}</Input.Error>
                 </Flex>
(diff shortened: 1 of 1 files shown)


nathanstitt:
### BUILD STATUS ✅

[Preview Link](https://pr119.dev.safeinsights.org/)

src/app/admin/invite/invite-form.tsx
```diff
@@ -36,7 +36,12 @@ export function InviteForm() {
         },
         onSuccess(info) {
             notifications.show({ message: `User invited successfully\nClerk ID: ${info.clerkId}`, color: 'green' })
-            studyProposalForm.setValues(initialValues())
+            // Keep the selected organization, reset other fields using initialValues
+            const currentOrgId = studyProposalForm.values.organizationId
+            const newInitialValues = initialValues()
+            newInitialValues.organizationId = currentOrgId // Restore the organization
+            studyProposalForm.setValues(newInitialValues)
+            studyProposalForm.clearErrors() // Clear validation state explicitly
```
chrisbendel, line 44: https://mantine.dev/form/use-form/

there is a method to do this simpler i think:
```
// Resets form values to `initialValues`,
// clears all validation errors,
// resets touched and dirty state
form.reset();

```

chrisbendel: Also initialValues can just be moved inline, rather than being a function up above ^

chrisbendel (changes requested):

src/app/admin/invite/invite-form.tsx
therealmarv: I think I've got it working with reset now. There was some weird behaviour (probably because I wanted to keep the organisation) but reset() works now in my testing.

nathanstitt (approved):

src/app/admin/invite/invite-form.tsx
```diff
@@ -36,7 +36,10 @@ export function InviteForm() {
         },
         onSuccess(info) {
             notifications.show({ message: `User invited successfully\nClerk ID: ${info.clerkId}`, color: 'green' })
-            studyProposalForm.setValues(initialValues())
+            // Keep the selected organization, reset other fields using initialValues
+            const oldOrgId = studyProposalForm.values.organizationId
+            studyProposalForm.reset()
+            studyProposalForm.setFieldValue('organizationId', oldOrgId)
```
chrisbendel, line 42: take a look at [the docs](https://mantine.dev/form/use-form/#values)

they propose using something like this 🤔 
```
// Sets initial values, used when form is reset
form.setInitialValues({ values: 'object' });
```

chrisbendel (approved):

PR #120, opened 2025-04-03
give a more meaningful error message when user email is already registered

therealmarv:
fix another QA issue of comment https://openstax.atlassian.net/browse/OTTER-61?focusedCommentId=15701

### DIFF
--- a/src/app/admin/invite/admin-users.actions.ts
+++ b/src/app/admin/invite/admin-users.actions.ts
@@ -16,7 +16,7 @@ export const adminInviteUserAction = adminAction(async (invite) => {
     const existingUsers = await client.users.getUserList({
         emailAddress: [invite.email],
     })
-    if (existingUsers.length > 0) {
+    if (existingUsers.data.length > 0) {
         throw new SanitizedError(`A user with the email address '${invite.email}' already exists.`)
     }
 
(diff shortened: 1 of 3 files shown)


src/app/admin/invite/admin-users.actions.ts
```diff
@@ -12,6 +12,14 @@ export const adminInviteUserAction = adminAction(async (invite) => {
...
     let clerkUserId = ''
 
+    // Check if a user with this email already exists in Clerk
+    const existingUsers = await client.users.getUserList({
+        emailAddress: [invite.email],
+    })
+    if (existingUsers.data.length > 0) {
+        throw new SanitizedError(`A user with the email address '${invite.email}' already exists.`)
+    }
+
```
nathanstitt, line 22: I get what you're doing here, but doesn't clerk does that for us when we submit a dupe.  I just tested and saw:

image

I do think we could do a little better error handling than a toast message though, maybe check what I just did in #121 

https://github.com/safeinsights/management-app/pull/121/files#diff-eaa2f04c5b07b4ec606e1fe838bf0c6247c6733918fb0ce89b5cd521cbfa277fR32-R55

nathanstitt (changes requested):
I like the concept of what you're proposing but don't think we need to pre-vet proposed email address instead we can just submit and use the clerk error that's returned

src/app/admin/invite/admin-users.actions.ts
therealmarv: I will try that and refactor, thanks! I'm also somehow mind-stuck by the weird testing errors here.

therealmarv:
I will open a new PR

PR #121, opened 2025-04-03
carefully handle all errors on sign-in/reset

nathanstitt:
Also moves login components alongside pages

### DIFF
--- a/src/app/account/signin/signin.tsx
+++ b/src/app/account/signin/signin.tsx
@@ -6,7 +6,7 @@ import { Loader } from '@mantine/core'
 import { type MFAState } from './logic'
 import { SigninComplete } from './complete'
 import { RequestMFA } from './mfa'
-import { EmailPWForm } from './email-pw'
+import { EmailPasswordForm } from './email-password-form'
 import { onUserSignInAction } from '@/server/actions/user.actions'
 import { reportError } from '@/components/errors'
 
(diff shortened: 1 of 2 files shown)


src/lib/errors.ts
```diff
@@ -29,6 +29,10 @@ export function isClerkApiError(error: unknown): error is ClerkAPIErrorResponse
     )
 }
 
+export function extractClerkCodeAndMessage(error: ClerkAPIErrorResponse) {
```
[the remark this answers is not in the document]
nathanstitt: good point, I changed the typedef to make it clear that `errors` will always have at least one object

src/app/account/signin/mfa.tsx
```diff
@@ -0,0 +1,75 @@
...
+        validate: {
+            code: isNotEmpty('Required'),
+        },
+    })
+
+    const { isPending, mutate: onMFASubmit } = useMutation({
+        async mutationFn(form: { code: string }) {
+            if (!isLoaded || !mfa) return
+
+            return mfa.signIn.attemptSecondFactor({
```
chrisbendel, line 28: Not urgent but when we upgrade we'll have access to [useReverification](https://clerk.com/docs/hooks/use-reverification) which might make this a bit nicer down the line

src/app/account/signin/signin.tsx
```diff
@@ -28,7 +31,7 @@ export function SignIn() {
     return (
         <>
             <SigninComplete />
-            <SignInForm onComplete={setPending} mfa={state} />
+            <EmailPWForm onComplete={setPending} mfa={state} />
```
chrisbendel, line 34: me being a stickler: usually preferring full names like <EmailPasswordForm> but not a show stopper either

chrisbendel (approved):

src/app/account/signin/signin.tsx
nathanstitt: yah I usually like to do that also, don't know why I used `PW` here, it's an easy rename so I'll do so now

PR #122, opened 2025-04-03
e2e test fest fixes

jbwilson8:
fixes for some of the [Test Fest](https://docs.google.com/document/d/1MvnixdU6x_AhPzovj9tc66W1nCxpOaXjE2YW8_xEmZw/edit?tab=t.0) findings (screeshots of resolved items below):

Screenshot 2025-04-03 at 2 13 46 PM

Screenshot 2025-04-03 at 2 14 24 PM

Screenshot 2025-04-03 at 2 15 52 PM

Screenshot 2025-04-03 at 2 18 24 PM

Screenshot 2025-04-03 at 2 29 31 PM

Screenshot 2025-04-03 at 2 47 03 PM

Screenshot 2025-04-03 at 2 47 50 PM


src/app/member/[memberIdentifier]/dashboard/page.tsx
```diff
@@ -24,6 +24,7 @@ import dayjs from 'dayjs'
 import Link from 'next/link'
 import { DisplayStudyStatus } from './display-study-status'
 import { UserName } from '@/components/user-name'
+import { theme } from '@/theme'
```
Comment (chrisbendel) [inline src/app/member/[memberIdentifier]/dashboard/page.tsx:27]: It looks like you're doing this cause `useMantineTheme` can't be used in a server component. we may just want to add one thing to the `theme` [per the docs here](https://mantine.dev/theming/theme-object/#access-theme-outside-of-components)

src/theme.ts
```diff
@@ -106,4 +106,5 @@ export const theme = createTheme({
     },
     primaryShade: 5,
     primaryColor: 'purple',
+    columnWidth: '20%',
```
chrisbendel, line 109: what's this for 🤔 it's a global setting so just want to make sure this is very intentional and thought out

nathanstitt:
### BUILD STATUS ❌



src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx
```diff
@@ -27,100 +39,113 @@ export const StudyProposalForm: FC<{
...
-            <Title order={4} c="purple">
-                Study Proposal
-            </Title>
+            <Title order={4}>Study Proposal</Title>
             <Divider my="sm" mt="sm" mb="md" />
-            <Text>
-                This section is key to your proposal, as it defines the analysis that will generate the results you’re
-                intending to obtain from the Member’s data. Upload any necessary files to support your analysis. In this
-                iteration, we currently support .r and .rmd files.
+            <Text mb="md">
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx]: i typically trend towards using a `<Stack>` component with the gap to handle spacing in a more "flex layout" approach. ie all these sibling items could have an even gap vs setting margins on them one by one

src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx
```diff
@@ -27,100 +39,113 @@ export const StudyProposalForm: FC<{
...
-                            key={studyProposalForm.key('agreementDocument')}
-                            {...studyProposalForm.getInputProps('agreementDocument')}
-                        />
-                        <Text size="xs" c="dimmed">
-                            Accepted formats: doc, docx, pdf
-                        </Text>
-                    </Stack>
-                </Group>
-            </Stack>
+            <Flex>
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx]: I typically try to avoid using `Flex` unless absolutely necessary, and instead prefer the more explicit `Group` or `Stack` depending on which way you want things to render. in this case i wonder if the <Flex> is even necessary at all 🤔

src/app/researcher/study/request/[memberIdentifier]/upload-study-job-code.tsx
```diff
@@ -12,80 +12,100 @@ import { StudyProposalFormValues } from '@/app/researcher/study/request/[memberI
...
+                                <Text size="xs" c="dimmed">
+                                    10MB max
+                                </Text>
+                            </Group>
+                        </Stack>
+                    </Dropzone>
+                    <Divider orientation="vertical" />
+                </GridCol>
+
+                <GridCol span={6} md={6}>
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberIdentifier]/upload-study-job-code.tsx]: it looks like these md properties are invalid 🤔

src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx
```diff
@@ -27,100 +39,113 @@ export const StudyProposalForm: FC<{
...
+                <Table variant="vertical" withRowBorders={false}>
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberIdentifier]/study-proposal.tsx]: I think using a <Table> for the form layout is a little overkill and can probably be done simpler 🤔 maybe something similar to how the `StudyDetails` component lays everything out 🤔

src/app/member/[memberIdentifier]/dashboard/page.tsx
Comment (chrisbendel) [reply src/app/member/[memberIdentifier]/dashboard/page.tsx:27]: Actually - im realizing we can just use `c='charcoal.1'` simple as that! think we were overcomplicating bringing in the full theme object. also the checks complain about `theme` being potentially undefined anyway 🤔

src/theme.ts
chrisbendel: i ended up removing this

PR #123, opened 2025-04-03
Display a popover with job id when job errors

nathanstitt:
I started to refactor in order to display other statuses like 'CODE-REJECTED' but then noticed that they're marked as POST-MVP

I held off implementing them but left the refactoring so it'll be easier to support them in the future

image


nathanstitt:
### BUILD STATUS ❌



chrisbendel (approved):

PR #124, opened 2025-04-03
a few fixes for prod

nathanstitt:
* db is not importable on aws lambda migrator
* popover needs a mantine component as child


src/app/member/[memberIdentifier]/dashboard/display-study-status.tsx
```diff
@@ -13,7 +13,9 @@ const JobIdPopover: PopOverComponent = ({ jobId }) => {
     return (
         <Popover width={200} position="bottom" withArrow shadow="md">
             <PopoverTarget>
-                <Info color="blue" />
+                <ActionIcon variant="transparent">
+                    <Info color="blue" />
+                </ActionIcon>
```
Comment (nathanstitt) [inline src/app/member/[memberIdentifier]/dashboard/display-study-status.tsx:18]: probably should have been using this from the start, its more accessible

chrisbendel (approved):

PR #125, opened 2025-04-07
correct auth role checks for researcher actions

nathanstitt:
We had the role check wrong on the researcher dashboard.  it was actually checking that the user was a reviewer, not researcher.


nathanstitt:
### BUILD STATUS ❌



chrisbendel (approved):
There just seems to be a merge conflict but good otherwise 👍

PR #127, opened 2025-04-08
Add programmatic emails

chrisbendel:
Waiting for mailgun template variables to be finalized


nathanstitt:
### BUILD STATUS ❌



src/app/researcher/study/request/[orgSlug]/actions.ts
```diff
@@ -17,12 +19,17 @@ const onCreateStudyActionArgsSchema = z.object({
 
 export const onCreateStudyAction = researcherAction(async ({ orgSlug, studyInfo }) => {
     const userId = await getUserIdFromActionContext()
+    const user = await currentUser()
+    if (!user) {
+        throw new Error('No user')
+    }
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[orgSlug]/actions.ts:25]: `researcherAction` will require a user, it calls userAction internally. Given that you can call `actionContext()` to get the SiUser which should have name, etc.

src/server/db/queries.ts
```diff
@@ -191,3 +192,22 @@ export async function getFirstOrganizationForUser(userId: string) {
...
+
+    if (role === 'researcher') {
+        query = query.where('orgUser.isResearcher', '=', true)
+    }
+
+    if (role === 'reviewer') {
+        query = query.where('orgUser.isReviewer', '=', true)
+    }
+
+    return await query.execute()
```
nathanstitt, line 212: https://kysely.dev/docs/recipes/expressions#conditional-expressions says not to do this, instead use filters

src/server/mailgun.ts
```diff
@@ -29,6 +34,16 @@ export async function mailGunConfig(): Promise<[null | ReturnType<Mailgun['clien
...
 
+const getStudyAndOrg = async (studyId: string) => {
+    const study = await getStudyAction(studyId)
+    // TODO DISCUSS in general, do we want to throw? or return undefined if no study found inside getStudyAction?
+    if (!study) throw new Error('Study not found')
+    const org = await getOrgFromIdAction(study.orgId)
+    if (!org) throw new Error('Org not found')
+
+    return { study, org }
+}
```
nathanstitt, line 45: I'm conflicted about:
 * do we want to call actions from server-side code?  While technically we can, they do have a bit more overhead because of the wrappers
 * Calling two functions that each run db queries seems inefficient, instead we should make a single query (maybe in db/queries?) to fetch the study and organization information at once.

src/server/mailgun.ts
```diff
@@ -43,10 +58,201 @@ export const sendWelcomeEmail = async (emailTo: string, fullName: string) => {
...
+
+    const { study, org } = await getStudyAndOrg(studyId)
+    const reviewersToNotify = await getUsersByRoleAndOrgId('reviewer', org.id)
+
+    for (const reviewer of reviewersToNotify) {
+        const email = reviewer.email
+        if (!email) continue
+
+        try {
+            await mg.messages.create(domain, {
```
nathanstitt, line 81: instead of creating a new message for each reviewer, can we make a single message and BCC each of them?  Or maybe we should use batch sending? https://help.mailgun.com/hc/en-us/articles/203068874-How-can-I-send-the-same-email-to-multiple-recipients

src/server/mailgun.ts
```diff
@@ -43,10 +58,201 @@ export const sendWelcomeEmail = async (emailTo: string, fullName: string) => {
...
+    }
+}
+
+export const sendStudyProposalApprovedEmail = async (studyId: string) => {
+    const [mg, domain] = await mailGunConfig()
+    if (!mg) return
+
+    const { study, org } = await getStudyAndOrg(studyId)
+
+    const researchersToNotify = await getUsersByRoleAndOrgId('researcher', org.id)
```
nathanstitt, line 106: I think we're only supposed to notify the researcher who created the study right? not every researcher in the org

src/server/mailgun.ts
```diff
@@ -43,10 +58,201 @@ export const sendWelcomeEmail = async (emailTo: string, fullName: string) => {
...
+    }
+}
+
+export const sendStudyProposalRejectedEmail = async (studyId: string) => {
+    const [mg, domain] = await mailGunConfig()
+    if (!mg) return
+
+    const { study, org } = await getStudyAndOrg(studyId)
+
+    const researchersToNotify = await getUsersByRoleAndOrgId('researcher', org.id)
```
nathanstitt, line 138: same as above, we only notify the study creator

src/server/mailgun.ts
```diff
@@ -43,10 +58,201 @@ export const sendWelcomeEmail = async (emailTo: string, fullName: string) => {
...
+        }
+    }
+}
+
+export const sendResultsReadyForReviewEmail = async (studyId: string) => {
+    const [mg, domain] = await mailGunConfig()
+    if (!mg) return
+
+    const { study, org } = await getStudyAndOrg(studyId)
+    const reviewersToNotify = await getUsersByRoleAndOrgId('reviewer', org.id)
```
nathanstitt, line 170: for this email we're only supposed to notify the reviewer who initially reviewed the study

src/server/mailgun.ts
```diff
@@ -43,10 +58,201 @@ export const sendWelcomeEmail = async (emailTo: string, fullName: string) => {
...
+        }
+    }
+}
+
+export const sendStudyResultsApprovedEmail = async (studyId: string) => {
+    const [mg, domain] = await mailGunConfig()
+    if (!mg) return
+
+    const { study, org } = await getStudyAndOrg(studyId)
+    const researchersToNotify = await getUsersByRoleAndOrgId('researcher', org.id)
```
nathanstitt, line 201: same, only study creator gets notified

nathanstitt (commented):
Looking good @chrisbendel but I think we have the logic on who recv's the emails a bit off

src/app/researcher/study/request/[orgSlug]/actions.ts
Comment (chrisbendel) [reply src/app/researcher/study/request/[orgSlug]/actions.ts:25]: user is totally unused here, i just realized we can remove it entirely :)

src/server/db/queries.ts
chrisbendel: i like the 2nd approach linked for sure
```
  .where((eb) => {
    const filters: Expression<SqlBool>[] = []
```

but i think its important to note that both approaches are valid and they dont say to not use it:
```
Having a set of optional filters you want to combine using and, is the most basic and common use case of conditional where expressions. Since the where, having and other filter functions are additive, most of the time this is enough:
```

src/server/mailgun.ts
chrisbendel: 👍 makes sense

src/server/mailgun.ts
chrisbendel: BCC won't work in it's current state as each email has a unique "Hello <your_name>" in the email body... ideally we could remove that and just BCC and our lives would be a lot easier. Batch sending seems to have some extra work in the templates around [recipient variables](https://documentation.mailgun.com/docs/mailgun/user-manual/sending-messages/#recipient-variables) and may be even more work to maintain on that end

chrisbendel: let me check with iris on this to see if we can get away with BCC and no personalized text in the body

src/server/mailgun.ts
chrisbendel: Checking with @iccgoncalves on this, we talked a while ago about the recipient rules and [the ticket](https://openstax.atlassian.net/browse/OTTER-100) says:

```Reviewer == all Reviewers in that organization

Researcher == all Researchers in that organization```

**Comment by chrisbendel** [comment]
@nathanstitt thanks for the review, waiting to hear iris's decision around the proper recipients & bcc notes 😄

File: src/server/mailgun.ts (hunk 4 above)
Comment (nathanstitt) [reply src/server/mailgun.ts:69]: ah, I'd missed that each email was personalized.  I think this is fine, we probably won't be sending even 100s at a time, but if the count does get larger than 10ish we'll need to start using background jobs to send them.  Something to keep in mind is that a network request is pending and the 'approve' button is spinning until this completes

Comment (chrisbendel) [reply src/server/mailgun.ts:69]: just thinking outside the box but what if we just didn't await any of the mailgun calls? they could fire off asynchronously without blocking the UI 🤔

File: src/server/db/queries.ts [hunk 9]
```diff
@@ -189,3 +190,71 @@ export async function getFirstOrganizationForUser(userId: string) {
...
+            'study.irbDocPath',
+            'study.reviewerId',
+            'study.agreementDocPath',
+            'researcher.fullName as researcherName',
+            'org.slug as orgSlug',
+            'org.name as orgName',
+        ])
+        .selectAll('org')
+        .where('study.id', '=', studyId)
+        .executeTakeFirst()
```
nathanstitt, line 251: not really important, but could use:
```suggestion
        .executeTakeFirstOrThrow(() => new Error('Study & Org not found'))
```

nathanstitt (approved):
looks great Chris, ty!

PR #129, opened 2025-04-08
add dbgate UI

therealmarv:


Makefile
chrisbendel, line 1: i like this idea of having some very commonly run commands that we can add/share for local dev, ie migrating, rebuilding, clearing out, etc

nathanstitt:
I'll let those who use docker to review this, but seems fine to me.   I suppose we could make it optional in the docker compose if people didn't want to run it all the time https://stackoverflow.com/questions/45680958/can-you-define-optional-docker-compose-services

jbwilson8:
Looks good and works for me. love not having to run the dbgate commands separately, thx marvin!

jbwilson8 (approved):

therealmarv:
I'll merge it for now but I will do a small parking lot on Monday if people want it optional or always (I want it always).

Technically I have no problem to make it optional. It only makes the docker commands very long from a user perspective.

In a past commit https://github.com/safeinsights/management-app/pull/129/commits/e2948eb67b5f8d237029373150893f47cfb53e41 I've used a Makefile for that reason but removed it at the end.

PR #131, opened 2025-04-09
WIP use s3 presigned URLs to upload study document/code files

chrisbendel:


next.config.ts
```diff
@@ -12,7 +12,7 @@ const nextConfig: NextConfig = async (phase: string) => {
         transpilePackages: ['si-encryption'],
         experimental: {
             serverActions: {
-                bodySizeLimit: '10mb',
+                bodySizeLimit: '6mb',
```
chrisbendel, line 15: Stick with 6mb explicitly since this is all lambda functions can handle

docker-compose.yml
```diff
@@ -33,6 +33,11 @@ services:
             - E2E_CLERK_ADMIN_EMAIL=$E2E_CLERK_ADMIN_EMAIL
             - E2E_CLERK_ADMIN_PASSWORD=$E2E_CLERK_ADMIN_PASSWORD
             - DATABASE_URL=postgres://mgmnt:mgmntpass@postgres:5432/mgmnt_dev
+            - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
```
chrisbendel, line 36: This would require local users to set up their aws credentials in the .env file

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -5,19 +5,126 @@ import { Button, Group, Stack, Text } from '@mantine/core'
...
 import { useRouter } from 'next/navigation'
+import { StudyDocumentType } from '@/lib/types'
+import { zodResolver } from 'mantine-form-zod-resolver'
+import { CodeReviewManifest } from '@/lib/code-manifest'
+import { PresignedPost } from '@aws-sdk/s3-presigned-post'
+import { signedUrlForCodeUploadAction, signedUrlForStudyFileUploadAction } from '@/server/actions/s3.actions'
 
-export const StudyProposal: React.FC<{ memberId: string }> = ({ memberId }) => {
+// TODO @nathan, should we talk about local vs s3 storage?
+//  Could we just use s3 locally and not have to switch on USING_S3 to simplify?
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:21]: discussion not here (commenting for visibility)

src/database/migrations/1742320602314_rename_member_user_public_key_table.ts
```diff
@@ -1,7 +1,6 @@
 import { sql, type Kysely } from 'kysely'
-import { type DB } from '@/database/types'
 
-export async function up(db: Kysely<DB>): Promise<void> {
+export async function up(db: Kysely<unknown>): Promise<void> {
```
chrisbendel, line 3: This shouldn't change any prior migrations but saw [in the docs](https://kysely.dev/docs/migrations#migration-files) they say to explicitly not use your `DB` as the typing, so i reverted these so noone looks at a prior migration file and tries to do the same thing

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -5,19 +5,126 @@ import { Button, Group, Stack, Text } from '@mantine/core'
...
+            descriptionDocument: null,
+            agreementDocument: null,
+            codeFiles: [],
+        },
+    })
+
     const { mutate: createStudy } = useMutation({
         mutationFn: async (formValues: StudyProposalFormValues) => {
-            return await onCreateStudyAction({ memberId, studyInfo: formValues })
+            const { studyId, studyJobId } = await onCreateStudyAction({ memberSlug, studyInfo: formValues })
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:83]: we can't send the formValues as-is into the action because they hold the files which potentially could be larger than lambda or nextjs action size limits.  We'll have to alter the action to accept only the text values and send that, then use the File objects as you have it below. 
You could also have the `onCreateStudyAction` return the urls that will be needed for uploading which would eliminate the need for additional action calls

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -5,19 +5,126 @@ import { Button, Group, Stack, Text } from '@mantine/core'
...
+    }
+
+    const manifestFile = new File([manifest.asJSON], 'manifest.json', { type: 'application/json' })
+    body.append(manifestFile.name, manifestFile)
+
+    await fetch(upload.url, {
+        method: 'POST',
+        body: body,
+    })
+}
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:63]: there doesn't seem to be any error handling here.  what happens if the upload does not succeed?

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -5,19 +5,126 @@ import { Button, Group, Stack, Text } from '@mantine/core'
...
+                const studyCodeUpload = await signedUrlForCodeUploadAction({
+                    memberSlug,
+                    studyId,
+                    studyJobId,
+                })
+
+                await uploadCodeFiles(formValues.codeFiles, studyCodeUpload, studyJobId)
+            } catch (error) {
+                // IF something goes wrong with study file uploads,
+                //  do we want to roll back and delete the study?
```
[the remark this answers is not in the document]
Comment (nathanstitt) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:125]: yes, I think we could:
* Retry uploads X times
* if that doesn't work, show a toast and call an action to delete the partially created study

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (nathanstitt) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:21]: I added to the dev agenda to discuss, I'm hesitant to require it because:
* I want to be able to develop without or with slow internet
* I don't think we want to have a requirement that all devs must have a AWS account

src/database/migrations/1742320602314_rename_member_user_public_key_table.ts
nathanstitt: ah interesting, I guess that makes sense

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -5,19 +5,126 @@ import { Button, Group, Stack, Text } from '@mantine/core'
...
+import { PresignedPost } from '@aws-sdk/s3-presigned-post'
+import { signedUrlForCodeUploadAction, signedUrlForStudyFileUploadAction } from '@/server/actions/s3.actions'
 
-export const StudyProposal: React.FC<{ memberId: string }> = ({ memberId }) => {
+// TODO @nathan, should we talk about local vs s3 storage?
+//  Could we just use s3 locally and not have to switch on USING_S3 to simplify?
+async function uploadFile(file: File, upload: PresignedPost) {
+    const body = new FormData()
+    for (const [key, value] of Object.entries(upload.fields)) {
+        body.append(key, value)
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:25]: what is `key` here,  is it the file name?  If so that means we control the path on the client and can use a single signed url for all the files, with each one setting it's own path here.

src/server/actions/s3.actions.ts
```diff
@@ -0,0 +1,33 @@
...
+            Bucket: s3BucketName(),
+            Expires: 3600,
+            Key: pathForStudyDocumentFile(studyInfo, documentType, filename),
+        })
+    },
+    z.object({
+        studyInfo: minimalStudyInfoSchema,
+        documentType: z.nativeEnum(StudyDocumentType),
+        filename: z.string(),
+    }),
```
nathanstitt, line 32: see my comment above about what `key` does when adding to the Form for uploading. If we can control the path, then we only need a single signed url generator and each file can set it's own path under the study

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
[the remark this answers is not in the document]
Comment (nathanstitt) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:125]: yes, I noted that above as well.  I think we should:
* retry uploads, probably only once?
* show toast and call action to delete the partially created study

nathanstitt (commented):
See what you think about using a single singed url returned by the study action and handling upload errors

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (chrisbendel) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:83]: right - this totally slipped my mind, good catch

src/app/researcher/study/request/[memberSlug]/study-proposal-form.tsx
```diff
@@ -45,80 +32,95 @@ export const StudyProposalForm: FC<{
...
+                            name="agreementDocument"
+                            aria-label="Upload Agreement Document"
+                            placeholder="Upload Agreement Document"
+                            clearable
+                            accept=".doc,.docx,.pdf"
+                            {...studyProposalForm.getInputProps('agreementDocument')}
+                        />
+                    </Group>
+                </Stack>
+            </Stack>
```
Comment (kswanson33) [inline src/app/researcher/study/request/[memberSlug]/study-proposal-form.tsx:123]: Yepyep these are similar changes to the ones on #150.

Similarities:
- Broad reshaping of form into `<Stack><Group><Text/><Input/></Group><Group>...</Group></Stack>`

Differences:
- Text & Input width handling (I do fixed minwidth for Text and per-element width for Inputs, you do 20% and 30% for Text & Input respectively)
- You added the SVG inside the `FileInput` components which I think is a good move
- Specific fixes for our cards

Where are the Figma designs you're trying to imitate? I'm just vibe coding myself lmao

Comment (chrisbendel) [reply src/app/researcher/study/request/[memberSlug]/study-proposal-form.tsx:123]: yeah figma files [are here](https://www.figma.com/design/icLzqqjan9qZuK7T1JnZyk/AE-Beluga-simple-2025?node-id=3151-5236&t=VUCadDafa9OKD3rk-4) but semi-useless cause they're out of date and we've moved the code files section around

im aslo pushing for just using the native fileinput and not styling it as a link no response from devika though, so im planning on just keeping it as the file upload. we can always modify styles later, but this keeps everything lined up properly
Screenshot 2025-04-18 at 12 22 03 PM

Comment (chrisbendel) [reply src/app/researcher/study/request/[memberSlug]/study-proposal-form.tsx:123]: im also not married to the width percentages, i dont want to have fixed widths because of responsiveness but maybe theres a better way... i originally had a grid layout which seemed to work okay, but 🤔

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -5,19 +5,130 @@ import { Button, Group, Stack, Text } from '@mantine/core'
...
+                'irbDocument',
+                'codeFiles',
+            ])
+
+            const valuesWithFilenames = {
+                ...valuesWithoutFiles,
+                descriptionDocPath: formValues.descriptionDocument?.name || '',
+                agreementDocPath: formValues.agreementDocument?.name || '',
+                irbDocPath: formValues.irbDocument?.name || '',
+            }
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:91]: since the files are mandatory in the zod scheme, why would we not have filenames?  I think we can remove all the. `?.name || ''` bits, right?

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -5,19 +5,130 @@ import { Button, Group, Stack, Text } from '@mantine/core'
...
+
+            const { studyId, studyJobId } = await onCreateStudyAction({
+                memberSlug,
+                studyInfo: valuesWithFilenames,
+            })
+
+            if (formValues.irbDocument) {
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:98]: same as above, we'll always have a `irbDocument`, can remove the`if` here and below

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -5,19 +5,130 @@ import { Button, Group, Stack, Text } from '@mantine/core'
...
+            const studyCodeUpload = await signedUrlForCodeUploadAction({
+                memberSlug,
+                studyId,
+                studyJobId,
+            })
+
+            const ok = await uploadCodeFiles(formValues.codeFiles, studyCodeUpload, studyJobId)
+            if (!ok) return { studyId, studyJobId }
+
+            return { studyId, studyJobId }
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:131]: is it possible to shorten all this, having this much code in the body of a component gives me shakes.  At first I was going to suggest exacting it but instead I think we can make a `uploadFiles` function that accepts an array of Files & paths and cut all this down to:
```suggestion
        uploadFiles(signedUrlFromStudySave, [
            { path: pathForStudyDocuments({ studyId, memberSlug }, StudyDocumentType.IRB),  file: formValues.irbDocument },
            {  const path = pathForStudyDocuments({ studyId, memberSlug }, StudyDocumentType.AGREEMENT), file:  formValues.agreementDocument }
            ...etc
])            
```

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -5,19 +5,130 @@ import { Button, Group, Stack, Text } from '@mantine/core'
...
+            }
+
+            const { studyId, studyJobId } = await onCreateStudyAction({
+                memberSlug,
+                studyInfo: valuesWithFilenames,
+            })
+
+            if (formValues.irbDocument) {
+                const path = pathForStudyDocuments({ studyId, memberSlug }, StudyDocumentType.IRB)
+                const upload = await signedUrlForStudyFileUploadAction(path)
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:100]: I thought we were going to return the url in the study save action and re-use it?  We shouldn't call multiple actions one after another because they are all fetch calls over the network.

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -28,23 +139,13 @@ export const StudyProposal: React.FC<{ memberId: string }> = ({ memberId }) => {
...
-            piName: '',
-            irbDocument: null,
-            descriptionDocument: null,
-            agreementDocument: null,
-            codeFiles: [],
+            if (!context) return
+            const deletePath = pathForStudy({ memberSlug: memberSlug, studyId: context.studyId })
+            const deleteStudyFilesURL = await signedUrlForDeletingStudyFilesAction(deletePath)
+            await fetch(deleteStudyFilesURL)
+            await onDeleteStudyAction({ studyId: context.studyId, studyJobId: context.studyJobId })
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:148]: same as above, we ~can't~ shouldn't call actions back-to-back.    Is the delete path different from the upload path though? Seems like it should not be?  If it does need to be, then have the study save action return that as well.

src/lib/paths.ts
```diff
@@ -10,7 +10,10 @@ export const pathForStudyJobCode = (parts: MinimalJobInfo) => `${pathForStudyJob
 export const pathForStudyJobCodeFile = (parts: MinimalJobInfo, fileName: string) =>
     `${pathForStudyJobCode(parts)}/${sanitizeFileName(fileName)}`
 
-export const pathForStudyDocuments = (parts: MinimalStudyInfo, docType: StudyDocumentType, fileName: string) =>
+export const pathForStudyDocuments = (parts: MinimalStudyInfo, docType: StudyDocumentType) =>
+    `${pathForStudy(parts)}/docs/${docType}`
+
+export const pathForStudyDocumentFile = (parts: MinimalStudyInfo, docType: StudyDocumentType, fileName: string) =>
     `${pathForStudy(parts)}/docs/${docType}/${sanitizeFileName(fileName)}`
```
nathanstitt, line 17: the `pathForStudyDocumentFile` can call `pathForStudyDocuments` and append the `${sanitizeFileName(fileName)` to it's return value

src/server/actions/s3.actions.ts
```diff
@@ -0,0 +1,35 @@
...
+        Conditions: [['starts-with', '$key', path]],
+        Key: path + '/${filename}', // single quotes are intentional, S3 will replace ${filename} with the filename
+    })
+}, z.string())
+
+export const signedUrlForDeletingStudyFilesAction = userAction(async (Key: string) => {
+    return await getSignedUrl(getS3Client(), new DeleteObjectCommand({ Bucket: s3BucketName(), Key }), {
+        expiresIn: 3600,
+    })
+}, z.string())
```
nathanstitt, line 35: I don't think these should be actions,  they can just be functions that the the study save will call to generate all these and send to the client after the study's saved.  

I also see why we need a different DELETE url now  so disregard that comment above

src/server/actions/study.actions.ts
```diff
@@ -189,14 +189,15 @@ export const approveStudyProposalAction = memberAction(async (studyId: string) =
 
         let status: StudyJobStatus = 'CODE-APPROVED'
 
-        if (USING_S3_STORAGE) {
+        // if we're not using s3 then containers will never build so just mark it ready
```
nathanstitt, line 192: not really s3 related, this could be:
```suggestion
        // if we're not connected to AWS codebuild,  containers will never build so just mark it ready
```

src/server/storage.ts
```diff
@@ -70,24 +48,13 @@ export async function fetchCodeManifest(info: MinimalJobInfo) {
...
 
 export async function storeStudyResultsFile(info: MinimalJobResultsInfo, file: File) {
     return await storeFile(pathForStudyJobResults(info), info, file)
 }
 
-export async function storeStudyDocumentFile(info: MinimalStudyInfo, fileType: StudyDocumentType, file: File) {
-    await storeFile(pathForStudyDocuments(info, fileType, file.name), info, file)
+export async function fetchStudyResultsFile(info: MinimalJobResultsInfo) {
+    return await fetchFile(pathForStudyJobResults(info))
 }
```
nathanstitt, line 60: this can be left for later, but the all storage is now s3, so these methods could be moved into s3.ts, OR s3.ts into here

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (chrisbendel) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:91]: Maybe there is some cleaner way to approach this with zod, but the reason for this is:
- The zod schema is of type `File | null` because we can't seed the initial form values with an empty file for example. omitting it is effectively saying undefined too, so the reason we need this `?.` is because of the typing. i dont think its a huge deal but maybe theres a clever way around it simply using the zod schema, i spent a little time trying to figure that out but didn't come up with a solution

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (nathanstitt) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:131]: Ooh, a further thought:  we can have the save study action generate the exact url for each file and return something like:
```
{
 irb:  { uploadUrl: <signed url>, deleteUrl: <signed url> }
 agreement:  { uploadUrl: <signed url>, deleteUrl: <signed url>  }
 codeFiles: [{  uploadUrl: <signed url>, deleteUrl: <signed url> }]
}
```

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (chrisbendel) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:98]: see above comment about the schema ^

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (nathanstitt) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:131]: a FINAL thought: maybe we don't need delete urls returned by study save.  if uploads fail, we'll need to call a 2nd `deleteStudyAction` and that could return the delete urls

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (chrisbendel) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:148]: yeah disregarding per your later comment, but the fetch needs to happen on the client side so we would need both calls here (just confirming if that sounds right)

src/server/storage.ts
chrisbendel: yea lets clean this stuff up later to reduce scope in this

chrisbendel: I added a TODO

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (nathanstitt) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:91]: oh, I see, that makes sense.  I wonder if we can use an empty file to make the initial values happy, I might test that out.

Comment (chrisbendel) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:91]: yea ill try that out now and see - it might work

src/app/researcher/study/request/[memberSlug]/study-proposal-form.tsx
Comment (kswanson33) [reply src/app/researcher/study/request/[memberSlug]/study-proposal-form.tsx:123]: I had the same problem with `FileInput` for accessibility; got permission from Iris to use the native component https://openstax.atlassian.net/browse/OTTER-149?focusedCommentId=16567

src/server/actions/s3.actions.ts
nathanstitt: I don't think these are used any longer right?

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -28,23 +111,17 @@ export const StudyProposal: React.FC<{ memberId: string }> = ({ memberId }) => {
...
-            descriptionDocument: null,
-            agreementDocument: null,
-            codeFiles: [],
+            const deleteStudyFilesURL = await onDeleteStudyAction({
+                memberSlug: memberSlug,
+                studyId: context.studyId,
+                studyJobId: context.studyJobId,
+            })
+
+            await fetch(deleteStudyFilesURL)
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:124]: have you tested the fetch(deleteStudyFilesURL) removes all the study files?  as I recall, S3 requires looping over each file in a path and won't delete everything  in one call like this.  

It also occurs to me that we could fire the deletes from the server inside the `onDeleteStudyAction` but that doesn't matter too much

src/server/actions/s3.actions.ts
chrisbendel: ah yes! great catch i forgot to delete this file

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (chrisbendel) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:124]: oh right - because they're not sending over a giant payload ... good idea!

src/server/aws.ts
```diff
@@ -111,11 +116,54 @@ export const storeS3File = async (
...
+    }
+
+    const objectsToDelete = listedObjects.Contents.map(({ Key }) => ({ Key }))
+
+    const deleteCommand = new DeleteObjectsCommand({
+        Bucket: s3BucketName(),
+        Delete: { Objects: objectsToDelete },
+    })
+
+    await getS3Client().send(deleteCommand)
```
nathanstitt, line 164: I think we should add a limit to the amount of files that can be deleted, just to ensure that the code doesn't mistakenly pass `/studies` into this and delete EVERYTHING.  something like this maybe?
```suggestion
    if (objectsToDelete.length > 10) { throw new Error('cowardly refusing to delete more than 10 files at once') }
    await getS3Client().send(deleteCommand)
```

nathanstitt (approved):
looks great @chrisbendel thanks for sticking with this through all the suggestions!  A few final comments and  then I'm done and you can merge whenever you want 😁

PR #132, opened 2025-04-14
Fix a few issues encountered when updating dev

nathanstitt:
Mainly clerk needs the publishable key provided on the react context provider


src/components/layout/app-shell.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+export function AppShell({ children }: Props) {
+    const theme = useMantineTheme()
+    return (
+        <MantineAppShell footer={{ height: 60 }} navbar={{ width: 250, breakpoint: 'xs' }} padding="md">
+            <RequireUser />
+            <RequireMFA />
+
+            <Notifications position="top-right" />
+            <AppShellNavbar bg={theme.colors.purple[8]}>
```
chrisbendel, line 35: you can just do `'purple.8'` here right? no need for useMantineTheme i think

chrisbendel (approved):

chrisbendel (approved):

PR #136, opened 2025-04-14
return jobs that have the LATEST status RUNNING/READY

nathanstitt:
The old query returned jobs that had the RUNNING/READY but also might have other statuses, such as COMPLETE.  Once it has a later status the job shouldn't appear in the list


sparksam (approved):
Looks good to me.

PR #139, opened 2025-04-14
style OrganizationSwitcher (white text)

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-154


therealmarv:
it works, I know it's not the most beautiful way https://clerk.com/docs/customization/overview#use-custom-css-classes-to-style-clerk-components but when we have our own organisation switcher this "ugly" workaround will be gone and no global CSS needs to be cleaned along with it.

nathanstitt:
looks good, thanks @therealmarv !

PR #143, opened 2025-04-15
Show approved/rejected regardless of results

chrisbendel:


nathanstitt (approved):

package.json
```diff
@@ -78,7 +78,7 @@
         "react-icons": "^5.5.0",
         "remeda": "^2.21.0",
         "server-only": "^0.0.1",
-        "si-encryption": "github:safeinsights/encryption",
+        "si-encryption": "github:safeinsights/encryption#8b6467a921278408e1f9e4159446e766f1d55e2e",
```
chrisbendel, line 81: Do you want to update this after merging the other PR?

nathanstitt: yes I was planning  to remove the locked sha, but now I think it makes sense to leave it in.  

My reasoning for that is that if the encryption lib ever introduces an incompatible change we would still want the ability to checkout this sha in the management app, run `nom install` and everything still work.  If we allow this version to float to a hypothetical incompatible is-encryption, then that would break.

PR #144, opened 2025-04-15
Otter 108 minio integration

chrisbendel:


src/server/actions/study.actions.ts
```diff
@@ -189,14 +189,16 @@ export const approveStudyProposalAction = memberAction(async (studyId: string) =
 
         let status: StudyJobStatus = 'CODE-APPROVED'
 
-        if (USING_S3_STORAGE) {
+        // TODO Is code build part of s3? do we wanna skip this on local/ci?
+        //  maybe we want to use SIMULATE_IMAGE_BUILD ?? its unused right now
+        if (DEV_ENV) {
+            status = 'JOB-READY' // if we're not using s3 then containers will never build so just mark it ready
+        } else {
```
nathanstitt, line 196: code build is not really part of s3, it's a codebuild job that reads the code files from the bucket.  I do think we can use SIMULATE_IMAGE_BUILD vs DEV_ENV as you note

src/server/storage.ts
```diff
@@ -70,24 +53,13 @@ export async function fetchCodeManifest(info: MinimalJobInfo) {
 }
```
nathanstitt, line 53: I think this can be just `urlForFile` or something like that, we don't need the `content` part any longer, that was just for when files were stored in `tmp` in local dev

src/server/aws.ts
```diff
@@ -20,20 +19,24 @@ export function objectToAWSTags(tags: Record<string, string>) {
...
+
+// For Presigned URLs and client calls
+let _s3BrowserClient: S3Client | null = null
+export const getS3BrowserClient = () =>
+    _s3BrowserClient ||
+    (_s3BrowserClient = new S3Client({
+        region: process.env.AWS_REGION || 'us-east-1',
+        forcePathStyle: true,
+        endpoint: process.env.S3_BROWSER_ENDPOINT || process.env.S3_ENDPOINT,
         credentials: process.env.AWS_PROFILE ? fromIni({ profile: process.env.AWS_PROFILE }) : undefined,
```
nathanstitt, line 40: I don't fully understand the ramifications of using 2 s3Clients or what the `forcePathStyle: true,` does, it seem harmless though so as long as it works I'm ok with it

nathanstitt (approved):
Seems great, I'd say lets deploy it and see how it does

src/server/aws.ts
chrisbendel: yeah the main motivation for doing this was something I believe we saw when testing on a call together, when we tried to change the URL. It will reject it if the URL is different than what it's originally signed with, and the URL comes from the client so it made sense to me when thinking about it that way 🤣

PR #148, opened 2025-04-17
fix tab select for mfa option buttons

kswanson33:
image

### DIFF
--- a/src/app/account/mfa/panel.tsx
+++ b/src/app/account/mfa/panel.tsx
@@ -3,7 +3,7 @@
 import React from 'react'
 import { useUser } from '@clerk/nextjs'
 import { ButtonLink, Link } from '@/components/links'
-import { Container, Flex, Stack, Text } from '@mantine/core'
+import { Container, Group, Stack, Text } from '@mantine/core'
 import { Panel } from '@/components/panel'
 import { notifications } from '@mantine/notifications'
 import { redirect } from 'next/navigation'
@@ -46,14 +46,14 @@ export function ManageMFAPanel() {
                     <Text size="md" mb="md">
                         Feel free to opt in to use either SMS verification OR Authenticator App verification.
                     </Text>
-                    <Flex gap="md">
+                    <Group gap="md">
                         <ButtonLink href="/account/mfa/sms" m={1}>
                             SMS Verification
                         </ButtonLink>
                         <ButtonLink href="/account/mfa/app" m={1}>
                             Authenticator App Verification
                         </ButtonLink>
-                    </Flex>
+                    </Group>
                 </Stack>
             </Panel>
         </Container>


src/app/account/mfa/panel.tsx
```diff
@@ -46,12 +46,10 @@ export function ManageMFAPanel() {
...
                         Feel free to opt in to use either SMS verification OR Authenticator App verification.
                     </Text>
-                    <Link href="/account/mfa/sms">
-                        <Button>SMS Verification</Button>
-                    </Link>
-                    <Link href="/account/mfa/app">
-                        <Button>Authenticator App Verification</Button>
-                    </Link>
+                    <Flex gap="md">
+                        <ButtonLink href="/account/mfa/sms" styles={{ root: { margin: 1 } }}>SMS Verification</ButtonLink>
```
chrisbendel, line 50: ```styles={{ root: { margin: 1 } }}```
this scares me 😨

nathanstitt: yes, I see that ButtonLink supports all the props that Button takes, so pretty sure you can use mantine style props: https://mantine.dev/styles/style-props/ on it.  try something like this:
```suggestion
                        <ButtonLink href="/account/mfa/sms" m="xs">SMS Verification</ButtonLink>
```

that won't quite be a 1px, but pretty close.  we much prefer to use those over raw styles.

nathanstitt: although, why are we setting the margin on the buttons at all? shouldn't the gap on the flex that you added handle that? I'm guessing I'm mis-understanding what this is intended to do, sorry.

kswanson33: sorry :( what do you normally do?

kswanson33: the `:focus` border is touching the button if the margin isn't set. adding a margin makes it more visible

kswanson33: the mantine props worked out! thanks nathan and chris

src/app/account/mfa/panel.tsx
```diff
@@ -46,12 +46,14 @@ export function ManageMFAPanel() {
...
-                        <Button>Authenticator App Verification</Button>
-                    </Link>
+                    <Flex gap="md">
+                        <ButtonLink href="/account/mfa/sms" m={1}>
+                            SMS Verification
+                        </ButtonLink>
+                        <ButtonLink href="/account/mfa/app" m={1}>
+                            Authenticator App Verification
+                        </ButtonLink>
+                    </Flex>
```
chrisbendel, line 56: Just got around to testing this out locally, i think the margin is a red herring and you don't need it. Try this out (it seems to be tabbing fine for me now)

```
                    <Group>
                        <Button component={Link} href="/account/mfa/sms">
                            SMS Verification
                        </Button>
                        <Button component={Link} href="/account/mfa/app">
                            Authenticator App Verification
                        </Button>
                    </Group>
```

chrisbendel (changes requested):
Try out the suggestion i made using <Group> and no margins 🤔

src/app/account/mfa/panel.tsx
kswanson33: Looking at the [definition](https://github.com/safeinsights/management-app/blob/main/src/components/links.tsx#L31) of `ButtonLink`, I believe `<Button component={Link}>` is about the same thing. Is there a reason to prefer the import vs. defining inline? I can't see a strong reason one way or another, other than "if the thing is exported, might as well use it" 😆 

When I try your way I'm still seeing that the tab-border around the element is close enough to the element that it's difficult to see.

image


Here it is with `m={1}`. It's a little more visible.

image

The `m={1}` thing is minor, though. If you'd rather avoid custom styles, I get ya.

nathanstitt: thinking through this a bit more, I'm assuming we're going to want to add a focus to ALL buttons eventually.  @chrisbendel do you know if we can adjust the them to do the m={1} thing globally?

kswanson33: so yeah tl;dr it's ambiguous to me what to do here. I've changed the `Flex` to a `Group` bc I agree that's better. Otherwise it's all about the same to me. I like the margins but I can let them go 🥲

nathanstitt: heh, I actually prefer using `Flex` everywhere because it's easier for me to map to `flexbox` css styles and I don't have to recall whether Stack or Group does what.  it doesn't really matter tho

chrisbendel: and @kswanson33 maybe im misunderstanding but the ticket ([otter-151](https://openstax.atlassian.net/browse/OTTER-151)) is about double focusing, and not whether the blue border is more or less visible, so i dont think we need to worry about the margin thing

kswanson33: - I like keeping `Group` instead of `Flex` too, I think it's clearer
- For consistency, I'd rather remove `ButtonLink` everywhere it's used, or use it here. No preference on which.

Re: margins, you're right & it's not on the card, I just happened to notice it. I should've mentioned, I did ask Iris about it and she says to leave it (I documented the convo on the card): https://openstax.atlassian.net/browse/OTTER-151?focusedCommentId=16569

chrisbendel (approved):
i suppose its fine to add a margin of 1 here, i dont think its a good idea in general to add a global margin to all the buttons (unless we REALLY think this is important and have some UX backed reason to do so) and having it here would just make it inconsistent with any of the other buttons and begs the larger question of is this really accessible/inaccessible or is it just someone's personal preference

anyway lets get this merged so we can not think about this again for a while 🤣 because the original intent of the ticket was just the tab order anyway

PR #156, opened 2025-04-22
Fixed under review styling

sparksam:


src/components/study/display-study-status.tsx
```diff
@@ -57,7 +57,15 @@ const StatusBlock: React.FC<StatusLabels & { jobId?: string | null }> = ({ type,
                     <InfoComponent jobId={jobId} />
                 </Flex>
             ) : (
-                <Text>{label}</Text>
+                <Text
```
chrisbendel, line 60: This looks fine, the only suggestion I have is this component is getting a bit complicated and "fancy" with the logic. ie returning a ternary and then having the "else" condition on the ternary also doing some conditional logic on the `c` prop. I think this is fine for now, if we end up having to make any more changes or add any more conditions i'm just making a mental note that we should simplify the logic so that this component is more readable down the line

src/components/study/display-study-status.tsx
```diff
@@ -45,6 +45,12 @@ const StatusLabels: Partial<Record<AllStatus, StatusLabels>> = {
...
 }
 
+const styleStatus = (label: string) => {
+    if ([StatusLabels['RUN-COMPLETE']?.label, StatusLabels['PENDING-REVIEW']?.label].indexOf(label) > -1)
+        return '#C70000'
+    else return '#000000'
+}
+
 const StatusBlock: React.FC<StatusLabels & { jobId?: string | null }> = ({ type, label, jobId, InfoComponent }) => {
     return (
```
chrisbendel, line 55: Maybe instead of the function just declare it as a variable in the method 🤔  like

```
const color = ([StatusLabels['RUN-COMPLETE']?.label, StatusLabels['PENDING-REVIEW']?.label].indexOf(label) > -1) ? '#C70000' : '#000000'
```

i think this is splitting hairs at this point, but we prob dont need the function 🤔

nathanstitt: Thanks for tackling this @sparksam!  I don't think we should using random hex values in code because they have no information attached, who really knows what color `#C70000` is 🤷     if we REALLY need to use those exact RGB values vs something like `red.8`, then lets move them to `theme.ts` and refer to them by a name here, like maybe `runs.complete` & `runs.pending`.   I'll comment on the card for this as well

nathanstitt: fyi the mantine built-in colors chart is here: https://mantine.dev/theming/colors/#default-colors

chrisbendel: oh yeah @sparksam nathan brings up some great points, i totally missed the hex value thing. we can use our own theme colors [we've defined here](https://github.com/safeinsights/management-app/blob/main/src/theme.ts) and like nathan said we can just do `c='colors.charcoal.8` or whatever the designs dictate

src/components/study/display-study-status.tsx
```diff
@@ -45,10 +47,16 @@ const StatusLabels: Partial<Record<AllStatus, StatusLabels>> = {
     'PENDING-REVIEW': { type: 'Proposal', label: 'Under Review' },
 }
 
+const styleStatus = (label: string) => {
+    if ([StatusLabels['RUN-COMPLETE']?.label, StatusLabels['PENDING-REVIEW']?.label].indexOf(label) > -1)
+        return theme.colors.red[10]
+    else return 'dark.8'
+}
+
```
nathanstitt, line 55: a final comment, you don't need to return the actual `them.colors.red[10]`, you can return a string `red.10` and then you wouldn't need to import the theme object.

PR #158, opened 2025-04-22
let already selected items stay active

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-133


src/components/layout/navbar-items.tsx
```diff
@@ -19,35 +20,35 @@ export const NavbarItems: FC = () => {
...
 
-            <Button
-                fullWidth
-                className={styles.hover}
-                justify="flex-start"
-                variant="transparent"
+            <NavLink
+                label="Settings"
+                leftSection={<Gear />}
                 onClick={() => openUserProfile()}
```
chrisbendel, line 45: we will want to add the active prop here too just for styling (it seems everything is white on hover without it)
```active={pathname === currentDashboardUrl}```

File: src/components/layout/navbar-items.tsx [hunk 2]
```diff
@@ -61,17 +62,13 @@ export const NavbarItems: FC = () => {
...
                 />
             </Group>
-            <Button
-                fullWidth
-                className={styles.hover}
-                justify="flex-start"
-                variant="transparent"
+            <NavLink
+                label="Logout"
+                leftSection={<SignOut />}
```
chrisbendel, line 67: we will want to add the active prop here too just for styling (it seems everything is white on hover without it)
```active={pathname === currentDashboardUrl}```

File: src/components/layout/navbar-items.tsx (hunk 1 above)
Comment (chrisbendel) [reply src/components/layout/navbar-items.tsx:44]: yeah good point - it is quite weird 🤔 css should be good 👍 thanks!!

File: src/components/layout/navbar-items.tsx [hunk 3]
```diff
@@ -19,34 +21,31 @@ export const NavbarItems: FC = () => {
         return '/'
     }
 
+    const currentDashboardUrl = dashboardURL()
```
nathanstitt, line 24: I know you didn't create this (it might even have been me?), but why are we creating a function and then calling it immediately?  also what does `current` mean in the variable name prefix? I don't see it changing, so maybe it's just `dashboardUrl`?

Instead it seems like we can just inline the code:
```suggestion
    let dashboardUrl = '/'
    if (isMember) dashboardUrl = '/member/openstax/dashboard'
    if (isResearcher) dashboardUrl = '/researcher/dashboard'
    if (isAdmin) dashboardUrl = '/admin/dashboard'
```

therealmarv: good catch and I totally agree.

chrisbendel (approved):

PR #167, opened 2025-04-28
run minio locally

nathanstitt:
This adds instructions and a `./bin/local-minio` script to run minio locally w/out Docker

It also changes the port and usernames to non-default values to provide a _tiny_ bit of protection from network scanners.

Additionally it fixes a bug noted during testing where the study submit would happily proceed to the dashboard even if all the files failed to upload.
 
Pinging @stetradis to review this along with @chrisbendel since she'd expressed interest in running app locally.  Please don't bother reviewing until specs are passing since they'll probably be more tweaks needed


src/database/types.ts
```diff
@@ -3,7 +3,7 @@
  * Please do not edit it manually.
  */
 
-import type { ColumnType } from "kysely";
+import type { ColumnType } from "kysely"
```
chrisbendel, line 6: only thing with this and removing the reference in `eslint.config.mjs` is that next time the types are re-generated (ie after some migration) it will put the `;` back in automatically. ideally there's a way to use the kysely ctl library to control that when the migrations run and the `types` file is generated, however i couldn't find anything. if we do want to be manually removing the `;` from the file each time a migration is ran, we will probably want to remove the entry in `prettier.ignore` as well cause that is currently ignoring this file.

overall technically we shouldn't be "editing" this file (even though its not an edit of any substance), but just throwing this out there as to why it was originally included in the eslint/prettier ignores

chrisbendel (approved):
Looks good, out of curiosity why the port change?

src/database/types.ts
nathanstitt: Yes, it does put the `;` back (as well as a bunch of other changes),  but we then re-run eslint/prettier to auto-fix the files right afterwards in ./dev/migrate-dev-db

nathanstitt:
so that if someone scans your laptop they don't find an open port that minio is known to use as a default.  Hopefully everyone's running a  firewall that would block it regardless though

PR #171, opened 2025-04-29
Verify researcher is a member of this organization

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-117

### DIFF
--- a/src/app/researcher/study/request/[memberSlug]/actions.ts
+++ b/src/app/researcher/study/request/[memberSlug]/actions.ts
@@ -1,5 +1,6 @@
 'use server'
 
+import { AccessDeniedError } from '@/lib/errors'
 import { codeBuildRepositoryUrl, deleteFolderContents, signedUrlForStudyUpload } from '@/server/aws'
 import { studyProposalApiSchema } from './study-proposal-form-schema'
 import { db } from '@/database'
@@ -18,6 +19,18 @@ export const onCreateStudyAction = researcherAction(async ({ memberSlug, studyIn
     const userId = await getUserIdFromActionContext()
 
     const member = await getMemberFromSlugAction(memberSlug)
+
+    // Verify researcher is a member of this organization
+    const memberUser = await db
+        .selectFrom('memberUser')
+        .select('id')
+        .where('userId', '=', userId)
+        .where('memberId', '=', member.id)
+        .executeTakeFirst()
+
+    if (!memberUser) {
+        throw new AccessDeniedError('You are not a member of this organization')
+    }
 
     const studyId = uuidv7()
 


nathanstitt:
@therealmarv I'm a bit confused on why we're checking the database for the membership, when clerk gives us as part of the session, right?  can we check the memberSlug from the params against the clerk current organization?

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -101,20 +101,56 @@ export const StudyProposal: React.FC<{ memberSlug: string }> = ({ memberSlug })
...
+
+            let title = 'Submission Failed' // Default title
+            let message = 'An unexpected error occurred.' // Default message
+            let needsCleanup = false
+            let rawErrorMessage = error instanceof Error ? error.message : 'Unknown error details'; // Capture raw message
+
+            // Check for specific access denied error
+            if (error instanceof Error && error.message.includes('You are not a member of this organization')) {
+                title = 'Access Denied'
+                message = 'You can only submit study proposals to organizations you are a member of.'
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:115]: isn't this case a non-issue? the `page.tsx` file already ensures the user is a member of the org 🤔

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -101,20 +101,56 @@ export const StudyProposal: React.FC<{ memberSlug: string }> = ({ memberSlug })
...
+                needsCleanup = false
+            } else if (context) {
+                // Use the requested title for upload/processing errors
+                title = 'Failed to upload file'
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:119]: the `uploadFile` method already throws an error, we can probably just display those errors with a notification.show on line 30:
```
    if (!response.ok) {
        throw new Error(`failed to upload file ${await response.text()}`)
    }
```

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
```diff
@@ -101,20 +101,56 @@ export const StudyProposal: React.FC<{ memberSlug: string }> = ({ memberSlug })
...
             })
-            if (!context) return
 
-            await onDeleteStudyAction({
-                memberSlug: memberSlug,
-                studyId: context.studyId,
-                studyJobId: context.studyJobId,
-            })
+            // Keep the cleanup logic and its error handling from 'mb'
+            if (needsCleanup && context) {
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:137]: `needsCleanup` might be irrelevant, we always need to clean up here if something goes wrong with the study creation, because this is the onError handler aka something went wrong creating the study 🤔 

I don't believe we'd want orphan studies lingering around if something went wrong with creating them 🤔

chrisbendel (changes requested):

src/app/researcher/study/request/[memberSlug]/study-proposal.tsx
Comment (therealmarv) [reply src/app/researcher/study/request/[memberSlug]/study-proposal.tsx:137]: Thanks, I will refactor this whole part and testing everything right now. I maybe have been over ambitious here.

therealmarv:
Just a sample screenshot if changing URL manually to openstax (although researcher is from a different org): 
image

therealmarv:
I struggled with generate any sort of good looking and working unit test code with this multiple Clerk orgs<->DB<->UI interaction. It's harder than the solution itself.

As a note (but maybe a weak excuse): There are also no existing unit test code for the files I've touched in this PR:
* [action.ts](https://github.com/safeinsights/management-app/blob/main/src/app/researcher/study/request/%5BmemberSlug%5D/actions.ts)
* [page.tsx](https://github.com/safeinsights/management-app/blob/main/src/app/researcher/study/request/%5BmemberSlug%5D/page.tsx)

On the positive news: **The code works.**

Test case which I will also add to the card:
* Create a new organization in `admin/organization`
* Create a new researcher for this new organization in `admin/invite`
* When logged in as this new researcher the URL `http://localhost:4000/researcher/study/request/openstax` should not be reachable with "access denied" (it was reachable and fully functional before)

src/app/researcher/study/request/[memberSlug]/actions.ts
```diff
@@ -20,6 +21,18 @@ export const onCreateStudyAction = researcherAction(async ({ memberSlug, studyIn
 
     const member = await getMemberFromSlugAction(memberSlug)
 
+    // Verify researcher is allowed to submit studies to this organization
+    const memberUser = await db
+        .selectFrom('memberUser')
+        .select('id')
+        .where('userId', '=', userId)
+        .where('memberId', '=', member.id)
+        .executeTakeFirst()
```
Comment (chrisbendel) [inline src/app/researcher/study/request/[memberSlug]/actions.ts:30]: nitpicking here but we can extract this query out (as an identical query is in the `page.tsx` file) into `queries.ts` but not a showstopper

chrisbendel (approved):
This looks good! @nathanstitt i think in general im curious about how UX wants to approach these instances where you try to access a page but the user lacks permissions. Do we want to show the error boundary or do we want to redirect them to a homepage with a notification. something to bring up next sprint maybe

chrisbendel:
@therealmarv there are a couple merge conflicts from nathan's org renaming PR FYI

nathanstitt:
hey Marvin #175 modified a new `researcherAction` to optionally accept a orgSlug property, which if present will require the user to be a researcher in that organization. I think it's usage will simplify your code a lot

src/app/researcher/study/request/[memberSlug]/actions.ts
```diff
@@ -20,6 +21,18 @@ export const onCreateStudyAction = researcherAction(async ({ memberSlug, studyIn
 
     const member = await getMemberFromSlugAction(memberSlug)
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[memberSlug]/actions.ts:22]: I don't see where this is being called, but if you rename `memberSlug` to be `orgSlug` and call it like:
`onCreateStudyAction({ orgSlug, studyInfo })` 

`researcherAction` will now will automatically check that the user is a member of that organization

therealmarv:
The `researcherAction` wrapper keeps sure that only researchers with access to organisation can propose studies in the organisation. So my additional checks in actions.ts are not needed anymore since the Thursday refactoring.

To fulfill OTTER-117 card this PR only needs visual feedback as in screenshot https://github.com/safeinsights/management-app/pull/171#issuecomment-2843419634

So this PR is purely visual for user feedback to fulfill OTTER-117 and `researcherAction` wrapper keeps behind the scenes sure that no studies are posted for orgs where the researcher has no access.

jbwilson8 (approved):

PR #172, opened 2025-04-30
Addressed first half of OTTO-171.

Rod0352:
Added new entry point at bottom of the side panel,
 Created user profile avatar component along with a test.
 Added new style for onHover with blue.8 bg


src/components/user-avatar.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+}))
+
+describe('UserAvatar', () => {
+    it('displays the user profile image when logged in', async () => {
+        vi.mocked(useUser).mockReturnValue({
+            user: {
+                profileImageUrl: 'https://example.com/profile.jpg',
+            },
+        } as UseUserReturn)
+
```
nathanstitt, line 21: we have mocks for clerk already setup for all tests in `tests/vitest.setup.ts`, and also have a lot of ways to mock the current user, see `tests/unit.helpers.tsx`, but I do not see where we're returning a profileImageUrl from the `mockClerkSession`.   Adding that seems like a natural addition though, can you add it there and then use it in this test?

nathanstitt (commented):
Looking good Rodney, thank you.  made a small suggestion to add to our existing mocks vs adding a new one.

src/components/layout/navbar-items.tsx
```diff
@@ -66,6 +68,19 @@ export const NavbarItems: FC = () => {
...
                 className={styles.navLinkHover}
             />
+            <NavLink
+                label={
+                    <>
+                        Hi! <UserName />
+                    </>
+                }
+                leftSection={UserAvatar()}
+                rightSection={<CaretUp />}
```
chrisbendel, line 78: Interesting the designs have a caret right, but the ticket specifies caret up 🤔 probably worth checking with UX to see what they want

src/components/layout/navbar-items.module.css
```diff
@@ -1,3 +1,7 @@
 .navLinkHover:hover {
     background-color: var(--mantine-color-blue-9);
 }
+
+.navLinkProfileHover:hover {
+    background-color: var(--mantine-color-blue-8);
```
chrisbendel, line 6: I wonder if all the hover colors should be the same color, again worth checking with UX to confirm, i dont imagine we want two separate colors on hover in the navbar 🤔

src/components/layout/navbar-items.tsx
```diff
@@ -66,6 +68,19 @@ export const NavbarItems: FC = () => {
...
+                c="white"
+                className={styles.navLinkProfileHover}
+                pos={'absolute'}
```
chrisbendel, line 81: Rather than forcing this with absolute positioning, take a look at [this example](https://mantine.dev/app-shell/?e=NavbarSection&s=code) they have a way to utilize flex so its a bit more responsive (in case we ever have a super long navbar or for mobile viewports) https://mantine.dev/core/app-shell/#examples

chrisbendel (changes requested):

src/components/layout/navbar-items.tsx
```diff
@@ -66,6 +68,19 @@ export const NavbarItems: FC = () => {
                 c="white"
                 className={styles.navLinkHover}
             />
+            <NavLink
+                label={
+                    <>
+                        Hi! <UserName />
+                    </>
+                }
+                leftSection={UserAvatar()}
```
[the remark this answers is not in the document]
Rod0352: definitely should be, will adjust this

src/components/layout/navbar-items.tsx
Rod0352: thanks for that tip, this makes more sense to me

src/components/layout/navbar-items.tsx
Rod0352: I did confirm it was supposed to be caret right, thanks

src/components/layout/navbar-profile-menu.tsx
```diff
@@ -0,0 +1,27 @@
+'use client'
+
+import { NavLink, AppShellSection } from '@mantine/core'
+import { UserAvatar } from '@/components/user-avatar'
+import { UserName } from '@/components/user-name'
+import { CaretRight } from '@phosphor-icons/react/dist/ssr'
+import { useUser } from '@clerk/nextjs'
+import styles from './navbar-items.module.css'
+
+export function NavbarProfileMenu() {
```
chrisbendel, line 10: do you think this component will get more complex/additional logic down the line?

Only thing i think about is if this is too DRY, main thoughts are:

1. There is no conditional logic or control flows in here
2. it only seems to be used in one place right now

chrisbendel (approved):

src/components/layout/navbar-profile-menu.tsx
Rod0352: I do anticipate some more logic being added with the addition of the collapsable navigation items - left those out as to not clutter this pr.

PR #174, opened 2025-05-01
rename Member to Org

nathanstitt:
renames all references to 'member' with 'org'


tests/configure-test-env.ts
```diff
@@ -17,27 +17,26 @@ async function setupUsers() {
...
     }
 
     for (const clerkId of CLERK_RESEARCHER_TEST_IDS) {
         const userId = await findOrCreateSiUserId(clerkId, {
             firstName: 'Test Researcher User',
             lastName: 'Test Researcher User',
-            isResearcher: true,
         })
-        findOrCreateOrgMembership({ userId, slug: 'openstax', isReviewer: false })
+        findOrCreateOrgOrgship({ userId, slug: 'openstax', isReviewer: false, isResearcher: true })
```
chrisbendel, line 35: this got a little wonky with the replacement

src/server/mutations.ts
```diff
@@ -33,38 +27,41 @@ export const findOrCreateSiUserId = async (clerkId: string, attrs: SiUserOptiona
     return user.id
 }
 
-export async function findOrCreateOrgMembership({
+export async function findOrCreateOrgOrgship({
```
chrisbendel, line 30: weird naming here

src/database/migrations/1746044928902_reviewer-to-member.ts
```diff
@@ -0,0 +1,18 @@
+import type { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema.alterTable('member').renameTo('org').execute()
```
chrisbendel, line 4: I maybe would prefer `organization` instead of just org 🤔

src/server/api-context.ts
```diff
@@ -14,19 +14,19 @@ export function actionContext() {
...
+    if (!org) throw new AccessDeniedError('No org in context')
 
-    return member
+    return org
 }
 
-export function requestingUser() {
-    const user = localStorageContext.getStore()?.user
-    if (!user) throw new AccessDeniedError('No user in context')
+// export function requestingUser() {
```
chrisbendel, line 24: can remove this?

chrisbendel (approved):
some small nitpicks and naming stuff which we can either deal with now, or later if we capture it in a ticket

chrisbendel:
@nathanstitt it looks like you also might need to update the github CI env variables to match the changes in this PR?

bin/debug/request.ts
```diff
@@ -10,8 +10,8 @@ export class DebugRequest {
 
     constructor(public path = '') {
         this.program
-            .option('-m, --member <member>', 'member slug')
-            .option('-o, --origin <origin>', 'base URL to send the request to')
```
rnathuji, line 14: Small nit: This change to origin need to be reflected where the option is read below.

nathanstitt: ah, yep good point. ty!  I'll address this in a follow up.

PR #176, opened 2025-05-01
add table for org admin to list users

nathanstitt:


stetradis:
Looks good to me! Discussed the duplicated role-mapping logic in Slack, that can be addressed later when we have a clear answer. Some suggestions for UX:

- Add pagination to the `DataTable` if we expect this list to be lengthy.
- Show a friendly empty state (e.g. “No users found”) when the table has no records, so it doesn’t look like a loading error.

stetradis (approved):

nathanstitt:
Good points @stetradis, but I do have empty message, that's the `noRecordsText` property.  as for pagination, I'm doubting that any orgs will have more than 10ish users, but datatable makes it pretty easy for us to add if we need to later

PR #180, opened 2025-05-05
Event auditing

sparksam:
Audit logs are currently recorded for Studies approval and rejection.

### DIFF
--- a/src/database/migrations/1746109057988_audit.ts
+++ b/src/database/migrations/1746109057988_audit.ts
@@ -1,6 +1,5 @@
 import { type Kysely, sql } from 'kysely'
 
-// biome-ignore lint/suspicious/noExplicitAny: `any` is required here since migrations should be frozen in time. alternatively, keep a "snapshot" db interface.
 export async function up(db: Kysely<unknown>): Promise<void> {
     await db.schema.createType('audit_event_type').asEnum(['APPROVED', 'REJECTED']).execute()
     await db.schema.createType('audit_record_type').asEnum(['STUDY']).execute()
@@ -15,7 +14,6 @@ export async function up(db: Kysely<unknown>): Promise<void> {
         .execute()
 }
 
-// biome-ignore lint/suspicious/noExplicitAny: `any` is required here since migrations should be frozen in time. alternatively, keep a "snapshot" db interface.
 export async function down(db: Kysely<unknown>): Promise<void> {
     await db.schema.dropTable('audit').execute()
 }
--- a/src/database/migrations/1746109057988_audit.ts
+++ b/src/database/migrations/1746109057988_audit.ts
@@ -1,0 +1,21 @@
+import { type Kysely, sql } from 'kysely'
+
+// biome-ignore lint/suspicious/noExplicitAny: `any` is required here since migrations should be frozen in time. alternatively, keep a "snapshot" db interface.
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema.createType('audit_event_type').asEnum(['APPROVED', 'REJECTED']).execute()
+    await db.schema.createType('audit_record_type').asEnum(['STUDY']).execute()
+    await db.schema
+        .createTable('audit')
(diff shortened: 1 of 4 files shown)


src/database/migrations/1746109057988_audit.ts
```diff
@@ -0,0 +1,21 @@
...
+    await db.schema.createType('audit_event_type').asEnum(['APPROVED', 'REJECTED']).execute()
+    await db.schema.createType('audit_record_type').asEnum(['STUDY']).execute()
+    await db.schema
+        .createTable('audit')
+        .addColumn('id', 'uuid', (col) => col.defaultTo(sql`v7uuid()`).primaryKey())
+        .addColumn('created_at', 'timestamp', (col) => col.defaultTo(sql`now()`).notNull())
+        .addColumn('event_type', sql`audit_event_type`, (col) => col.notNull())
+        .addColumn('record_id', 'text', (col) => col.notNull())
+        .addColumn('record_type', sql`audit_record_type`, (col) => col.notNull())
+        .addColumn('userId', 'text', (col) => col.notNull())
```
nathanstitt, line 14: the record_id and user_id should both be type of uuid

src/server/actions/study.actions.ts
```diff
@@ -255,6 +257,7 @@ export const rejectStudyProposalAction = orgAction(
                 .executeTakeFirstOrThrow()
         })
 
+        await audit('REJECTED', userId, 'STUDY', studyId)
```
nathanstitt, line 260: I would find it easier to follow if the `audit` table accepted an object:
```suggestion
        await audit({ eventType: 'REJECTED', userId,  recordType: 'STUDY', recordId: studyId })
```

src/lib/audit.ts
```diff
@@ -0,0 +1,21 @@
...
+    console.log(`Auditing ${eventType}: ${recordType}/${recordId}`)
+    await db
+        .insertInto('audit')
+        .values({
+            userId: userId,
+            eventType: eventType,
+            recordType: recordType,
+            recordId: recordId,
+        })
+        .execute()
```
nathanstitt, line 20: I'd suggest `audit` function accept an object that could be passed as-is into `values`

sparksam: Thanks for the suggestions @nathanstitt! I addressed the review comments.

src/lib/audit.ts
```diff
@@ -0,0 +1,23 @@
...
+
+type AuditEntry = {
+    eventType: AuditEventType
+    userId: string
+    recordType: AuditRecordType
+    recordId: string
+}
+
+export const audit = async (entry: AuditEntry): Promise<void> => {
+    console.log(`Auditing ${entry.eventType}: ${entry.recordType}/${entry.recordId}`)
```
chrisbendel, line 13: do we need console.log here?

src/database/migrations/1746109057988_audit.ts
```diff
@@ -0,0 +1,21 @@
+import { type Kysely, sql } from 'kysely'
+
+// biome-ignore lint/suspicious/noExplicitAny: `any` is required here since migrations should be frozen in time. alternatively, keep a "snapshot" db interface.
```
chrisbendel, line 3: is this needed or just some editor specific thing?

chrisbendel (approved):

src/database/migrations/1746109057988_audit.ts
nathanstitt: yes good catch @chrisbendel - kysely migrate adds these automatically.    I've been meaning to investigate `biome` since it does sounds like a neat tool vs our eslint+prettier

@sparksam can you remove the comments and replace the `any` with `unknown` please?

sparksam: I replaced the `any` with `unknown`, will remove the auto generated comment.

nathanstitt (approved):
looks great, thank you so much @sparksam

PR #182, opened 2025-05-06
Consolidate event handling and add some user related events

nathanstitt:
sets us up to track last user activity by selecting most recent event for a user from the event table


jbwilson8 (approved):
looks, good!

PR #183, opened 2025-05-06
Otter 168 - Part 1

stetradis:
- Creates new modal for the invite form
- Adds migration for pending users
- Updates user actions for inviting users/pending users
- Updates user invite tests


src/app/admin/invite/invite-form.tsx
```diff
@@ -1,112 +1,169 @@
...
-                visible
-                placeholder="Enter password"
-                mb="sm"
-                {...studyProposalForm.getInputProps('password')}
-            />
-            <Button type="submit" mt="md" fullWidth loading={isPending}>
-                Invite
+            {studyProposalForm.errors.email && (
+                <Flex align="center" gap={4} my={2}>
+                    <WarningCircle size={16} color="var(--mantine-color-error)" />
```
jbwilson8, line 109: We should use the following import `import { theme } from `'@/theme'`` and use `color={theme.colors.red[7]}` for the WarningCircle component. :-)

src/app/admin/invite/invite-form.tsx
```diff
@@ -1,112 +1,169 @@
...
+                        <Radio value="researcher" label="Researcher (can submit studies and access results)" />
+                    </Flex>
+                </Radio.Group>
+            </div>
+
+            <Button type="submit" mt="sm" loading={isInviting} disabled={!studyProposalForm.isValid()}>
+                Send invitation
             </Button>
+
+            <hr style={{ margin: '32px 0', color: '#D9D9D9' }} />
```
jbwilson8, line 134: can the Mantine` <Divider />` component be used here instead?  Then you could use `c="red.7"`for the color since this is a mantine component.

src/components/modal.tsx
```diff
@@ -0,0 +1,57 @@
...
+            title={title}
+            size={size}
+            centered={centered}
+            closeOnClickOutside={closeOnClickOutside}
+            trapFocus={trapFocus}
+            closeButtonProps={{ size: 'md' }}
+            styles={{
+                header: {
+                    padding: '0px 40px',
+                    backgroundColor: '#F1F3F5',
```
jbwilson8, line 41: I don't see this color `#F1F3F5` in the `theme.ts` file. it's probably a good idea to check with UX to see if there's an existing color that can be used in it's place, otherwise we could add it.

src/components/modal.tsx
```diff
@@ -0,0 +1,57 @@
...
+                },
+                body: {
+                    padding: '40px 80px',
+                },
+                close: {
+                    backgroundColor: '#8C8C8C',
```
jbwilson8, line 47: we can use `{theme.colors.charcoal[4]}` for the `backgroundColor` here

src/components/modal.tsx
```diff
@@ -0,0 +1,57 @@
...
+                    color: '#FFFFFF',
```
jbwilson8, line 48: It's probably ok to leave this one, I know it's hardcoded in other places.

jbwilson8 (changes requested):
Left a few style suggestions. Otherwise, things look great! :-)

src/app/admin/invite/admin-users.actions.ts
```diff
@@ -32,60 +40,77 @@ export const adminInviteUserAction = adminAction(async (invite) => {
...
+
+export const deletePendingUserAction = adminAction(async ({ id }: { id: string }) => {
+    try {
+        await db.deleteFrom('pendingUser').where('id', '=', id).execute()
+        return { success: true }
+    } catch (error) {
+        logger.error(error)
+        return { success: false }
+    }
+})
```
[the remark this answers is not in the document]
stetradis: This isn't used yet (will be part of PR part 2) to delete the user from the pending table after they sign up. I can remove it for now and include it in the next PR

src/app/admin/invite/invite-form.tsx
```diff
@@ -1,112 +1,169 @@
...
-            <Button type="submit" mt="md" fullWidth loading={isPending}>
-                Invite
+            {studyProposalForm.errors.email && (
+                <Flex align="center" gap={4} my={2}>
+                    <WarningCircle size={16} color="var(--mantine-color-error)" />
+                    <Text c="red" size="xs">
+                        {studyProposalForm.errors.email}
+                    </Text>
+                </Flex>
+            )}
```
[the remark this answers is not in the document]
stetradis: Not that I've seen, I noticed it in this design - https://www.figma.com/design/YxDg7G4oLIOevrA2JuxvS7/Team-management-April-2025?node-id=289-2446&t=V0sMKAMOx4cmxo0z-4

Although it now appears to have changed to a filled circle since I last checked

src/components/modal.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+interface AppModalProps extends Omit<ModalProps, 'opened'> {
+    isOpen: boolean
+    onClose: () => void
+    children: React.ReactNode
+    title: React.ReactNode
+}
+
+/**
+ * A wrapper around Mantine's Modal component to promote reusability and consistent styling.
```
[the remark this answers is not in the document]
stetradis: My thinking is that we should enforce our modal defaults (e.g. closeOnClickOutside: true, centered: true) in one place so we maintain consistent styling across the app instead of repeating overrides. I’ll check with Iris to see if all existing modals should share this style, if not I’m happy to pull it back out

src/app/admin/invite/invite-form.tsx
```diff
@@ -1,112 +1,169 @@
...
+                                <Button
+                                    variant="outline"
+                                    size="xs"
+                                    onClick={() => reInviteUser({ email: user.email })}
+                                    loading={isReinviting && reinvitingEmail === user.email}
+                                    disabled={isReinviting}
+                                >
+                                    Re-invite
+                                </Button>
+                            </Flex>
```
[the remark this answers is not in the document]
stetradis: Great point! This should definitely be extracted

src/app/admin/invite/invite-form.tsx
stetradis: Thanks for the suggestions on the Mantine theme! These points have been very helpful in getting up to speed on our styling approach

src/components/modal.tsx
stetradis: This one might be better to add to the theme as it appears to be hardcoded in several places

src/database/migrations/1746319371011_pending_users.ts
```diff
@@ -0,0 +1,20 @@
...
+        .addColumn('organization_id', 'uuid', (col) => col.references('org.id').notNull())
+        .addColumn('org_slug', 'text', (col) => col.notNull())
+        .addColumn('email', 'text', (col) => col.notNull())
+        .addColumn('isResearcher', 'boolean', (col) => col.notNull())
+        .addColumn('isReviewer', 'boolean', (col) => col.notNull())
+        .addColumn('created_at', 'timestamp', (col) => col.defaultTo(sql`now()`).notNull())
+        .execute()
+
+    await db.schema.createIndex('pending_user_org_slug_idx').on('pending_user').column('org_slug').execute()
+}
```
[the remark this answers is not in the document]
stetradis: removed it, this simplifies the logic 💯

src/app/admin/invite/invite-form.tsx
stetradis: Updated to use the mantine component and theme color!

src/app/admin/invite/invite-form.tsx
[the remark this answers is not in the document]
stetradis: Created an InputError component in the errors file

nathanstitt (approved):
looks good @stetradis thank you!

PR #184, opened 2025-05-06
Org Admin (internally) - Org Admin Entry Points

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-166

### DIFF
--- a/src/components/auth.tsx
+++ b/src/components/auth.tsx
@@ -1,18 +1,47 @@
 import { useAuth as clerkUseAuth } from '@clerk/nextjs'
 import { CLERK_ADMIN_ORG_SLUG } from '@/lib/types'
+import { useEffect, useState } from 'react'
+import { getCurrentOrgUserAdminStatusAction } from '@/server/actions/user.actions'
 
 export const useAuthInfo = () => {
-    const { isLoaded, userId, orgSlug } = clerkUseAuth()
-    const isAdmin = orgSlug == CLERK_ADMIN_ORG_SLUG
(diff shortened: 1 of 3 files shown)


nathanstitt (changes requested):
Hey @therealmarv I don't think we need any of this, as I understand it we only want to show the Admin link if the user is an admin, right? I think clerk does this for us with the Protect component: https://clerk.com/docs/components/protect

why don't we add a AdminDashBoardLink component that looks roughly like:

```
const AdminDashBoardLink = () => (
   <Protect
      permission="org:admin"
>
    <NavLink
                     label="Admin"
                     leftSection={<Gear />}
                     component={Link}
                     href={orgAdminPageURL}
                     active={pathname.startsWith(`/organization/${orgSlug}/admin`)}
                     c="white"
                     color="blue.7"
                     variant="filled"
                     className={styles.navLinkHover}
                 />
</Protect>
)
```

therealmarv:
> Hey @therealmarv I don't think we need any of this, as I understand it we only want to show the Admin link if the user is an admin, right? I think clerk does this for us with the Protect component: https://clerk.com/docs/components/protect
> 

Yes that would simplify things a lot! It boils basically down to my parking lot [for tomorrow in Slack].(https://openstax.slack.com/archives/C07H80W7656/p1746564752012219).

therealmarv:
I will close this PR because it's going in the wrong direction. New one will be this one https://github.com/safeinsights/management-app/pull/187

PR #187, opened 2025-05-07
Org Admin - Org Admin Entry Points

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-166


src/components/layout/navbar-items.tsx
```diff
@@ -34,18 +52,76 @@ export const NavbarItems: FC = () => {
...
+                            active={pathname === `${orgAdminBaseUrl}/settings`}
+                            c="white"
+                            color="blue.7"
+                            variant="filled"
+                            className={styles.navLinkHover}
+                            pl="xl"
+                        /> */}
+                    </NavLink>
+                </Protect>
+            )}
```
nathanstitt, line 100: this is too much code to put inside a jsx `{}` conditional block.  please extract it to an AdminNavLink (or something named like that) component as previously suggested, then the

nathanstitt (changes requested):

src/components/layout/navbar-items.tsx
therealmarv: okay, putting it into a component

PR #191, opened 2025-05-08
truncate text for study detail fields. 

Rod0352:
truncate study detail fields if character count exceeds 20+ chars and added tooltip if fields are truncated

### DIFF
--- a/src/components/study/study-details.tsx
+++ b/src/components/study/study-details.tsx
@@ -7,37 +7,44 @@ import { StudyDocumentType } from '@/lib/types'
 import { studyDocumentURL } from '@/lib/paths'
 import { truncate } from '@/lib/string'
 
+interface BadgeWithDescriptionProps {
+    path?: string | null
+    type: StudyDocumentType
+    studyId: string
+}
+
+const BadgeWithDescription: FC<BadgeWithDescriptionProps> = ({ path, type, studyId }) => {
+    if (!path) return null
+
+    const truncatedText = truncate(path)
+    const needsTooltip = path.length > 20
+
+    const badge = (
+        <Badge
+            key={path}
+            color="#D4D1F3"
+            c="black"
+            component="a"
+            href={studyDocumentURL(studyId, type, path)}
+            target="_blank"
+            rightSection={<Download />}
+            style={{ cursor: 'pointer' }}
+        >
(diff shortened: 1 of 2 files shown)


src/components/study/study-details.tsx
```diff
@@ -12,6 +12,40 @@ export const StudyDetails: FC<{ studyId: string }> = ({ studyId }) => {
...
+                c="black"
+                component="a"
+                href={studyDocumentURL(studyId, docType, docPath)}
+                target="_blank"
+                rightSection={<Download />}
+                style={{ cursor: 'pointer' }}
+            >
+                {truncatedText}
+            </Badge>
+        )
```
nathanstitt, line 40: would css `text-overflow: ellipsis` work for this instead of clipping it with JS? https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

nathanstitt: if not, lets move `truncate` to `lib/string`

Rod0352: `text-overflow: ellipsis` would work but we'd have to define a max width property in that case. i'll move truncate over to lib/string for now and check up on desired width specs with ux

src/components/study/study-details.tsx
```diff
@@ -32,60 +60,15 @@ export const StudyDetails: FC<{ studyId: string }> = ({ studyId }) => {
...
-                                    )}
-                                    target="_blank"
-                                    rightSection={<Download />}
-                                    style={{ cursor: 'pointer' }}
-                                >
-                                    {study.descriptionDocPath}
-                                </Badge>
-                            )}
+                            {study.descriptionDocPath &&
+                                renderBadge(study.descriptionDocPath, StudyDocumentType.DESCRIPTION, study.id)}
```
nathanstitt, line 64: love this abstraction, how about taking it one step further and making `renderBadge` be a component that accepts a path that may be a string or null / undefined.  if it's not present it renders null, otherwise the same thing it renders now.  then this could all be:
```
<BadgeWithDescription path={study.descriptionDocPath} type={StudyDocumentType.DESCRIPTION} studyId={study.id} />
```

nathanstitt (approved):

PR #193, opened 2025-05-08
Otto 176 padding consistency

Rod0352:
padding consistency across study details page


nathanstitt (changes requested):
thanks for investigating this @Rod0352 , but we only want to use fixed values as a last resort and instead use mantine defined sizes.  xl looks like it's 44 and lg is 34, maybe lower to that?   https://mantine.dev/styles/variants-sizes/

nathanstitt:
actually after investigating, it looks like xl is only 32px (2rem), after code freeze lets investigate making the defined mantine sizes larger

PR #197, opened 2025-05-12
Don't include studies pending review in ready/route

chrisbendel:


src/app/api/studies/ready/route.ts
```diff
@@ -28,6 +28,7 @@ export const GET = wrapApiOrgAction(async () => {
                     .on('latestStatusChange.status', 'in', ['JOB-READY', 'JOB-RUNNING'])
                     .onRef('latestStatusChange.studyJobId', '=', 'studyJob.id'),
         )
+        .where('study.status', '!=', 'PENDING-REVIEW')
```
nathanstitt, line 31: I'm curious how studies are having runs that are approved but the study is not, that shouldn't currently be possible to achieve using the UI.  However it will in the future when we support multiple runs.

regardless, I think what we really care about is that the study is approved, so this should be:
```suggestion
        .where('study.status', '=', 'APPROVED')
```

nathanstitt (changes requested):

src/app/api/studies/ready/route.ts
chrisbendel: good call, i think that's simpler

chrisbendel: @philschatz @lwtchu @kswanson33 does this logic align with the setup apps needs? we only want to be pulling approved studies from the mgmnt app - correct?

philschatz: Yep, just approved studies (it'd be dangerous otherwise)

nathanstitt (approved):

PR #200, opened 2025-05-12
extend mfa spec e2e tests

therealmarv:
small e2e fix


therealmarv:
It's only a small fix in attempt to fix some weird e2e errors we have since Monday.

PR #201, opened 2025-05-13
fontsize inconsistency

Rod0352:
fix  fontsize inconsistency in study detail headers

### DIFF
--- a/src/app/researcher/study/[studyId]/review/page.tsx
+++ b/src/app/researcher/study/[studyId]/review/page.tsx
@@ -34,14 +34,18 @@ export default async function StudyReviewPage(props: { params: Promise<{ studyId
             <Title order={1}>Study Details</Title>
             <Paper bg="white" p="xl">
                 <Stack>
-                    <Title fz={'xl'}>Study Details</Title>
+                    <Title order={4} fz={'xl'}>
+                        Study Details
+                    </Title>
                     <StudyDetails studyId={studyId} />
                 </Stack>
             </Paper>
 
             <Paper bg="white" p="xl">
                 <Stack>
-                    <Title fz="xl">Study Code</Title>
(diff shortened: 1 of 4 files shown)


src/app/researcher/study/[studyId]/review/page.tsx
```diff
@@ -34,22 +34,22 @@ export default async function StudyReviewPage(props: { params: Promise<{ studyId
             <Title order={1}>Study Details</Title>
             <Paper bg="white" p="xl">
                 <Stack>
-                    <Title order={4}>Study Details</Title>
+                    <Title fz={'xl'}>Study Details</Title>
```
Comment (chrisbendel) [inline src/app/researcher/study/[studyId]/review/page.tsx:37]: I think we should keep the `order` props because of screen readers/accessibility/seo stuff

see here: https://mantine.dev/core/title/#size we can set the size to a couple different things, but would prefer using the `size` over `fz`

chrisbendel (changes requested):

src/app/researcher/study/[studyId]/review/page.tsx
Comment (Rod0352) [reply src/app/researcher/study/[studyId]/review/page.tsx:37]: thanks for that tip! ill opt for adding both order and fz (to ensure 20px)

Comment (chrisbendel) [reply src/app/researcher/study/[studyId]/review/page.tsx]: do we not want to use `size` instead of fz? size should proportionately control any other attributes, fz just covers the font size itself, example:
```
  return (
    <>
      <Title order={3} size="h1">
        H3 heading with h1 font-size
      </Title>
      <Title size="h4">H1 heading with h4 font-size</Title>
      <Title size={16}>H1 heading with 16px size</Title>
      <Title size="xs">H1 heading with xs size</Title>
    </>
  );
```

chrisbendel (approved):

PR #205, opened 2025-05-13
Otto 176 padding consistency

Rod0352:


src/app/sizes.css
```diff
@@ -0,0 +1,3 @@
+:root {
+    --mantine-spacing-2xl: calc(2.5rem * var(--mantine-scale));
```
nathanstitt, line 2: I like this a lot but I think Martine uses `xxl` for 2x in a few other definitions so we should probably follow that:
```suggestion
    --mantine-spacing-xxl: calc(2.5rem * var(--mantine-scale));
```

Rod0352: this actually didnt work. the mantine type definition is as follows:
`type MantineSpacing = number | "xl" | "xs" | "sm" | "md" | "lg" `
instead ive changed it to be 'xl' instead

nathanstitt: ah, ok that makes sense.

nathanstitt: hm, have you investigated adding the xxl size to the types?  this has an example: https://mantine.dev/guides/typescript/

Rod0352: thanks for this tip! i took that approach instead so we're now extending that type in the theme. (removed the sizes.css)

nathanstitt (approved):
awesome ty @Rod0352 !

PR #207, opened 2025-05-13
Otter 178 org admin settings page

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-178


src/schema/org.ts
```diff
@@ -4,9 +4,25 @@ import { Org as DefinedOrg } from '@/database/types'
...
-    name: z.string().min(1, { message: 'Name must be provided' }),
+    name: z
+        .string()
+        .trim()
+        .min(1, 'Name is required')
+        .max(50, 'Name cannot exceed 50 characters')
+        .refine((val) => /\p{L}/u.test(val), {
+            message: 'Name must contain at least one alphabetic character',
+        })
+        .refine((val) => !/^\d+$/.test(val) || /\p{L}/u.test(val), {
```
nathanstitt, line 15: these could be combined into a single `refine`:
```suggestion
        .refine((val) => {
           if (/\p{L}/u.test(val)) return { message: 'Name must contain at least one alphabetic character' }
           if (!/^\d+$/.test(val)) return { message: 'Name cannot be only numbers' }
           }
```

src/server/actions/org.actions.ts
```diff
@@ -62,6 +64,55 @@ export const getReviewerPublicKeyAction = userAction(async () => {
...
+    } catch (error) {
+        console.error(`Failed to update organization name in Clerk for ${orgSlug}:`, error)
+        // Revert the database change
+        try {
+            await db
+                .updateTable('org')
+                .set({ name: originalName })
+                .where('slug', '=', orgSlug)
+                .executeTakeFirstOrThrow()
+            console.warn(`Successfully reverted organization name in DB for ${orgSlug} after Clerk update failure.`)
```
nathanstitt, line 97: we should replace console.warn with a new logger that calls sentry.captureMessage as well as logging the message.  Maybe we can extend lib/logger to do that for error and warning?

src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx
```diff
@@ -0,0 +1,140 @@
...
+            form.resetDirty({ name: variables.name, description: variables.description ?? '' })
+            queryClient.invalidateQueries({ queryKey: ['org', orgSlug] })
+        },
+        onError: (error) =>
+            notifications.show({
+                title: 'Error updating settings',
+                message: (error as Error).message || 'An unexpected error occurred.',
+                color: 'red',
+            }),
+    })
```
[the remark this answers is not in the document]
Comment (nathanstitt) [reply src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx:58]: also since I see you're throwing ActionError, you can use `handleMutationErrorsWithForm` [example](https://github.com/safeinsights/management-app/blob/main/src/app/account/invitation/%5BinviteId%5D/account-form.tsx#L57)

src/server/actions/org.actions.ts
```diff
@@ -62,6 +64,55 @@ export const getReviewerPublicKeyAction = userAction(async () => {
...
+        orgSlug: z.string(),
+    })
+    .merge(orgSchema.pick({ name: true, description: true }))
+
+export const updateOrgSettingsAction = orgAdminAction(async ({ orgSlug, name, description }) => {
+    // Fetch the current organization
+    const currentOrg = await db.selectFrom('org').select('name').where('slug', '=', orgSlug).executeTakeFirst()
+    if (!currentOrg) {
+        throw new ActionFailure({ org: `Organization with slug ${orgSlug} not found.` })
+    }
```
nathanstitt, line 78: the `orgAdminAction` wrapper ensures that the org is loaded and and stores it in the context.  example: https://github.com/safeinsights/management-app/blob/main/src/app/organization/%5BorgSlug%5D/admin/admin-users.actions.ts#L21

nathanstitt (changes requested):

src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+    const { setValues } = form
+
+    // Effect to update form values when org data changes while in edit mode
+    useEffect(() => {
+        if (isEditing && org) {
+            setValues({ name: org.name, description: org.description ?? '' })
+        }
+    }, [isEditing, org, setValues])
```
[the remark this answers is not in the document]
Comment (therealmarv) [reply src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx:42]: This is overengineered for org changes from other si/org admins. I agree it does not make any sense or is too much.

src/server/actions/org.actions.ts
therealmarv: I will replace all console functions here with logger.ts functions. I agree with the sentry captureMessage. But I will leave that for a new PR after this one (because it requires some refactoring of logger.ts)

src/app/organization/[orgSlug]/admin/settings/settings-form.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const FormLabel = ({ label, required = false }: { label: string; required?: boolean }) => (
+    <Text fw={600} size="sm">
+        {label}
+        {required && (
+            <Text span c="red" ml={4}>
+                *
+            </Text>
+        )}
+    </Text>
```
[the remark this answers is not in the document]
Comment (therealmarv) [reply src/app/organization/[orgSlug]/admin/settings/settings-form.tsx]: I've introduced a new component `FormFieldLabel` which addresses all concerns

src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx
```diff
@@ -0,0 +1,117 @@
...
+                    type="submit"
+                    form="organization-settings-form"
+                    disabled={!form.isDirty() || !form.isValid()}
+                    loading={isOrgUpdating}
+                >
+                    Save
+                </Button>
+            </Group>
+        )
+        content = <OrganizationSettingsEdit form={form} onFormSubmit={onFormSubmit} />
```
Comment (nathanstitt) [inline src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx:97]: form and onFormSubmit is only used by the OrganizationSettingsEdit component, they can be moved there to lessen the responsibility of this component.

Comment (nathanstitt) [reply src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx:97]: this is looking much better, ty!  One final optimization, why don't you have each file export a `Buttons` and `Body`, each accepting identical props, then this becomes something like:
```
import { EditBody, EditButtons } from './edit'
import { DisplayBody, DisplayButtons } from './display'
const Editing = {
  true: { Body: editBody, Display: editDisplay }
  false: { Body: displayBody, Buttons: displayButtons }
}

const OrganizationSettingsManager = () => {
  const [isEditing, setEditing] = useState(false);
  
 const Current = Editing[isEditing]

  return (
<Paper shadow="xs" p="xl" mb="xl">
             <Flex direction="row" justify={'space-between'} align="center" mb="lg">
                 <Title order={3}>About organization</Title>
                 <Current.Buttons onEdit={() => setEditting(true)}  onEditComplete={() => setEditting(false)}  />
             </Flex>
             <Divider mb="lg" />
             <Content.Body />
         </Paper>
  )
}
```

src/app/organization/[orgSlug]/admin/settings/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+    return (
+        <Stack p="md">
+            <RequireOrgAdmin />
+            <Breadcrumbs>{items}</Breadcrumbs>
+            <Divider />
+            <Title order={1} mb="xl">
+                Settings
+            </Title>
+
+            <OrganizationSettingsManager orgSlug={orgSlug} initialOrg={initialOrg} />
```
Comment (nathanstitt) [inline src/app/organization/[orgSlug]/admin/settings/page.tsx:33]: why is the edit and display components loading the org again instead of using  `initialOrg`?

src/components/form-field-label.tsx
```diff
@@ -0,0 +1,56 @@
...
+                )}
+            </Title>
+        )
+    }
+    return (
+        <Input.Label htmlFor={inputId} className={className} style={style}>
+            {labelContent}
+        </Input.Label>
+    )
+}
```
nathanstitt, line 56: I like this a lot, thank you!

src/server/actions/wrappers.ts
```diff
@@ -107,7 +109,18 @@ export function userAction<S extends Schema, F extends WrappedFunc<S>>(func: F,
...
+                            for (const key in fieldErrors) {
+                                if (fieldErrors[key]) {
+                                    sanitizedErrors[key] = (fieldErrors[key] as string[]).join(', ')
+                                }
+                            }
+                            reject(new ActionFailure(sanitizedErrors))
+                        } else {
+                            reject(error)
+                        }
                     }
```
nathanstitt, line 124: I do like this as a fallback, but really we shouldn't ever get to this state.  the form should be validating with the zod schema and won't even call the onSubmit if it's invalid

nathanstitt (commented):
Looking good @therealmarv thank you for splitting the component up, made one file suggestion on it to remove some more of the logic from the manager.

src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx
Comment (nathanstitt) [reply src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx:97]: hey, after thinking this a bit more, that's also over complicating it.  IT's better to duplicate the 4 or so lines of JSX and have everything completely separate:
```
import { Edit } from './edit'
import { Display } from './display'

const OrganizationSettingsManager = () => {
  const [isEditing, setEditing] = useState(false);
  return (
      <Paper shadow="xs" p="xl" mb="xl">
            {isEditing ? <Edit /> : <Display />}
      </Paper>
}
```

Comment (therealmarv) [reply src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx:97]: That was my initial approach. I still have it locally on my branch. It caused flickering on switching from display to edit mode and after some research I even tried to put a transition animation there to reduce this flickering. I also know the flickering is absolutely not critical here 😄 but I could not get rid of it completely because it replaces a big part of the page. The current solution is like "half-way" separated, something in between but "performance" wise nice. It's definitely an interesting problem... there must be other people who want to replace some parts in between only to not re-render all the other things too.

I can resurrect my local branch today and you can try it out to see the flickering.

src/app/organization/[orgSlug]/admin/settings/organization-settings-edit.tsx
```diff
@@ -0,0 +1,64 @@
...
+                <Grid align="flex-start">
+                    <Grid.Col span={orgSettingsLabelSpan}>
+                        <FormFieldLabel label="Description" inputId={form.key('description')} variant="orgset" />
+                    </Grid.Col>
+                    <Grid.Col span={orgSettingsValueSpan}>
+                        <Textarea
+                            id={form.key('description')}
+                            aria-label="Description"
+                            maxLength={250}
+                            key={form.key('description')}
```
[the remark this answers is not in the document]
Comment (therealmarv) [reply src/app/organization/[orgSlug]/admin/settings/organization-settings-edit.tsx:51]: I used key={form.key('field')} on TextInput and TextArea so that whenever the form is reset or values are updated, the input will completely re-mount and show the new value. This is needed in Mantine’s (preferred) uncontrolled form mode to make sure the input always matches the form’s latest values. It is explained in the Mantine uncontrolled mode documentation https://mantine.dev/form/uncontrolled/#uncontrolled-mode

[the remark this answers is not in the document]
Comment (therealmarv) [reply src/app/organization/[orgSlug]/admin/settings/organization-settings-edit.tsx:51]: I used key={form.key('field')} on TextInput and TextArea so that, whenever the form is reset or values are updated, the input will completely re-mount and show the new value (it's a intended side effect).

This is especially needed in Mantine’s uncontrolled mode (not used here) to make sure the input always matches the latest form values, but you can use it in controlled mode too if you want to force a re-mount. The Mantine documentation explains this best for [uncontrolled mode](https://mantine.dev/form/uncontrolled/#uncontrolled-mode).

src/app/organization/[orgSlug]/admin/settings/page.tsx
Comment (therealmarv) [reply src/app/organization/[orgSlug]/admin/settings/page.tsx:33]: display is updating org data in background but I'm removing that functionality.

src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const handleSaveSuccess = () => {
+        cancelEdit()
+    }
+
+    return (
+        <Paper shadow="xs" p="xl" mb="xl">
+            {isEditing ? (
+                <OrganizationSettingsEdit
+                    org={org}
+                    onSaveSuccess={handleSaveSuccess}
```
Comment (nathanstitt) [inline src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx:25]: since `handleSaveSuccess` only calls `cancelEdit` this could be:
```suggestion
                    onSaveSuccess={cancelEdit}
```

nathanstitt (approved):
approved but if you can accept the suggested edit first that'll clean it up a tiny bit

src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx
Comment (therealmarv) [reply src/app/organization/[orgSlug]/admin/settings/organization-settings-manager.tsx:25]: yes, I'm also seeing that right now in my editor. Makes no sense to use that method

PR #208, opened 2025-05-14
Wait for items to load?

chrisbendel:


nathanstitt:
thanks for the PR @chrisbendel,  I've moved the speeeeling fix into #219 where I'm also working on getting specs working again

PR #211, opened 2025-05-15
fix org admin selector; disable flaky specs

nathanstitt:


src/components/layout/org-admin-dashboard-link.tsx
```diff
@@ -36,7 +36,7 @@ export const OrgAdminDashboardLink: FC<OrgAdminDashboardLinkProps> = ({ orgSlug,
     }
 
     return (
-        <Protect role="org:admin">
+        <Protect role="admin">
```
therealmarv, line 39: got confused for a moment but I see it's not Clerk's Protect component anymore.

PR #213, opened 2025-05-15
create a context to track user inactivity. 

Rod0352:
added wrapper, event listeners


nathanstitt:
sorry I'm confused: why are you tracking this client side?  did you investigate https://clerk.com/docs/authentication/configuration/session-options#maximum-lifetime

src/components/activity-context.tsx
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            if (remaining !== null && remaining <= 0) {
+                notifications.show({
+                    title: 'Session Expired',
+                    message: "You've been logged out due to inactivity. Log back in to resume your work.",
+                    withCloseButton: true,
+                    autoClose: false,
+                })
+                signOut()
```
nathanstitt, line 63: does clerk redirect once we call `signOut`? in that case the toast may not display

nathanstitt (approved):
looks good but had small question on calling `signOut`

src/components/activity-context.tsx
Rod0352: it does redirect and from testing it seems the toast stays around until closed

nathanstitt: ok, that sounds great then.

PR #215, opened 2025-05-15
Small change README

therealmarv:
I tried to comment that on Rodney's PR but that comment was "pending" and pending comments are only visible for me 🤦‍♂️ when not finishing the review (argh).

Postgres is not a direct requirement when it's handled by docker locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ It's responsible for:
 
 Either:
 
-- **Option 1:** Node.js + Docker and Docker Compose (handles PostgreSQL automatically)
+- **Option 1:** Node.js + Docker and Docker Compose (installs PostgreSQL in container)
 - **Option 2:** Node.js + PostgreSQL installed locally
 
 ## Development
--- a/README.md
+++ b/README.md
@@ -14,9 +14,14 @@ It's responsible for:
 
 ## Requirements
 
+### Production Environment
 - Node.js version 20.x or higher
 - PostgreSQL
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -14,9 +14,17 @@ It's responsible for:
...
+
 - Node.js version 20.x or higher
 - PostgreSQL
-- Docker and Docker Compose (optional, for containerized development)
+
+### Development Environment
+
+Either:
+
+- **Option 1:** Node.js + Docker and Docker Compose (recommended, handles PostgreSQL automatically)
```
nathanstitt, line 26: personal preference, but I don't think we should recommend docker.  while it does "solve" pg, it also introduces a lot of complexity with people not being able to access the app directly and needing to recreate it frequently.

Personally  `brew install postgres` seems _much_ simpler than dealing with docker all the time 😁

therealmarv: I will make it more neutral. I think that's more in line with the both options explained further down the README.

nathanstitt (approved):

nathanstitt:
looks like the README needs to be listed

therealmarv:
It's weird, I cannot replicate the linting error locally and cannot fix it because I don't know what's wrong. All okay locally. It's not the first time I see the CI linting being more strict... I just don't understand: why?

Things I tried so far:
* Reinstalled node_modules
* fresh clone
* replicate ci environment in docker

PR #216, opened 2025-05-18
Otter 181 - reviewer key

stetradis:
- Refactors reviewer key page
- Fixes for performance optimization/hydration error/overlay display not showing


src/components/auth.tsx
```diff
@@ -25,9 +36,9 @@ export const Protect: React.FC<ProtectProps> = ({ role, orgSlug, children }) =>
 
     if (!isLoaded) return null
 
-    if (role == 'admin' && org.isAdmin) return children
-    if (role == 'researcher' && org.isResearcher) return children
-    if (role == 'reviewer' && org.isReviewer) return children
+    if (role == AuthRole.Admin && org.isAdmin) return children
+    if (role == AuthRole.Researcher && org.isResearcher) return children
+    if (role == AuthRole.Reviewer && org.isReviewer) return children
```
nathanstitt, line 42: why'd this file get updated? I don't see any changes, but am probably missing them.  it does seem like maybe good cleanups tho

src/components/require-reviewer-keys.tsx
```diff
@@ -1,24 +1,27 @@
...
                 router.push('/account/keys')
             }
         })
-    }, [pathname, router])
+    }, [router, pathname])
 
-    return null
+    if (!hasKey) {
+        return <LoadingOverlay visible />
+    }
```
nathanstitt, line 26: this component is intended to be invisible and silently redirect if key is needed,  I don't think we should render a spinner here.

is using useState really needed?  was there issues with the  original logic ?

src/server/actions/user-keys.actions.ts
```diff
@@ -32,12 +32,19 @@ export const setReviewerPublicKeyAction = userAction(async ({ publicKey, fingerp
...
+            .values({
+                userId,
+                publicKey: Buffer.from(publicKey),
+                fingerprint,
+            })
+            .execute()
+    } catch (error) {
+        console.error('Error setting reviewer public key:', error)
+        throw error
+    }
```
nathanstitt, line 47: this is a bad pattern, instead we try to use `executeTakeFirstOrThrow` which can take a function if we need to log or something.  there is also a `ActionFailure` error which pairs nicely with useMutation. For an example see: 
throwing an error: https://github.com/safeinsights/management-app/blob/main/src/server/actions/org.actions.ts#L53-L56

and handling it: https://github.com/safeinsights/management-app/blob/main/src/app/organization/%5BorgSlug%5D/admin/users-table.tsx#L64

nathanstitt (commented):
looking good @stetradis have a few questions about why some of the changes

src/app/account/keys/generate-keys.tsx
```diff
@@ -1,25 +1,42 @@
...
-    const [opened, { open, close }] = useDisclosure(false)
+    const theme = useMantineTheme()
+    const [keys, setKeys] = useState<Keys>()
+    const [confirmationOpened, { open: openConfirmKeyCopied, close: closeConfirmKeyCopied }] = useDisclosure(false)
+    const [isKeyCopied, setIsKeyCopied] = useState<boolean>(false)
 
     const onGenerateKeys = async () => {
         const { privateKeyString, fingerprint, exportedPublicKey } = await generateKeyPair()
-        const privateKeyLines = (privateKeyString.match(/.{1,64}/g) || []).join('\n')
+        const privateKeyLines = (privateKeyString.match(/.{1,112}/g) || []).join('\n')
```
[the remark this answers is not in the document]
stetradis: this was a formatting choice made so that the key display matches the design. From what I understand, a PEM parser should strip out whitespace and newlines before decoding so the chunk size shouldn't matter but happy to revert to 64 if you'd like to stick to the convention, it just won't exactly match the design

src/components/require-reviewer-keys.tsx
stetradis: I’ve noticed that redirects are sluggish, so users without a key can briefly access and interact with the dashboard. Showing a spinner during the redirect prevents unintended access and gives a clear visual cue that loading is in progress

src/app/(root)/user-nav.tsx
```diff
@@ -15,13 +15,13 @@ export const UserNav = () => {
         if (auth.isResearcher) {
             router.push('/researcher/dashboard') // Redirect to the Researcher dashboard
         } else if (auth.isReviewer) {
-            router.push(`/reviewer/${auth.orgSlug}/dashboard`) // Redirect to the Member dashboard
+            router.push('/account/keys') // Redirect to account keys page to generate a reviewer key
```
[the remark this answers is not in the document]
stetradis: Ahh good catch! Leftover from a revert, definitely should be changed

src/app/account/keys/generate-keys.tsx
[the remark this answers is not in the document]
nathanstitt: as long as we can use whatever is sent here to review later, I guess it doesn't really matter

src/components/require-reviewer-keys.tsx
nathanstitt: I see. in that case I think the spinner should probably be shown in the layout, not in the `requireXYZ` components.  maybe we extend UserLayout in components/layouts/user-layout to do that?

src/server/actions/user-keys.actions.ts
stetradis: Thanks for sharing these functions! I've updated it but found that `executeTakeFirstOrThrow` is more appropriate for select statements where we would want to return a row. Sticking with `execute()` for now unless we want to add a `.returning()`

src/app/account/keys/generate-keys.tsx
```diff
@@ -28,91 +47,111 @@ export const GenerateKeys: FC = () => {
...
+                <ConfirmationModal onClose={closeConfirmKeyCopied} isOpen={confirmationOpened} keys={keys} />
             </Container>
         )
     }
+}
+
+const ConfirmationModal: FC<{ onClose: () => void; isOpen: boolean; keys: Keys }> = ({ onClose, isOpen, keys }) => {
+    const router = useRouter()
+
+    const { mutate: saveReviewerKey } = useMutation({
```
nathanstitt, line 127: capture pending for use on button below:
```suggestion
    const { mutate: saveReviewerKey, isPending: isSavingKey } = useMutation({
```

src/app/account/keys/generate-keys.tsx
```diff
@@ -28,91 +47,111 @@ export const GenerateKeys: FC = () => {
...
+        mutationFn: () =>
+            setReviewerPublicKeyAction({
+                publicKey: keys.binaryPublicKey,
+                fingerprint: keys.fingerprint,
+            }),
+        onError: reportMutationError,
+        async onSuccess() {
```
nathanstitt, line 134: doesn't need to be async, router.push is not:
```suggestion
       onSuccess() {
```

src/components/require-reviewer-keys.tsx
nathanstitt: I think you extended the layout, but haven't removed it from here yet 😁

src/server/actions/user-keys.actions.ts
nathanstitt: heh no, `executeTakeFirstOrThrow` is poorly named, its for all statements, not just select with returning. in fact their insert example shows it being used: https://kysely-org.github.io/kysely-apidoc/classes/InsertResult.html

nathanstitt (commented):
looking good @stetradis, just a few final comments

stetradis:
@nathanstitt resolved all comments on this thread 👍

PR #222, opened 2025-05-21
Update user management table styles

nathanstitt:
Adds tooltips and adjusts the styles to better match mockups.   Depends on #219


jbwilson8:
@nathanstitt  nice additions!

jbwilson8 (approved):

PR #223, opened 2025-05-22
OTTO-180 nav panel accessibility

Rod0352:
- abstracted hotkeys hook. 
- created ref wrapper for non-refable components

**Adresses the following acceptance criteria**

- Keyboard navigation must be supported to ensure accessible access to all audiences

- All buttons must be focusable when navigating using the keyboard (e.g., via Tab key) or using arrow keys (e.g., up/down)


nathanstitt (approved):
thanks @Rod0352

PR #224, opened 2025-05-22
Adds feature for regenerating a reviewer key Otter 182

stetradis:
- Adds feature for regenerating a reviewer key

### DIFF
--- a/src/app/account/keys/layout.tsx
+++ b/src/app/account/keys/layout.tsx
@@ -10,7 +10,6 @@ export default async function ReviewerKeysPageLayout({
     children: ReactNode
 }>) {
     const publicKey = await getReviewerPublicKeyAction()
-    const clerkPublishableKey = process.env.NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY || ''
 
     if (publicKey) {
         // User has a key - show UserLayout
@@ -21,6 +20,6 @@ export default async function ReviewerKeysPageLayout({
         )
     } else {
         // No key, first-time generation - show AnonLayout
-        return <AnonLayout clerkPublishableKey={clerkPublishableKey}>{children}</AnonLayout>
+        return <AnonLayout>{children}</AnonLayout>
     }
 }
--- a/src/app/account/keys/layout.tsx
+++ b/src/app/account/keys/layout.tsx
@@ -1,0 +1,26 @@
+import { getReviewerPublicKeyAction } from '@/server/actions/user-keys.actions'
+import { UserLayout } from '@/components/layout/user-layout'
+import { AnonLayout } from '@/components/layout/anon-layout'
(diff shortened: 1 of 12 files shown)


src/app/account/keys/layout.tsx
```diff
@@ -0,0 +1,26 @@
...
+        return (
+            <UserLayout>
+                <RegenerateKeys />
+            </UserLayout>
+        )
+    } else {
+        // No key, first-time generation - show AnonLayout
+        return <AnonLayout clerkPublishableKey={clerkPublishableKey}>{children}</AnonLayout>
+    }
+}
```
nathanstitt, line 26: future work: I think we're also reading the `process.env.NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` in other layouts, lets split AnonLayout so part can be 'use server' and read the env var itself, that way we're not repeating this everywhere

src/components/layout/navbar-profile-menu.tsx
```diff
@@ -99,20 +99,22 @@ export function NavbarProfileMenu() {
...
+                        }}
+                        c="white"
+                        className={styles.navLinkProfileHover}
+                        role="menuitem"
+                        aria-label="Reviewer Key"
+                        tabIndex={opened ? 0 : -1}
+                        ref={reviewerKeyMenuItemRef}
+                    />
+                )}
                 <NavLink
```
nathanstitt, line 118: can you extract this into it's own `<ReviewerKeyLink />` component which has a `if (!auth.isReviewer) return null` in it?  That will eliminate the large `{` `}` block

src/server/actions/user-keys.actions.ts
```diff
@@ -35,4 +36,22 @@ export const setReviewerPublicKeyAction = userAction(async ({ publicKey, fingerp
...
+
+    await db
+        .updateTable('userPublicKey')
+        .set({
+            publicKey: Buffer.from(publicKey),
+            fingerprint,
+            updatedAt: new Date(),
+        })
+        .where('userId', '=', userId)
+        .executeTakeFirstOrThrow(() => new ActionFailure({ message: 'Failed to update reviewer public key.' }))
```
nathanstitt, line 54: the property in the ActionFailure object will look for a form input with a matching name to display errors on, or if not will capitalize it and make a string shown in a toast.  This will end up being something like: `Message Failed to update reviewer public key.` which reads a little awkward.  Maybe it'd be better to use `error: key did not update in database` so it reads `Error key did not update..`
```suggestion
        .executeTakeFirstOrThrow(() => new ActionFailure({ error: 'public key was not saved.' }))
```

nathanstitt (approved):
looks great @stetradis a few comments but I'll approve so you can merge after addressing them

PR #226, opened 2025-05-22
check keys as part of server rendered layout

nathanstitt:


src/app/reviewer/layout.tsx
```diff
@@ -1,39 +1,16 @@
...
 import { UserLayout } from '@/components/layout/user-layout'
+import { redirect } from 'next/navigation'
 
-type Props = { children: ReactNode }
+type Props = { children: React.ReactNode }
 
-export default function ReviewerLayout({ children }: Props) {
-    const pathname = usePathname()
-    const router = useRouter()
-    const [loading, setLoading] = useState(!pathname.startsWith('/account'))
```
stetradis, line 13: Showing  the overlay while the app redirects to account/keys depends on determining a loading state here, without it we end up with a blank flickering screen between the redirect

stetradis (approved):

PR #228, opened 2025-05-24
OTTER-202 fix: study rejection not working

stetradis:
- Passes orgSlug to query to fetch study job


src/server/actions/study.actions.ts
```diff
@@ -232,7 +232,7 @@ export const rejectStudyProposalAction = orgAction(
                 .where('id', '=', studyId)
                 .execute()
 
-            const latestJob = await latestJobForStudy(studyId, { userId }, trx)
+            const latestJob = await latestJobForStudy(studyId, { orgSlug, userId }, trx)
```
nathanstitt, line 235: nice catch!

PR #232, opened 2025-05-27
Org Admin - Invite People Flow - Logic Redesign

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-199

### DIFF
--- a/src/app/organization/[orgSlug]/admin/admin-users.actions.ts
+++ b/src/app/organization/[orgSlug]/admin/admin-users.actions.ts
@@ -5,9 +5,25 @@ import { z, inviteUserSchema } from './invite-user.schema'
 import { orgActionContext, orgAdminAction } from '@/server/actions/wrappers'
 import { sendInviteEmail } from '@/server/mailer'
 import { onUserInvited } from '@/server/events'
+import { ActionFailure } from '@/lib/errors'
 
 export const orgAdminInviteUserAction = orgAdminAction(
     async ({ invite }) => {
+        const { org } = await orgActionContext()
+
+        // Check if the user (by email) is already inside this organization
+        const existingOrgUser = await db
+            .selectFrom('user')
+            .innerJoin('orgUser', 'user.id', 'orgUser.userId')
+            .select('user.id')
+            .where('user.email', '=', invite.email)
+            .where('orgUser.orgId', '=', org.id)
+            .executeTakeFirst()
+
+        if (existingOrgUser) {
+            throw new ActionFailure({ email: 'This team member is already in this organization.' })
+        }
+
         // Check if the user already exists in pending users, resend invitation if so
         const existingPendingUser = await db
             .selectFrom('pendingUser')
@@ -18,7 +34,6 @@ export const orgAdminInviteUserAction = orgAdminAction(
             await sendInviteEmail({ emailTo: invite.email, inviteId: existingPendingUser.id })
             return
         }
-        const { org } = await orgActionContext()
         const record = await db
             .insertInto('pendingUser')
             .values({


src/app/organization/[orgSlug]/admin/admin-users.actions.ts
```diff
@@ -5,9 +5,25 @@ import { z, inviteUserSchema } from './invite-user.schema'
...
+            .innerJoin('orgUser', 'user.id', 'orgUser.userId')
+            .select('user.id')
+            .where('user.email', '=', invite.email)
+            .where('orgUser.orgId', '=', org.id)
+            .executeTakeFirst()
+
+        if (existingOrgUser) {
+            throw new ActionFailure({ email: 'This team member is already in this organization.' })
+        }
+
```
Comment (nathanstitt) [inline src/app/organization/[orgSlug]/admin/admin-users.actions.ts:26]: what happens if the user is a member of a _different_ organization?

src/app/account/invitation/[inviteId]/invitation-handler.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    useEffect(() => {
+        if (isLoaded && isSignedIn) {
+            setIsLoadingAction(true)
+            claimInviteAction({ inviteId })
+                .then((result) => {
+                    if (result.success && result.organizationName) {
+                        setOrgName(result.organizationName)
+                        setNewOrgSlug(result.orgSlug || null)
+                        // The success message is now part of the SuccessPanel
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/invitation-handler.tsx:62]: using the state above this could become `setState({...state, ...result })`

src/app/account/invitation/[inviteId]/invitation-handler.tsx
```diff
@@ -0,0 +1,128 @@
...
+                        message: 'An unexpected error occurred.',
+                        color: 'red',
+                    })
+                })
+                .finally(() => setIsLoadingAction(false))
+        }
+    }, [isLoaded, isSignedIn, inviteId])
+
+    useEffect(() => {
+        if (!newOrgSlug || !user) return
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/invitation-handler.tsx:86]: this is the 3rd useEffect, can they be combined? or this be completed after the org is set?

src/app/account/invitation/[inviteId]/invitation-handler.tsx
```diff
@@ -0,0 +1,128 @@
...
+            setActive({ organization: newOrgSlug }).catch(() => {})
+        }
+    }, [newOrgSlug, user, setActive])
+
+    // fetch whether this email is already in our SI user table
+    const { data: userExists } = useQuery({
+        queryKey: ['userExists', invitedEmail],
+        queryFn: () => userExistsAction(invitedEmail),
+        enabled: isLoaded,
+    })
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/invitation-handler.tsx:100]: can this all be done as part of the `claimInviteAction`  and return error if it's already in use?

src/app/account/invitation/[inviteId]/invitation-handler.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    useEffect(() => {
+        if (!newOrgSlug || !user) return
+
+        const orgMeta = user.publicMetadata?.orgs?.find((o) => o.slug === newOrgSlug)
+        if (orgMeta?.isReviewer) {
+            // update Clerk’s active organization so useDashboardUrl picks it up
+            setActive({ organization: newOrgSlug }).catch(() => {})
+        }
+    }, [newOrgSlug, user, setActive])
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/invitation-handler.tsx:93]: we're not going to redirect to dashboard for new users, why do we care about this here?  we can set it at the end of the process if we need to

src/app/account/invitation/[inviteId]/invitation-handler.tsx
```diff
@@ -0,0 +1,128 @@
...
+    // remember the inviteId so that once they sign in we auto-claim
+    useEffect(() => {
+        window.localStorage.setItem('pendingInviteId', inviteId)
+    }, [inviteId])
+
+    useEffect(() => {
+        if (isLoaded && isSignedIn) {
+            setIsLoadingAction(true)
+            claimInviteAction({ inviteId })
+                .then((result) => {
```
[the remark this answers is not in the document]
Comment (nathanstitt) [reply src/app/account/invitation/[inviteId]/invitation-handler.tsx:58]: Actually shouldn't `claimInviteAction` be called using useMutation?  then you get the loading and error handling as part of it

src/app/account/invitation/[inviteId]/invitation-handler.tsx
```diff
@@ -0,0 +1,128 @@
...
+        enabled: isLoaded,
+    })
+
+    if (!isLoaded || isLoadingAction) {
+        return <LoadingMessage message="Processing your invitation..." />
+    }
+
+    // returning user: prompt them to sign in instead of sign up
+    if (userExists && !isSignedIn) {
+        return <SignIn />
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/invitation-handler.tsx:108]: what happens after they signin?  shouldn't we delete the invitation at this point and redirect to sign page instead of rendering it?

src/app/account/invitation/[inviteId]/invitation-handler.tsx
```diff
@@ -0,0 +1,128 @@
...
+    if (isSignedIn) {
+        if (error) {
+            return <Text color="red">{error}</Text>
+        }
+        if (orgName) {
+            return (
+                <SuccessPanel title={`You're now a member of ${orgName}!`} onContinue={() => router.push(dashboardUrl)}>
+                    Visit your dashboard
+                </SuccessPanel>
+            )
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/invitation-handler.tsx:120]: do you know the user exists at this point? I may be misreading all the if statements, but I don't think you're checking that here

src/app/account/invitation/[inviteId]/invitation-handler.tsx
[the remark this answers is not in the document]
Comment (nathanstitt) [reply src/app/account/invitation/[inviteId]/invitation-handler.tsx:58]: another thought, why aren't we checking some of this in the `Page` on the server before we even render this component?  We can fetch the org and things there and pass them as props to this

src/app/account/invitation/[inviteId]/new-user-account-form.tsx
```diff
@@ -1,25 +1,34 @@
...
-const Success: FC = () => {
+const Success: FC<{ inviteId: string }> = ({ inviteId }) => {
     const router = useRouter()
 
     const onContinue = () => {
-        router.push('/account/mfa')
+        // Store inviteId for use after MFA setup
+        localStorage.setItem('pendingInviteId', inviteId)
+        // Redirect to generic MFA page
+        router.push(`/account/mfa/app`)
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/new-user-account-form.tsx:31]: instead of using localStorage, we could pass the invite id as a query parameter:
```suggestion
        router.push(`/account/mfa/app?inviteId=${inviteId}`)
```
And then in the mfa page you'd read it using `useSearchParams`

src/app/account/invitation/[inviteId]/new-user-account-form.tsx
```diff
@@ -53,25 +71,30 @@ const SetupAccountForm: FC<InviteProps & { onComplete(): void }> = ({ inviteId,
...
     const { mutate: createAccount, isPending: isCreating } = useMutation({
-        mutationFn: (form: FormValues) => onCreateAccountAction({ inviteId, form }),
+        mutationFn: (formData: FormValues) => onCreateAccountAction({ inviteId, email, form: formData }), // Pass email
         onError: handleMutationErrorsWithForm(form),
-        async onSuccess(_, vals) {
+        async onSuccess(clerkUserId, vals) {
+            // clerkUserId is returned by onCreateAccountAction
             if (!signIn) {
-                reportError('unable to signin')
+                reportError(new Error('SignIn object not available'), 'SignIn not available post account creation')
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/new-user-account-form.tsx:79]: `reportError` will accept and display a string:
```suggestion
                reportError('SignIn not available post account creation')
```

src/app/account/invitation/[inviteId]/new-user-account-form.tsx
```diff
@@ -53,25 +71,30 @@ const SetupAccountForm: FC<InviteProps & { onComplete(): void }> = ({ inviteId,
...
+                return
+            }
+            if (!clerkUserId) {
+                reportError(new Error('Clerk User ID not returned from account creation'), 'Clerk User ID missing')
                 return
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/new-user-account-form.tsx:84]: can the action throw an exception if it didn't get a clerkUserId and then that'd be handled by the onError of the useMutation

src/app/account/invitation/[inviteId]/page.tsx
```diff
@@ -11,18 +10,16 @@ export default async function AcceptInvitePage({ params }: { params: Promise<{ i
 
     const invite = await db
         .selectFrom('pendingUser')
-        .select('email')
+        .select(['email', 'orgId']) // Fetch orgId to potentially get orgName later if needed
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/page.tsx:13]: this should fetch the org and other information needed by InviteHandler so that it doesn't need to call any actions to fetch the data.
```suggestion
        .join('org', 'pendingUser.orgId', 'org.id')
        .select(['email', 'orgId', 'org.name as orgName', 'org.slug as orgSlug'])
```

src/app/account/invitation/[inviteId]/page.tsx
```diff
@@ -11,18 +10,16 @@ export default async function AcceptInvitePage({ params }: { params: Promise<{ i
...
 
-    if (!invite) return <ErrorAlert error="Invalid invitation" />
+    if (!invite) return <ErrorAlert error="Invalid or already claimed invitation." />
 
     return (
         <Container>
-            <Panel title="Welcome To SafeInsights">
-                <AccountPanel inviteId={inviteId} email={invite.email} />
-            </Panel>
+            <InvitationHandler inviteId={inviteId} invitedEmail={invite.email} />
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/page.tsx:22]: why did you add orgId if you didn't pass it on here:
```suggestion
            <InvitationHandler invite={invite} />
```

nathanstitt (commented):
only partially done, but submitting so comments are visible

src/app/account/invitation/[inviteId]/new-user-account-form.tsx
Comment (therealmarv) [reply src/app/account/invitation/[inviteId]/new-user-account-form.tsx:31]: I was thinking about that solution too. The difference with localStorage is that the email link is not needed when the mfa has not been done fully and user returns in same browser.

I can switch it to a query parameter but then the email link is mandatory or the server makes an assumption which invite is accepted (in the edge case that more than one invite is sent out to same user). Will think about that... 🤔

src/server/actions/invite.actions.ts
```diff
@@ -0,0 +1,192 @@
...
+        logger.error({ message: 'Non-Clerk error during user creation', error, email })
+        throw new ActionFailure({ form: 'Could not create account due to an unexpected error.' })
+    }
+}, CreateAccountSchema)
+
+const ClaimInviteSchema = z.object({
+    inviteId: z.string(),
+})
+
+export const claimInviteAction = userAction(async ({ inviteId }) => {
```
nathanstitt, line 83: this is called after the MFA setup is complete, correct?  If that's the case then it seems like all it should do is look update any invitations that are for a lowercase email matching the current user's email to be claimed.

All the other checks you have here should be front-loaded so the user hits them before going through the acceptance flow

src/server/actions/invite.actions.ts
```diff
@@ -0,0 +1,192 @@
...
+        isResearcher: pendingUser.isResearcher,
+        isReviewer: pendingUser.isReviewer,
+        isAdmin: false, // Default for invitees
+    })
+
+    try {
+        const client = await clerkClient()
+        // 2. Add to Clerk Organization (or update role if already member)
+        try {
+            await client.organizations.createOrganizationMembership({
```
nathanstitt, line 134: `findOrCreateOrgMembership` does this

src/server/actions/invite.actions.ts
```diff
@@ -0,0 +1,192 @@
...
+                logger.info(
+                    `User ${clerkUserId} already member of Clerk org ${clerkOrg.id}. Role update might be needed separately if changed.`,
+                )
+            } else {
+                throw e // Re-throw if it's another error
+            }
+        }
+
+        // 3. Update Clerk User Public Metadata
+        const clerkUserToUpdate = await client.users.getUser(clerkUserId)
```
nathanstitt, line 150: this should be done when account is created, not after mfa

src/server/actions/user.actions.ts
```diff
@@ -52,7 +52,12 @@ export const onUserSignInAction = anonAction(async () => {
                 isAdmin: md?.isAdmin,
                 isReviewer: md?.isReviewer,
             })
-        } catch {}
+        } catch (e) {
+            logger.warn(
+                `During login, user ${user.id} was found to be a member of clerk org ${org.organization.slug}, which was not found in the SI database. Skipping membership creation.`,
+                e,
+            )
```
nathanstitt, line 59: how do you know the error is caused by this?

src/server/clerk.ts
```diff
@@ -11,27 +11,34 @@ type ClerkOrganizationProps = {
     slug: string
 }
 
-export const findOrCreateClerkOrganization = async ({ name, slug, adminUserId }: ClerkOrganizationProps) => {
+export const findClerkOrganization = async ({ slug }: { slug: string }) => {
     const client = await clerkClient()
+    // .getOrganization throws an error if not found, which is desired behavior.
+    return await client.organizations.getOrganization({ slug })
+}
```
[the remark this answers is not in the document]
therealmarv: good point, I was thinking too complicated and can reverse the whole clerk.ts refactoring and use that one liner in the invite flow.

src/server/clerk.ts
```diff
@@ -11,27 +11,34 @@ type ClerkOrganizationProps = {
...
+    })
+}
 
-        return clerkOrg
+export const findOrCreateClerkOrganization = async (props: ClerkOrganizationProps) => {
+    try {
+        return await findClerkOrganization({ slug: props.slug })
+    } catch {
+        // if we get an error, we assume it's because the org doesn't exist.
+        return await createClerkOrganization(props)
```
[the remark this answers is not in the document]
therealmarv: findOrCreateClerkOrganization is used in SI admin (but this would be too much for invite flow). I wanted only findClerkOrganization but I will use that one liner above directly.

src/server/actions/user.actions.ts
```diff
@@ -84,3 +89,44 @@ export const updateUserRoleAction = orgAdminAction(
...
+export const userExistsAction = anonAction(async (email: string) => {
+    const row = await db.selectFrom('user').select('id').where('email', '=', email).executeTakeFirst()
+    return Boolean(row)
+}, z.string())
+
+// Action to check for a pending invite for a user who needs to set up MFA.
+// This is used as a fail-safe in the sign-in flow to prevent users from getting
+// stuck if they sign up via invite but then sign in from a different browser
+// before completing MFA setup.
+export const checkPendingInviteForMfaUserAction = anonAction(async (email: string) => {
```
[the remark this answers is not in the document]
therealmarv: I will remove that part and simplify then.

src/server/actions/user.actions.ts
```diff
@@ -84,3 +89,44 @@ export const updateUserRoleAction = orgAdminAction(
         isReviewer: z.boolean(),
     }),
 )
+
+// Action to let the client know if an email already exists in SI
+export const userExistsAction = anonAction(async (email: string) => {
```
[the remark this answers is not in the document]
therealmarv: This will be removed definitely. Thanks for pointing this out... this should not be here. I used it in past commits I think.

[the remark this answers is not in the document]
therealmarv: The userExistsAction server action is used within the invitation flow to determine if an invited user already has an account. Based on the result, it routes them to either the sign-in page or the account creation
form.

However I've refactored it to not use emails but invite ids for the checkup.

src/server/actions/invite.actions.ts
therealmarv: done

src/server/actions/invite.actions.ts
nathanstitt: yes, but for now we are ignoring the case of a user possibly being in multiple orgs.  We do not want to spend any effort past what we have right now on them, because we are _likely_ to stop using Clerk orgs, as well as the clerk org switcher  altogether very soon once the permissions work progresses further

src/app/account/signin/sign-in-form.tsx
```diff
@@ -44,9 +52,34 @@ export const SignInForm: FC<{
...
+                        form.setErrors({
+                            email: 'To complete your account setup, please use the link from your invitation email.',
+                        })
+                        return
+                    }
+                }
+
+                // If we're here, either the localStorage invite was valid, or there was no pending invite.
+                // It's safe to proceed to MFA.
                 const state = await signInToMFAState(attempt)
```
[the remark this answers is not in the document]
therealmarv: This check handles an edge case: a new user accepts an invitation and creates a password, but quits before setting up Multi-Factor Authentication (MFA).

When they later try to sign in, Clerk requires MFA. Without this check, they would set up MFA, but we would lose the original invitation ID, leaving the invite stuck in a "pending" state forever.

Our code detects this specific scenario. It checks for a pending invite when MFA is required at sign-in. If found, it blocks the login and directs the user back to their invitation link. This ensures the invite is
properly claimed after they complete MFA.

src/app/account/signin/mfa.tsx
```diff
@@ -42,7 +42,10 @@ export const RequestMFA: React.FC<{ mfa: MFAState; onReset: () => void }> = ({ m
             if (signInAttempt?.status === 'complete' && setActive) {
                 await setActive({ session: signInAttempt.createdSessionId })
                 await onUserSignInAction()
-                router.push('/')
+                const pendingInviteId = typeof window !== 'undefined' ? localStorage.getItem('pendingInviteId') : null
+                if (!pendingInviteId) {
+                    router.push('/')
+                }
```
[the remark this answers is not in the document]
therealmarv: Using router.push('/') is a deliberate choice here because it elegantly solves a data synchronization timing issue (between Clerk and Server).

Here's the situation: When a user completes MFA, their client-side session becomes active instantly. However, at the same moment, we trigger a server action (onUserSignInAction) to update their role information in
Clerk's metadata.

The useDashboardUrl hook needs this role information to determine the correct URL. If we called it immediately within the MFA component, it would likely run with stale data before the server action has finished,
causing it to redirect to the wrong place.

By redirecting to `/`, we delegate the final routing decision to our `<UserNav />` component. This component is specifically designed to wait until the authentication state is fully loaded and synchronized (including the
updated roles). Once all the correct data is available on the client, `<UserNav />` reliably calculates the correct dashboard URL and performs the final redirect.

src/server/actions/invite.actions.ts
therealmarv: done

src/app/account/invitation/[inviteId]/invitation-handler.tsx
```diff
@@ -0,0 +1,128 @@
...
+    const dashboardUrl = useDashboardUrl()
+
+    // remember the inviteId so that once they sign in we auto-claim
+    useEffect(() => {
+        window.localStorage.setItem('pendingInviteId', inviteId)
+    }, [inviteId])
+
+    useEffect(() => {
+        if (isLoaded && isSignedIn) {
+            setIsLoadingAction(true)
```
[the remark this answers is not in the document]
Comment (therealmarv) [reply src/app/account/invitation/[inviteId]/invitation-handler.tsx]: if we do that we get two new issues:

 1. A less accurate loading message for existing users. An existing user who is not signed in would see "Processing your invitation..." instead of the more specific "Redirecting to sign-in..." message while they are being redirected.
 2. A broken experience for new users. A new user would get stuck on the "Processing your invitation..." screen. This is because the logic to set the loading state back to false is inside a useEffect that only runs for
   authenticated (isSignedIn) users. A new user is never signed in at this stage, so the loading flag would never be turned off.

src/app/account/invitation/[inviteId]/invitation-handler.tsx
```diff
@@ -0,0 +1,128 @@
...
+            claimInviteAction({ inviteId })
+                .then((result) => {
+                    if (result.success && result.organizationName) {
```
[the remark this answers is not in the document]
Comment (therealmarv) [reply src/app/account/invitation/[inviteId]/invitation-handler.tsx]: yes, that should never happen, it's a safeguard (maybe not needed but will not hurt). In the unlikely case we would need this safeguard and it's not implemented we would see a `"You're now a member of !"` message.

src/app/account/invitation/[inviteId]/new-user-account-form.tsx
Comment (therealmarv) [reply src/app/account/invitation/[inviteId]/new-user-account-form.tsx:78]: The server action now throws an error itself if it doesn't get a user ID.

This simplifies the client code because the onSuccess handler is now only for true successes, and the onError handler automatically catches this failure.

src/app/account/invitation/[inviteId]/page.tsx
Comment (therealmarv) [reply src/app/account/invitation/[inviteId]/page.tsx:22]: this has been an oversight from an old approach. I've removed it now.

src/app/account/invitation/[inviteId]/invitation-handler.tsx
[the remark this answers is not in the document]
Comment (therealmarv) [reply src/app/account/invitation/[inviteId]/invitation-handler.tsx]: I've opted for claimInviteAction using useMutation

src/server/actions/user.actions.ts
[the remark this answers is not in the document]
therealmarv: update: this is still used on signin, I thought I've removed it :/

The problem is the claiming here. Without such a check after or during sign in the claims for an invite could be never fulfilled. I will fix that with hardening this method or move it to the sign in.

PR #244, opened 2025-06-02
Test for validity of key

chrisbendel:


chrisbendel:
@nathanstitt for context this was around [this ticket](https://openstax.atlassian.net/browse/SHRMP-100), we were trying to figure out a way to basically sanity check that the keys were valid before kicking things off

maybe there's some context we were missing around how this came up, but it sounded like maybe something related to a migration? maybe theres a more elegant way to solve this problem too, curious about your thoughts!! thanks for the thoughts on this 🙏

PR #245, opened 2025-06-02
update job review button label spacing to match study review button label spacing

jbwilson8:
[OTTER-134](https://openstax.atlassian.net/browse/OTTER-134)

**Issue**: approve/reject label spacing is not consistent

**Solution**: update job review button label spacing to match study review button label spacing

_Before:_ 
Screenshot 2025-06-02 at 2 44 27 PM

_After:_
Screenshot 2025-06-02 at 2 43 32 PM

[OTTER-134]: https://openstax.atlassian.net/browse/OTTER-134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbwilson8:
@chrisbendel I tried using each of the mantine sizes but none of them satisfied the UX spec. So it was my understaning that if a mantine size doesn't work we should/could use a hardcoded `em` or `rem` value. Otherwise, we'd need to redefine mantine sizes in the `theme.ts` file right? (which I think we decided against).

PR #257, opened 2025-06-07
OTTER-206 - mfa phone input

stetradis:
- Implement phone number component with country code
- Fixes bug where sms is only sent to the first saved number and not what is captured in the form input


src/app/account/mfa/sms/panel.tsx
```diff
@@ -81,10 +87,18 @@ export function ManageSMSMFAPanel() {
...
+                notifications.show({
+                    message: 'Please enter a valid phone number.',
+                    color: 'red',
+                })
+            } else {
+                notifications.show({
+                    message: errorMessage,
+                    color: 'red',
+                })
+            }
```
nathanstitt, line 101: I know you didn't start this, but seems like we should probably add the error to the form so it displays inline vs a notification

PR #270, opened 2025-06-18
Otter 208 log in flow v.2 refactor

Rod0352:
Fix sign-in error handling to show consistent "Invalid login credentials" message for all auth failures instead of different messages for email vs password errors.

### DIFF
--- a/src/app/account/signin/sign-in-form.tsx
+++ b/src/app/account/signin/sign-in-form.tsx
@@ -53,25 +53,16 @@ export const SignInForm: FC<{
         } catch (err: unknown) {
             reportError(err, 'Failed Signin Attempt')
             if (isClerkApiError(err)) {
-                const emailError = err.errors?.find((error) => error.meta?.paramName === 'email_address')
-                if (emailError) {
-                    form.setFieldError('email', emailError.longMessage)
-                    return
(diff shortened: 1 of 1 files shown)


src/app/account/signin/sign-in-form.tsx
```diff
@@ -53,25 +53,16 @@ export const SignInForm: FC<{
...
-
                 const authError = err.errors?.find(
                     (error) =>
                         error.code === 'form_password_incorrect' ||
-                        error.code === 'session_invalid' ||
-                        error.code === 'form_identifier_not_found',
+                        error.code === 'form_param_format_invalid' ||
+                        error.code === 'form_identifier_not_found' ||
+                        error.meta?.paramName === 'email_address' ||
+                        error.meta?.paramName === 'password',
```
nathanstitt, line 62: instead of attempting to enumerate all possible clerk errors, we need to display the error message that clerk sends.  

I'd missed that we werent' doing so before but we need to have a fallback where we do that.  we have a errorToString method in lib/errors.ts that we can use for this.  If we do want to special case any errors, we should move the logic into that method so all code benefits

nathanstitt (approved):

PR #275, opened 2025-06-24
feat: add confirmation modal for study proposal submission

Rod0352:


src/app/researcher/study/request/[orgSlug]/confirmation-modal.tsx
Comment (chrisbendel) [inline src/app/researcher/study/request/[orgSlug]/confirmation-modal.tsx:1]: You know... the more time I spend reading through mantine docs the more I realize they have solved a lot of problems right out of the box 🤣 check this out and see what you think: https://mantine.dev/x/modals/#confirm-modal

maybe this is even simpler 🙃

Comment (Rod0352) [reply src/app/researcher/study/request/[orgSlug]/confirmation-modal.tsx:1]: thanks for the tip chris! this approach helped simplify the study proposal page. I did opt to kept the theme.tsx the same to keep the modal consistency, is this what you had in mind?

Comment (Rod0352) [reply src/app/researcher/study/request/[orgSlug]/confirmation-modal.tsx:1]: Oh, I see! I could definitely call this inline instead of having it in its own file. I was just worried about cluttering that page with 40+ new lines. The additions in the theme keep the rest of the modals the same as the old appmodal component styling, but the styling [here](https://www.figma.com/file/qU4yMlayj7iIpMqlWOX7ee?node-id=7119%3A2741&mode=dev) differs from those slightly, hence the inline styling above.

Comment (chrisbendel) [reply src/app/researcher/study/request/[orgSlug]/confirmation-modal.tsx:1]: @Rod0352 gotcha i wonder if with this new discovery, we could push back on UI/UX a bit on the designs to see if they can be consistent with what the lib offers us, that way we don't have two diverging styles of modals 🤔 hopefully its a non issue on their end and it would keep things consistent for both devs and users

Comment (Rod0352) [reply src/app/researcher/study/request/[orgSlug]/confirmation-modal.tsx:1]: @chrisbendel After confirming with ux i've added more props to the modal theming. I think this newest implementation works with what you had in mind 😁

PR #278, opened 2025-06-25
update icons to have Icon suffix

nathanstitt:
The un-suffixed versions are deprecated

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -12,6 +12,8 @@ const nextConfig: NextConfig = async (phase: string) => {
         output: 'standalone',
         transpilePackages: ['si-encryption'],
         experimental: {
+            // https://github.com/phosphor-icons/react?tab=readme-ov-file#nextjs-specific-optimizations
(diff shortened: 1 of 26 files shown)


next.config.ts
```diff
@@ -12,6 +12,8 @@ const nextConfig: NextConfig = async (phase: string) => {
         output: 'standalone',
         transpilePackages: ['si-encryption'],
         experimental: {
+            // https://github.com/phosphor-icons/react?tab=readme-ov-file#nextjs-specific-optimizations
```
chrisbendel, line 15: Just noticing in the docs linked here the section below: https://github.com/phosphor-icons/react?tab=readme-ov-file#react-server-components-and-ssr

looks like we can potentially go from doing this:
`'@phosphor-icons/react/dist/ssr'`
to this
`'@phosphor-icons/react/ssr'` and remove the `dist`? not sure if that makes a huge difference but might clean things up a bit if it does work 🤔

PR #283, opened 2025-06-27
Prevent dupe results

nathanstitt:
we discovered that if results were recorded twice it puts the app into an invalid state where the earlier results could not be viewed.

https://openstax.slack.com/archives/C07H80W7656/p1751058793960479


nathanstitt:
ug, this is against staging, I'll probably close and cherry just the the commit

nathanstitt:
card is https://openstax.atlassian.net/browse/OTTER-232

chrisbendel (approved):

PR #286, opened 2025-07-01
Otter 224 auto timeout dashboard

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-224

 # Sign-in automatic redirect fix

 ## Problem
 After an automatic sign-out (session idle timeout), entering credentials again signed the user in
 successfully in Clerk, **but the browser remained on `/account/signin` showing a blank blue page**
 until it was refreshed manually.

 ## Root cause
 The sign-in flow did not instruct the Next JS App Router to navigate away from the sign-in route
 after Clerk created the new session, so the client component stayed mounted with stale state.

 ## Fix implemented on this branch
 | Scenario | File & callback | What happens now |
 |----------|-----------------|------------------|
 | Password-only sign-in succeeds | `src/app/account/signin/sign-in-form.tsx` – `onSubmit` | `router.push('/')` after `setActive` + `onUserSignInAction()` |
 | MFA sign-in succeeds           | `src/app/account/signin/mfa.tsx` – `onSuccess`         | Same `router.push('/')` after session activation |

 Because every successful path now ends with `router.push('/')`, the user is immediately redirected
 to the dashboard, eliminating the blank screen and the need for a manual refresh.

### DIFF
--- a/src/app/account/signin/sign-in-form.tsx
+++ b/src/app/account/signin/sign-in-form.tsx
@@ -18,7 +18,7 @@ type SignInFormData = z.infer<typeof signInSchema>
 
 export const SignInForm: FC<{
     mfa: MFAState
-    onComplete: (state: MFAState) => void
+    onComplete: (state: MFAState) => Promise<void>
 }> = ({ mfa, onComplete }) => {
     const { setActive, signIn } = useSignIn()
     const { isSignedIn } = useUser()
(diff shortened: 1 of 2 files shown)


nathanstitt:
the code you have here looks good, but another thought occurs to me.  would it be possible to record the page that a user was on before they were logged out, and then redirect to that page once they sign in again?  not sure if we can record that or not but is worth investigating

nathanstitt (approved):
I'm not sure we'll want to redirect to previous page on things like keys and mfa, but doesn't seem like a bad idea.  Thanks Marvin!

PR #289, opened 2025-07-03
use blob object for download links

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-237

The old code put the entire file's content directly into the download link. This crashed the browser with large files.

The fix changes this. Now, the code creates a file-like object in the browser's memory (a Blob) and generates a short, temporary blob: URL that points to it. This new link is used for the download, which works for much larger files without crashing.

### Difference from data: URLs

* data: URL: The link is the file content. It becomes very long and is limited by the browser (often around 2MB).
* blob: URL (the fix): The link is just a short reference to the file data stored in the browser. The link itself is short, so it avoids the browser's size limit.

New file size limits show be in the several hundred megabytes or gigabytes or limited by users virtual memory (depends on browser).

### DIFF
--- a/src/app/reviewer/[orgSlug]/study/[studyId]/review/study-results.test.tsx
+++ b/src/app/reviewer/[orgSlug]/study/[studyId]/review/study-results.test.tsx
@@ -57,6 +57,16 @@ describe('View Study Results', () => {
     })
 
     it('decrypts and displays the results', async () => {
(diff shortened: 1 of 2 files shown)


chrisbendel (approved):
This will be great to have in #281 thanks!

nathanstitt (approved):

PR #291, opened 2025-07-03
auto refresh researcher dashboard

nathanstitt:
when it has studies that are pending

### DIFF
--- a/src/app/reviewer/[orgSlug]/dashboard/table.tsx
+++ b/src/app/reviewer/[orgSlug]/dashboard/table.tsx
@@ -1,0 +1,108 @@
+'use client'
+
+import { ActionReturnType } from '@/lib/types'
+import dayjs from 'dayjs'
+import { fetchStudiesForOrgAction } from '@/server/actions/study.actions'
+import { DisplayStudyStatus } from '@/components/study/display-study-status'
+import {
+    Divider,
+    Stack,
+    Title,
+    Table,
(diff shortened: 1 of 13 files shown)


src/app/reviewer/[orgSlug]/dashboard/table.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const {
+        data: studies,
+        refetch,
+        isRefetching,
+    } = useQuery({
+        enabled: false,
+        initialData: initialStudies,
+        queryKey: ['org-studies', orgSlug],
+        queryFn: async () => await fetchStudiesForOrgAction({ orgSlug }),
+    })
```
Comment (chrisbendel) [inline src/app/reviewer/[orgSlug]/dashboard/table.tsx:70]: Something like...

```suggestion
    const {
        data: studies,
        refetch,
        isRefetching,
    } = useQuery({
        enabled: false,
        initialData: initialStudies,
        queryKey: ['org-studies', orgSlug],
        queryFn: async () => await fetchStudiesForOrgAction({ orgSlug }),
          refetchInterval: (query) => {
            const hasInitiatedStudies = query.state.data.some(s => FINAL_STATUS.includes(s.latestJobStatus || 'INITIATED')
            if (!hasInitiatedStudies) {
              return false
            }

            // keep polling
            return 1000;
      },
    })
```

i wrote this up in the comment editor so syntax may be off, but i wonder if this covers our use case enough 🤔

PR #292, opened 2025-07-04
OTTER-233 hide reviewer key option from non-reviewer context

stetradis:
Identifies when a user is in a researcher context (when no organization is actively selected) and adjusts their permissions accordingly, ensuring that users only have access to UI elements appropriate for their currently active role. This fixes a bug where users with multiple roles could see the Reviewer Key option from the user profile navbar.

### DIFF
--- a/src/components/org-info.tsx
+++ b/src/components/org-info.tsx
@@ -15,19 +15,34 @@ export function useOrgInfo(orgSlugFallback = '') {
     const { orgSlug: paramOrgSlug } = useParams<{ orgSlug: string }>()
 
     return useMemo(() => {
+        const isSiAdmin = !!user?.organizationMemberships.find(
+            (membership) => membership.organization.slug === CLERK_ADMIN_ORG_SLUG,
+        )
+
         const orgs = user?.publicMetadata?.orgs || []
-        const orgSlug = paramOrgSlug || orgSlugFallback || currentOrgSlug || orgs[0]?.slug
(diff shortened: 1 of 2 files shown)


src/components/org-info.tsx
```diff
@@ -15,19 +15,34 @@ export function useOrgInfo(orgSlugFallback = '') {
...
+                isLoaded,
+                orgSlug: orgs[0]?.slug, // default to the first org if no org slug is provided
+                preferredOrgSlug: null,
+                org: {
+                    isResearcher: true,
+                    isReviewer: false, // ensure a researcher does not have access to reviewer context such as the reviewer key nav option
+                    isAdmin: isSiAdmin,
+                },
+            }
+        }
```
nathanstitt, line 37: I think I see what you're doing here, but am not sure it's a good idea.  What will happen here if a user has been set as only a reviewer, not researcher in the team admin page?   I think this code will lie and say they're a researcher but they'll _hopefully_ get errors as soon as they attempt to do any researcher activities since the server side will check the DB and reject it.

stetradis: I don't think this would affect reviewer roles as a reviewer should always be associated with an active org, right? Is there a use case I'm not considering here?

My testing on this [PR](https://pr292.qa.safeinsights.org/):
https://github.com/user-attachments/assets/5ded15c4-8059-40b5-9d6d-d29b1ba814ce

nathanstitt (approved):

PR #293, opened 2025-07-06
OTTER-240 Bug fix: redirect reviewer to key generation on login

Rod0352:
This change introduces a check after user sign-in to verify if the user is a reviewer and if they possess a public key. If the user is a reviewer but lacks a public key, they are redirected to the `/account/keys` page to generate one. This ensures that reviewers have the required keys to perform their duties effectively.


chrisbendel (approved):
Looks good - not sure if this is relevant for a general approach but we can [redirect](https://nextjs.org/docs/app/api-reference/functions/redirect) in nextjs server actions too, not sure if that cleans it up vs passing the metadata/return values

nathanstitt (approved):

PR #296, opened 2025-07-07
Otter 223 empty org admin list

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-223

### DIFF
--- a/src/server/actions/org.actions.ts
+++ b/src/server/actions/org.actions.ts
@@ -3,6 +3,7 @@
 import { db } from '@/database'
 import { orgSchema } from '@/schema/org'
 import { findOrCreateClerkOrganization } from '../clerk'
+import { findOrCreateOrgMembership } from '../mutations'
 import {
     siAdminAction,
     getUserIdFromActionContext,
@@ -24,6 +25,15 @@ export const upsertOrgAction = siAdminAction(async (org) => {
         if (duplicate) {
             throw new Error('Organization with this name already exists')
         }
+
+        const adminUserId = await getUserIdFromActionContext()
+        await findOrCreateOrgMembership({
+            userId: adminUserId,
+            slug: org.slug,
+            isAdmin: true,
+            isResearcher: false, // by default not a researcher. can be changed later
+            isReviewer: false, // by default not a reviewer. can be changed later
+        })
     }
 
     const results = await db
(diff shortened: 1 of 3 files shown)


src/server/actions/org.actions.ts
```diff
@@ -24,6 +25,15 @@ export const upsertOrgAction = siAdminAction(async (org) => {
...
+
+        const adminUserId = await getUserIdFromActionContext()
+        await findOrCreateOrgMembership({
+            userId: adminUserId,
+            slug: org.slug,
+            isAdmin: true,
+            isResearcher: false, // by default not a researcher. can be changed later
+            isReviewer: false, // by default not a reviewer. can be changed later
+        })
     }
```
nathanstitt, line 37: I don't think we want to set the SiAdmin as a member of newly created orgs, I think what we'll need to do is update the new org form to also ask for a user's information and setup their account here.  That doesn't need to be part of these fixes though

src/server/actions/wrappers.ts
```diff
@@ -252,8 +252,18 @@ export function orgAdminAction<S extends OrgActionSchema, F extends WrappedFunc<
...
+                    userId: user?.id,
+                    orgSlug: arg.orgSlug,
+                    resolvedOrgContext: org,
+                },
+                errorMessage,
+            )
             throw new AccessDeniedError({
-                user: `${user?.id} is not an admin of organization ${arg.orgSlug} ${org.slug}`,
+                user: errorMessage,
             })
```
nathanstitt, line 267: I think we need to throw a ActionFailure here so it can be read on the front-end.  then update the front-end to display it

therealmarv: I will enhance that by tomorrow.

src/server/actions/org.actions.ts
therealmarv: yes, I suspected that. I know it's not directly associated with the Sentry reported problem (which I'm still investigating)

The issue card description is kinda wrong or misleading then. I've also asked [Iris for clarification there](https://openstax.atlassian.net/browse/OTTER-223?focusedCommentId=18981&atlOrigin=eyJpIjoiNjQyYTk0ODE0YmQ1NDc0Njg5NTcxN2M2ZGZiNzZlOTAiLCJwIjoiaiJ9)

therealmarv:
I will close this PR after discussion with Iris.

I think I can pin-point the issue now and also replicate it (testing it now). The issue could be a discrepancy between Clerk and the DB for the org admin role and then this weird issue happens. I will open a new PR.

PR #304, opened 2025-07-11
Implement permissions and sessions w/out Clerk

nathanstitt:
This adds a:
* Session object that is constructed from details stored in clerk user's publicMetadata with per-env keys so staging/qa/prod do not share details
* implements permissions checks using `casl` library

### DIFF
--- a/src/app/account/invitation/[inviteId]/create-account.action.test.ts
+++ b/src/app/account/invitation/[inviteId]/create-account.action.test.ts
@@ -1,17 +1,81 @@
-import { describe, expect, it } from 'vitest'
-import { mockSessionWithTestData } from '@/tests/unit.helpers'
+import { describe, expect, it, vi, beforeEach, type Mock } from 'vitest'
+import { auth as clerkAuth, clerkClient } from '@clerk/nextjs/server'
+import { faker, insertTestOrg, insertTestUser } from '@/tests/unit.helpers'
 import { onCreateAccountAction } from './create-account.action'
 import { db } from '@/database'
+import { v7 } from 'uuid'
 
 describe('Create Account Actions', () => {
+    let org = { id: '', slug: '' }
+
+    beforeEach(async () => {
+        org = await insertTestOrg()
+        const client = clerkClient as unknown as Mock
+        const auth = clerkAuth as unknown as Mock
+        auth.mockResolvedValue({
+            userId: null,
+            sessionClaims: null,
+        })
+
+        client.mockResolvedValue({
+            users: {
+                updateUserMetadata: vi.fn(async () => ({})),
+                getUserList: vi.fn(async () => ({
+                    totalCount: 1,
+                    data: [
+                        {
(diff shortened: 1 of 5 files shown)


src/lib/ability.test.ts
```diff
@@ -0,0 +1,86 @@
...
+            isResearcher: false,
+            isReviewer: false,
+            ...(roles || {}),
+        },
+    }
+    return { ability: defineAbilityFor(session), session }
+}
+
+test('researcher role', () => {
+    const { ability, session } = createAbilty({ isReviewer: true })
```
therealmarv, line 30: test is incorrectly configured to test reviewer's permissions for researcher role. Probably needs other test case name.

src/app/dl/study-code/[jobId]/[fileName]/route.ts
```diff
@@ -9,14 +10,13 @@ export const GET = async (_: Request, { params }: { params: Promise<{ jobId: str
         return NextResponse.json({ error: 'no job id or file name provided' }, { status: 400 })
     }
 
-    const canUserAccessJob = await checkUserAllowedJobView(jobId)
+    const job = await jobInfoForJobId(jobId)
 
-    if (!canUserAccessJob) {
+    const session = await sessionFromClerk()
+    if (!session?.can('read', 'StudyJob')) {
```
Comment (therealmarv) [inline src/app/dl/study-code/[jobId]/[fileName]/route.ts:16]: This is a general check for the ability to read any StudyJob.

However
https://github.com/safeinsights/management-app/blob/facb5277633c987b8c3a232583ba542fec9d78a9/src/lib/permissions.ts#L58 
defines the permission conditionally with a study org id.

Does not the check requires a specific StudyJob subject with an orgId to be evaluated correctly?

src/lib/permissions.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    // action all users can perform
+    permit('update', 'User', { id: session.user.id })
+    permit('claim', 'PendingUser')
+
+    permit('read', 'Team', { orgSlug: session.team.slug })
+    permit('read', 'Team', { id: session.team.id })
+
+    if (isResearcher || isReviewer || isTeamAdmin) {
+        permit('view', 'Study', { orgId: session.team.id })
```
therealmarv, line 41: the use of orgId is different than here:

https://github.com/safeinsights/management-app/blob/facb5277633c987b8c3a232583ba542fec9d78a9/src/lib/permissions.ts#L57

top level `orgID` usage vs `study.orgId` usage. Can this lead to errors?

It seems e.g.:
`getStudyAction` (used by researcher/.../review/page.tsx), the subject passed to the ability check is shaped like `{ study: { orgId: '...' } }` and this check will fail here probably?

src/app/dl/study-code/[jobId]/[fileName]/route.ts
Comment (nathanstitt) [reply src/app/dl/study-code/[jobId]/[fileName]/route.ts:16]: yes it should use the more specific version, updated. ty!

PR #307, opened 2025-07-13
OTTER-252 fix mobile dashboard table style

stetradis:
Added missing table props on the reviewer dashboard table and introduced the `TableScrollContainer` component to enhance table rendering by adding horizontal scrolling for tables with a minimum width of 768px.

### DIFF
--- a/src/app/reviewer/[orgSlug]/dashboard/table.tsx
+++ b/src/app/reviewer/[orgSlug]/dashboard/table.tsx
@@ -80,7 +80,7 @@ export const StudiesTable: FC<{ studies: Studies; orgSlug: string }> = ({ studie
                 <Refresher isEnabled={needsRefreshed} refresh={refetch} isPending={isRefetching} />
             </Flex>
             <Divider c="charcoal.1" />
-            <Table>
+            <Table layout="fixed" verticalSpacing="md" striped="even" highlightOnHover stickyHeader>
                 <TableThead>
                     <TableTr>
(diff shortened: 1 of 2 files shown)


src/app/reviewer/[orgSlug]/dashboard/table.tsx
```diff
@@ -80,7 +80,7 @@ export const StudiesTable: FC<{ studies: Studies; orgSlug: string }> = ({ studie
                 <Refresher isEnabled={needsRefreshed} refresh={refetch} isPending={isRefetching} />
             </Flex>
             <Divider c="charcoal.1" />
-            <Table>
+            <Table layout="fixed" verticalSpacing="md" striped="even" highlightOnHover stickyHeader>
```
Comment (nathanstitt) [inline src/app/reviewer/[orgSlug]/dashboard/table.tsx:83]: doesn't `fixed` layout need the cell widths to be set explicitly and not then not resize based on their contents?  I also think it will clip the contents if they overflow which I don't think we want here.  perhaps we should set the cell contents to `overflow-wrap` ?

Comment (stetradis) [reply src/app/reviewer/[orgSlug]/dashboard/table.tsx]: Turns out that tables always look terrible on mobile 😅 I've opted for wrapping the table in a `TableScrollContainer` element which keeps all the content and styling intact and simply makes the table scrollable at a set breakpoint. 

Found a few examples of this being used here and thought it might be a nicer UX - https://ui.mantine.dev/category/tables/

PR #309, opened 2025-07-15
Otter 248 - Clerk - Must not be able to invite an account to an organization they're already a part of

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-248


nathanstitt (approved):
looks good, thanks @therealmarv

therealmarv:
I will close this because not needed anymore by recent refactorings from Stella. I will do an analysis of the issue in card https://openstax.atlassian.net/browse/OTTER-248 and write my findings in there (I think it's fixed).

PR #311, opened 2025-07-16
Otter 223, allow org editing for SI/super admins even when another org is selected

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-223

## Problem

User is SI/super admin but not org admin.

The "Manage Team" page appeared empty for a super admin if their active organization in the Clerk UI was not the designated admin organization (CLERK_ADMIN_ORG_SLUG).

This occurred because the authorization logic only checked the user's active organization slug from their session to determine super admin status. When a super admin switched away from the admin organization, they lost their universal admin privileges, causing backend actions to fail with an "Access Denied" error and resulting in an empty user table.

## Fix

The authorization logic in the orgAction wrapper was updated to correctly identify super admins regardless of their active organization.

The fix modifies `src/server/actions/wrappers.ts` to:

1. Fetch the user's full list of organization memberships directly from Clerk using clerkClient.
2. Check if the user is a member of the admin organization (CLERK_ADMIN_ORG_SLUG) in any capacity.

If the user is a member of the admin organization, they are granted universal admin rights for the action, ensuring they can always view and manage any organization's team page.

## What is not modified

There is another wrapper, siAdminAction, which protects global, super-admin-only actions (like creating or deleting organizations). This wrapper was not modified and still relies on checking the user's active organization slug to be `SafeInsights (admin)` organization.

### DIFF
--- a/src/server/actions/wrappers.ts
+++ b/src/server/actions/wrappers.ts
@@ -1,5 +1,5 @@
 import logger from '@/lib/logger'
-import { auth as clerkAuth } from '@clerk/nextjs/server'
+import { auth as clerkAuth, clerkClient } from '@clerk/nextjs/server'
 import { CLERK_ADMIN_ORG_SLUG } from '@/lib/types'
 import { UnknownKeysParam, z, ZodObject, ZodString, ZodTypeAny, type Schema } from 'zod'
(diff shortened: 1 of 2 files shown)


therealmarv:
this PR will be closed in favour of https://github.com/safeinsights/management-app/pull/320

PR #312, opened 2025-07-17
Otter 251 all audiences log in flow ux bug fixing

Rod0352:


src/app/account/signin/sign-in-form.tsx
```diff
@@ -69,7 +70,7 @@ export const SignInForm: FC<{
 
             // any other clerk error
             setClerkError({
-                title: 'Sign-in Error',
+                title: errorMessage?.includes('locked') ? 'Account Locked' : 'Sign-in Error',
                 message: errorMessage || 'An error occurred during sign-in. Please try again.',
             })
         }
```
nathanstitt, line 76: a few notes:
* what happens if errorMessage does not include 'password' ?   in that case it seems like it's ignored?
* our `errorToString` func in lib/errors has support for custom messages now by using an object as 2nd argument, suggest we use that but you'll need to find the clerk error code vs string matching.  For an example see the mfa/sms/panel

src/app/account/signin/sign-in-form.tsx
```diff
@@ -94,32 +95,28 @@ export const SignInForm: FC<{
...
                     <Link
+                        c="blue.7"
+                        fw={600}
+                        w="fit-content"
+                        size="xs"
                         href={`/account/reset-password${searchParams.get('redirect_url') ? `?redirect_url=${searchParams.get('redirect_url')}` : ''}`}
                     >
                         Forgot password?
                     </Link>
+                    {clerkError && <SignInError clerkError={clerkError} setClerkError={setClerkError} />}
```
nathanstitt, line 110: instead of wrapping in `{clerkError &&` you can have `SignInError` return null if clerkError is falsy

src/app/account/signin/sign-in-error.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <Stack justify="space-between" gap="xs">
+                <Flex direction="row" justify="space-between" align="flex-start">
+                    <Text ta="left" c="red" fw="bold">
+                        {clerkError.title}
+                    </Text>
+                    <CloseButton c="red" aria-label="Close password reset form" onClick={() => setClerkError(null)} />
+                </Flex>
+                <Text>{clerkError.message}</Text>
+            </Stack>
+        </Paper>
```
nathanstitt, line 39: we have a little used `ErrorAlert` in @/components/errors - lets update it to use these styles and then render it here.

src/app/account/signin/sign-in-form.tsx
Rod0352: if the error does not include "password" or "Couldn't find your account" then its because of something other than incorrect login credentials. the design asks that we show any credential related errors under the password textfield and any other error in the error component. You cant see if here but that IF block on line 62 has a return statement. Will look into that errorToString func, thank you!

PR #321, opened 2025-07-26
OTTER-199 invite flow update

stetradis:
### Enhancements to invitation handling:

* `src/app/account/invitation/[inviteId]/create-account.action.ts`: Added logic to mark an invitation as claimed by a user, preventing it from appearing in pending lists.
* `src/app/account/invitation/[inviteId]/page.tsx`: Added checks to detect claimed invitations and redirect users to the sign-in page if applicable. Updated variable names for clarity.

### Improvements to user experience:

* `src/app/account/invitation/[inviteId]/join-team/page.tsx`: Simplified the flow by removing the `hasJoined` state and automatically joining the team when the organization is loaded. Enhanced error handling by redirecting users who are already team members and refining error messages. Updated UI to show a confirmation message and a button to visit the dashboard.

### Data filtering improvements:

* `src/app/admin/team/[orgSlug]/admin-users.actions.ts`: Added a filter to exclude claimed invitations when fetching pending users for the admin view.


src/app/account/invitation/[inviteId]/create-account.action.ts
```diff
@@ -89,6 +89,14 @@ export const onJoinTeamAccountAction = new Action('onJoinTeamAccountAction')
...
         onUserAcceptInvite(siUser.id)
 
+        // mark invite as claimed by this user so it no longer shows in pending lists
+        await db
+            .updateTable('pendingUser')
+            .set({ claimedByUserId: siUser.id })
+            .where('id', '=', inviteId)
+            .where('claimedByUserId', 'is', null)
+            .executeTakeFirst()
+
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/create-account.action.ts:99]: 🎉 I was hoping it was something simple like this that was missed

Comment (nathanstitt) [reply src/app/account/invitation/[inviteId]/create-account.action.ts:99]: the same thing probably need to happen in the other two flows, right?  the "added to other org" and "already a member" ones?  We should de-active the invite in all those cases.  Maybe they're already doing this though

src/app/account/invitation/[inviteId]/join-team/page.tsx
```diff
@@ -24,30 +23,33 @@ const AddTeam: FC<InviteProps> = ({ params }) => {
 
     const { mutate: joinTeam, isPending: isJoining } = useMutation({
         mutationFn: () => onJoinTeamAccountAction({ inviteId }),
-        onError: reportMutationError('Unable to join team'),
-        onSuccess() {
-            setHasJoined(true)
+        onError: (err) => {
+            const failure = extractActionFailure(err)
+            if (failure?.team === 'already a member') {
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/join-team/page.tsx:28]: matching exact strings like this makes me nervous. Is it possible to throw a different object with a code that's safer to match?

src/app/account/invitation/[inviteId]/join-team/page.tsx
```diff
@@ -24,30 +23,33 @@ const AddTeam: FC<InviteProps> = ({ params }) => {
...
+            reportMutationError(failure?.team ?? 'Unable to join team')
         },
     })
 
+    useEffect(() => {
+        if (org && !isLoading) {
+            joinTeam()
+        }
+    }, [org, isLoading]) // eslint-disable-line react-hooks/exhaustive-deps
+
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/join-team/page.tsx:41]: I do not think we should join a org automatically, especially since we don't have a way for users to leave an org.  A user may not want to be associated with an organization at all.  That said, I don't care that much.  Are there any requests/mockups saying we should do this?

src/app/account/invitation/[inviteId]/page.tsx
```diff
@@ -27,13 +26,20 @@ export default async function AcceptInvitePage({ params }: { params: Promise<{ i
...
-        return <AlertNotFound title="Unable to signup" message="unable to find invitation" />
+    const claimedInvite = await db
+        .selectFrom('pendingUser')
+        .where('claimedByUserId', 'is not', null)
+        .where('id', '=', inviteId)
+        .executeTakeFirst()
+
+    // if the invite has been claimed by a user, redirect to the signin page
+    if (claimedInvite) {
+        redirect(`/account/signin`, RedirectType.replace)
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/page.tsx:37]: should we show an error here?  We probably don't want to say "invite has been used ", but could just say "Invite not found" like the earlier message

src/app/account/invitation/[inviteId]/join-team/page.tsx
Comment (stetradis) [reply src/app/account/invitation/[inviteId]/join-team/page.tsx:41]: Yep! This was outlined in the ticket description [here](https://openstax.atlassian.net/jira/software/projects/OTTER/boards/24?selectedIssue=OTTER-199#:~:text=Automatically%20add%20them%20to%20the%20new%20organization%20in%20the%20background%20AND%20remove%20them%20from%20the%20%22Pending%20Invitations%22%20list). 

Happy to change it if this isn't the correct approach!

Comment (iccgoncalves) [reply src/app/account/invitation/[inviteId]/join-team/page.tsx:41]: Thanks, @nathanstitt !

You're right - a user should only join an organization when they actively accept the invite. We're currently working with UX on designing an invitation modal that appears after the user clicks the email link. The goal is to give them a clear point of action to either accept or reject the invitation before they're added to the organization.

src/app/account/invitation/[inviteId]/create-account.action.ts
Comment (stetradis) [reply src/app/account/invitation/[inviteId]/create-account.action.ts:99]: I believe this fix should cover all cases:

'new user invite' - already does this
'already a member, invite to new org' - this case
'already a member, invited to same org' - I think this case will be cleaned up by [OTTER-270](https://openstax.atlassian.net/browse/OTTER-270) where we won't allow invites to be sent to members who are already a part of the org. I can take that one on next

src/app/account/invitation/[inviteId]/join-team/page.tsx
Comment (stetradis) [reply src/app/account/invitation/[inviteId]/join-team/page.tsx]: removed the string match here as a temp fix but I'll be cleaning this up in the follow up [ticket](https://openstax.atlassian.net/browse/OTTER-270) which prevents sending invites to users who are already members of the org

PR #325, opened 2025-07-30
Fix admin updating organizations

nathanstitt:
This was potentially a pretty bad bug.  Luckily it would always fail because the id had a unique index and an update would create dups.  If it wasn't for that this would be updating all rows 😭


stetradis (approved):
Good catch! Just needs a lint fix

PR #328, opened 2025-08-02
OTTER-270 prevent invites to same org

stetradis:
### Enhancements to User Invitation Workflow

*  Added a check to prevent inviting users who are already members of the organization, throwing an appropriate error if detected. 
* Improved response handling for already-invited users by returning a flag (`alreadyInvited`) and resending the invitation email when applicable.
* Simplified the logic by removing redundant checks for `matchingTeam`, handling this upstream by preventing invites.
* Removed redundant error handling logic for joining a team.
* Removed the `onRevokeInviteAction` and its handler logic, which is no longer used.


src/app/account/invitation/[inviteId]/create-account.action.ts
```diff
@@ -30,16 +30,6 @@ export const getOrgInfoForInviteAction = new Action('getOrgInfoForInviteAction')
...
-export const onRevokeInviteAction = new Action('onRevokeInviteAction')
-    .params(
-        z.object({
-            inviteId: z.string(),
-        }),
-    )
-    .handler(async function ({ params: { inviteId }, db }) {
-        await db.deleteFrom('pendingUser').where('id', '=', inviteId).executeTakeFirstOrThrow()
-    })
-
```
Comment (nathanstitt) [inline src/app/account/invitation/[inviteId]/create-account.action.ts:42]: I wasn't aware we had this, looks like it wasn't connected to the UI for some reason?

Comment (stetradis) [reply src/app/account/invitation/[inviteId]/create-account.action.ts:42]: Should we add this? Maybe an X button in the pending users list?

nathanstitt (approved):

PR #329, opened 2025-08-03
OTTER-276 update researcher job errored study status

stetradis:
* Researchers no longer see "Job Errored" status before file review is completed; the status falls back to the study status until files are approved or rejected.
* Added a test to confirm that researchers do not see the errored status before file review, and the status defaults to study status.

### DIFF
--- a/src/app/researcher/study/[studyId]/review/job-results-status-message.tsx
+++ b/src/app/researcher/study/[studyId]/review/job-results-status-message.tsx
@@ -52,18 +52,14 @@ const Errored: FC<ErroredProps> = ({ isApproved, isRejected, jobId }) => {
     }
 
     return (
-        <>
-            {message && (
-                <Stack>
-                    <Text>{message}</Text>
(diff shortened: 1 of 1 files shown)


src/app/researcher/study/[studyId]/review/job-results-status-message.tsx
```diff
@@ -46,15 +46,24 @@ const Errored: FC<ErroredProps> = ({ isApproved, isRejected, jobId }) => {
...
     return (
         <>
-            {message && <Text>{message}</Text>}
-            <Group justify="flex-start" align="center">
-                <Text size="xs" fw="bold">
-                    Job ID:
-                </Text>
-                <CopyingInput value={jobId} tooltipLabel="Copy" />
-            </Group>
+            {message && (
```
Comment (nathanstitt) [inline src/app/researcher/study/[studyId]/review/job-results-status-message.tsx:56]: you know message is defined here because otherwise it would have returned earlier

nathanstitt (approved):
looks good, except don't think we need the `message &&` bit ;)

PR #333, opened 2025-08-06
Otter 213 improvements

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-213

improvement branch:
* make the checks more fail safe
* add tests for accounts who do not have second factor set up
* do not use `@clerk/clerk-react` imports, use `@clerk/nextjs` !
* add `ActivityContext` for account pages

### DIFF
--- a/src/components/layout/focused-layout-shell.test.tsx
+++ b/src/components/layout/focused-layout-shell.test.tsx
@@ -1,19 +1,12 @@
-import { it, vi } from 'vitest'
+import { it } from 'vitest'
 import { FocusedLayoutShell } from './focused-layout-shell'
 import { render, screen } from '@testing-library/react'
-import { useClerk, useSession } from '@clerk/nextjs'
+import { mockClerkSession } from '@/tests/unit.helpers'
 
 import { TestingProviders } from '@/tests/providers'
 
 it('renders without a user', async () => {
-    vi.mocked(useSession).mockReturnValue({
-        session: null,
-        isLoaded: true,
-        isSignedIn: false,
-    })
-    vi.mocked(useClerk).mockReturnValue({
-        signOut: vi.fn(),
-    } as unknown as ReturnType<typeof useClerk>)
+    mockClerkSession(null)
 
     render(
         <FocusedLayoutShell>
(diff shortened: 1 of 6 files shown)


src/components/layout/focused-layout-shell.test.tsx
```diff
@@ -1,10 +1,20 @@
...
 it('renders without a user', async () => {
+    vi.mocked(useSession).mockReturnValue({
+        session: null,
+        isLoaded: true,
+        isSignedIn: false,
+    })
+    vi.mocked(useClerk).mockReturnValue({
+        signOut: vi.fn(),
+    } as unknown as ReturnType<typeof useClerk>)
+
```
nathanstitt, line 17: lets add these to the existing clerk mocks in `tests/unit.helpers.jsx` and then call `mockClerkSession()` here

src/components/require-mfa.test.tsx
```diff
@@ -1,25 +1,57 @@
...
-    router.setCurrentUrl('/')
-})
+describe('RequireMFA', () => {
+    describe('when MFA is not enabled', () => {
+        beforeEach(() => {
+            router.setCurrentUrl('/')
+        })
+
+        it('redirects if mfa is not set', async () => {
+            ;(useUser as Mock).mockImplementation(() => ({ user: { twoFactorEnabled: false } }))
```
nathanstitt, line 19: another advantage of mockClerkSession is that it returns the mocks, so this could then call mockImplementation on them

src/components/require-mfa.tsx
```diff
@@ -10,7 +10,7 @@ export const RequireMFA = () => {
     const router = useRouter()
 
     useLayoutEffect(() => {
-        if (user?.twoFactorEnabled === false && !pathname.startsWith('/account/mfa')) {
+        if (user?.twoFactorEnabled !== true && !pathname.startsWith('/account/mfa')) {
```
nathanstitt, line 13: I think this will redirect while clerk is loading, in that case user is null and there's a `isLoading` returned from useUser.  That's why we were previously testing that it was `false`.  https://clerk.com/docs/nextjs/hooks/use-user

nathanstitt (commented):
looks nice Marvin, the specs are good to have.  I think we do need to either check isLoading and remove the `.?` or leave the check on twoFactorEnabled to test for false.

nathanstitt (approved):

PR #334, opened 2025-08-07
feat: Add test image support and deletion safeguards for base images

Rod0352:
Changes:

Test Image Support: Org admins can now choose to run approved studies against test base images via a checkbox on the study review page
Deletion Protection: Prevents deletion of the last non-test base image to ensure orgs always have at least one production image

### DIFF
--- a/src/components/surety-guard.tsx
+++ b/src/components/surety-guard.tsx
@@ -7,7 +7,8 @@ export const SuretyGuard: FC<{
     children?: React.ReactNode
     message?: string
     onConfirmed: () => void | Promise<void>
-}> = ({ message, children, onConfirmed }) => {
+    blockAction?: boolean
+}> = ({ message, children, onConfirmed, blockAction }) => {
     const [opened, { close, open }] = useDisclosure(false)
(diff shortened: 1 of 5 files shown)


src/components/surety-guard.tsx
```diff
@@ -21,7 +22,7 @@ export const SuretyGuard: FC<{
     return (
         <Popover width={300} trapFocus withArrow shadow="md" opened={opened} onChange={close} closeOnClickOutside>
             <Popover.Target>
-                <ActionIcon size="sm" variant="subtle" color="red" onClick={open} disabled={opened}>
+                <ActionIcon size="sm" variant="subtle" color="red" onClick={open} disabled={opened || blockAction}>
```
nathanstitt, line 25: the `blockAction` prop seems superfluous to me.  Instead of relying on the SuretyGuard to not call the functions we provided, we can either not render it, or render it but have our function be a no-op. 

Addendum: as I was typing this, I realized that I guess there is a case for wanting to render the button in a disabled state.  Maybe we rename the property to be `disabled`?

PR #336, opened 2025-08-10
OTTER-273 add invite permissions

stetradis:
This pull request introduces support for assigning admin permissions to invited users. It adds a new `isAdmin` property to pending users, updates the invitation flow to handle permissions, and enhances the invite form to allow admins to specify permissions when inviting new users. The changes are covered by updates to the database schema, backend logic, and tests.

**Database and Data Model Updates:**

* Added a new `is_admin` column to the `pending_user` table with a default of `false` and updated the `PendingUser` interface to include the `isAdmin` property. 

**Backend Logic:**

* Updated account creation actions to set `isAdmin` based on the invitation, rather than always defaulting to `false`. 
* Modified the admin invite user action to set `isAdmin` according to the new `permission` field in the invite. 

**Frontend and Validation:**

* Enhanced the invite user schema and form to include a `permission` field, allowing selection between "contributor" and "admin" roles. 

**Testing:**

* Updated tests to cover the new `isAdmin` property and permission logic for invitations.

### DIFF
--- a/src/app/account/invitation/[inviteId]/create-account.action.test.ts
+++ b/src/app/account/invitation/[inviteId]/create-account.action.test.ts
@@ -57,6 +57,7 @@ describe('Create Account Actions', () => {
                 email: faker.internet.email({ provider: 'test.com' }),
                 isResearcher: true,
                 isReviewer: true,
+                isAdmin: false,
             })
             .returningAll()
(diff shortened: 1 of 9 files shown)


nathanstitt (approved):
Looks great, ty!

PR #345, opened 2025-08-21
OTTER-270 fix invite error message

stetradis:
- Removes the thrown `ActionFailure` error from the server component. Next.js appears to mask this custom error in production builds, showing the default 'An error occurred' toast instead of the inline form error. 

See my confirmation below, using 2 PR builds to show the current error and the fix that this PR introduces: 

https://github.com/user-attachments/assets/d3da541a-766e-4309-97ab-86129aaea285


stetradis:
Closing in favor of global fix

PR #347, opened 2025-08-21
Use all status changes to determine labels

nathanstitt:
statuses can arrive out of order or be repeated by the TOA

Hopefully fixes https://openstax.atlassian.net/browse/OTTER-271


stetradis:
@nathanstitt pushed a small fix for reviewer studies. The JOB-ERRORED status doesn't seem to be displaying, looking into it

nathanstitt:
Thanks for your help on this @stetradis !

PR #350, opened 2025-08-24
allow keyboard navigation for user profile menu

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-287

### DIFF
--- a/src/components/layout/navbar-profile-menu.tsx
+++ b/src/components/layout/navbar-profile-menu.tsx
@@ -18,7 +18,7 @@ export function NavbarProfileMenu() {
     const [opened, { toggle, close }] = useDisclosure()
     const router = useRouter()
     const menuRef = useClickOutside<HTMLDivElement>(() => opened && close())
-    const firstMenuItemRef = useRef<HTMLDivElement>(null)
+    const firstMenuItemRef = useRef<HTMLAnchorElement>(null)
 
     const closeAndCall = (fn: () => void) => () => {
         fn()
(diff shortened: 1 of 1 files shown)


src/components/layout/navbar-profile-menu.tsx
```diff
@@ -18,7 +18,7 @@ export function NavbarProfileMenu() {
     const [opened, { toggle, close }] = useDisclosure()
     const router = useRouter()
     const menuRef = useClickOutside<HTMLDivElement>(() => opened && close())
-    const firstMenuItemRef = useRef<HTMLDivElement>(null)
+    const firstMenuItemRef = useRef<HTMLButtonElement>(null)
 
     const closeAndCall = (fn: () => void) => () => {
```
nathanstitt, line 23: it's not clear to me why you added all the stopPropagation below, but a cleaner way to do so would be in the closeAndCall.  If the reason is to prevent nav, you also should check that the user didn't right click and open in new tab.  the typical way is to:

```suggestion
    const closeAndCall = (fn: () => void) => (e: React.MouseEvent) => {
         // if any of these are the case then the user likely is attempting to open in new tab
        if (!(event.button === 1 || event.ctrlKey || event.metaKey || event.shiftKey)) {
            e.stopPropagation() // stopPropagation to prevent <reason>
        }
```

src/components/layout/org-admin-dashboard-link.tsx
```diff
@@ -80,6 +75,7 @@ export const OrgAdminDashboardLink: FC<OrgAdminDashboardLinkProps> = ({ isVisibl
                     component={Link}
                     href={`${orgAdminBaseUrl}/settings`}
                     active={pathname === `${orgAdminBaseUrl}/settings`}
+                    aria-current={pathname === `${orgAdminBaseUrl}/settings` ? 'page' : undefined}
```
nathanstitt, line 78: https://mantine.dev/core/nav-link/#active says that the pathname should be taken into account and active be set automatically, is that not occurring?

nathanstitt (commented):
a few suggestions to DRY it up a bit

src/components/layout/navbar-profile-menu.tsx
therealmarv: Good point. I refactored closeAndCall to include e.stopPropagation(), which is needed to prevent the click from bubbling up and re-toggling the menu. This cleans up the repeated code.

The modifier key check isn't needed to my understanding because these are buttons, not links, so there's no 'open in new tab' browser action to preserve.

PR #351, opened 2025-08-24
Otter 289 show tooltips on keyboard selection too

therealmarv:
see https://openstax.atlassian.net/browse/OTTER-289

* The InfoIcon component has been refactored to use forwardRef, allowing it to correctly receive focus.
* The Tooltip component has been configured to trigger on focus events, ensuring it displays for both mouse hover and keyboard navigation.
* An aria-label was added to the icon for better screen reader support.

### DIFF
--- a/src/components/icons.tsx
+++ b/src/components/icons.tsx
@@ -2,20 +2,21 @@
 
 import { InfoIcon as SVGIcon } from '@phosphor-icons/react/dist/ssr'
 import { ActionIcon, ActionIconProps, useMantineTheme } from '@mantine/core'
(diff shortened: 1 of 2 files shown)


therealmarv:
The ref is being used implicitly. The Mantine `<Tooltip>` component needs to pass a ref to its child for positioning. The change to InfoIcon uses forwardRef to accept this implicit ref and forward it to the underlying `<ActionIcon>`. This is the standard React pattern and allows for the removal of the non-standard `refProp="innerRef"` from the `<Tooltip>`.

PR #352, opened 2025-08-25
Enforce Auto Refresh on Status Change - OTTER-294

Rod0352:
This PR ensures the dashboard automatically reflects status changes, eliminating the need for a manual refresh.

The Problem:
After a user performed a status-changing action (e.g., approving a study), the dashboard would navigate but show stale data until the page was manually refreshed.

Solution:
client-side cache invalidation strategy using tanstack-query (react-query).

1. From Server Components to Client-Side Queries: The dashboards, which were previously fetching data on the server, have been refactored. They now contain client-side components that use the useQuery hook to fetch and manage their own data.

2. Cache Invalidation on Mutation: The components that trigger status changes (e.g., the "Approve" and "Submit" buttons) now use queryClient.invalidateQueries in their onSuccess handlers.

The Flow:
   1. A user performs an action (e.g., approves a study).
   2. The mutation succeeds.
   3. The onSuccess callback invalidates the query key associated with the dashboard's data.
   4. The user is navigated to the dashboard via router.push.
   5. When the dashboard mounts, tanstack-query sees the invalidated query and automatically refetches the data, ensuring the user sees the latest status.


src/app/researcher/study/request/[orgSlug]/study-proposal.tsx
```diff
@@ -154,6 +154,7 @@ export const StudyProposal: React.FC<{ orgSlug: string }> = ({ orgSlug }) => {
                 color: 'green',
             })
             router.push(`/researcher/dashboard`)
+            window.location.assign(`/researcher/dashboard`)
```
Comment (nathanstitt) [inline src/app/researcher/study/request/[orgSlug]/study-proposal.tsx:157]: ???  I don't think we want to do this, but if we did, we should remove the router.push before it

src/app/reviewer/[orgSlug]/study/[studyId]/review/job-review-buttons.tsx
```diff
@@ -45,7 +44,7 @@ export const JobReviewButtons = ({
         },
         onError: reportMutationError('Failed to update study job status'),
         onSuccess: () => {
-            router.push('/')
+            window.location.assign(`/reviewer/${orgSlug}/dashboard`)
```
Comment (nathanstitt) [inline src/app/reviewer/[orgSlug]/study/[studyId]/review/job-review-buttons.tsx:47]: again, don't think we should perform raw navigation, router.push will be much faster

src/app/reviewer/[orgSlug]/study/[studyId]/review/study-review-buttons.tsx
```diff
@@ -35,7 +34,7 @@ export const StudyReviewButtons: FC<{ study: SelectedStudy }> = ({ study }) => {
             return rejectStudyProposalAction({ orgSlug, studyId: study.id })
         },
         onError: reportMutationError('Failed to update study status'),
-        onSuccess: () => router.push(backPath),
+        onSuccess: () => window.location.assign(backPath),
```
Comment (nathanstitt) [inline src/app/reviewer/[orgSlug]/study/[studyId]/review/study-review-buttons.tsx:37]: sorry I'm pretty confused, why did you add all these?

nathanstitt (changes requested):
I don't understand what you're doing here?  I'm guessing it's an attempt to clear the browser state?

Are we:
  * calling revalidatePath in the action
  * using a queryKey on whatever useQuery that you're trying to clear that gets invalidated?  Or you can import queryClient and manually clear it as well if that's the case

PR #371, opened 2025-09-14
OTTER-222 add custom error boundary and not found pages

stetradis:
This pull request introduces a global error boundary to the application, improving error handling and user experience. It also adds dedicated pages for both error testing and "not found" scenarios, ensuring users receive clear feedback when issues arise.

**Error Handling Improvements:**

* Added a new `ErrorBoundary` class component in `src/components/error-boundary.tsx` that catches rendering errors, reports them to Sentry, and displays a user-friendly error page with options to refresh or return home.
* Wrapped all children in the `Providers` component with the new `ErrorBoundary`, making error handling consistent across the app.

**User Experience Enhancements:**

* Created an error demo page (`src/app/error-demo/page.tsx`) for testing the error boundary by intentionally throwing an error on mount.
* Implemented a custom "not found" page in `src/app/not-found.tsx` with improved design and navigation options for users who access invalid URLs.


nathanstitt (approved):
looks good @stetradis,    can you update the packages to fix the `trivy` failure as well?

nathanstitt (changes requested):
Hey @stetradis I'm glad I checked the Jenkins build because it's actually failing due to issues with the PR for once 😭 

The new not_found page appears to be loading pages that are using clerk, which we can't do.  Doing so will embed the clerk key into the build which will work fine on QA+staging, but then fail disastrously on prod because prod uses a different clerk account and key.

This is why only the (auth) layout uses clerk and the other pages at the root do not.  Can we move it up there?  
I think you can replicate the build failure by removing any .env files you have locally and running `nom run build`
https://jenkins.security.safeinsights.org:8443/job/safeinsights/job/management-app/job/PR-371/5/pipeline-overview/?start-byte=0&selected-node=22#log-298

stetradis:
@nathanstitt updated to remove the clerk code and only keep the UI elements

PR #629, opened 2026-04-08
fix date for old studies

jbwilson8:
## Summary

Falls back to `createdAt` when `submittedAt` is null in the "Submitted On" column, so older studies submitted before the migration still display a date instead of `-`.


nathanstitt:
thanks @jbwilson8 but the failure on QA was caused by the migration failing due to the dupe migration filename.  I fixed it manually and QA looks fine now.

PR #631, opened 2026-04-09
Otter 413 agreements page bugfix

Rod0352:
fixed: The redirect was incorrectly firing for all post-submission states (code approved, errored, results ready) because CODE-SUBMITTED is always in the job history. The gate now only applies when code hasn't been reviewed yet (!codeAlreadyReviewed).


src/app/[orgSlug]/study/[studyId]/agreements/page.tsx
```diff
@@ -27,8 +27,14 @@ export default async function StudyAgreementsRoute(props: { params: Promise<{ or
...
-        if (!codeSubmitted) {
+        const statuses = study.jobStatusChanges.map((s) => s.status)
+        const latestSubmittedIdx = statuses.findIndex((s) => s === 'CODE-SUBMITTED')
+        const latestReviewedIdx = statuses.findIndex((s) => s === 'CODE-APPROVED' || s === 'CODE-REJECTED')
+        // statuses is ordered newest-first; lower index = more recent
+        // Show agreements only when code is awaiting review (submitted more recently than any review decision)
+        const awaitingCodeReview =
+            latestSubmittedIdx !== -1 && (latestReviewedIdx === -1 || latestSubmittedIdx < latestReviewedIdx)
+        if (!awaitingCodeReview) {
             redirect(Routes.studyReview({ orgSlug, studyId }))
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/agreements/page.tsx:47]: This logic isn't quite correct.   All we care about is if the job has status of 'CODE-SUBMITTED' and has not been approved or rejected.

if the user re-submits code, that will be a new job and a fresh set of statuses.  In that case we need to check if the job is not the first one and not re-display agreements page.

nathanstitt (changes requested):

nathanstitt:
I'm sorry @Rod0352 I got confused and somehow opened a new pr with this one's commits + some more fixes in #636 

I'll close this one in favor of that

PR #632, opened 2026-04-09
Fix generating invalid Coder names with multiple `-`

rnathuji:


nathanstitt (approved):
LGTM, thanks @rnathuji

PR #633, opened 2026-04-10
OTTER-382-default-dashboartd-bug -fixes

Rod0352:
Fix default dashboard view and breadcrumb navigation for dual-role users

Dual-role users (researcher + reviewer) always landed on the Reviewer tab on My Dashboard, and study breadcrumbs didn't return them to the correct dashboard. Both are now fixed — the default tab is Researcher for dual-role users, org dashboards pass audience context back to My Dashboard, and studies opened from an org dashboard breadcrumb back to that org rather than My Dashboard.


nathanstitt (approved):
thanks for your work on this, @Rod0352 - I think at this point we should not use only a /dashboard url, it's getting to complicated to determine the users intent and we're fighting to indicate it with query params.  I'm going to approve this so we can move forward and also make a card to note the confusion and suggest we only use the per-role dashboards

PR #639, opened 2026-04-16
Enhance ProposalSection with orgSlug prop and add tests 

Rod0352:
Summary

Wires ProposalReviewFeatureFlag into review/page.tsx so OpenStax orgs in spy mode see the new NewProposalReviewView; everyone else keeps the legacy ProposalReviewView. Completes the OTTER-488 scaffold's flag integration.
Enhances the two PI/Researcher ProposalSection tests to actually hover and assert the popover opens, per OTTER-490's "info icons trigger popovers" criterion.


src/app/[orgSlug]/study/[studyId]/review/page.test.tsx
```diff
@@ -11,6 +11,8 @@ import {
 } from '@/tests/unit.helpers'
 import StudyReviewPage from './page'
 import { CodeReviewView } from './code-review-view'
+import { NewProposalReviewView } from './new-proposal-review-view'
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/page.test.tsx]: lets not name things `New`, very soon this will not be the "new" flow, it'll just be the ProposalReview.  if we can't find better names, lets name the pre-existing one 'Old' to free up the name ProposalReviewView

Personally, I also dislike tacking View on the end but that's a pre-existing issue

src/app/[orgSlug]/study/[studyId]/review/proposal-section.tsx
```diff
@@ -1,17 +1,187 @@
...
+    orgSlug,
+    isExpanded,
+    onToggle,
+    getPopoverProps,
+}: {
+    study: StudyForReview
+    orgSlug: string
+    isExpanded: boolean
+    onToggle: () => void
+    getPopoverProps: ReturnType<typeof usePopover>['getPopoverProps']
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/proposal-section.tsx]: Nit: minor prop drilling — `isExpanded`, `onToggle`, and `getPopoverProps` are all threaded from `ProposalSection` (see line 172) down through `SectionHeader` and `ProposalBody`. Since both subcomponents live in this same file and are always rendered together, `ProposalSection` could render them without passing these through — each could call `useProposalSection()` itself, or the whole section could be one component. Not a blocker, just noise in the signatures.

src/app/[orgSlug]/study/[studyId]/review/proposal-section.tsx
```diff
@@ -1,17 +1,187 @@
...
+    if (!date) return null
+    return dayjs(date).format('MMM DD, YYYY')
+}
+
+function ToggleLink({ isExpanded, onToggle }: { isExpanded: boolean; onToggle: () => void }) {
+    const label = isExpanded ? 'Hide full initial request' : 'Show full initial request'
+    const Icon = isExpanded ? CaretUpIcon : CaretDownIcon
+
+    return (
+        <Anchor component="button" onClick={onToggle} c="blue" size="sm" data-testid="proposal-toggle">
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/proposal-section.tsx]: `ToggleLink` is rendered twice (once in `SectionHeader`, once in `ProposalBody`), so two DOM nodes end up with `data-testid="proposal-toggle"`. The tests rely on `getAllByTestId(...)[0]` to pick the header toggle — works, but fragile if someone reorders the render. Consider accepting a `testId` prop here (or `location: "header" | "body"`) so callers can assign distinct ids like `proposal-toggle-header` / `proposal-toggle-body`.

nathanstitt (approved):
looks good, just a few minor comments about names 😁

PR #640, opened 2026-04-17
OTTER-417 temp fix, do not check cert for now

therealmarv:
Disable strict TLS cert verification on the RDS connection as a hotfix, because enabling `ssl: true` started rejecting AWS's RDS CA which Node doesn't trust by default.
Traffic stays encrypted; proper CA bundling will follow.


nathanstitt (approved):
lgtm, lets give it a try. thank you!

PR #641, opened 2026-04-17
OTTER 492 Flag - DO - Proposal Review - Decision radio group, submit button logic & respective modals

therealmarv:
see [OTTER-492](https://openstax.atlassian.net/browse/OTTER-492)

- Added a decision radio group (Approve, Needs clarification, Reject) with instructional text showing the submitting lab's name.
- Wired the submit button with feedback-and-decision gating, plus confirmation modals for approve/needs-clarification and a destructive reject modal.
- Connected approve and reject to existing server actions with redirect on success; needs-clarification is a placeholder pending backend work.
- Kept the "Proceed to Step 2" path intact when reviewers return from the agreements step.
- Added unit coverage for the new flow, modals, gating, status changes, and double-submit prevention.

[OTTER-492]: https://openstax.atlassian.net/browse/OTTER-492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


therealmarv:
this is ready for review. But: I would feel more comfortable to merge this after OTTER-491 (there are some touching points). However this PR stands for it own in this form.

src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx
```diff
@@ -1,11 +1,19 @@
 'use client'
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx]: lets not name files/components as `new`. someday soon this will not be the new flow, it'll be the standard flow and then the name won't fit.

if we can't think of better names, lets rename the old one as `old` and re-use the name for this.  That's fine because the "old" one will be removed upon removal of the feature flag

src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx
```diff
@@ -17,56 +25,223 @@ type NewProposalReviewViewProps = {
...
     const router = useRouter()
+    const queryClient = useQueryClient()
+    const [confirmOpen, { open: openConfirm, close: closeConfirm }] = useDisclosure(false)
+    const [rejectOpen, { open: openReject, close: closeReject }] = useDisclosure(false)
 
     const canSubmit = feedback.isValid && decision.selected !== null
+    const backPath = Routes.orgDashboard({ orgSlug })
+
+    const { mutate: submitReview, isPending } = useMutation({
+        mutationFn: async () => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx]: This `useMutation` block (mutationFn / onSuccess / onError / `router.push(backPath)` / `invalidateQueries(['org-studies', orgSlug])`) is a near-verbatim copy of `src/hooks/use-proposal-review-mutation.ts`, which is still in use by the old `ProposalReviewView`. The meaningful differences are only: (a) branching on `decision.selected` strings vs. taking a `StudyStatus` arg, and (b) the needs-clarification placeholder.

Suggest extending the existing hook (or extracting a shared one) so the new view composes it instead of rebuilding. Otherwise a future change to redirect or error handling has two places to update and they will drift.

Not blocking — but worth addressing before this view replaces the old one.

src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx
```diff
@@ -17,56 +25,223 @@ type NewProposalReviewViewProps = {
...
+            if (decision.selected === 'approve') {
+                return approveStudyProposalAction({ orgSlug, studyId })
+            }
+            if (decision.selected === 'reject') {
+                return rejectStudyProposalAction({ orgSlug, studyId })
+            }
+            // TODO(OTTER-493): replace with submitProposalReviewAction({ decision: 'needs-clarification', feedback })
+            // once the `proposal change requested` status and feedback storage land.
+            console.warn('[OTTER-492] Needs clarification submission is blocked pending OTTER-493 backend work')
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx]: I get why we're doing this but lets not add console.warn, I'm afraid it'll not get removed later

src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx
```diff
@@ -17,56 +25,223 @@ type NewProposalReviewViewProps = {
...
+                </Button>
+                <Button onClick={onConfirm} loading={isPending}>
+                    Yes, submit review
+                </Button>
+            </Group>
+        </Stack>
+    </AppModal>
+)
+
+const RejectReviewModal: FC<SubmitReviewModalProps> = ({ isOpen, onClose, onConfirm, isPending }) => (
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx]: `SubmitReviewModal` and `RejectReviewModal` share the "Please confirm you are ready to submit your review…" paragraph, Cancel/confirm layout, and `closeOnClickOutside` / `closeOnEscape` / `withCloseButton` pending-gating verbatim. The only real divergences are the title, the extra red warning paragraph, and the confirm-button label + color.

Suggest parameterizing a single modal — e.g. `variant: 'default' | 'destructive'`, optional `warning` node, `title`, `confirmLabel`

src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.test.tsx
```diff
@@ -1,20 +1,62 @@
...
 import { useParams } from 'next/navigation'
-import { beforeEach, describe, expect, it } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { NewProposalReviewView } from './new-proposal-review-view'
 
+vi.mock('@/hooks/use-review-feedback', async (importOriginal) => {
+    const original = await importOriginal<typeof import('@/hooks/use-review-feedback')>()
+    return {
+        ...original,
+        useReviewFeedback: vi.fn(() => ({
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.test.tsx]: lets not mock any of our own components or actions.    I think mock is working around the fact that `ReviewFeedbackSection` currently renders only a `<Skeleton />`, so there's no real textarea to drive `isValid`.

Instead of mocking it now though, let's not test it here since we don't have the code to do so.  we can add this test as part of the ReviewFeedbackSection work

As written, this suite is effectively tautological against the placeholder `ReviewFeedbackSection` — it tests the mock, not the UI.

src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.test.tsx
```diff
@@ -1,20 +1,62 @@
...
+    waitFor,
     type Mock,
 } from '@/tests/unit.helpers'
+import { memoryRouter } from 'next-router-mock'
 import { useParams } from 'next/navigation'
-import { beforeEach, describe, expect, it } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { NewProposalReviewView } from './new-proposal-review-view'
 
+vi.mock('@/hooks/use-review-feedback', async (importOriginal) => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.test.tsx]: Project convention (CLAUDE.md): *"do not mock out any of our own components or actions."*

`@/hooks/use-review-feedback` is our own code, so mocking it here violates that rule. When we mock our own code, tests become tautological — they verify the mock, not the behavior — and they don't catch real regressions in the mocked module.

The right shape for this file is to exercise `useReviewFeedback` through the real UI (the textarea that `ReviewFeedbackSection` will eventually render) and drive it with `userEvent.type`, the same way the other tests in this repo drive hooks.

If the UI isn't available yet because `ReviewFeedbackSection` is currently a `<Skeleton />`, defer the gating tests to the PR that lands the real textarea rather than papering over the gap with a hook mock.

(This comment is the general principle; my earlier comment on line 23 is the concrete instance.)

src/app/[orgSlug]/study/[studyId]/review/proposal-review-view.test.tsx
```diff
@@ -10,14 +10,25 @@ import {
...
+const spyModeState = { isSpyMode: false }
+
+vi.mock('@/components/spy-mode-context', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('@/components/spy-mode-context')>()
+    return {
+        ...actual,
+        useSpyMode: () => ({ isSpyMode: spyModeState.isSpyMode, toggleSpyMode: vi.fn() }),
+    }
+})
+
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/proposal-review-view.test.tsx]: good idea, lets move this or something like it to helpers, I think i'd be useful on other specs

nathanstitt (commented):
Looking good, mainly a few comment on naming

src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/new-proposal-review-view.tsx]: this was added because suggested by the issue card ;) I will remove it.

PR #642, opened 2026-04-20
Redesign study code upload screen (Card 54)

nathanstitt:
## Summary

Split the study code upload screen into two distinct states per the Figma redesign ([Card 54](https://www.figma.com/design/jqhZqgwEmGiaWg7eaZnXi0/Card-54-Study-code-redesign?node-id=350-3238)):

- **Empty state** — two prominent cards, "Write and test your code in IDE (recommended)" above + "Upload your files" below, separated by a centered "OR" divider. Starter-code link lives inside the upload card.
- **Review state** — compact header with "Edit files in IDE" + "Upload files" buttons, then a "Review files" table. Main file is now chosen by clicking a star (no auto-pick).

The pivot is driven by `ide.showEmptyState` (`files.length === 0`); `StudyCodePanel` becomes a thin router between `StudyCodeEmptyView` and `StudyCodeReviewView`. `useIDEFiles` behavior is unchanged except for removing the auto-pick fallbacks on `mainFile`.

- Design spec: `docs/superpowers/specs/2026-04-20-study-code-upload-redesign-design.md`
- Implementation plan: `docs/superpowers/plans/2026-04-20-study-code-upload-redesign.md`

## Test Plan

- [ ] `pnpm typecheck` passes (verified locally)
- [ ] `pnpm lint` passes (verified locally)
- [ ] Full unit suite passes except the pre-existing `s3-integration.test.ts` failures that also fail on `main` (SeaweedFS not running locally)
- [ ] Manual: open a draft researcher study → code page; verify the empty view (violet IDE card + OR + upload card)
- [ ] Manual: upload a file or launch the IDE; verify the review view flips in with the file table and header buttons
- [ ] Manual: click a star in the table; Submit enables
- [ ] Manual: delete the last file; view flips back to the empty state
- [ ] Manual: verify long filenames truncate with a tooltip showing the full name
- [ ] Manual: regression-check the resubmit flow (same `StudyCodePanel` is used there)


jbwilson8 (approved):

src/components/study/file-review-table.tsx
```diff
@@ -24,6 +24,32 @@ function formatModified(mtime: string, jobCreatedAt: string | null): string {
...
+interface MainFileStarProps {
+    fileName: string
+    isSelected: boolean
+    onSelect: (fileName: string) => void
+}
+
+function MainFileStar({ fileName, isSelected, onSelect }: MainFileStarProps) {
+    return (
+        <Tooltip label={isSelected ? 'Main file' : 'Set as main file'}>
+            <UnstyledButton
```
therealmarv, line 37: not a blocker I think:

swapping Mantine Radio.Group/Radio for UnstyledButton elements with role="radio" regresses radio accessibility. The new control has none of the expected keyboard radio behavior or roving focus management, so keyboard and assistive-tech users lose the semantics the old implementation had.

therealmarv (approved):

src/components/study/file-review-table.tsx
nathanstitt: are you sure? a radio button by itself is accessible, I'm not seeing what mantine gives use here above it?  got a reference or something?

therealmarv: Upfront: this is super nit-picky and strictly ARIA related and not a blocker in my opinion. Happy to leave it if the tradeoff feels right; flagging it only because I've discovered it.

A very/too detailed worst-case scenario:

> **Where the concern lives**
> 
> It's about the table as a whole, not the individual star. The current implementation declares a radio group contract on the `<Table>` (`role="radiogroup"`, `aria-label="Main file"`) and then puts `role="radio"` + `aria-checked` on each star button. Once we advertise that contract to the browser and assistive tech, we're on the hook for the keyboard behavior that comes with it.
> 
> The contract is defined here: WAI-ARIA APG, Radio Group Pattern: https://www.w3.org/WAI/ARIA/apg/patterns/radio/
> 
> The APG requires (for non-toolbar radio groups):
> 
> 1. Tab / Shift+Tab moves focus into and out of the group (the whole group is one tab stop)
> 2. When focus enters the group, it lands on the checked radio, or on the first radio if none is checked
> 3. Right/Down arrow moves focus to the next radio and checks it (unchecking the previous)
> 4. Left/Up arrow does the same in reverse
> 5. Space checks the focused radio
> 
> Of those five, only 5) is met in the current code. 1) fails because each `UnstyledButton` is its own tab stop (nothing sets `tabindex="-1"` on unselected stars). 2) only appears to work by coincidence of DOM order. 3) and 4) don't do anything, since there's no key handler.
> 
> **The worst-case scenario**
> 
> Multiple files uploaded, no main file picked yet (`mainFile === ""`). This is also the first thing a new user encounters, so the UX tax lands on the first interaction.
> 
> Screen reader / keyboard user flow today:
> 
> 1. Focus enters the table. SR announces "Main file, radio group, N items."
> 2. User does what the radiogroup contract told them to do: press Down arrow to hear the first option. Nothing happens. No focus move, no announcement.
> 3. User tries Right arrow. Still nothing.
> 4. User falls back to Tab and walks through every star one at a time, hearing "Main file: foo.R, radio button, not checked" ... "Main file: bar.R, radio button, not checked" ... etc. Tab visits every row instead of entering/exiting the group once.
> 5. User presses Space on the file they want. That part works.
> 
> It's functional, but the radiogroup role actively misled the user at step 2, which is exactly the moment they were trying to survey the options. A sighted mouse user never notices, which is why this is easy to miss in review.
> 
> **Minimal fix if we want to clear it**
> 
> The cheapest honest option is to stop advertising the radio group contract and treat each star as a toggle button:
> 
> - Remove `role="radiogroup"` and `aria-label="Main file"` from the `<Table>`
> - On the star, drop `role="radio"`, replace `aria-checked` with `aria-pressed`, and phrase the `aria-label` as an action (e.g. `Set foo.R as main file`)
> 
> That makes the ARIA match what the keyboard actually does (a row of independent toggle buttons). We lose the "one of many" semantic a real radio group conveys, but the visual pattern (only one filled star) already makes exclusivity obvious for sighted users, and SR users discover it through interaction.
> 
> The fuller fix is to revert to Mantine `Radio.Group` / `Radio` with custom star styling, so we get native radio semantics and keyboard handling for free. More churn, more correct.

nathanstitt: hm, yes I see tab working for me but can't hurt to add the aria attributes as you suggest.  I've done so

PR #644, opened 2026-04-20
fix(OTTER-514): let researcher reach proposal after code submit

nathanstitt:
## Summary

Jira: [OTTER-514](https://openstax.atlassian.net/browse/OTTER-514)

After a researcher submitted their code, clicking **Previous** ping-ponged between the study-details page and the agreements page — there was no path back to the proposal. Reproduced for all post-submission statuses: Code under review, Code approved, Result under review, Result ready.

## Root cause

1. `view/page.tsx` always routed to `CodeOnlyView` (study-details) once any `CODE-SUBMITTED` job existed — so `ResearcherProposalView` became unreachable.
2. `agreements/page.tsx` sent Previous to the org dashboard when `hasJobActivity` was true, skipping the proposal entirely.
3. `CodeOnlyView` Previous targeted agreements. Combined, the researcher was trapped between study-details ↔ agreements ↔ dashboard, never the proposal.

## Fix

- `view/page.tsx`: when `searchParams.from === 'agreements'`, render `ResearcherProposalView` even if code has been submitted. The `showProposalView` condition now includes `codeSubmitted`, so the proposal is reachable across all post-submission statuses.
- `agreements/page.tsx`: researcher Previous always targets `studyView?from=agreements`, restoring the backward chain **Proposal ← Agreements ← Study Details**.

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] Added unit test: agreements Previous click lands on `studyView?from=agreements` when code is submitted
- [x] Added unit test: `view` page renders `ResearcherProposalView` (with Proceed to Step 3) when `?from=agreements` and code is submitted
- [x] All 137 study-related unit tests pass
- [ ] Manual: as a researcher, submit code and confirm you can click Previous on the study-details page → agreements → Previous → proposal view; clicking Proceed to Step 3 returns to agreements

[OTTER-514]: https://openstax.atlassian.net/browse/OTTER-514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/agreements/page.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/agreements/page.tsx
@@ -75,6 +75,10 @@ export default async function StudyAgreementsRoute(props: {
         redirect(Routes.studyView({ orgSlug: study.submittedByOrgSlug, studyId }))
     }
 
+    // Preserve org-scoped breadcrumb context through the Previous → proposal → agreements roundtrip
+    const returnToSuffix = searchParams.returnTo === 'org' ? '&returnTo=org' : ''
+    const previousHref = `${Routes.studyView({ orgSlug: study.submittedByOrgSlug, studyId })}?from=agreements${returnToSuffix}`
+
     return (
         <Stack p="xl" gap="xl">
             <ResearcherBreadcrumbs crumbs={{ orgSlug, studyId, current: 'Agreements' }} />
@@ -83,7 +87,7 @@ export default async function StudyAgreementsRoute(props: {
                 isReviewer={false}
                 studyId={studyId}
                 proceedHref={Routes.studyCode({ orgSlug: study.submittedByOrgSlug, studyId })}
-                previousHref={`${Routes.studyView({ orgSlug: study.submittedByOrgSlug, studyId })}?from=agreements`}
+                previousHref={previousHref}
                 previousLabel="Previous"
             />
         </Stack>
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,3 +27,4 @@
 - Ask before creating new migrations
 - Ask before modifying permission rules in `src/lib/permissions.ts`
 - Ask before changing route definitions in `src/lib/routes/definitions.ts`
+- Do not commit planning files unless explicity instructed to do so
--- a/src/app/[orgSlug]/study/[studyId]/agreements/page.test.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/agreements/page.test.tsx
@@ -1,11 +1,13 @@
 import { beforeEach, describe, it, expect, vi } from 'vitest'
 import { redirect } from 'next/navigation'
+import * as RouterMock from 'next-router-mock'
 import {
     insertTestStudyJobData,
     insertTestStudyOnly,
     mockSessionWithTestData,
     renderWithProviders,
     screen,
+    userEvent,
 } from '@/tests/unit.helpers'
 import { db } from '@/database'
 import StudyAgreementsRoute from './page'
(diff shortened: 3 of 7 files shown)


jbwilson8 (approved):

src/app/[orgSlug]/study/[studyId]/agreements/page.tsx
```diff
@@ -59,10 +59,10 @@ export default async function StudyAgreementsRoute(props: {
         ? Routes.studyView({ orgSlug: study.submittedByOrgSlug, studyId })
         : Routes.studyCode({ orgSlug: study.submittedByOrgSlug, studyId })
     const proceedLabel = hasJobActivity ? 'Back to Study Details' : undefined
-    const previousHref = hasJobActivity
-        ? Routes.orgDashboard({ orgSlug: study.submittedByOrgSlug })
-        : `${Routes.studyView({ orgSlug: study.submittedByOrgSlug, studyId })}?from=agreements`
-    const previousLabel = hasJobActivity ? undefined : 'Previous'
+    // Always send Previous back to the proposal view (with ?from=agreements so the view
+    // page renders the proposal even when the study has an active job submission).
+    const previousHref = `${Routes.studyView({ orgSlug: study.submittedByOrgSlug, studyId })}?from=agreements`
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/agreements/page.tsx]: This new [`previousHref`](src/app/[orgSlug]/study/[studyId]/agreements/page.tsx#L64) drops the existing `returnTo=org` query param. We use that param to keep researcher breadcrumbs pointing back to the org dashboard, but after this navigation the view page falls back to the global dashboard instead.
I think the same issue applies to [`agreementsHref`](src/app/[orgSlug]/study/[studyId]/view/page.tsx#L36), since that step also loses `returnTo=org`.
Can we preserve `returnTo=org` through this new flow and add a regression test for it?

Worst case:
A researcher who opens the study from an org-scoped dashboard will silently lose that org context as soon as they use this new back-and-forth flow.
Concretely:
1. They start from an org dashboard link that includes [`?returnTo=org`](src/components/dashboard/studies-table/study-action-link.tsx#L30).
2. The agreements page correctly reads that param and sets breadcrumbs back to the org dashboard ([`agreements/page.tsx`](src/app/[orgSlug]/study/[studyId]/agreements/page.tsx#L67)).
3. They click **Previous**, but [`previousHref`](src/app/[orgSlug]/study/[studyId]/agreements/page.tsx#L64) only sends them to `view?from=agreements`, so `returnTo=org` is lost.
4. The view page now falls back to the global dashboard breadcrumb because [`dashboardHref`](src/app/[orgSlug]/study/[studyId]/view/page.tsx#L23) is derived from `searchParams.returnTo`, and that param is now missing.
5. If they click **Proceed to Step 3**, [`agreementsHref`](src/app/[orgSlug]/study/[studyId]/view/page.tsx#L36) sends them back to agreements without `returnTo=org`, so the org-dashboard breadcrumb is gone for the rest of the flow.

Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/agreements/page.tsx]: Good catch — verified the concern is accurate. I traced the flow:

1. Researcher opens study from an org-scoped dashboard → URL carries `?returnTo=org`
2. They land on `/view` with `returnTo=org` → `dashboardHref` resolves to the org dashboard ✓
3. They navigate to `/agreements` and click **Previous** → `previousHref` was `${studyView}?from=agreements` with no `returnTo=org` → back on `/view`, `dashboardHref` silently falls back to the global dashboard ✗
4. Clicking **Proceed to Step 3** from the proposal view → `agreementsHref` was `Routes.studyAgreements({orgSlug, studyId})` with no `returnTo` → context gone for the rest of the chain ✗

Worth noting this bug actually pre-dates PR 644 — the recently-merged PR 636 agreements restructure already had this gap in `previousHref`, so the rebase didn't introduce it, just preserved it. Regardless, fix pushed in 55b9bfdd:

- `agreements/page.tsx`: append `&returnTo=org` to `previousHref` when `searchParams.returnTo === 'org'`
- `view/page.tsx`: pass `returnTo` through to `Routes.studyAgreements(...)`
- `routes/definitions.ts`: extended `studyAgreements` to accept an optional `returnTo` query param (built via `URLSearchParams` so `from` + `returnTo` compose cleanly)
- Added regression tests on both pages that click through and assert `returnTo=org` survives

20/20 unit tests in the affected files pass locally.

Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/agreements/page.tsx]: oh thanks. Interesting info!

therealmarv (approved):

PR #647, opened 2026-04-22
set up crate workspaces with system and language claude.md

kswanson33:


workspace_contexts/python.md
```diff
@@ -0,0 +1,23 @@
...
+- Load query results with `pandas.read_sql()` or `polars.read_database()`; for large pulls use `chunksize=` or a streaming cursor instead of materializing the full result set.
+- Close cursors and connections with `with` blocks (or `try` / `finally`) so they're released on error.
+
+## Packages
+
+Only the packages already available in the environment can be used. No new packages can be installed in the enclave. Use `pip list` to get the list of installed packages.
+
+## Producing the output artifact
+
+The deliverable is a **CSV** uploaded to the TOA endpoint.
```
lwtchu: Won't always be a CSV since we've allowed people to export whatever

workspace_contexts/system.md
```diff
@@ -0,0 +1,57 @@
+# Study environment — system context
```
lwtchu: `You are a data science coding assistant specialized in R, Python, and statistical analysis.

Focus on:
- Clean, reproducible code
- Proper data validation and error handling
- Statistical best practices
- Clear documentation and comments
- Performance optimization for large datasets

Always suggest vectorized operations where possible and include proper error checking.`;

add this somewhere

kswanson33:
I have read the CLA Document and I hereby sign the CLA?

src/app/admin/safeinsights/claude-context.tsx
```diff
@@ -0,0 +1,68 @@
...
+import React, { useRef, useState } from 'react'
+
+type ContextProps = { name: string; orgId: string | null }
+
+function ClaudeContextEditor(
+    {name, orgId, initialContent}:  ContextProps & { initialContent: string}
+) {
+    const [content, setContent] = useState(initialContent)
+
+    const onSubmit = async () => {
```
chrisbendel: You might want to look into the `useMutation` pattern elsewhere in the code for doing stuff like this, it should clean things up a little too. happy to look through together too if its confusing

src/server/actions/claude-context.actions.ts
```diff
@@ -0,0 +1,49 @@
+'use server'
+
+import { Action, ActionFailure } from './action'
+import { z } from 'zod'
+import { sql } from 'kysely'
+
+export const writeClaudeContextAction = new Action('writeClaudeContext', { performsMutations: true })
+    .params(z.object({ content: z.string(), orgId: z.string().uuid().nullable(), name: z.string().min(1)})) // todo better types
+    // .middleware(async ({ session }) => ({ id: session?.user.id }))
```
chrisbendel: just flagging this in case it was missed

src/lib/paths.ts
```diff
@@ -26,6 +26,11 @@ export const pathForStarterCodePrefix = (parts: MinimalCodeEnvInfo) => `${pathFo
 export const pathForStarterCode = (parts: MinimalCodeEnvInfo & { fileName: string }) =>
     `${pathForStarterCodePrefix(parts)}/${parts.fileName}`
 
+export const pathForContextPrefix = () => `claude-context`
+
+export const pathForContext = (parts: { fileName: string }) =>
+    `${pathForContextPrefix()}/${parts.fileName}`
+
```
chrisbendel: should we be explicit/consistent with these method names `claudeContext` or just generalize the path to `context`?

src/database/types.ts
```diff
@@ -81,6 +81,15 @@ export interface Audit {
     userId: string
 }
 
+export interface ClaudeContext {
```
chrisbendel: deferring to team on this but we are referencing `claude` a lot in this PR (which is expected) - just thinking out loud do we want to be generic in where we store our "AI context" files or are we all in on claude naming 🤔

src/app/admin/safeinsights/page.tsx
```diff
@@ -10,9 +12,16 @@ export default async function OrgsAdministration() {
...
 
+    await queryClient.prefetchQuery({
+        queryKey: ['claudeContext', 'system', 'null'],
+        queryFn: () => getClaudeContextAction({name: 'system', orgId: null})
+    })
+
     return (
         <HydrationBoundary state={dehydrate(queryClient)}>
             <OrgsAdminTable />
+            <br />
```
chrisbendel: minor but i think we generally want to rely on a `<Stack gap>` for spacing so its flex-y

src/database/types.ts
kswanson33: good note, flagging the naming things problem -- could replace `claude` with something else. i'd vote `LLM` unless anyone has a better word for "generic type of AI model that can assist with coding"

src/server/actions/agent-context.actions.ts
```diff
@@ -0,0 +1,51 @@
...
+    })
+
+export const getClaudeContextAction = new Action('getClaudeContextAction')
+    .params(
+        z.object({
+            name: z.enum(CONTEXT_NAMES),
+            orgId: z.string().uuid().nullable(),
+        }),
+    )
+    .handler(async ({ db, params: { name, orgId } }) => {
```
nathanstitt, line 39: same as above, we need a `requireAbilityTo` - I think this one should be callable by anyone who can view the orgs studies, so `requireAbilityTo('view', 'orgStudies')` is probably what you want.,  

If that's not quite right we can setup our own permission check in `lib/permissions.ts`

src/server/coder/workspaces.ts
```diff
@@ -233,4 +236,24 @@ const initializeWorkspaceCodeFiles = async (studyId: string): Promise<void> => {
         await fs.writeFile(targetFilePath, Buffer.from(await fileData.arrayBuffer()))
         await fs.utimes(targetFilePath, pastDate, pastDate)
     }
+
+    // Initialize claude.md
+    const workspaceContexts: ContextName[] = ['SYSTEM', codeEnv.language]
+
+    let combinedContextString = ''
+    for (const contextName of workspaceContexts) {
+        const response = await getClaudeContextAction({ name: contextName, orgId: null })
```
nathanstitt: This is a server-only code path that already has DB access, so going through a server action here is doing extra Zod-validation + middleware work for the same query. If we add an auth gate to `getClaudeContextAction` (per the comment on its file), this call site loses access since it has no Clerk session — which would silently break workspace creation. A plain `getClaudeContext(db, { name, orgId })` helper in `lib/claude-context.ts` that both the action and `initializeWorkspaceCodeFiles` call would solve both problems and decouple workspace init from the action's signature.

src/server/coder/workspaces.ts
```diff
@@ -233,4 +236,24 @@ const initializeWorkspaceCodeFiles = async (studyId: string): Promise<void> => {
         await fs.writeFile(targetFilePath, Buffer.from(await fileData.arrayBuffer()))
         await fs.utimes(targetFilePath, pastDate, pastDate)
     }
+
+    // Initialize claude.md
+    const workspaceContexts: ContextName[] = ['SYSTEM', codeEnv.language]
```
nathanstitt, line 264: If a new value is added to the `Language` enum (e.g., `'JULIA'`) without a matching entry in `CONTEXT_NAMES`/`CONTEXT_LABELS`, the `getClaudeContextAction` call below will fail Zod validation (`z.enum(CONTEXT_NAMES)` is `['SYSTEM','R','PYTHON']`). That bubbles up as a thrown error from `initializeWorkspaceCodeFiles` and kills the whole workspace creation. Soft-failing here (skip the missing context, log a warning, write CLAUDE.md with just SYSTEM) would let a new Language land without breaking workspaces until someone fills in its context row.

src/server/coder/workspaces.ts
```diff
@@ -233,4 +236,24 @@ const initializeWorkspaceCodeFiles = async (studyId: string): Promise<void> => {
...
+
+    let combinedContextString = ''
+    for (const contextName of workspaceContexts) {
+        const response = await getClaudeContextAction({ name: contextName, orgId: null })
+        if (isActionError(response)) {
+            throw new Error(errorToString(response))
+        }
+        combinedContextString = combinedContextString + response.content + '\n'
```
nathanstitt: Minor: we append `\n` even when `response.content === ''` (empty row case), which leaves blank lines in `CLAUDE.md`. `if (response.content) combinedContextString += response.content + '\n'` (or filtering before the loop) would tidy it up.

src/app/admin/safeinsights/claude-context.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (isLoading) return null
+
+    if (error) {
+        return (
+            <Text c="red">
+                Failed to load context for {name}: {error.message}
+            </Text>
+        )
+    }
+    return <ClaudeContextEditor name={name} orgId={orgId} initialContent={data ? data.content : ''} />
```
nathanstitt: Two things worth a look here:

1. `data` from `useQuery` carries whatever `getClaudeContextAction` returns, which can be an `ActionError` (`{ error: '...' }`) — the `error` branch above only catches *thrown* errors, not the action's error-return shape. So an `ActionFailure` from the server will silently render an empty textarea with no notice. An `isActionError(data)` check (same pattern as the actions test) would surface it.

2. The `data ? data.content : ''` fallback is effectively dead: when the query succeeds, `data` is always `{ content: string }` (the action's `?? ''` already provides the empty-row case). Replacing it with `isActionError(data) ? '' : data.content` (plus a notification on error) would make the intent clearer.

src/lib/claude-context.ts
```diff
@@ -0,0 +1,15 @@
+import { Language } from '@/database/types'
+
+export const CONTEXT_NAMES = ['SYSTEM', 'R', 'PYTHON'] as const
```
nathanstitt: `CONTEXT_NAMES` (the literal tuple) and `ContextName` (derived from `Language`) can drift: if `Language` gains `'JULIA'`, `ContextName` widens but `CONTEXT_NAMES` doesn't, and TypeScript won't catch the gap. `CONTEXT_LABELS` would fail to compile (good), but the Zod `z.enum(CONTEXT_NAMES)` in the action would silently keep rejecting `'JULIA'`. A `satisfies readonly ContextName[]` on `CONTEXT_NAMES`, or deriving the tuple from `Language` + `'SYSTEM'`, would force them to stay in sync.

src/app/admin/safeinsights/claude-context.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { errorToString } from '@/lib/errors'
+import { getClaudeContextAction, writeClaudeContextAction } from '@/server/actions/claude-context.actions'
+import { Stack, Title, Button, Textarea, Text, Group, Paper } from '@mantine/core'
+import { notifications } from '@mantine/notifications'
+import { useState } from 'react'
+
+type ContextProps = { name: ContextName; orgId: string | null }
+
+function ClaudeContextEditor({ name, orgId, initialContent }: ContextProps & { initialContent: string }) {
+    const [content, setContent] = useState(initialContent)
```
nathanstitt: Worth pulling this onto Mantine's `useForm` — it's the codebase convention for any form with a dirty check + mutation + submit (see `src/app/[orgSlug]/admin/settings/organization-settings-edit.tsx` for a near-identical shape, also `code-review-client.tsx`, `study-request-context.tsx`, `proposal-form.tsx`, the recovery-code flow…). Both `useForm` and `zodResolver` are re-exported from `@/common` already, same place this file pulls `useMutation` from.

The wins are concrete here: `useForm` replaces both `useState` calls *and* the manual `isDiff` comparison with `form.isDirty()`, gives you proper input wiring via `form.getInputProps`, has `form.resetDirty(values)` for the on-success state sync, and integrates with `handleMutationErrorsWithForm` for surfacing field-level `ActionFailure` errors (which would also resolve the other comment about `useQuery` data possibly being an `ActionError`). Adding Zod validation later (e.g. a max-length on content) is then a one-line change.

This also assumes you throw errors in the action so you can use onError directly without needing the explicitly return {success:true} you have in the action

Rough shape, mirroring the existing pattern:

```tsx
import { useForm, useMutation, useQueryClient, zodResolver, z } from '@/common'
import { handleMutationErrorsWithForm } from '@/components/errors'
// ...

const contextFormSchema = z.object({ content: z.string() })
type ContextFormValues = z.infer<typeof contextFormSchema>

function ClaudeContextEditor({ name, orgId, initialContent }: ContextProps & { initialContent: string }) {
    const queryClient = useQueryClient()

    const form = useForm<ContextFormValues>({
        initialValues: { content: initialContent },
        validate: zodResolver(contextFormSchema),
    })

    const { mutate: updateContext, isPending } = useMutation({
        mutationFn: writeClaudeContextAction,
        onSuccess: (_, variables) => {
            form.resetDirty({ content: variables.content })
            queryClient.invalidateQueries({ queryKey: ['claudeContext', name, orgId] })
        },
        onError: handleMutationErrorsWithForm(form),
    })

    return (
        <form onSubmit={form.onSubmit(({ content }) => updateContext({ name, content, orgId }))}>
            <Stack gap="md">
                <Textarea
                    autosize
                    minRows={8}
                    maxRows={20}
                    label={CONTEXT_LABELS[name].label}
                    description={CONTEXT_LABELS[name].description}
                    key={form.key('content')}
                    {...form.getInputProps('content')}
                />
                <Group justify="flex-end">
                    <Button type="submit" loading={isPending} disabled={!form.isDirty()}>
                        Submit
                    </Button>
                </Group>
            </Stack>
        </form>
    )
}
```

Note the switch from `action={...}` to `onSubmit={form.onSubmit(...)}` — Mantine's `onSubmit` wrapper does validation + prevents default before calling the handler, which is what the rest of the codebase expects.

src/database/migrations/1778079447233_add_agent_context_table.ts
```diff
@@ -0,0 +1,22 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema
+        .createTable('claude_context')
+        .addColumn('id', 'uuid', (col) => col.primaryKey().defaultTo(sql`gen_random_uuid()`))
+        .addColumn('name', 'text', (col) => col.notNull())
```
nathanstitt, line 7: given the schema below I think name is supposed to really be 'language' ?  if so we have a pg type `Language` enum we can restrict this to, see how other migrations do this.

src/server/actions/claude-context.actions.test.ts
```diff
@@ -0,0 +1,104 @@
...
+                content: 'nope',
+                orgId: null,
+            })
+            expect(isActionError(result)).toBe(true)
+            expect(errorToString(result)).toMatch('Validation')
+        })
+    })
+    describe('getClaudeContextAction', () => {
+        beforeEach(async () => {
+            await db.deleteFrom('claudeContext').execute()
```
nathanstitt: same as above, don't think we need this, all modifications to the db are cleared on every test run (they run in transactions)

src/server/actions/claude-context.actions.test.ts
```diff
@@ -0,0 +1,104 @@
+import { describe, expect, it, beforeEach } from 'vitest'
+import { db, mockSessionWithTestData } from '@/tests/unit.helpers'
+import { writeClaudeContextAction, getClaudeContextAction } from './claude-context.actions'
+import { errorToString, isActionError } from '@/lib/errors'
+
+describe('context actions', () => {
+    describe('writeClaudeContextAction', () => {
+        beforeEach(async () => {
+            await db.deleteFrom('claudeContext').execute()
```
nathanstitt: don't think we need this, all db modifications are reset on each test

src/server/actions/agent-context.actions.ts
```diff
@@ -0,0 +1,51 @@
+'use server'
+
+import { Action, ActionFailure } from './action'
+import { z } from 'zod'
+import { sql } from 'kysely'
+import { CONTEXT_NAMES } from '@/lib/claude-context'
+
+export const writeClaudeContextAction = new Action('writeClaudeContextAction', { performsMutations: true })
+    .params(z.object({ content: z.string(), orgId: z.string().uuid().nullable(), name: z.enum(CONTEXT_NAMES) }))
+    .handler(async ({ session, db, params: { name, content, orgId } }) => {
```
nathanstitt, line 11: allmost all actions must have a `.requireAbilityTo` check - that's much preferred over the manual checks you have below.  since this is an org admin only actions, it'd probably be:
`.requireAbilityTo('update', 'Org')`

once you add this, you can remove the below check

src/server/actions/claude-context.actions.ts
```diff
@@ -0,0 +1,51 @@
...
+            .onConflict((onconflict) =>
+                onconflict.columns(['orgId', 'name']).doUpdateSet({
+                    content: content,
+                    updatedBy: userId,
+                    updatedAt: sql`now()`,
+                }),
+            )
+            .execute()
+
+        return { success: true }
```
nathanstitt: this isn't really needed.  it's better to throw an error which the front-end can check for with a onError in useMutation

nathanstitt (commented):
Thanks for working through this KC, there's a lot here for someone who's not super familiar with the codebase

src/database/migrations/1778079447233_add_agent_context_table.ts
kswanson33: name can be eg. `{ <language>, 'SYSTEM', 'DATA_TYPE' }`, so I think `Language` would be reductive. what do u think?

src/server/actions/claude-context.actions.test.ts
kswanson33: i was getting issues with checks for empty DB lines bc what i had put in my localhost experiments was being returned instead (modifications to the db within a test are cleared, but not what was previously there). so this test will succeed on the branch without this `beforeEach` line, but might fail locally. ok to keep it in for the local case?

src/server/actions/claude-context.actions.test.ts
kswanson33: same as ^

src/server/actions/agent-context.actions.ts
kswanson33: I made custom permissions, because I believe there needs to be an `Ability` that matches the params for the action. LMK if there's a better way!

src/server/actions/agent-context.actions.ts
kswanson33: see comment on other requireAbilityTo check

src/server/coder/workspaces.ts
kswanson33: I think I'd rather keep the failure and require that values are added to the `Language` enum, `CONTEXT_NAMES`, and `CONTEXT_LABELS` together. I can't imagine wanting to add language support and having it be a bad thing to add a form field for the context too!

`typecheck` will ping for `CONTEXT_LABELS` getting out of sync with the `Language` enum. I added a few lines to enforce `CONTEXT_NAMES` parity as well.

src/app/admin/safeinsights/claude-context.tsx
kswanson33: added the check for `ActionError`, thanks!

I tried a couple of ways to satisfy `typecheck` about whether `data` is defined (yes, `data` will always be defined at this point, no, typescript isn't picking this up), and the only thing that worked was still `data ? data.content : ''`. if you have any other suggestions lmk

src/app/admin/safeinsights/claude-context.tsx
```diff
@@ -63,10 +63,10 @@ function ClaudeContextDataLoader({ name, orgId }: ContextProps) {
 
     if (isLoading) return null
 
-    if (error) {
+    if (error || isActionError(data)) {
```
nathanstitt: data will never be an action error.  we've modified usQuery to convert thrown errors into the `error` and not have it on data.  You only need to use `isActionError` if you call a action directly (not through useQuery/useMutation)

src/app/admin/safeinsights/claude-context.tsx
kswanson33: thanks!! didn't know about the `useForm` stuff, that's super cool. guess I was reinventing the wheel 🛞

src/app/admin/safeinsights/claude-context.tsx
kswanson33: oh i did this bc you told me to, haha 😅  https://github.com/safeinsights/management-app/pull/647#discussion_r3269597438

kswanson33: looked into `hooks/query-wrappers.ts` and verified that you're correct here, the check for the `isActionError` is handled by the `processResponse` function. I'm taking out the check again! 😮‍💨

src/database/types.ts
jpslav: `AgentContext`?

src/database/migrations/1778079447233_add_claude_context_table.ts
```diff
@@ -0,0 +1,22 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema
+        .createTable('claude_context')
+        .addColumn('id', 'uuid', (col) => col.primaryKey().defaultTo(sql`gen_random_uuid()`))
```
nathanstitt: Hey sorry, I'd missed this on the earlier review. We use uuidv7, so the `gen_random_uuid()` here should be: `v7uuid()` 


```suggestion
        .addColumn('id', 'uuid', (col) => col.primaryKey().defaultTo(sql`v7uuid()`))
```

src/lib/permissions.ts
```diff
@@ -29,6 +29,7 @@ export function defineAbilityFor(session: UserSession) {
     permit('claim', 'PendingUser')
     permit('reset', 'MFA')
     permit('view', 'Orgs')
+    permit('view', 'ClaudeContext')
```
nathanstitt: `view ClaudeContext` is granted in the "all users can perform" block, so every authenticated user can call `getClaudeContextAction` and read the system/language context, while create/update stay admin-gated. 

That asymmetry may well be intentional , but maybe we want to only admins and researchers to view? probably doesn't _really_ matter

src/server/coder/workspaces.ts
```diff
@@ -233,4 +236,24 @@ const initializeWorkspaceCodeFiles = async (studyId: string): Promise<void> => {
...
+    // Initialize claude.md
+    const workspaceContexts: ContextName[] = ['SYSTEM', codeEnv.language]
+
+    let combinedContextString = ''
+    for (const contextName of workspaceContexts) {
+        const response = await getClaudeContext(database.db, { name: contextName, orgId: null })
+        if (isActionError(response)) {
+            throw new Error(errorToString(response))
+        }
+        if (response.content) combinedContextString += response.content + '\n'
```
nathanstitt, line 272: Because this whole block lives inside `initializeWorkspaceCodeFiles`, which early-returns when `studyDirHasFiles(studyDir)` is true, CLAUDE.md only gets written on the very first init of a workspace. If an admin later edits the context, existing workspaces won't pick up the change. Probably fine for now but we might consider moving it up if wanted to update long-running studies

src/lib/agent-context.ts
```diff
@@ -0,0 +1,31 @@
+import type { Language } from '@/database/types'
+import { DBConn } from '@/database'
+
+export type ContextName = 'SYSTEM' | Language
+export const CONTEXT_NAMES = ['SYSTEM', 'R', 'PYTHON'] as const satisfies readonly ContextName[]
+
+type _MissingContextNames = Exclude<ContextName, (typeof CONTEXT_NAMES)[number]>
+const _contextNamesExhaustive: [_MissingContextNames] extends [never] ? true : _MissingContextNames = true
```
nathanstitt, line 8: Good defensive typing, this should catch the case that someone adds a new `Language` to the union but forgets to list it in `CONTEXT_NAMES`, the build breaks instead of silently dropping a language.

src/server/coder/workspaces.ts
nathanstitt: sounds good, long as you've thought of it 😁

src/lib/permissions.ts
kswanson33: yeah, i guess i was thinking about researchers clicking that "launch coder" button, but since we're not using an action there anymore, i think the `get` permissions can be more gated, since i expect they'll only be used on pages that admins can access anyways

src/app/admin/safeinsights/agent-context.tsx
```diff
@@ -0,0 +1,107 @@
...
+                <Textarea
+                    autosize
+                    minRows={8}
+                    maxRows={20}
+                    label={CONTEXT_LABELS[name].label}
+                    description={CONTEXT_LABELS[name].description}
+                    {...form.getInputProps('content')}
+                />
+                <Group justify="flex-end">
+                    <Button type="submit" loading={isPending} disabled={!form.isDirty()}>
```
nathanstitt: nitpick:  it's _slightly_ non-standard to disable the form when nothing has changed, instead we can no-op the save if it was expensive.  It's confusing to users when a submit button is disabled with no errors on the screen, makes them think something is broken.

nathanstitt (commented):
Looking really good KC, ty! I think the only things that are really blockers are the db uuid and form disable btn

src/lib/agent-context.ts
kswanson33: thanks, my agent helped 🐬

src/app/admin/safeinsights/agent-context.tsx
kswanson33: gotcha! yeah without formal designs for this form i kind of guessed about what would be good. i was thinking of the button being enabled/disabled as a way of confirming visually that a save had been successful. is there a better way to do that?

nathanstitt (approved):
Looks great, @kswanson33 - I'm fine with the submit button, up to you if you wanna do anything there

PR #648, opened 2026-04-22
Automated code summaries for reviewers

chrisbendel:
depends on https://github.com/safeinsights/iac/pull/142

Decision was made to incorporate the [crate-do-review-agent](https://github.com/safeinsights/crate-do-review-agent) directly into the management app, so this PR also does that work

Flow of the UI:

https://github.com/user-attachments/assets/7d34e147-10f8-446c-b245-4bd853f2dbd2


chrisbendel:
I have read the CLA Document and I hereby sign the CLA

src/server/claude/code-summary.ts
```diff
@@ -0,0 +1,122 @@
...
+    answer: string
+}
+
+const SUMMARY_QUESTIONS = [
+    { id: 'q1', question: 'What is this code doing at a high level?' },
+    { id: 'q2', question: 'What data sources does this code access or use?' },
+    { id: 'q3', question: 'What outputs or results does this code produce?' },
+    { id: 'q4', question: 'What external libraries or packages does this code rely on?' },
+    { id: 'q5', question: 'Are there any potential data privacy or security concerns in this code?' },
+] as const
```
chrisbendel: These should be reviewed carefully and probably refined by product and/or devs

src/server/events.ts
```diff
@@ -45,6 +46,10 @@ export const onStudyCreated = deferred(async ({ studyId, userId }: StudyEvent) =
     await email.sendStudyProposalEmails(studyId)
 })
 
+export const onCodeSummaryRequested = deferred(async ({ studyJobId }: { studyJobId: string }) => {
```
chrisbendel: Do we want to investigate adding something like `pgboss` (tool TBD) to handle queues/async jobs?

package.json
```diff
@@ -40,6 +40,7 @@
         "update-all-packages": "npx npm-check-updates -u"
     },
     "dependencies": {
+        "@anthropic-ai/sdk": "^0.90.0",
```
nathanstitt, line 43: I thought we were supposed to be using https://github.com/safeinsights/crate-do-review-agent for this work?

chrisbendel: hah yes this is what i wanted to check in with you about and why this is still a draft 😎

src/database/migrations/1775200000000_study_review.ts
```diff
@@ -0,0 +1,16 @@
+import { type Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema
+        .createTable('study_review')
+        .addColumn('id', 'uuid', (col) => col.primaryKey().defaultTo(sql`gen_random_uuid()`))
```
rnathuji: This is as much a question as a comment 😅 : When creating a migration in [my own PR](https://github.com/safeinsights/management-app/pull/664), I actually wasn't sure whether to use `gen_random_uuid()` or `v7uuid()`. I went with the latter as it seemed to be used in most tables, but it doesn't seem 100% consistent (or, if there's a reason why sometimes one is selected vs the other, it was not obvious to me 🤷‍♂️ ).

src/database/migrations/1775200000000_study_review.ts
```diff
@@ -0,0 +1,16 @@
...
+        .addColumn('study_job_id', 'uuid', (col) => col.notNull().references('study_job.id').onDelete('cascade'))
+        .addColumn('generated_at', 'timestamptz', (col) => col.notNull())
+        .addColumn('report', 'jsonb', (col) => col.notNull())
+        .addColumn('created_at', 'timestamptz', (col) => col.notNull().defaultTo(sql`now()`))
```
rnathuji, line 9: This is as much a question as a comment 😅 : Similar to the above, I noticed that `created_at`  is not tz-aware in **most** tables but is in some. I also observed `updated_at` was always unaware. I figured that the general (but not consistent 🤷‍♂️ ) convention was no tz in the DB and assume UTC, but again, that was my best guess 😂 .

rnathuji (commented):
I'll review more of this later today, but I had a couple of comments specific to the DB migration which I (selfishly 😅 ) thought I'd ask now as the answers might impact the PR I am working on as well.

src/server/aws.ts
```diff
@@ -349,14 +349,13 @@ export const storeS3File = async (
...
-    const [_, upStream] = body.tee()
     const uploader = new Upload({
         client: getS3Client(),
         tags: objectToAWSTags(info),
         params: {
             Bucket: s3BucketName(),
             Key,
-            Body: upStream,
+            Body: body,
         },
```
nathanstitt, line 359: I'm not sure why we had this, but doing so means that the body can no longer be read from, is this intentional?

src/server/actions/study-request.ts
```diff
@@ -277,6 +277,7 @@ export const finalizeStudySubmissionAction = new Action('finalizeStudySubmission
                 .values({ studyJobId: latestJob.id, userId, status: 'CODE-SUBMITTED' })
                 .execute()
             triggerCodeScan(latestJob.id, orgSlug, studyId)
+            onStudyReviewRequested({ studyJobId: latestJob.id })
```
nathanstitt, line 395: we're calling  onStudyReviewRequested from three places (finalizeStudySubmissionAction, addJobToStudyAction,  submitStudyCodeAction). The deferred(...) runner doesn't deduplicate. If a code submission flow retries (network blip,
  user double-clicks, etc.) you get two reviews per job stored. The runner inserts unconditionally:

  await db.insertInto('studyReview').values({ studyJobId, ... }).execute()

  getStudyReviewForJob then orderBy('createdAt', 'desc').limit(1) masks the duplicate but we'd still be sending out to the API

  Recommendations (pick one):
  - Add a unique(studyJobId) constraint on study_review and use upsert semantics.
  - Check for an existing row at the top of generateAndStoreStudyReview and bail if present.
  - Make the event idempotent at the call site (store an event-fired marker on studyJob).

src/server/agents/review-agent/runner.ts
```diff
@@ -0,0 +1,138 @@
+import { db } from '@/database'
+import logger from '@/lib/logger'
+import { extractTextFromLexical } from '@/lib/word-count'
+import { generateAnalysis } from './agent'
+import type { ReviewContent } from './types'
+import { getConfigValue } from '@/server/config'
+import { fetchFileContents } from '@/server/storage'
+
+const MAX_FILE_SIZE_BYTES = 100_000
```
nathanstitt, line 9: something we should consider is having a MAX_REVIEW_SIZE or something.  right now we're not capping the number of files and if they upload 10 or so files at this size, some napkin math puts that at 250k tokens which would be around $1.00 per review

src/server/agents/review-agent/README.md
```diff
@@ -0,0 +1,32 @@
...
+- `resultsSummary` — interpretation of researcher-supplied test results (optional)
+- `alignmentCheck` — does the code match the proposal? findings if not
+- `complianceCheck` — does the submission violate org rules? findings if so
+
+## Files
+
+- `agent.ts` — `ReviewAgent` class. Constructor takes API key or injected client. `generateAnalysis()` returns the report; `chat()` reserved for future follow-up Q&A (descoped now, planned by Oct 2026).
+- `types.ts` — `ReviewContent`, `ReviewAgentConfig`, `AnalysisReport`, `ReferenceDocs`.
+- `prompts.ts` — `DEFAULT_SYSTEM_INSTRUCTION`, `DEFAULT_ANALYSIS_PROMPT_TEMPLATE`, single-pass `buildAnalysisPrompt(...)` (placeholder injection-safe).
+- `index.ts` — public exports.
```
nathanstitt: slightly out of date, index.ts doesn't exist

src/app/[orgSlug]/study/[studyId]/review/study-review-section.tsx
```diff
@@ -0,0 +1,162 @@
...
+    initialReview: StudyReviewWithMeta | null
+}
+
+export function StudyReviewSection({ studyJobId, initialReview }: StudyReviewSectionProps) {
+    const { data: review } = useQuery({
+        queryKey: ['study-review', studyJobId],
+        queryFn: () => getStudyReviewAction({ studyJobId }),
+        initialData: initialReview,
+        // Poll while review is still pending; stop once the row arrives.
+        refetchInterval: (query) => (query.state.data ? false : POLL_INTERVAL_MS),
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/study-review-section.tsx]: suggest we add a cap on number of retries and go to an error state if it takes too long

src/database/migrations/1775200000000_study_review.ts
```diff
@@ -0,0 +1,16 @@
+import { type Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema
+        .createTable('study_review')
+        .addColumn('id', 'uuid', (col) => col.primaryKey().defaultTo(sql`v7uuid()`))
+        .addColumn('study_job_id', 'uuid', (col) => col.notNull().references('study_job.id').onDelete('cascade'))
+        .addColumn('generated_at', 'timestamptz', (col) => col.notNull())
```
nathanstitt: should this default to now(), and if so, then does created_at really add anything?

nathanstitt (approved):
looks pretty good, a few ideas around not (potentially) firing multiple requests.  I'll approve and you can decide if they're valid or not.

src/server/aws.ts
chrisbendel: this was a stashed change that got brought forward from [this PR](https://github.com/safeinsights/management-app/pull/654/changes#diff-729c016d12e1c7c0da7edf4acc10e506580818ea202c575d2aaf316d2119753eR352) which basically cleaned up the unused call to the old `csStream` and left the original body, tee essentially forks the 1 readable stream into two separate readable streams, but we're only using this one so it should be fine - just leftover cleanup, sorry it got popped into here!
```
    /**
     * The **`tee()`** method of the ReadableStream interface tees the current readable stream, returning a two-element array containing the two resulting branches as new ReadableStream instances.
     *
     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ReadableStream/tee)
     */
    tee(): [ReadableStream<R>, ReadableStream<R>];
```

src/app/[orgSlug]/study/[studyId]/review/study-review-section.tsx
Comment (chrisbendel) [reply src/app/[orgSlug]/study/[studyId]/review/study-review-section.tsx]: good call, since the study review job might take an unknown amount of time, it might take 10 seconds to complete or it might take 30+... thats happening in background, so this is really just polling to see if the field is populated yet. will add an error condition but setting a timeout might be tough (we can solve this later with the queue stuff) - also we can get UX feedback on this behavior once they have a chance to interact with it

src/server/agents/review-agent/runner.ts
chrisbendel: yeah good idea, putting a max/cap of 10 files for now, can followup with product if needs change

chrisbendel:
@nathanstitt see updates/comments from last round, thanks for the review!!

nathanstitt:
hey @chrisbendel I think this has a small bug.  It's showing a spinner in place of the review panel outside of the feature flag.  When users are not using the feature flag we should show the same UI as previously.

PR #649, opened 2026-04-23
OTTER-493 Flag - All - Backend: "Needs clarification" status and feedback storage

therealmarv:
see [OTTER-493](https://openstax.atlassian.net/browse/OTTER-493)

- Add a unified proposal review submission action for approve, needs-clarification, and reject
- Add `CHANGE-REQUESTED` status, labels, and clarification audit event
- Persist each submitted review in the new `study_proposal_comment` table with decision, reviewer, and Lexical-compatible feedback, validated server-side
- Prevent duplicate/stale submissions by only allowing reviews from `PENDING-REVIEW`
- Keep legacy approve/reject paths intact, with the new clarification flow gated behind `ProposalReviewFeatureFlag`

[OTTER-493]: https://openstax.atlassian.net/browse/OTTER-493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/proposal-review.ts
```diff
@@ -0,0 +1,20 @@
...
+export function toReviewDecision(decision: Decision): ReviewDecision {
+    return decision.toUpperCase() as ReviewDecision
+}
+
+export const FEEDBACK_MIN_WORDS = 50
+export const FEEDBACK_MAX_WORDS = 500
+
+export const SUBMITTED_PROPOSAL_REVIEW_STATUSES = [
+    'APPROVED',
+    'PROPOSAL-CHANGE-REQUESTED',
```
nathanstitt: the `PROPOSAL-` prefix feels out of place to me.  we're not saying `PROPOSAL-APPROVED` for the other statuses, so it seems like this should be only: `CHANGE-REQUESTED`

src/database/migrations/1776200000001_study_proposal_feedback_table.ts
```diff
@@ -0,0 +1,35 @@
...
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema.createType('review_decision').asEnum(['APPROVE', 'NEEDS-CLARIFICATION', 'REJECT']).execute()
+    await db.schema.createType('proposal_feedback_author_role').asEnum(['REVIEWER', 'RESEARCHER']).execute()
+    await db.schema
+        .createType('proposal_feedback_entry_type')
+        .asEnum(['REVIEWER-FEEDBACK', 'RESUBMISSION-NOTE'])
+        .execute()
+
+    await db.schema
+        .createTable('study_proposal_feedback')
```
nathanstitt: since this holds two way discussion between reviewer / researcher, it seems like more of a 'discussion' or 'comments' than 'feedback', which would be only from reviewer to researcher.

src/database/migrations/1776200000000_add_proposal_change_requested_status.ts
```diff
@@ -0,0 +1,8 @@
+import { type Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await sql`ALTER TYPE study_status ADD VALUE 'PROPOSAL-CHANGE-REQUESTED'`.execute(db)
```
nathanstitt: see below but feels to me like this should be 'CHANGE-REQUESTED'

nathanstitt (commented):
code looks good, just comments around naming things

PR #651, opened 2026-04-26
OTTER-491: Add collaborative editor for proposal feedback

stetradis:
## Summary                                                                                                    
                                                                                                                  
  - Add a real-time collaborative rich text editor (Lexical + Yjs + Hocuspocus) to the proposal review feedback   
  section, allowing multiple reviewers to edit simultaneously with live cursor presence                           
  - Set up database persistence for the editor service with a new `yjs_document` table and Postgres-backed storage
   in the Hocuspocus server                                                                                       
  - Add save status indicator (with debounced "Saving progress…" / "Last saved X ago" states) and active editors
  list showing who else is currently editing                                                                      
  - Update `useReviewFeedback` hook to count words from Lexical editor JSON instead of plain text               
  - Remove unused editor-demo and error-demo pages                                                                
                                                                                                                  
  ## Changes                                                                                                      
                                                                                                                  
  - **Editor service** (`services/editor/`): Add `@hocuspocus/extension-database` with Postgres persistence via   
  `pg` pool; documents are upserted on save
  - **Database**: New migration for `yjs_document` table (name, data, updated_at)                                 
  - **`CollaborativeEditor` component**: Lexical-based editor with collaboration plugin, toolbar, save status     
  tracking, active editors badges, and placeholder support                                                        
  - **`ReviewFeedbackSection`**: Replaced skeleton placeholder with the collaborative editor, added instructional 
  copy and minimum word count guidance                                                                            
  - **`useReviewFeedback`**: Switched from `countWords` to `countWordsFromLexical` for Lexical JSON content     
  - **Server actions**: New `getYjsDocumentUpdatedAtAction` with study view permission check                      
  - **Cleanup**: Removed `/editor-demo` and `/error-demo` pages

## Save status — UI behavior

Three rendered states: `idle` (nothing), `Saving progress…`, `Last saved {ago}`.

**Rules**
- "Saving progress…" only appears for the user whose own keystrokes triggered the save.
- "Last saved X ago" refreshes for every connected client whenever the document settles (local edit, remote edit, or 30s heartbeat during continuous typing).
- The 2s debounce is shared across local + remote events on each client, so concurrent multi-user typing only resolves the indicator after activity globally pauses for 2s — matching when the server actually persists.

**Per-client behavior by scenario**

| Scenario | Active typist sees | Other connected clients see |
|---|---|---|
| Solo edit, then 2s pause | `Saving progress…` → `Last saved just now` | n/a |
| You type, others idle | `Saving progress…` → `Last saved just now` | `Last saved just now` (no "Saving…" flash) |
| Others type, you idle | `Last saved just now` (no "Saving…" flash) | `Saving progress…` → `Last saved just now` |
| Concurrent typing | `Saving progress…` → `Last saved just now` after **all** typists pause 2s | same |
| Continuous typing past 30s | `Last saved just now` heartbeat every 30s, no "Saving…" flash; full choreography on final pause | `Last saved just now` heartbeat every 30s |
| Nobody typing | `{ago}` text auto-refreshes every 30s ("just now" → "less than a minute ago" → …) | same |

**Timing constants** (kept aligned with the editor service's `Server` config):
- `TYPING_DEBOUNCE_MS = 2000` — wait after activity stops before showing the save indicator
- `PERSIST_DELAY_MS = 1000` — buffer between "Saving…" and "Last saved"
- `MAX_SAVE_INTERVAL_MS = 30_000` — heartbeat cap during continuous typing


src/server/config.ts
```diff
@@ -32,6 +32,7 @@ export const PROD_BUILD = process.env.NODE_ENV === 'production'
 export const PROD_ENV = ENVIRONMENT_ID == 'production'
 export const STAGING_ENV = ENVIRONMENT_ID == 'staging'
 export const DEPLOYED_ENV = !CI_ENV && ENVIRONMENT_ID !== 'local'
+export const WS_URL = DEPLOYED_ENV ? '/ws' : 'ws://localhost:1234'
```
nathanstitt, line 35: hm, shouldl 1234 port mapping be set in ENV (probably as fallback is fine though)

src/components/editable-text/collaborative-editor.tsx
```diff
@@ -0,0 +1,368 @@
...
+import { Toolbar } from './toolbar'
+import { EscapeFocusPlugin } from './escape-focus-plugin'
+
+// These align with the editor service's Server config (debounce: 2000, maxDebounce: 30_000).
+// TYPING_DEBOUNCE_MS waits for typing to stop before showing "Saving progress…".
+// PERSIST_DELAY_MS is the brief window while the server writes to the database.
+// MAX_SAVE_INTERVAL_MS triggers a save indicator during continuous typing.
+const TYPING_DEBOUNCE_MS = 2000
+const PERSIST_DELAY_MS = 1000
+const MAX_SAVE_INTERVAL_MS = 30_000
```
nathanstitt: can we extract these to a shared file, `services/editor/constants.ts` or something? that way they stay in sync

src/database/migrations/1776000000000_yjs_document.ts
```diff
@@ -0,0 +1,14 @@
+import { type Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema
+        .createTable('yjs_document')
+        .addColumn('name', 'text', (col) => col.primaryKey())
+        .addColumn('data', 'bytea', (col) => col.notNull())
+        .addColumn('updated_at', 'timestamp', (col) => col.defaultTo(sql`now()`).notNull())
+        .execute()
+}
```
nathanstitt, line 11: is it possible to pass any metadata to the hocus-pocus server like studyId as json, which we could save?

services/editor/server.ts
```diff
@@ -1,8 +1,31 @@
...
+    connectionString: process.env.DATABASE_URL,
+})
+
 const server = new Server({
     port: 1234,
+    debounce: 2000,
+    maxDebounce: 30_000,
+    extensions: [
+        new Database({
+            fetch: async ({ documentName }) => {
```
nathanstitt, line 57: where does documentName come from?  what sets it?  as I noted on the migration I'm wondering if it's possible to save metadata like studyId on the record so we could match state

nathanstitt (commented):
code looks good, just a few questions on the table data

nathanstitt (approved):
On second thought I think we should merge this and then I'll make a follow-up. I see some additional work might be needed to get it deployed securely

services/editor/server.ts
stetradis: `documentName` is set by the `id` prop that is passed to the `CollaborativeEditor`, in this case it is `review-feedback-{studyId}`. I structured it this way to support multiple editors for different fields on the same study, but I think having a `studyId` column could also be useful for easier querying - I'll add it

PR #654, opened 2026-04-27
Remove checksum call

chrisbendel:
Upload from @aws-sdk/lib-storage does multipart above ~5MB. ChecksumSHA256 in params is single-object semantic — S3 expects it to match one PutObject body. Multipart computes per-part checksums and a composite ETag instead. Passing whole-body SHA256 to multipart = mismatch = upload fails.


nathanstitt (approved):
sounds fine, but lint check is failing for some reason

chrisbendel:
@nathanstitt thanks! looks like it was just some unused imports/variables 👍

PR #655, opened 2026-04-27
remove stub ReviewProgressBar from feature-flagged proposal review

therealmarv:
- Removed the stub progress bar from the top of the new feature-flagged proposal review page, since it was a non-functional placeholder hard-coded to step 1 of 5.
- Cleaned up the unused step definitions, type, and test assertion that only the progress bar relied on.


nathanstitt (approved):
thanks Marvin

PR #663, opened 2026-05-02
OTTER-495: Consolidate initial request views into shared component

stetradis:
- Extracted shared `ProposalRequest` component (src/components/study/proposal-initial-request.tsx) to consolidate the duplicated initial request UI between the researcher submitted view and the DO review view                      
- Both `ProposalSubmitted` and `ProposalSection` now delegate to ProposalRequest, passing view-specific props (stepLabel, heading, banner)                                             
- Fixed title wrapping: added `wrap="nowrap"` to the title/date row and `whiteSpace: 'nowrap'` on the submitted date so it stays pinned to the right                                          
- Capped title width at `105ch` (~120 characters in Open Sans)

### DIFF
--- a/src/components/study/proposal-initial-request.tsx
+++ b/src/components/study/proposal-initial-request.tsx
@@ -2,7 +2,7 @@
 
 import { useState, type ReactNode } from 'react'
 import { Anchor, Collapse, Divider, Group, Paper, Stack, Text, Title } from '@mantine/core'
-import { CaretRightIcon } from '@phosphor-icons/react/dist/ssr'
+import { CaretRightIcon, CaretUpIcon } from '@phosphor-icons/react/dist/ssr'
 import dayjs from 'dayjs'
 import { stringifyJson } from '@/lib/string'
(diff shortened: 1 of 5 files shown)


src/components/study/proposal-initial-request.tsx
```diff
@@ -0,0 +1,131 @@
...
+            component="button"
+            size="sm"
+            fw={700}
+            onClick={onClick}
+            display="inline-flex"
+            style={{ alignItems: 'center', gap: 4 }}
+            data-testid={testId}
+        >
+            {isExpanded ? 'Hide full initial request' : 'View full initial request'}
+            {isExpanded ? <CaretUpIcon size={12} weight="bold" /> : <CaretRightIcon size={12} weight="bold" />}
```
nathanstitt: not a huge deal but I think we lost the  CSS transform: rotate(-90deg) transition for a smooth 200ms   rotation animation when toggling.  probably easy to add back here?

nathanstitt (approved):
nice cleanup!

PR #664, opened 2026-05-04
Add support for multiple data source URLs and corresponding descriptions

rnathuji:


src/database/migrations/1776461687255_org_data_source_doc.ts
```diff
@@ -0,0 +1,40 @@
...
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema
+        .createTable('org_data_source_document')
+        .addColumn('id', 'uuid', (col) => col.primaryKey().defaultTo(sql`v7uuid()`))
+        .addColumn('org_data_source_id', 'uuid', (col) =>
+            col.notNull().references('org_data_source.id').onDelete('cascade'),
+        )
+        .addColumn('url', 'text')
+        .addColumn('description', 'text')
+        .addColumn('created_at', 'timestamp', (col) => col.notNull().defaultTo(sql`now()`))
```
rnathuji: Note for self: Per discussion [here](https://openstax.slack.com/archives/C06NR64JF40/p1777912293599869), update this to use `timestamptz`.

chrisbendel (approved):

src/app/[orgSlug]/admin/settings/data-sources.actions.ts
```diff
@@ -79,23 +110,72 @@ export const updateOrgDataSourceAction = new Action('updateOrgDataSourceAction',
...
+                .where('id', '=', doc.id)
+                .returning(['id', 'url', 'description'])
+                .executeTakeFirstOrThrow(throwNotFound(`Data source document with id ${doc.id}`))
+
+            dataSourceDocResults.push(updatedDoc)
+        }
+
+        // Delete docs that are no longer associated with the data source
+        for (const doc of docsToDelete) {
+            await db.deleteFrom('orgDataSourceDocument').where('id', '=', doc.id).execute()
```
Comment (nathanstitt) [inline src/app/[orgSlug]/admin/settings/data-sources.actions.ts]: could delete all at once:

```suggestion
            await db.deleteFrom('orgDataSourceDocument').where('id', 'in', docsToDelete.map(d=>d.id)).execute()
```

src/app/[orgSlug]/admin/settings/use-data-source-form.ts
```diff
@@ -38,7 +78,30 @@ export function useDataSourceForm(dataSource: DataSource | undefined, onComplete
         onError: reportMutationError(isEditMode ? 'Failed to update data source' : 'Failed to add data source'),
     })
 
-    const onSubmit = form.onSubmit((values) => save(values))
+    const onSubmit = form.onSubmit(({ newDocumentUrl, newDocumentDescription, ...values }) => {
+        // Handle the case where a user may have document data in-progress. We proactively add the
+        // document and then validate to surface errors. If everything is fine, we save().
+        if (newDocumentUrl !== '' || newDocumentDescription !== '') {
+            addDocument()
```
Comment (nathanstitt) [inline src/app/[orgSlug]/admin/settings/use-data-source-form.ts]: mutating the form while submitting makes me nervous.  maybe check how the other forms handle not-yet-added fields by grabbing the values and including in the request without actually adding them to the form

nathanstitt (commented):
This mentioned 'documents' a lot, but we're not dealing with documents, it's URLs or maybe "paths".  Reading documents made me think we were uploading files at first.

Also should we update the coder/workspace to send these to coder?  I think that'll likely fail right now

rnathuji:
@nathanstitt  - I used terminology "docs" after talking with @lwtchu about how this might generalize in PDM vision. It may be too future proof, and I'm okay with changing nouns, but my understanding was today they may have URLs, but in the future :man_shrugging: . I'll let y'all hash it out and then make changes as needed :sweat_smile: .

As it pertains to Coder -- These aren't supposed to end up there AFAIK. This change is a first step towards injecting into Claude context for DOs. Maybe I'm confused, though :man_shrugging: .

src/app/[orgSlug]/admin/settings/use-data-source-form.ts
Comment (rnathuji) [reply src/app/[orgSlug]/admin/settings/use-data-source-form.ts]: Yeah, TBH, I stole this form style from [here](https://github.com/safeinsights/management-app/blob/769b292e674c4aaa831950501d8a8c9da394d170/src/app/%5BorgSlug%5D/admin/settings/use-code-env-form.ts#L206-L211) where we add environment variables on submit if present (I'm assuming because it's easy for users to otherwise get confused that it was not added). But, to me that was kinda broken because those fields don't get validated, so users can inadvertently add bad values (even worse, those values then prevent subsequent saving w/o a clear error as to why :grimacing: ). So, I went this route as a lesser evil (e.g. it "auto adds to the form", but validation is consistent and it'll nudge the user right away before saving if that row is problematic). I'm okay with another approach, but probably we should address both forms in the same way :man_shrugging: .

rnathuji:
@nathanstitt - Oh, as an added note on the naming, I think I also [suggested](https://openstax.atlassian.net/browse/SHRMP-220?focusedCommentId=40571) `org_data_source_document` to @lwtchu since the original column we were shifting from 1:1 to 1:N was called `documentation_url`.  Assuming it even makes sense to call the new table `org_data_source_{foo}` , I am totally open to what `{foo}` should be if `document` creates an incorrect mental model for the future.

src/database/migrations/1776461687255_org_data_source_doc.ts
rnathuji: Addressed in [this commit](https://github.com/safeinsights/management-app/pull/664/changes/83d4a1ed56ff1de0c2d31ccbd115c26612f2f699).

src/app/[orgSlug]/admin/settings/data-sources.actions.ts
Comment (rnathuji) [reply src/app/[orgSlug]/admin/settings/data-sources.actions.ts]: Addressed in [this commit](https://github.com/safeinsights/management-app/pull/664/changes/6758284254dd0c0612165a2c4a54c561b14bb657)

rnathuji:
> This mentioned 'documents' a lot, but we're not dealing with documents, it's URLs or maybe "paths". Reading documents made me think we were uploading files at first.

Addressed in [this commit](https://github.com/safeinsights/management-app/pull/664/changes/259474c71a0c3d4094673f8868bd68922a13c940)

src/app/[orgSlug]/admin/settings/use-data-source-form.ts
Comment (nathanstitt) [reply src/app/[orgSlug]/admin/settings/use-data-source-form.ts]: Ah, I see what you were attempting there.  In the env code though notice it only mutates the form values which is a plan JS object that is copied from the form and modifying them doesn't affect the form itself.  (Even so it would probably have been clearer if that code did an outright copy to be save but 🤷)

What you're doing is calling `form.setValues(...)`  which causes the form to re-render.  That's what seemed dangerous to me.

Instead why don't we do something like:

```
 const onSubmit = form.onSubmit(({ newDocumentUrl, newDocumentDescription, ...values }) => {
        // Handle the case where a user may have document data in-progress. We proactively add the
        // document and then validate to surface errors. If everything is fine, we save().
        if (newDocumentUrl !== '' || newDocumentDescription !== '') {
            values = {
                 ...values,
                 documents: [
                ...form.values.documents,
                { url: form.values.newDocumentUrl, description: form.values.newDocumentDescription },
            ],
            // validity checks are not needed since onSubmit handler only gets called if form is valid, and we've validated that newDocumentUrl is valid.  to be super efficient we could have zod use a function that we re-use in the if statement, but this is probably fine since it's simple.
            // form.validate() 
            //if (!form.isValid()) {
            //    return
            //}
        }
        save(values)
    })
```

Comment (nathanstitt) [reply src/app/[orgSlug]/admin/settings/use-data-source-form.ts]: also to confirm your suspicions, the need for this is because it's really easy to type something in the last box, but not hit + to add it to documents.  Then if you save the form without this that entry is lost

Comment (rnathuji) [reply src/app/[orgSlug]/admin/settings/use-data-source-form.ts]: > which causes the form to re-render. That's what seemed dangerous to me.

I see the point / hesitation with that approach. I've incorporated your suggestion in [this commit](https://github.com/safeinsights/management-app/pull/664/changes/b9f1f2e3aa6647907510d3f24053ef64a86b425f).

PR #666, opened 2026-05-04
Normalize UUID defaults to v7uuid() and timestamp columns to timestamptz

nathanstitt:
## Summary
- Switches three legacy `gen_random_uuid()` id defaults to `v7uuid()` (`org_data_source.id`, `pending_user.id`, `researcher_position.id`) so all tables use time-ordered v7 UUIDs for better index locality
- Converts 24 `timestamp` (without time zone) columns to `timestamptz` across `audit`, `job_status_change`, `org`, `org_code_env`, `org_user`, `pending_user`, `researcher_position`, `researcher_profile`, `study`, `study_job`, `study_job_file`, `user`, `user_public_key`, and `yjs_document`
- Documents both conventions as Hard Rules in `CLAUDE.md` so future migrations follow the pattern

The migration uses `AT TIME ZONE 'UTC'` when converting, treating existing naked timestamps as UTC. `down()` reverses both changes.

## Test plan
- [ ] `npm run db:migrate` runs cleanly on a fresh DB
- [ ] After migration, no `public` columns have `gen_random_uuid()` default or `timestamp without time zone` type (verified via `information_schema.columns`)
- [ ] Down migration reverts to prior state
- [ ] Existing unit tests still pass (S3 + lexical-seed failures observed locally are pre-existing, unrelated to this change)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -9,6 +9,8 @@
 - Comments: Only add comments that explain "why", not "what". Avoid trivial comments like `// Delete users` before `deleteFrom('user')` or `// Create profile` before `insertInto('profile')`. If the code is self-explanatory, no comment is needed.
 - Testing: Write unit tests for new features. Only mock using helpers in tests/unit.helpers.tsx as needed, do not mock out any of our own components or actions.
 - E2E timeouts: Do not set inline timeouts in Playwright tests. Timeouts are configured globally in `playwright.config.ts` using constants from `tests/common.helpers.ts`.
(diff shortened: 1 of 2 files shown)


chrisbendel (approved):

therealmarv:
I just want to mention it:

This PR makes the strong assumption that the server which **saved values in the DB is running with UTC clock.** Are we 100% sure about that? If (non time zone) time values saved are not UTC time we get clock shifts with this DB migration!

Postgres does not normalise to UTC for timestamp (no time zone).

(maybe it's also not important at this stage)

nathanstitt:
@therealmarv I think it's the other way around - the plain `timestamp` columns assumed everything was UTC, now we'll be storing the actual timezone information so we can convert from one to the other as needed.

PR #667, opened 2026-05-04
OTTER 497: more tests

therealmarv:
This PR increases test coverage of [OTTER-497](https://openstax.atlassian.net/browse/OTTER-497)

Only tests are modified and added here.

This PR tries to address these multi editor test gaps with unit tests (a real end to end test requires multiple clerk test users in same org which we don't have + Hocuspocus editor service running):

1. **Lab-collaboration visibility is locked in.** Tests prove that lab teammates can see and edit each other's DRAFT and CHANGE-REQUESTED studies, and that cross-lab access is denied. Today only the enclave branch is exercised, so the core OTTER-497 user claim ("my teammate can pick up where I left off") is unverified.
2. **The Y.js form bridge gets its first real coverage.** Two-peer round-trip, cold seed, warm load from persisted CRDT state, remote-vs-local origin tagging. This is the closest unit-level substitute for "user A types, user B sees", minus the WebSocket layer.
3. **Broadcast wiring on both sides is contract-tested.** `useSubmitProposal` (researcher submits) and `useProposalReviewMutation` (reviewer decides) both push stateless events; today nothing catches a silent rename like `submitterFullName -> submitter_name` that would break the redirect listener in prod.
4. **`getDraftStudyAction` goes from zero tests to four.** Original creator, lab teammate, cross-lab rejection, and non-editable status, all paths.

[OTTER-497]: https://openstax.atlassian.net/browse/OTTER-497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/use-proposal-review-mutation.test.ts
+++ b/src/hooks/use-proposal-review-mutation.test.ts
@@ -1,0 +1,244 @@
+import { vi } from 'vitest'
+import {
+    act,
+    afterEach,
+    beforeEach,
+    buildFeedback,
+    createTestQueryWrapper,
+    db,
+    describe,
+    expect,
+    faker,
+    insertTestStudyJobData,
+    it,
+    mockSessionWithTestData,
+    renderHook,
+    resetSpyMode,
+    setSpyMode,
+    setTestStudyStatus,
+    spyModeState,
+    waitFor,
+    type Mock,
+} from '@/tests/unit.helpers'
+import { memoryRouter } from 'next-router-mock'
+import { notifications } from '@mantine/notifications'
+import { Routes } from '@/lib/routes'
+import { useProposalReviewMutation } from './use-proposal-review-mutation'
+
+vi.mock('@/components/spy-mode-context', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('@/components/spy-mode-context')>()
+    return {
+        ...actual,
+        useSpyMode: () => ({ isSpyMode: spyModeState.isSpyMode, toggleSpyMode: vi.fn() }),
+    }
+})
+
+type Listener = (...args: unknown[]) => void
+
+type ProviderHandle = {
+    sendStateless: ReturnType<typeof vi.fn>
+}
+
+vi.mock('@hocuspocus/provider', () => {
+    const constructed: ProviderHandle[] = []
+
+    class HocuspocusProvider {
+        sendStateless = vi.fn()
+        private syncListeners: Listener[] = []
+
+        constructor() {
+            constructed.push({ sendStateless: this.sendStateless })
+        }
+        attach() {}
+        on(event: string, fn: Listener) {
+            if (event === 'synced') this.syncListeners.push(fn)
+        }
+        off(event: string, fn: Listener) {
+            if (event === 'synced') this.syncListeners = this.syncListeners.filter((l) => l !== fn)
+        }
+        destroy() {}
+    }
+
+    class HocuspocusProviderWebsocket {
+        constructor(_opts?: unknown) {}
+        destroy() {}
+    }
+
+    return {
+        HocuspocusProvider,
+        HocuspocusProviderWebsocket,
+        __constructed: constructed,
+    }
+})
+
+import * as HocuspocusModule from '@hocuspocus/provider'
+
+const constructed = (HocuspocusModule as unknown as { __constructed: ProviderHandle[] }).__constructed
+
+const validFeedback = buildFeedback(60)
+
+describe('useProposalReviewMutation', () => {
+    let tabSessionId: string
+
+    beforeEach(() => {
+        tabSessionId = faker.string.uuid()
+        constructed.length = 0
+        memoryRouter.setCurrentUrl('/start')
+        ;(notifications.show as Mock).mockClear()
+    })
+
+    afterEach(() => {
+        resetSpyMode()
+    })
+
+    it('approve decision broadcasts and navigates when the flag is ON', async () => {
+        const { user, org } = await mockSessionWithTestData({ orgSlug: 'openstax', orgType: 'enclave' })
+        const { study } = await insertTestStudyJobData({
+            org,
+            researcherId: user.id,
+            studyStatus: 'PENDING-REVIEW',
+        })
+        setSpyMode(true)
+
+        const { result } = renderHook(
+            () => useProposalReviewMutation({ studyId: study.id, orgSlug: org.slug, tabSessionId }),
+            { wrapper: createTestQueryWrapper() },
+        )
+
+        await waitFor(() => expect(constructed).toHaveLength(1))
+        const handle = constructed[0]
+
+        await act(async () => {
+            result.current.submitReview({ decision: 'approve', feedback: validFeedback })
+        })
+        await waitFor(() => expect(result.current.isSuccess).toBe(true))
+
+        const updated = await db
+            .selectFrom('study')
+            .select('status')
+            .where('id', '=', study.id)
+            .executeTakeFirstOrThrow()
+        expect(updated.status).toBe('APPROVED')
+
+        expect(handle.sendStateless).toHaveBeenCalledTimes(1)
+        const payload = JSON.parse(handle.sendStateless.mock.calls[0][0] as string)
+        expect(payload.type).toBe('proposal-review-submitted')
+        expect(payload.studyId).toBe(study.id)
+        expect(payload.submittedByTabId).toBe(tabSessionId)
+        expect(typeof payload.submittedByName).toBe('string')
+        expect(payload.submittedByName.length).toBeGreaterThan(0)
+
+        await waitFor(() =>
+            expect(memoryRouter.asPath).toBe(Routes.studyReview({ orgSlug: org.slug, studyId: study.id })),
+        )
+    })
+
+    it.each([
+        { decision: 'needs-clarification' as const, expectedStatus: 'CHANGE-REQUESTED' as const },
+        { decision: 'reject' as const, expectedStatus: 'REJECTED' as const },
+    ])('$decision decision broadcasts with the same wiring (flag ON)', async ({ decision, expectedStatus }) => {
+        const { user, org } = await mockSessionWithTestData({ orgSlug: 'openstax', orgType: 'enclave' })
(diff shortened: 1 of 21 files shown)


src/hooks/use-proposal-review-mutation.test.ts
```diff
@@ -0,0 +1,244 @@
...
+    spyModeState,
+    waitFor,
+    type Mock,
+} from '@/tests/unit.helpers'
+import { memoryRouter } from 'next-router-mock'
+import { notifications } from '@mantine/notifications'
+import { Routes } from '@/lib/routes'
+import { useProposalReviewMutation } from './use-proposal-review-mutation'
+
+vi.mock('@/components/spy-mode-context', async (importOriginal) => {
```
nathanstitt: **Two issues here:**

1. **Module-mutable `spyModeState` cross-file leak risk.** This factory reads from `spyModeState` at module scope (defined in `tests/unit.helpers.tsx`). The `afterEach(resetSpyMode)` below mitigates within this file, but any future test file that calls `setSpyMode(true)` and forgets to reset will leak the flag into the next test in the same worker. Consider moving `resetSpyMode()` into the global `afterEach` in `tests/vitest.setup.ts` so every test starts from `isSpyMode: false` regardless of file. Alternatively, drop the pattern entirely in favour of `vi.spyOn(spyModeContext, 'useSpyMode').mockReturnValue(...)` per-test, which auto-resets via `vi.restoreAllMocks()`.

2. **Mocking two layers deep.** This targets `useSpyMode`, which is one layer below the real flag (`useProposalCollaborationFeatureFlag → useOpenStaxFeatureFlag → useSpyMode`). If the feature-flag plumbing ever switches mechanism (real GrowthBook, etc.), every test using `setSpyMode` breaks at once even though the contract under test is just "flag on/off." Mocking `useProposalCollaborationFeatureFlag` directly removes a layer of indirection and decouples the tests from the spy-mode/openstax flag implementation.

src/hooks/use-proposal-review-mutation.test.ts
```diff
@@ -0,0 +1,244 @@
...
+    }
+})
+
+type Listener = (...args: unknown[]) => void
+
+type ProviderHandle = {
+    sendStateless: ReturnType<typeof vi.fn>
+}
+
+vi.mock('@hocuspocus/provider', () => {
```
nathanstitt: **Duplicated Hocuspocus mock factory.** `use-yjs-form-map.test.ts` ships a similar but slightly different `vi.mock('@hocuspocus/provider', ...)` factory (that one carries a real `Y.Doc` on each provider; this one doesn't). Two divergent fakes for the same module risks them drifting out of sync with the real Hocuspocus API.

**Recommendation:** extract a `tests/hocuspocus.mock.ts` helper exporting a factory like `createHocuspocusMock({ withYDoc?: boolean })` returning `{ HocuspocusProvider, HocuspocusProviderWebsocket, __constructed }`. Both test files apply via `vi.mock('@hocuspocus/provider', () => createHocuspocusMock())`.

src/hooks/use-yjs-form-map.test.ts
```diff
@@ -0,0 +1,296 @@
...
+    formInitial,
+    enabled = true,
+}: {
+    studyId: string
+    formInitial: Partial<ProposalFormValues>
+    enabled?: boolean
+}) => {
+    const { result: formResult } = buildProposalForm(formInitial)
+    const form = formResult.current
+    const websocketProvider = new (HocuspocusModule.HocuspocusProviderWebsocket as unknown as new () => InstanceType<
```
nathanstitt: **Awkward constructor cast.** This works but is opaque at the call site. A small helper hides the cast:

```ts
const newWebsocketProvider = () => new HocuspocusModule.HocuspocusProviderWebsocket(undefined as never)
```

Then `setupCollabHook` reads cleanly. Cosmetic but worth doing while the file is fresh.

src/hooks/use-yjs-form-map.test.ts
```diff
@@ -0,0 +1,296 @@
...
+        expect(fieldsMap!.get('datasets')).toEqual(['ds-1'])
+        expect(fieldsMap!.get('piName')).toBe('PI')
+        expect(fieldsMap!.get('piUserId')).toBe(piUserId)
+    })
+
+    it('warm load: applies persisted CRDT state to the form when a yjsDocument row already exists', async () => {
+        const { studyId } = await createDraftStudy('warm', 'OriginalForm')
+
+        await db
+            .insertInto('yjsDocument')
```
nathanstitt, line 101: **The `yjsDocument` insert isn't load-bearing.** This warm-load test inserts a row into `yjsDocument` and then *additionally* does `Y.applyUpdate(handle.document, ...)` to simulate the server pushing persisted state. Since the mocked Hocuspocus provider never reads from the DB, the `Y.applyUpdate` call is what actually drives the assertion — the DB insert is unused state that misleadingly suggests "this is testing the warm-load DB path."

**Recommendation:** drop the `db.insertInto('yjsDocument')` block. If we want to verify the actual server-side load path, that's a real Hocuspocus integration test (out of scope for this PR per the description).

tests/unit.helpers.tsx
```diff
@@ -598,6 +614,49 @@ export async function mockSessionWithTestData(options: MockSessionWithTestDataOp
...
+// Builds the canonical OTTER-497 fixture shape: enclave + lab + lab-member session +
+// DRAFT study where `submittedByOrgId` is the lab and `orgId` is the enclave. Use this
+// instead of `insertTestStudyOnly` for collaboration tests, which collapse both ids to
+// the same org and do not match the production submitting-lab vs reviewing-enclave split.
+//
+// `onSaveDraftStudyAction` is lazy-imported because eagerly importing it from this
+// always-loaded helper file would pull `@/server/aws` into every test's module graph,
+// which races with module-level mocks like `aws.test.ts` does for `./config`.
+export async function createTestProposalDraft({ enclaveSlug, studyInfo = {} }: CreateTestProposalDraftOptions) {
+    const { onSaveDraftStudyAction } = await import('@/server/actions/study-request')
```
nathanstitt: **Add a removal criterion to the lazy import.** The comment above explains the workaround for the `@/server/aws` self-mock race. It's load-bearing — without a concrete trigger, future maintainers may either remove the laziness without understanding why (re-introducing the race) or keep it forever as cargo culting.

**Recommendation:** add a `TODO: switch to a top-level import once @/server/aws no longer self-mocks at module scope` so the next person has a clear removal criterion.

src/server/actions/study-request.test.ts
```diff
@@ -541,44 +530,172 @@ describe('Request Study Actions', () => {
...
+            const { enclave, studyId } = await createTestProposalDraft({
+                enclaveSlug: 'getdraft-cross-enclave',
+                studyInfo: { title: 'LabA Draft' },
+            })
+            const labB = await insertTestOrg({ slug: `${enclave.slug}-lab-b`, type: 'lab' })
+
+            await mockSessionWithTestData({ orgSlug: labB.slug, orgType: 'lab' })
+            const result = await getDraftStudyAction({ studyId })
+            expect(result).toMatchObject({
+                error: expect.objectContaining({ permission_denied: expect.any(String) }),
```
nathanstitt: **Vague error-shape assertion.** `expect.any(String)` accepts any error message. If the action's failure path ever returns a different `permission_denied` string (or a different shape entirely), this test still passes silently.

**Recommendation:** match the exact message the action throws today, or — if the goal is to be resilient to copy changes — assert on a stable error code rather than free text. Same applies to other `permission_denied: expect.any(String)` matchers in this file.

src/server/actions/study-request.test.ts
```diff
@@ -541,44 +530,172 @@ describe('Request Study Actions', () => {
...
+                enclaveSlug: 'getdraft-pending-enclave',
+                studyInfo: { title: 'Soon-PR' },
+            })
+
+            await setTestStudyStatus(studyId, 'PENDING-REVIEW')
+
+            const result = await getDraftStudyAction({ studyId })
+            // Middleware filters status before CASL check; falls into not-found.
+            expect(result).toMatchObject({
+                error: expect.objectContaining({ user: expect.stringMatching(/not found/i) }),
```
nathanstitt: **Brittle non-editable-status assertion.** The implementation comment says "Middleware filters status before CASL check; falls into not-found." If middleware/CASL ordering ever changes, the result becomes a `permission_denied` instead of a `not found`, and the regex flips. The test would catch the *change* but the assertion is too vague to tell us whether the new behaviour is correct.

**Recommendation:** match the exact string the action throws, or assert on a stable error code rather than the `/not found/i` regex.

nathanstitt (commented):
Test-coverage PR with high leverage on the OTTER-497 gaps (lab visibility, Yjs bridge, broadcaster contracts, `getDraftStudyAction`). The Yjs hook tests are particularly good — driving a real `Y.Doc` through a structurally-mocked provider gets us closest to a true two-peer round-trip without the WebSocket layer.

One should-fix concern (module-mutable `spyModeState` cross-file leak risk) plus a handful of nice-to-haves around test infrastructure cohesion. Inline below.

src/hooks/use-yjs-form-map.test.ts
nathanstitt: also see the other comment about using built-in mock, not sure if that'll work here as well

tests/unit.helpers.tsx
therealmarv: Good point. I ended up fixing the root cause instead of adding the TODO.

The real issue was not only `@/server/aws` self-mocking. `tests/vitest.setup.ts` imported `createTempDir` from `tests/unit.helpers.tsx`, which meant the whole helpers module and its transitive imports were evaluated before test-file `vi.mock` calls had a chance to register. That made the lazy import load-bearing.

I moved `createTempDir` into a small leaf helper, so setup no longer preloads `unit.helpers`. After that, `createTestProposalDraft` can safely use a normal top-level import again.
That exposed a few partial mocks that were depending on the old accidental load order, so I converted those to the Vitest `importOriginal` partial mock pattern. Full checks and unit tests pass now.

PR #668, opened 2026-05-04
feat: Implement edit and resubmit functionality for study proposals

Rod0352:
Adds the researcher-side "Edit and resubmit" flow for studies in CHANGE-REQUESTED. New /edit-and-resubmit page with:

proposal form in edit mode
feedback and notes section
mandatory resubmission note (50–300 words)
auto-save
confirm modal on submit

On submit: writes a studyProposalComment row (entryType: RESUBMISSION-NOTE) and flips the study to PENDING-REVIEW. Gated behind the OpenStax / spy-mode flag.
Tests: schema bounds, auto-save indicator labels, resubmission-note section, DB-backed action tests.


src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/types.ts
```diff
@@ -0,0 +1,20 @@
+import type {
+    Json,
+    ReviewDecision,
+    StudyProposalCommentAuthorRole,
+    StudyProposalCommentEntryType,
+} from '@/database/types'
+
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/types.ts]: Good news — `ProposalFeedbackEntry` is already exported from `@/server/actions/study.actions` (line 611 on main), so this local mirror isn't needed anymore! In fact `form.tsx` is already importing the canonical one from there. Once the dead `feedback-and-notes-section.tsx` is removed, this file can probably go too.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const study = await getStudyAction({ studyId })
+
+    if ('error' in study) return notFound()
+    if (study.status !== 'CHANGE-REQUESTED') return notFound()
+
+    // Inlined here rather than importing OTTER-501's getProposalFeedbackForStudyAction
+    // so OTTER-521 ships independently. When 501 lands, swap to that action.
+    const entries: ProposalFeedbackEntry[] = await db
+        .selectFrom('studyProposalComment')
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx]: Quick win here: `getProposalFeedbackForStudyAction` is already on main in `study.actions.ts` (OTTER-501 landed!), so we can swap this inline query for the action and drop ~16 lines. The action also runs `requireAbilityTo('view', 'Study')` which we'd otherwise be bypassing — using it tightens the auth story for free.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx
```diff
@@ -0,0 +1,80 @@
...
+                            : undefined,
+                        projectSummary: study.projectSummary ? JSON.stringify(study.projectSummary) : undefined,
+                        impact: study.impact ? JSON.stringify(study.impact) : undefined,
+                        additionalNotes: study.additionalNotes ? JSON.stringify(study.additionalNotes) : undefined,
+                    }}
+                >
+                    <EditResubmitForm
+                        orgName={displayOrgName(enclaveOrg?.name ?? '')}
+                        members={memberOptions}
+                        researcherName={study.createdBy}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx:58]: Quick sanity check — is `study.createdBy` here the user's display name, or their id? The prop is named `researcherName` so I assume `SelectedStudy.createdBy` aliases to `fullName`, just want to make sure we're not displaying a UUID on the page.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/feature-flag-gate.tsx
```diff
@@ -0,0 +1,34 @@
...
+// When off, redirects to the existing /view page so the researcher still has a path forward.
+export function FeatureFlagGate({ children, orgSlug, studyId }: FeatureFlagGateProps) {
+    const isOptIn = useEditAndResubmitProposalFeatureFlag()
+    const router = useRouter()
+
+    if (isOptIn) return <>{children}</>
+
+    // Schedule a client-side redirect; render a fallback in the meantime.
+    if (typeof window !== 'undefined') {
+        router.replace(Routes.studyView({ orgSlug, studyId }))
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/feature-flag-gate.tsx]: `router.replace` is being called during render — React would prefer this happen inside a `useEffect` so it doesn't schedule updates while another component renders.

src/server/actions/study-request.ts
```diff
@@ -606,3 +607,109 @@ export const submitStudyCodeAction = new Action('submitStudyCodeAction', { perfo
...
+            proposalUpdatableFields.filter((k) => studyInfo[k] !== undefined).map((k) => [k, studyInfo[k]]),
+        )
+
+        if (Object.keys(updateValues).length > 0) {
+            await db
+                .updateTable('study')
+                .set(updateValues)
+                .where('id', '=', studyId)
+                .where('status', '=', 'CHANGE-REQUESTED')
+                .where('researcherId', '=', userId)
```
nathanstitt, line 666: Worth a look — this `where('researcherId', '=', userId)` filter means if it doesn't match we silently no-op (success returned, zero rows touched). If we trust `requireAbilityTo('update', 'Study')` to scope correctly we can drop the filter; if CASL doesn't already scope by researcher, the filter is load-bearing and we probably want to throw on no-match instead of silently succeeding. Either way feels worth nailing down.

src/server/actions/study-request.ts
```diff
@@ -606,3 +607,109 @@ export const submitStudyCodeAction = new Action('submitStudyCodeAction', { perfo
...
+    .middleware(async ({ params: { studyId } }) => await getInfoForStudyId(studyId))
+    .requireAbilityTo('update', 'Study')
+    .handler(async ({ db, params: { studyId, studyInfo, resubmissionNote }, session, orgSlug }) => {
+        const userId = session.user.id
+
+        const study = await db
+            .selectFrom('study')
+            .select(['id', 'status'])
+            .where('id', '=', studyId)
+            .where('researcherId', '=', userId)
```
nathanstitt, line 695: Same auth question as `onUpdateClarifiedProposalAction` — we've got both `requireAbilityTo('update', 'Study')` and a manual `where('researcherId', '=', userId)` ownership check. One of them is redundant. Worth picking a consistent pattern across both actions so future-us isn't second-guessing which check is doing the work.

src/server/actions/study-request.ts
```diff
@@ -606,3 +607,109 @@ export const submitStudyCodeAction = new Action('submitStudyCodeAction', { perfo
...
+
+// Final resubmission: writes the latest proposal edits, records the
+// resubmission note as a study_proposal_comment row, and transitions
+// CHANGE-REQUESTED -> PENDING-REVIEW.
+export const resubmitProposalAction = new Action('resubmitProposalAction', { performsMutations: true })
+    .params(
+        z.object({
+            studyId: z.string(),
+            studyInfo: draftStudyApiSchema,
+            resubmissionNote: z.string().min(1),
```
nathanstitt: Tiny nitpick — server-side we only enforce `min(1)` on the resubmission note, but the client enforces 50–300 words. Mirroring the word-count bound here would be nice defense in depth (a clever client could submit a 1-word note today).

src/hooks/use-auto-save.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        if (timerRef.current) clearTimeout(timerRef.current)
+        timerRef.current = setTimeout(() => {
+            void saveDraft({ silent: true })
+        }, DEBOUNCE_MS)
+
+        return () => {
+            if (timerRef.current) clearTimeout(timerRef.current)
+        }
+    }, [isDirty, isSaving, saveDraft])
```
nathanstitt: Two things to think about with this effect setup:

1. The deps include `isSaving`, which flips on every save — so each successful save tears down and rebuilds the timer even when the user hasn't touched anything.
2. The 45s interval effect below has the same dep array, so the interval restarts on every save too.

Simplest fix: read `isSaving` via a ref (or just check `isDirty` inside the timer callback) so the deps stay stable. Functionally correct today, just a bit more thrash than needed.

src/contexts/proposal/hooks/use-save-draft.ts
```diff
@@ -24,15 +24,23 @@ interface UseSaveDraftOptions {
 }
 
 export function useSaveDraft({ studyId, form }: UseSaveDraftOptions) {
+    const [lastSavedAt, setLastSavedAt] = useState<Date | null>(null)
+    // Auto-save calls saveDraft({ silent: true }) so the toast doesn't fire every 2.5s.
+    // Explicit clicks still get the toast confirmation.
+    const isSilentRef = useRef(false)
```
nathanstitt: Thought experiment — if a user clicks "Save as draft" while an auto-save is in flight, the second `saveDraft({ silent: false })` call overwrites `isSilentRef.current` *before* the first call's `onSuccess` runs. So the auto-save would end up firing the toast (or vice versa). Passing `silent` through as a mutation variable would dodge the race and be a bit cleaner overall.

src/components/study/auto-save-indicator.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+        expect(screen.queryByText(/last saved/i)).not.toBeInTheDocument()
+    })
+
+    it('shows "just now" right after a save', () => {
+        renderWithProviders(<AutoSaveIndicator isSaving={false} lastSavedAt={new Date()} />)
+        expect(screen.getByText(/last saved just now/i)).toBeInTheDocument()
+    })
+
+    it('shows minutes elapsed when more than a minute has passed', () => {
+        const fiveMinAgo = new Date(Date.now() - 5 * 60_000)
```
nathanstitt: Tiny flake risk — `new Date(Date.now() - 5 * 60_000)` then a synchronous render is fine 99% of the time but could land just over a minute boundary on slow CI and assert `6 minutes ago`. Injecting `now` (or using `vi.useFakeTimers`) would make this rock solid.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/edit-initial-request-section.tsx
```diff
@@ -0,0 +1,201 @@
...
+}
+
+const EditableField: FC<{
+    field: EditableTextField
+    value: string
+    error: string | undefined
+    onChange: (val: string) => void
+    onBlur: () => void
+}> = ({ field, value, error, onChange, onBlur }) => {
+    const [wordCount, setWordCount] = useState(0)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/edit-initial-request-section.tsx]: we should initialize this with the current count, some how collaborative-proposal-text-field does it

src/server/actions/resubmit-proposal.test.ts
```diff
@@ -0,0 +1,143 @@
+// OTTER-521: DB-backed tests for resubmitProposalAction and
```
nathanstitt, line 1: Tests look great — happy path + status guard + comment row creation all covered. One small gap: no test for the silent ownership check (e.g. user A tries to update user B's study) — would catch regressions if we ever loosen the `where researcherId` filter or the CASL rule. Not blocking, just a nice-to-have.

src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx
```diff
@@ -64,7 +64,7 @@ export const ProposalFooter: FC<ProposalFooterProps> = ({ researcherName, resear
                         size="md"
                         disabled={!form.isDirty() || isBusy}
                         loading={isSaving}
-                        onClick={saveDraft}
+                        onClick={() => saveDraft()}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx]: Heads up — wrapping `saveDraft` in `() => saveDraft()` is needed because the new signature accepts `opts` and Mantine's `Button` would otherwise pass the click event through. Functionally correct! A named handler (`handleSaveDraftClick`) might make intent more obvious to the next reader, but totally optional.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/feature-flag-gate.tsx
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    return (
+        <AlertNotFound
+            title="Page not available"
+            message="The edit & resubmit flow is not yet enabled for your organization."
+        />
+    )
+}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/feature-flag-gate.tsx:22]: A few things.  We should name this with 'ProposalReviewFeatureFlag' so it's findable by the feature.  Also: how does this differ from `ProposalReviewFeatureFlag`?  seems almost identical. 

seems like this could be:
```
<ProposalReviewFeatureFlag optInContent={children} defaultContent={<AlertNotFound ... />} />
```

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/feedback-and-notes-section.tsx
```diff
@@ -0,0 +1,92 @@
+'use client'
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/feedback-and-notes-section.tsx]: this file seems duplicated?   form.tsx actually imports FeedbackAndNotesSection from     @/components/study/feedback-and-notes (the existing shared component), so this local copy isn't wired up to anything.   the shared one uses ReadOnlyLexicalContent so it preserves formatting, while this version flattens to plain text

src/components/study/auto-save-indicator.tsx
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    if (!lastSavedAt) return null
+
+    return (
+        <Text size="xs" c="charcoal.7" aria-live="polite">
+            {formatMinutesAgo(lastSavedAt, now)}
+        </Text>
+    )
+}
```
nathanstitt: is this specced somewhere?  We are _always_ autosaving, so I don't see why we're lying and saying it's only every 5 minutes

src/components/openstax-feature-flag.tsx
```diff
@@ -62,4 +62,7 @@ export {
     // OTTER-497: Multi-user proposal collaboration
     useOpenStaxFeatureFlag as useProposalCollaborationFeatureFlag,
     OpenStaxFeatureFlag as ProposalCollaborationFeatureFlag,
+    // OTTER-521: Edit & resubmit proposal (RL) — testing 2026-05-11
+    useOpenStaxFeatureFlag as useEditAndResubmitProposalFeatureFlag,
+    OpenStaxFeatureFlag as EditAndResubmitProposalFeatureFlag,
```
nathanstitt: I think we can use the existing name, even if it's slightly inaccurate  we will remove all flags at the same time so it's fine

nathanstitt (commented):
I'm pausing review so we can discuss the auto-save implementation you have here. I don't think it's needed but will check with PMs and get back to you

src/components/study/auto-save-indicator.tsx
Rod0352: Pulling auto-save out of this PR — wasn't actually scoped to 521; Iris flagged it as an open question for the proposal flow more broadly. Better as its own card.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx:58]: Yes, createdBy is the full name

nathanstitt:
HI @Rod0352 thanks for working through this.  We've decided _not_ to auto-save in the MA because the editor service does this for us.  UX is working on a updated design that removes the little "Autosaving..." message as well.

src/lib/routes/definitions.ts
```diff
@@ -109,6 +109,11 @@ export const Routes = {
 
     studyResubmit: makeRoute(({ orgSlug, studyId }) => `/${orgSlug}/study/${studyId}/resubmit`, StudyParams),
 
+    studyEditAndResubmit: makeRoute(
```
nathanstitt, line 112: Is this used? I didn't see antthing that actually navigates to `studyEditAndResubmit`. A `grep -rn '/edit-and-resubmit\|studyEditAndResubmit'` only matches the new files themselves

src/server/actions/study-request.ts
```diff
@@ -606,3 +611,124 @@ export const submitStudyCodeAction = new Action('submitStudyCodeAction', { perfo
...
+        const updateValues = Object.fromEntries(
+            proposalUpdatableFields.filter((k) => studyInfo[k] !== undefined).map((k) => [k, studyInfo[k]]),
+        )
+
+        await db
+            .updateTable('study')
+            .set({
+                ...updateValues,
+                status: 'PENDING-REVIEW',
+                submittedAt: new Date(),
```
nathanstitt: Setting `submittedAt: new Date()` on every resubmit overwrites the original first-submission timestamp.  It seems like we should retain the original date and maybe use comments or add a lastSubmittedAt column if we really care about this?

We could rely on `studyProposalComment` rows as the timeline source of truth instead?

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx
```diff
@@ -0,0 +1,63 @@
...
+}) {
+    const { studyId, orgSlug } = await props.params
+
+    const study = await getStudyAction({ studyId })
+
+    if ('error' in study) return notFound()
+    if (study.status !== 'CHANGE-REQUESTED') return notFound()
+
+    const entriesResult = await getProposalFeedbackForStudyAction({ studyId })
+    const entries = 'error' in entriesResult ? [] : entriesResult
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx]: If `getProposalFeedbackForStudyAction` fails we silently render the page with an empty feedback list. That puts the researcher in an awkward spot — they're being asked to *address* feedback but we've hidden the fact that we couldn't load it. Could we `notFound()` (or render an inline error state) instead of `[]`? At least surface something so they don't unknowingly resubmit without addressing the DO's notes.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/edit-initial-request-section.tsx
```diff
@@ -0,0 +1,201 @@
...
+                    <FormFieldLabel label="Study title" required inputId="title" />
+                    <Text size="xs" c="charcoal.7" mb="xs">
+                        Give your study a short, clear title.
+                    </Text>
+                    <TextInput
+                        id="title"
+                        aria-label="Study Title"
+                        placeholder="Ex. Impact of highlighting on student learning outcomes."
+                        {...form.getInputProps('title')}
+                        value={form.values.title === DEFAULT_DRAFT_TITLE ? '' : form.values.title}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/edit-initial-request-section.tsx:101]: The display value is masked when it equals `DEFAULT_DRAFT_TITLE`, but `form.values.title` is still the default sentinel until the user types. If a user opens the page and immediately clicks Resubmit, does `proposalFormSchema` reject the literal default title? If not, they could resubmit with the placeholder title intact. Worth a quick check — if the schema doesn't already guard against this, we should reject `DEFAULT_DRAFT_TITLE` explicitly here.

nathanstitt (approved):
Looks good Rodney, most of the comments are pretty minor, feel to disregard if you disagree

PR #671, opened 2026-05-05
OTTER 493: fix duplicated proposal prefix in change-requested status label

therealmarv:
fixes issue described [here](https://openstax.atlassian.net/browse/OTTER-492?focusedCommentId=41223)

- Fixed a bug where the change-requested status badge displayed the word "Proposal" twice and caused text to overlap neighboring content in the reviewer dashboard.
- Added a test suite that pins the exact rendered text for every status across both reviewer and researcher audiences to prevent regressions.

### DIFF
--- a/src/components/study/display-study-status.test.tsx
+++ b/src/components/study/display-study-status.test.tsx
@@ -25,7 +25,9 @@ describe('DisplayStudyStatus', () => {
         it.each(cases)('renders %s as "%s"', (statusKey, expectedText) => {
             const status = REVIEWER_STATUS_LABELS[statusKey]!
             renderWithProviders(<DisplayStudyStatus status={status} />)
-            expect(screen.getByText(expectedText)).toBeDefined()
+            const textEl = screen.getByText(expectedText)
+            expect(textEl).toBeDefined()
+            expect(textEl.parentElement?.textContent?.trim()).toBe(expectedText)
         })
     })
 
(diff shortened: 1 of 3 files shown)


src/components/study/display-study-status.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+            ['JOB-ERRORED', 'Code errored'],
+            ['RUN-COMPLETE', 'Result needs review'],
+            ['FILES-APPROVED', 'Result ready'],
+            ['FILES-REJECTED', 'Result rejected'],
+        ]
+
+        it.each(cases)('renders %s as "%s"', (statusKey, expectedText) => {
+            const status = REVIEWER_STATUS_LABELS[statusKey]!
+            renderWithProviders(<DisplayStudyStatus status={status} />)
+            expect(screen.getByText(expectedText)).toBeDefined()
```
nathanstitt: a quibble, but this doesn't actually test that the bug was fixed.  It does test that the words 'Proposal change requested' are present, but nothing would stop it from actually being 'Proposal Proposal change requested'

I think this will work:
```suggestion
           const { container } = renderWithProviders(<DisplayStudyStatus status={status} />)
           expect(container.textContent?.trim()).toBe(expectedText)
```

nathanstitt (approved):

nathanstitt (approved):
looks good, but probably should update the test

src/components/study/display-study-status.test.tsx
therealmarv: After some research: `getByText` [is exact match by default](https://testing-library.com/docs/queries/about/#textmatch-examples), so it will fail in this case when the prefix is duplicated in one text node.

But your check catches a case mine misses: if someone refactors later and splits the prefix into a sibling (like `<span>Proposal</span><Text>Proposal change requested</Text>`), `getByText` still finds the `<Text>` and passes silently, even when the user sees `Proposal Proposal change requested`. So I added your assertion too (scoped to the badge root, otherwise it picks up Mantine's injected styles).

PR #682, opened 2026-05-06
OTTER 541: Adding download button to the view code modal

therealmarv:
see [OTTER-541](https://openstax.atlassian.net/browse/OTTER-541)

- Add a Download link inside the View Code modal so researchers can save a code file to their machine instead of copy-pasting from the viewer.
- Gate the link behind the existing OpenStax + spy-mode feature flag so only the targeted research lab cohort sees it during rollout.
- Cover the new behavior with unit tests for flag-on, flag-off, and non-OpenStax-org cases.

[OTTER-541]: https://openstax.atlassian.net/browse/OTTER-541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/study/study-code-panel.tsx
```diff
@@ -1,17 +1,56 @@
...
+        <Anchor
+            href={href}
+            download={name}
+            data-testid="download-link"
+            style={{ display: 'flex', alignItems: 'center' }}
+        >
+            Download <DownloadSimpleIcon size={16} style={{ marginLeft: 4 }} />
+        </Anchor>
+    )
+}
```
nathanstitt: `CodeDownloadLink` is nearly identical to `DownloadResultsLink` in `src/components/job-results.tsx` (and similar to the anchor in `encrypted-files-panel.tsx`). Same `useState('#')` + `useEffect` + `URL.createObjectURL` + revoke cleanup + Anchor/`DownloadSimpleIcon` JSX + `data-testid="download-link"`. Worth considering a shared `<DownloadBlobLink>` ?

src/components/study/study-code-panel.tsx
```diff
@@ -1,17 +1,56 @@
...
+import { useCodeDownloadFeatureFlag } from '@/components/openstax-feature-flag'
 import { StudyCodeEmptyView } from './study-code-empty-view'
 import { StudyCodeReviewView } from './study-code-review-view'
 
-function FilePreviewModal({ file, onClose }: { file: { name: string; contents: string } | null; onClose: () => void }) {
+const CodeDownloadLink: FC<{ name: string; contents: string }> = ({ name, contents }) => {
+    const [href, setHref] = useState('#')
+
+    useEffect(() => {
+        const blob = new Blob([contents], { type: 'text/plain' })
```
nathanstitt: Tiny thing on the MIME type — `text/plain` is fine since the `download` attribute forces a save in modern browsers, but for a code file you might consider leaving it off (defaults to `application/octet-stream`) or using something like `text/x-python` based on `highlightLanguageForFile`. Mostly cosmetic; users on weird configs occasionally see preview-instead-of-download with `text/plain`.

nathanstitt (approved):
Looks good, just a small suggestion about making a common download component

src/components/study/study-code-panel.tsx
therealmarv: makes sense, I will implement that!

PR #683, opened 2026-05-06
fix(realtime): review-page kick-out broken + editors dead after bfcache

nathanstitt:
## Summary
Two regressions in the YJS realtime layer on the proposal-review page.

OTTER-497

### Bug 1 — review-page kick-out is broken on a stable connection
`ReviewSubmissionListener` and the Lexical editor each constructed a `HocuspocusProvider` for the same document name (`review-feedback-\${studyId}`) on the shared websocket. `HocuspocusProviderWebsocket` keys providers by name in a `Map` ([HocuspocusProviderWebsocket.ts:212](https://github.com/ueberdosis/hocuspocus/blob/main/packages/provider/src/HocuspocusProviderWebsocket.ts#L212)), so the second \`attach()\` silently overwrites the first; inbound messages dispatch via \`providerMap.get(name)?.onMessage(...)\` (line 379) and the loser goes deaf for the \`proposal-review-submitted\` stateless event. \`useStudyStatusOnReconnect\` is *not* a poll — it only fires on reconnect transitions — so on a stable connection there's no fallback.

**Fix:** route the listener through a new \`ReviewFeedbackProviderProvider\` context that the editor publishes its provider into via a new \`onProviderReady\` prop on \`CollaborativeEditor\`. One \`HocuspocusProvider\` per document name on the shared websocket; both consumers share it.

### Bug 2 — editors silently dead after a bfcache back-button restore
\`YjsWebsocketProvider\` destroys+nulls the singleton socket on \`pagehide\` (correct, otherwise the page can't enter bfcache) but had no \`pageshow\` handler. After bfcache restore, React state still pointed at the destroyed socket: typing didn't sync, no "Working offline" banner, no reconnect. Only a hard refresh recovered.

**Fix:** add a \`pageshow\` handler that, on \`event.persisted\`, recreates the singleton via \`getOrCreateSharedSocket()\` and notifies live providers through a module-scoped subscriber set so React state updates to the new socket.

### Drive-by
\`vitest.config.ts\` now sets \`resolve.alias\` for \`yjs\` and \`y-protocols\` so tests load a single module instance. Without this, \`services/editor/node_modules/yjs\` and the root copy load as separate modules and any \`Doc.get('root', Y.XmlText)\` blows up with "Type with the name root has already been defined with a different constructor." Production is unaffected — \`services/editor\` and the Next app run as separate processes.

## Tests
- **bfcache restore**: pair verifying recreation on persisted \`pageshow\` and a no-op on non-persisted.
- **ReviewFeedbackProviderProvider publish/subscribe**: subscriber receives the published provider; sees null after editor unmount.

22 of 22 relevant tests pass locally.

## Test plan
- [x] \`npm run test:unit -- --run src/lib/realtime services/editor\` clean
- [x] \`npm run lint\` clean
- [ ] Manual: open the review page in two tabs, submit a decision in tab A, verify tab B shows the kick-out toast/redirect on a stable connection
- [ ] Manual: navigate away from the review page, hit Back, type in the editor — should reconnect and sync (was silently dead before)

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/proposal-review-view.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/proposal-review-view.tsx
@@ -7,7 +7,7 @@ import { useProposalReviewMutation } from '@/hooks/use-proposal-review-mutation'
 import { useReviewDecision } from '@/hooks/use-review-decision'
 import { useReviewFeedback } from '@/hooks/use-review-feedback'
 import { StudyKickOutProvider } from '@/hooks/use-study-status-on-reconnect'
-import { ReviewFeedbackProviderProvider } from '@/lib/realtime/review-feedback-provider-context'
+import { ReviewFeedbackProviderShare } from '@/lib/realtime/review-feedback-provider-context'
 import { isSubmittedProposalReviewStatus } from '@/lib/proposal-review'
 import { Routes } from '@/lib/routes'
 import { ReviewSubmissionListener } from './review-submission-listener'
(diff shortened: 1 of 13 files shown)


jbwilson8 (approved):

therealmarv (approved):
nice, thanks!

PR #687, opened 2026-05-07
fix: stop duplicating approved results in reviewer Study Status

nathanstitt:
## Summary

Testers reported approved study results appearing multiple times on the reviewer's Study Status panel ([OTTER-529](https://openstax.atlassian.net/browse/OTTER-529)). Two bugs combined to cause this:

- **`fileRows` in `use-encrypted-files-panel.ts` multiplied rows N×N.** When an encrypted archive contained N inner files, each `APPROVED-RESULT` row in `studyJobFile` re-emitted every approved file, so 2 results rendered as 4. Fixed by tracking `seenApprovedTypes` so each fileType is processed only once.
- **`<JobResults>` rendered a redundant legacy list below the unified table.** The new `EncryptedFilesPanel` table (PR #673) already shows approved files with View/Download per file; the older "Results: View | Download" block underneath was duplicating the same files. Removed it.

Repeat decrypt/view interactions are unaffected — the form is hidden after decryption (`shouldShowForm` flips false), and re-fetched approved files replace rather than append state.

## Test plan

- [x] Existing `EncryptedFilesPanel` and `StudyResults` unit tests pass (16 tests)
- [x] Related review-page tests pass (29 tests)
- [x] New regression test: approving a job with two `APPROVED-RESULT` files renders exactly one row per file
- [x] `tsc --noEmit` clean
- [ ] Manual: navigate to an approved study as a reviewer and confirm only one row per result file in the Study Status table, no list below

[OTTER-529]: https://openstax.atlassian.net/browse/OTTER-529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/study-results.test.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/study-results.test.tsx
@@ -9,7 +9,7 @@ import {
 } from '@/tests/unit.helpers'
 import { fireEvent, waitFor, screen } from '@testing-library/react'
 import { StudyResults } from './study-results'
-import { fetchEncryptedJobFilesAction } from '@/server/actions/study-job.actions'
+import { fetchApprovedJobFilesAction, fetchEncryptedJobFilesAction } from '@/server/actions/study-job.actions'
 import { latestJobForStudy } from '@/server/db/queries'
 import { ResultsWriter } from 'si-encryption/job-results/writer'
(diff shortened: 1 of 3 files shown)


therealmarv (approved):
👍

PR #690, opened 2026-05-11
align post-feedback view structure with Figma

Rod0352:
## Summary

Follow-up to the merged OTTER-501 PR addressing design review from Greg + Greg Fitch on the post-feedback view.

## Changes

- **Single proposal container** — Removed the separate `DecisionHeader` outside the proposal section. Page now matches Figma: breadcrumbs → "Study Proposal" h1 → one `ProposalRequest` paper (STEP 1, "Review initial request", title, decision-timestamp, divider, decision-banner, "View full initial request" toggle, expandable body) → feedback section → CTA. Fixes the duplicate "Review initial request" and "Title:" text.
- **Decision-overridden timestamp + banner** — Replaces the section's "Submitted on …" with "Approved on …" / "Clarification requested on …" / "Rejected on …" via a new `timestampDate` prop on `ProposalRequest`. Purple status banner is replaced by the decision-colored banner inline.
- **Banner contrast** — All three banners use `charcoal.9` text on `green.1` / `yellow.1` / `red.1` backgrounds for consistent readable contrast (fixes Greg Fitch's yellow-on-yellow note).

### DIFF
--- a/src/components/study/proposal-initial-request.tsx
+++ b/src/components/study/proposal-initial-request.tsx
@@ -17,6 +17,8 @@ type ProposalRequestProps = {
     banner: ReactNode
     initialExpanded?: boolean
     statusBadge?: string
+    /** Overrides study.submittedAt when rendering the timestamp (e.g. a decision date). */
+    timestampDate?: Date | string | null
 }
 
 function useProposalRequest(initialExpanded: boolean) {
@@ -59,8 +61,10 @@ export function ProposalRequest({
     banner,
     initialExpanded = true,
     statusBadge = 'Submitted on',
+    timestampDate,
 }: ProposalRequestProps) {
     const { expanded, toggle, collapse, getPopoverProps } = useProposalRequest(initialExpanded)
+    const renderedTimestamp = timestampDate ?? study.submittedAt
 
     return (
         <Stack gap="md" data-testid="proposal-section">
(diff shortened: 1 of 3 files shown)


src/components/study/proposal-initial-request.tsx
```diff
@@ -59,8 +61,10 @@ export function ProposalRequest({
     banner,
     initialExpanded = true,
     statusBadge = 'Submitted on',
+    timestampDate,
 }: ProposalRequestProps) {
     const { expanded, toggle, collapse, getPopoverProps } = useProposalRequest(initialExpanded)
+    const renderedTimestamp = timestampDate ?? study.submittedAt
```
nathanstitt, line 67: Because the prop type allows `null` (`Date | string | null`), `timestampDate={null}` will fall through to `study.submittedAt` rather than suppressing the timestamp — `??` only short-circuits `null`/`undefined` to the right-hand value. If a caller ever passes an explicit `null` meaning "don't show a timestamp," we'd silently render the submitted date instead. No caller does that today, but we could either drop `| null` from the prop type or branch on `timestampDate === null` explicitly so the intent is unambiguous.

nathanstitt (approved):
Looks great, thanks @Rod0352

src/components/study/proposal-initial-request.tsx
Rod0352: Good catch — dropped | null

PR #699, opened 2026-05-13
Fix/simplify study review malformed report

chrisbendel:
@nathanstitt I piggy-backed on your PR and tried to simplify a bit - I think the direction you went with the auto-retries + status tracking of the review (error/pending/malformed/finished/etc) might be better solved by implementing a more robust queue/job feature like we talked about in the dev roundtable - I think we should probably decide on something (SQS? Needs to support serverless right?) and this would be a great first candidate to implement/prove that out with

For now, dealing with those small drift/json output errors - i ended up learning about [structured outputs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs) - you can see in `agent.ts` how its working. That should get us the hardening we need for now without implementing a full job/queue for this iteration (hopefully 🤞)

Let me know what you think about all this, super open to ideas too but I wanted to keep the fix/implementation a bit smaller/simpler until the architecture change can do the heavy lifting

### DIFF
--- a/src/server/agents/review-agent/agent.ts
+++ b/src/server/agents/review-agent/agent.ts
@@ -4,10 +4,10 @@ import { analysisReportSchema } from './types'
 import type { AnalysisReport, AnalysisResult, ReviewAgentConfig, ReviewContent, ReviewMessage } from './types'
 
 const DEFAULT_MODEL = process.env.ANTHROPIC_MODEL ?? 'claude-sonnet-4-6'
-// Sized for ~3× the realistic worst-case review (2 narrative fields + ~10
-// findings per check at ~50 tokens each ≈ 4-5K tokens). The schema property
-// `description` strings below carry the actual length guidance to the model;
-// this cap exists only to bound runaway output if the model ignores them.
+// Realistic worst case is ~1.5K tokens (2 narrative fields × ~150 words ≈
+// 400 tokens, plus ~10 findings per check × 2 checks × ~50 tokens). 16K
+// gives ~10× headroom — schema property `description` strings carry the
+// actual length guidance to the model; this cap is just a runaway bound.
 const DEFAULT_MAX_TOKENS = 16_000
 const DEFAULT_MAX_RETRIES = 3
 
--- a/src/server/agents/review-agent/agent.ts
+++ b/src/server/agents/review-agent/agent.ts
@@ -40,13 +40,6 @@ const ANALYSIS_TOOL: Anthropic.Messages.Tool = {
     },
 }
 
-const RETRY_NUDGE =
-    'Your previous response did not call the `submit_analysis` tool with the exact required shape. ' +
-    'You MUST call `submit_analysis` exactly once. Every required field must be a real value of the ' +
-    'correct type — `alignmentCheck` and `complianceCheck` must each be objects with `isAligned`/' +
-    '`isCompliant` (boolean) and `findings` (array of strings). Do not return XML, parameter tags, ' +
-    'or string fragments — only the structured tool call.'
-
 function resolveClient(config: ReviewAgentConfig): Anthropic {
     if (config.client) return config.client
     if (config.apiKey) {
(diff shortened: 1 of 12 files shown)


src/server/agents/review-agent/runner.ts
```diff
@@ -2,10 +2,17 @@ import { db, jsonArrayFrom } from '@/database'
...
 import { generateAnalysis } from './agent'
-import type { ReviewContent } from './types'
+import type { AnalysisReport, ReviewContent } from './types'
 import { getConfigValue } from '@/server/config'
 import { fetchFileContents } from '@/server/storage'
 
+const DISABLED_REPORT: AnalysisReport = {
+    proposalSummary: 'AI code review is disabled for this environment — CLAUDE_API_KEY is not configured.',
+    codeExplanation: 'No automated review was generated.',
+    alignmentCheck: { isAligned: true, findings: [] },
```
nathanstitt: The disabled placeholder sets `isAligned: true` and `isCompliant: true`, which renders as green "Aligned" / "Compliant" badges in `CheckSection` (study-review-section.tsx:144). Combined with the "AI code review is disabled" proposal summary above, the UI ends up telling DOs that a study they never had reviewed passed both checks — the opposite of what we want. Could we either flip these to `false` with a single finding like "Automated review not run", or have the section render a distinct "disabled" treatment when it sees this sentinel report? Either way, green-badges-with-no-review feels worth not shipping.

src/server/agents/review-agent/runner.ts
```diff
@@ -138,11 +145,16 @@ export async function generateAndStoreStudyReview(studyJobId: string): Promise<v
         return
     }
 
+    const apiKey = await getConfigValue('CLAUDE_API_KEY', false)
```
nathanstitt, line 160: Nice catch reordering the API key check before `assembleReviewContent` — saves a round of DB / S3 work when the key is missing. One small behavior shift worth being aware of: previously a study with no code files was a silent skip (no row written); now if both the key is missing *and* there are no code files, we write the disabled placeholder for a study that had nothing to review anyway. Probably fine, but means the placeholder is no longer a strict "key missing" sentinel — it's "key missing, ignoring whatever else is wrong." Worth a sentence in the placeholder comment to set future-readers' expectations.

src/server/agents/review-agent/agent.ts
```diff
@@ -1,36 +1,72 @@
...
+import { analysisReportSchema } from './types'
 import type { AnalysisReport, AnalysisResult, ReviewAgentConfig, ReviewContent, ReviewMessage } from './types'
 
 const DEFAULT_MODEL = process.env.ANTHROPIC_MODEL ?? 'claude-sonnet-4-6'
-const DEFAULT_MAX_TOKENS = 4096
+// Sized for ~3× the realistic worst-case review (2 narrative fields + ~10
+// findings per check at ~50 tokens each ≈ 4-5K tokens). The schema property
+// `description` strings below carry the actual length guidance to the model;
+// this cap exists only to bound runaway output if the model ignores them.
+const DEFAULT_MAX_TOKENS = 16_000
```
nathanstitt, line 11: Comment math here doesn't quite line up — 2 narrative fields × ~150 words ≈ 400 tokens, plus 10 findings × 2 checks × ~50 tokens ≈ 1000 tokens, so realistic worst case is closer to 1.5K, making 16K more like ~10×. Not arguing for a lower cap (headroom is cheap and constrained-decoding has some overhead), just suggesting we drop or update the "~3×" claim so a future reader doesn't think the cap is tighter than it is.

src/server/agents/review-agent/agent.ts
```diff
@@ -1,36 +1,72 @@
...
+// model can no longer emit a shape that violates `input_schema` — the class of
+// "alignmentCheck came back as an XML-tag string fragment" prod bug is fixed
+// at the API boundary, not in our app. Requires `additionalProperties: false`
+// at every object level. Docs:
+//   https://platform.claude.com/docs/en/build-with-claude/structured-outputs
 const ANALYSIS_TOOL: Anthropic.Messages.Tool = {
     name: ANALYSIS_TOOL_NAME,
     description:
         'Submit the structured review of the research proposal. Always call this tool exactly once with the full report.',
+    strict: true,
```
nathanstitt, line 26: nice! I wasn't ware of this. much better than us doing it

nathanstitt (commented):
looks great, @chrisbendel ty!

nathanstitt (approved):

src/server/agents/review-agent/agent.ts
chrisbendel: yeah this def seems like the nice glue to hold this all together!! i missed this first round too

PR #700, opened 2026-05-13
OTTER-521: fix View as reviewer crash on edit-and-resubmit page

Rod0352:
Fixes:
"View as reviewer" crash on /edit-and-resubmit. 
ReviewerPreview was using useProposal() but the page is inside EditResubmitProvider. 
Refactored to take values + studyId as props; both callers updated.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx
@@ -87,6 +87,8 @@ export const ProposalFooter: FC<ProposalFooterProps> = ({ researcherName, resear
                 title="View as reviewer"
             >
                 <ReviewerPreview
+                    studyId={studyId}
+                    values={form.getValues()}
                     researcherName={researcherName}
                     researcherId={researcherId}
                     piUserId={piUserId}
--- a/src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx
@@ -13,11 +13,10 @@ import { ReviewerPreview } from './reviewer-preview'
 interface ProposalFooterProps {
     researcherName: string
     researcherId: string
-    piUserId: string
     enclaveOrgSlug?: string
 }
 
-export const ProposalFooter: FC<ProposalFooterProps> = ({ researcherName, researcherId, piUserId, enclaveOrgSlug }) => {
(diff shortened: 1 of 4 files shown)


src/app/[orgSlug]/study/[studyId]/proposal/reviewer-preview.tsx
```diff
@@ -6,29 +6,32 @@ import { useParams } from 'next/navigation'
...
-import { useProposal } from '@/contexts/proposal'
 import { useOrgDataSources } from '@/hooks/use-org-data-sources'
 import { usePopover } from '@/hooks/use-popover'
-import { DEFAULT_DRAFT_TITLE } from './schema'
+import { DEFAULT_DRAFT_TITLE, type ProposalFormValues } from './schema'
 import { editableTextFields } from './field-config'
 
 interface ReviewerPreviewProps {
+    studyId: string
+    values: ProposalFormValues
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/reviewer-preview.tsx:16]: `piUserId` is already a field on `ProposalFormValues`, so we're now accepting it both inside `values` and as a separate prop just below. 

We should remove it from one of them

src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx
```diff
@@ -87,6 +87,8 @@ export const ProposalFooter: FC<ProposalFooterProps> = ({ researcherName, resear
                 title="View as reviewer"
             >
                 <ReviewerPreview
+                    studyId={studyId}
+                    values={form.getValues()}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx]: Tiny consistency nit: this file already destructures `form.values` on line 30 and uses `form.values.*` elsewhere, then switches to `form.getValues()` here. Same content in practice, but mixing the two styles is noise — `values={form.values}` would match the surrounding code (and the same applies in the edit-and-resubmit footer).

src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx
```diff
@@ -87,6 +87,8 @@ export const ProposalFooter: FC<ProposalFooterProps> = ({ researcherName, resear
                 title="View as reviewer"
             >
                 <ReviewerPreview
+                    studyId={studyId}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx:89]: related note as above,  now that `ReviewerPreview` receives the full `values` object, the `piUserId` prop on `ProposalFooter` is only forwarded straight through  We could read it from `form.values` directly here

nathanstitt (approved):

PR #701, opened 2026-05-13
OTTER-522: Version proposal entries on the post-resubmission view

stetradis:
## Summary                                                                      
- Add a nullable `version` column to `study_proposal_comment`, backfill existing rows in chronological order (resubmission notes increment, reviewer feedback inherits the current version), and populate it on every insert from `resubmitProposalAction` and `insertReviewerProposalComment`.                   
- Derive `studyVersion` for the submitted page from the max entry version, and use it to drive:                                                                
    - the section heading (`Initial request` → `Initial request 2.0`, `3.0`, … on each resubmission),                                                             
    - the timestamp label (`Submitted on …` → `Resubmitted on …` once `studyVersion > 1`),                                                            
    - the yellow `PENDING-REVIEW` banner copy (revised-resubmission copy when `studyVersion > 1`),                                                            
    - per-entry titles (`Reviewer feedback (vN.0)` / `Resubmission note (vN.0)`). 
- Rework the Feedback & Notes entries: replace the Collapse/caret toggle with a `lineClamp`-driven `View more` / `View less` anchor in line with the design, and collapse resubmission notes by default even when they're the latest entry.

### DIFF
--- a/src/components/study/feedback-and-notes.tsx
+++ b/src/components/study/feedback-and-notes.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { useCallback, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import { Anchor, Box, Divider, Paper, Stack, Text, Title } from '@mantine/core'
 import { CaretRightIcon } from '@phosphor-icons/react'
 import dayjs from 'dayjs'
@@ -28,11 +28,13 @@ type FeedbackEntryProps = {
 function FeedbackEntry({ entry, isExpanded, onToggle }: FeedbackEntryProps) {
     const title = entryTitle(entry)
     const date = formatDate(entry.createdAt)
+    const bodyRef = useRef<HTMLDivElement>(null)
     const [isTruncated, setIsTruncated] = useState(false)
 
-    const bodyRef = useCallback((node: HTMLDivElement | null) => {
+    useEffect(() => {
+        const node = bodyRef.current
         if (node) setIsTruncated(node.scrollHeight > node.clientHeight)
-    }, [])
+    }, [isExpanded])
 
     const showToggle = isExpanded || isTruncated
 
--- a/src/components/study/feedback-and-notes.tsx
+++ b/src/components/study/feedback-and-notes.tsx
@@ -31,12 +31,28 @@ function FeedbackEntry({ entry, isExpanded, onToggle }: FeedbackEntryProps) {
     const bodyRef = useRef<HTMLDivElement>(null)
     const [isTruncated, setIsTruncated] = useState(false)
 
+    // scrollHeight vs clientHeight only reflects overflow while line-clamp is on; once expanded,
+    // that comparison stays stale — measure against a fixed collapsed height (lineHeight × clamp) instead.
     useEffect(() => {
         const node = bodyRef.current
-        if (node) setIsTruncated(node.scrollHeight > node.clientHeight)
+        if (!node) return
+
+        const check = () => {
+            const lineHeight = parseFloat(getComputedStyle(node).lineHeight)
+            if (Number.isFinite(lineHeight) && lineHeight > 0) {
+                setIsTruncated(node.scrollHeight > lineHeight * COLLAPSED_LINE_CLAMP + 1)
+                return
+            }
+            setIsTruncated(node.scrollHeight > node.clientHeight)
+        }
+
+        check()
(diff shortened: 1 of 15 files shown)


src/server/actions/study.actions.ts
```diff
@@ -510,6 +510,10 @@ async function insertReviewerProposalComment({
             entryType: 'REVIEWER-FEEDBACK',
             decision: toReviewDecision(decision),
             body: JSON.parse(body),
+            version: sql<number>`coalesce((
```
nathanstitt: Same `coalesce((select max(version) ...))` pattern as `resubmitProposalAction` — could be worth pulling into a small helper (e.g. `nextVersionForStudy({ db, studyId, increment })`) so the +1 vs +0 distinction lives in one place. Also picks up any future locking/uniqueness change from the race-condition note on the other call site.

src/database/migrations/1776400000000_add_version_to_study_proposal_comment.ts
```diff
@@ -0,0 +1,28 @@
+import { type Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema.alterTable('study_proposal_comment').addColumn('version', 'integer').execute()
```
nathanstitt, line 4: After the backfill in this migration completes, every existing row has a value, and both new insert paths (`resubmitProposalAction`, `insertReviewerProposalComment`) always populate `version`.  Should we add a call to set column to notNull?

src/components/study/feedback-and-notes.tsx
```diff
@@ -29,24 +28,21 @@ type FeedbackEntryProps = {
 function FeedbackEntry({ entry, isExpanded, onToggle }: FeedbackEntryProps) {
     const title = entryTitle(entry)
     const date = formatDate(entry.createdAt)
+    const bodyRef = useRef<HTMLDivElement>(null)
+    const [isTruncated, setIsTruncated] = useState(false)
+
+    useEffect(() => {
```
nathanstitt, line 36: Heads-up: this useEffect only re-measures when `isExpanded` flips, so `isTruncated` can go stale on viewport resize, or  ReadOnlyLexical re-renders that change content height.    Either of those would cause "View more" to disappear.    A `ResizeObserver` on `bodyRef.current` (or a window resize listener) would catch those cases. Probably fine for the initial version tho since the design is column-fixed

src/components/study/feedback-and-notes.tsx
```diff
@@ -29,24 +28,21 @@ type FeedbackEntryProps = {
...
-                    onClick={onToggle}
-                    c="blue"
-                    size="sm"
-                    data-testid={`feedback-toggle-${entry.id}`}
-                    aria-expanded={isExpanded}
-                >
-                    <ToggleCaret isExpanded={isExpanded} />
-                </Anchor>
-            </Group>
+        <Stack gap="sm" data-testid={`feedback-entry-${entry.id}`} aria-expanded={isExpanded}>
```
nathanstitt: Quick a11y note: `aria-expanded` is a state for a *control* (button, link, etc.) — putting it on a `Stack` (which renders a `<div>` with no `role`) doesn't convey anything to assistive tech. The previous version had it on the toggle Anchor, which is correct. Could keep it on the `Anchor` (renamed `feedback-toggle-${id}`) and tweak the tests to assert there. As-is it functions as a styling/test hook — the data-testid is doing the real work.

src/lib/studies.ts
```diff
@@ -7,3 +7,8 @@ type StudyWithJobStatuses = {
 export function studyHasJobStatus(study: StudyWithJobStatuses, status: StudyJobStatus): boolean {
     return study.jobStatusChanges.some((s) => s.status === status)
 }
+
+export function deriveStudyVersion(entries: { version: number | null }[]): number {
```
nathanstitt: Small thing: `e.version ?? 1` silently treats any null version as v1. After the backfill in the new migration nothing should be null in practice (and in future if we add notNull clause like I suggest).  Either way the `| null` seems unneeded.

src/server/actions/study-request.ts
nathanstitt, line 684: since this does a lot of select then insert/update, lets add a comment saying the `performsMutation: true` is important and MUST NOT be removed

nathanstitt (approved):
Looks great @stetradis this was a tricky one

PR #702, opened 2026-05-13
Fix OTTER-497 qa report

therealmarv:
see [OTTER-497 comment](https://openstax.atlassian.net/browse/OTTER-497?focusedCommentId=41671)

## Summary

Fixes the four OTTER-497 regressions reported by QA on the multi-user proposal review flow. All changes stay behind `ProposalCollaborationFeatureFlag`. Versioning enforcement lives in the management-app actions (Kysely); the editor service stays a thin store that only does read-auth and status checks.

| Issue | Cause | Fix |
|------|-------|-----|
| **1** R2 not redirected when R1 submits | `useProposalReviewMutation` constructed its own private `HocuspocusProvider`. Stateless flushes at WS-open before Hocuspocus auth populated `connection.context`, so the server's `onStateless` gate silently dropped the event. | Broadcast over the editor's already-authenticated provider via `useReviewFeedbackProvider()`. |
| **2** Round-2 editor pre-filled with round-1 text | No history surface in `ProposalReviewView`; `review/page.tsx` did not load prior entries on the editable branch. | Load entries + `reviewVersion` server-side, render prior rounds as read-only history via `FeedbackAndNotesSection`, round-aware heading. |
| **3** Post-submit pollution leaks into round 2 | Same Yjs doc name reused across rounds. | Version-key the Yjs doc as `review-feedback-${studyId}-v${N}`. Each round is now a separate Hocuspocus room with its own `Y.Doc` and persistence row, so stale `…-v1` writes cannot reach `…-v2`. |
| **4** Round-2 saved comment contains all prior text | Downstream of (3). | Falls out of (3); plus belt-and-suspenders `reviewVersion` Zod param on `submitProposalReviewAction`, recomputed server-side and rejected on mismatch with `ActionFailure({ review: 'stale review round … (current …)' })`. |

## What changed

- **Versioned doc name + parser** (`collaboration-documents.ts`, `services/editor/auth.ts`): `reviewFeedbackDocNameForVersion` is the only review-feedback name builder. The parser accepts only the versioned (`-v<n>`) shape, returning `{ kind: 'review-feedback', studyId, version: number }`. Unversioned names are rejected by `parseDocumentName`, so no Yjs server can persist them. Parsing is a pure naming convention; the editor no longer enforces policy on the result.
- **Helper**: `currentReviewVersion(studyId)` in `src/server/db/queries.ts` (Kysely). Reads `max(studyProposalComment.version)`, tie-immune across same-version reviewer rows.
- **Submit action** (`study.actions.ts`): new `reviewVersion` param, server recomputes + rejects on mismatch, deletes the versioned `yjs_document` row in-tx. Deferred safety-net purge takes the version.
- **Resubmit action** (`study-request.ts`): `resubmitProposalAction` now also deletes all `review-feedback-${studyId}%` `yjs_document` rows before committing. Closes the orphan-row gap: a stale R2 `…-v1` tab can no longer recreate the deleted row once status returns to `PENDING-REVIEW`.
- **Issue 1 client fix** (`use-proposal-review-mutation.ts`): private broadcast provider removed; `useReviewFeedbackProvider()` consumed instead, so the stateless event flushes through the editor's authenticated provider.
- **UI**: `review/page.tsx` editable branch loads entries (soft-fail) and `reviewVersion` (hard-fail, must stay in sync with the doc name the editor binds to); `ProposalReviewView` split into outer/inner so the mutation hook lives inside `ReviewFeedbackProviderShare`; `FeedbackAndNotesSection` returns `null` when empty so round 1 has no history surface; `ReviewFeedbackSection` binds the editor to the versioned doc name with a round-aware heading (`Round N review`).
- **Cleanup helper** (`yjs-cleanup.ts`): `purgeReviewFeedbackYjsDocBeforeAt` takes `version`.
- **Migration** (`1776500000000_purge_unversioned_review_feedback_yjs_docs.ts`): one-shot DELETE of any pre-version-keying unversioned `review-feedback-<uuid>` rows in `yjs_document`. Combined with the parser now rejecting the unversioned form, this guarantees no orphan unversioned rows exist post-deploy.

## What did NOT change (deliberately)

- The editor service no longer queries `study_proposal_comment`. `currentReviewVersionFromDb` and the version branches in `authenticate` / `shouldPersistDocument` are deleted. Round isolation rides entirely on the protocol-level boundary (Hocuspocus rooms are keyed by `documentName`).
- Versioning enforcement is the action layer's job: `submitProposalReviewAction` validates `reviewVersion` against `currentReviewVersion`, and `resubmitProposalAction` bumps the version + cleans up stale Yjs rows.
- Pre-existing editor patterns (read-auth membership SQL, status gates, `onStateless` defense-in-depth status check) are unchanged. Out of scope for a QA bugfix; can be a follow-up ticket for a broader "editor stops touching Postgres" architecture work.

## Test plan

- [x] `npm run checks` clean (typecheck + lint + validate-actions)
- [x] `./bin/docker-test` (full suite): 1148/1148 unit tests pass across 132 files
- [x] New direct unit tests for `currentReviewVersion` (cold round 1 returns 1, max across mixed entry types, tie-immune across same-version reviewer rows, no cross-study leak)
- [x] New round-2 happy-path test in `study.actions.test.ts`: PENDING-REVIEW study with seeded `REVIEWER-FEEDBACK@v1` + `RESUBMISSION-NOTE@v2`, submit with `reviewVersion: 2`, asserts a new `REVIEWER-FEEDBACK@v2` row contains only round-2 text and the v1 row body is untouched (directly covers QA Issue 4)
- [x] New resubmit cleanup test (`resubmit-proposal.test.ts`): seeds `review-feedback-${id}-v1` and `review-feedback-${id}-v2` Yjs rows, calls `resubmitProposalAction`, asserts both are gone (the wildcard pattern sweeps every version)
- [ ] Manual QA end-to-end:
    - [ ] Cold round 1, single reviewer, all three decision types (approve / needs-clarification / reject)
    - [ ] Two concurrent reviewers, R1 submits clarification, R2 redirected within ~1s with toast (Issue 1, client-side broadcaster)
    - [ ] Researcher resubmits, R1 opens round 2, empty editor + read-only history above, heading reads "Round 2 review" (Issue 2)
    - [ ] R1 submits round 2, new comment row contains only round-2 text, round-1 row body unchanged (Issue 4)
    - [ ] Stale `…-v1` tab open during round-2 submit: writes go into the dead `…-v1` room and are mopped up by the 5-second deferred purge; round 2 (`…-v2`) is unaffected
    - [ ] DB check after end-to-end: `SELECT name FROM yjs_document WHERE study_id = '<id>';` shows no `-v1` rows after resubmit, no `-v2` row after submit

## Notes for reviewer

- `page.tsx` loads `reviewVersion` eagerly (no soft-fail). A desync between the editor's doc name and the action's expected version would surface as a confusing "stale review round N (current M)" error on submit, which is worse UX than a 500 at page load.
- `d97565b2` introduced `ReviewFeedbackProviderShare` for the kick-out listener; this PR extends it to the broadcaster, which required splitting `ProposalReviewView` into outer (provider wrapper) / inner (hook consumer) so the mutation hook lives **inside** the share context.
- Migration `1776500000000_purge_unversioned_review_feedback_yjs_docs.ts` deletes any pre-version-keying unversioned `review-feedback-<uuid>` rows on deploy. The parser also rejects the unversioned form, so no future client can recreate one. No legacy code path remains in `yjs-cleanup.ts` or `study.actions.ts`.
- A stale `…-v1` tab open when the researcher resubmits is no longer kicked out at the editor level. The room itself is dead (Hocuspocus document names are namespaced per round) and the next `resubmitProposalAction` deletes any row the stale tab managed to recreate. Round 2 (`…-v2`) is independent. Confirm acceptable UX with product; if a hard kick-out is desired, that's a follow-up ticket against the broader editor architecture work.


therealmarv:
not to be merged before https://github.com/safeinsights/management-app/pull/701

nathanstitt (changes requested):
Hey @therealmarv and @stetradis - there's a lot here to think about.  

I do not think the editor should be responsible for enforcing logic, I think of it as a relatively dumb store that simply co-ordinates multiple user's edits.  It _should_ verify that those users have the rights to read the content, otherwise someone from org A could read reviews for org B, but it shouldn't be doing things like version restrictions and all this.

We definitely shouldn't be essentially rebuilding kysley with raw pg.pool sql calls 😁   Instead the normal management app actions should be saving content and enforcing versioning.

Happy to discuss further if ya'll feel different?

therealmarv:
I'm okay with refactoring this! Makes sense, thanks for feedback.

This branch is built on Stellar's branch because otherwise I would have even reinvented things she has already done for OTTER-522.

services/editor/auth.ts
```diff
@@ -84,7 +93,8 @@ export async function shouldPersistDocument(parsed: ParsedDocumentName, db: Pick
     const row = await db.query<{ status: StudyStatus }>('SELECT status FROM study WHERE id = $1', [parsed.studyId])
     const status = row.rows[0]?.status
     if (!status) return false
-    return editableStatusForKind(parsed.kind).includes(status)
+    if (!editableStatusForKind(parsed.kind).includes(status)) return false
+    return true
```
nathanstitt: very much a nitpick, but could be sipmleified:

```suggestion
    return editableStatusForKind(parsed.kind).includes(status)
```

src/lib/collaboration-documents.ts
```diff
@@ -30,18 +30,42 @@ export const proposalFieldsDocName = (studyId: string) => `${PROPOSAL_PREFIX}${s
...
+
+/**
+ * Legacy unversioned name kept only for the deferred safety-net purge so a
+ * partially migrated row can still be cleaned up. New call sites must use
+ * `reviewFeedbackDocNameForVersion` — no live client emits this form. The
+ * editor's `parseDocumentName` still recognizes it (returning `version:
+ * null`) so any straggler connection is gated by status checks rather
+ * than rejected outright.
+ */
+export const reviewFeedbackLegacyDocName = (studyId: string) => `${REVIEW_FEEDBACK_PREFIX}${studyId}`
```
nathanstitt: Given that this isn't used any longer, should we instead just purge all old docs?  it's not live yet

nathanstitt (approved):
Looks good, thank you @therealmarv

src/lib/collaboration-documents.ts
therealmarv: good idea! Was not thinking about that. I will implement that.

PR #703, opened 2026-05-14
OTTER-561: Handle invalid Lexical editor state

stetradis:
## Summary

Problem: 
- The error boundary page is thrown when saving/submitting partially complete drafts, and does not allow a user to view/edit the study further as access is blocked by the error state. "View as reviewer" modal is also crashing the page.

Changes:
- Fixes a crash in non-collaborative views (`EditableText`, `ReadOnlyLexicalContent`) when they tried to hydrate a Lexical editor from JSON with an empty root (`{"root":{"children":[]}}`). Lexical rejects an empty root on init and throws.           
- Root cause: `EditorChangePlugin` in the collaborative editor fired an update on mount before Yjs had synced. At that moment Lexical's root briefly has no children — a valid in-memory state, but invalid for re-hydration. That JSON propagated up through `onChange`, into form state, and (via "Save as draft") into `study.researchQuestions` / `projectSummary` / `impact` / `additionalNotes`. 
- Save-boundary fix: `EditorChangePlugin` now drops updates where `root.children` is empty, so the pre-sync state never leaves the editor. User-cleared input still keeps an empty paragraph in `children`, so this filter only catches the pre-sync case.
- Defensive read-side guards: `EditableText` and `ReadOnlyLexicalContent` use a new `isValidLexicalState` helper to detect empty-root JSON and fall back to Lexical's default empty state (or render nothing, for the read-only view). These guards cover legacy rows already persisted before this fix.                                
- Adds unit tests for `isValidLexicalState`, including the pre-sync vs. user-cleared distinction. 
- Renames `src/lib/word-count.ts` → `src/lib/lexical.ts` — the module already housed Lexical-specific helpers (`extractTextFromLexical`, `countWordsFromLexical`, `hasLexicalContent`, `lexicalJson`) and now also `isValidLexicalState`. Updates all imports.


src/components/editable-text/collaborative-editor.tsx
```diff
@@ -243,7 +243,15 @@ function EditorChangePlugin({ onChange }: { onChange: (json: string) => void })
...
-            onChange(JSON.stringify(editorState.toJSON()))
+            const json = editorState.toJSON()
+            // On mount, before Yjs syncs, Lexical's root briefly has no children.
+            // That state is valid in memory but Lexical rejects it on re-hydration,
+            // so persisting it would crash the non-collaborative views (EditableText,
+            // ReadOnlyLexicalContent) the next time the data is read. User-cleared
+            // input still keeps an empty paragraph in children, so this only filters
+            // the pre-sync state.
+            if (!json.root?.children?.length) return
+            onChange(JSON.stringify(json))
```
nathanstitt, line 254: does lexical provide any `isValid` checkers or anything we could run the json through?  I'm wondering if there are other possibly invalid json we might encounter?

nathanstitt (approved):
Looks good but wondered about a more comprehensive isValid check

src/components/editable-text/collaborative-editor.tsx
stetradis: Lexical doesn't provide a dedicated JSON validator. The closest option is `editor.parseEditorState(json)`, which throws on invalid input, but it re-parses the whole document -- effectively `serialize → stringify → parse the entire tree again` on every update

PR #706, opened 2026-05-15
OTTER-540: add Submitted code section to DO code review page

Rod0352:
- New SubmittedCodeSection wired into CodeReviewRedesignView with header, dataset pills, AI summary collapsible, security scan log, and code-files tabs + viewer.
- Inherits the existing CodeReviewFeatureFlag from page.tsx; no new flag.
- Datasets sourced from SelectedStudy.orgDataSources.
- AI Summary reads AnalysisReport.codeExplanation from studyReview.
- New latestCodeScanForStudy() query joins study -> orgCodeEnv -> codeScan.
- 28 unit tests cover every bullet in the ticket's Unit Tests list.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
@@ -1,0 +1,192 @@
+'use client'
+
+import { Box, Group, Stack, Text, UnstyledButton } from '@mantine/core'
+import { useState } from 'react'
+import type { StudyJobFileType } from '@/database/types'
+
+export type CodeFile = { name: string; fileType: StudyJobFileType }
+
+// 20–24 char ceiling per AC; midpoint chosen so neither extreme is the boundary.
+const MAX_TAB_CHARS = 22
+const MAX_VISIBLE_TABS_BEFORE_OVERFLOW = 4
+
+export function truncateFileName(name: string, max = MAX_TAB_CHARS): string {
+    if (name.length <= max) return name
+    return name.slice(0, max - 1) + '…'
+}
+
+export function splitVisibleFiles(files: CodeFile[]) {
+    if (files.length <= MAX_VISIBLE_TABS_BEFORE_OVERFLOW) {
+        return { visible: files, hiddenCount: 0 }
+    }
+    // When overflowing, the last visible slot becomes the "+N more files" indicator,
+    // so we keep three real tabs and roll the remainder into the overflow count.
+    const visibleSlots = MAX_VISIBLE_TABS_BEFORE_OVERFLOW - 1
+    return { visible: files.slice(0, visibleSlots), hiddenCount: files.length - visibleSlots }
+}
+
+function useAiSummaryToggle() {
+    const [isExpanded, setIsExpanded] = useState(false)
+    return { isExpanded, toggle: () => setIsExpanded((v) => !v) }
+}
+
+function AiSummaryBody({ isVisible, summary }: { isVisible: boolean; summary: string }) {
+    if (!isVisible) return null
+    return (
+        <Text size="sm" data-testid="ai-summary-body" style={{ whiteSpace: 'pre-wrap' }}>
+            {summary}
+        </Text>
+    )
+}
+
+type AiSummaryProps = { summary: string | null }
+
+export function AiSummaryCollapsible({ summary }: AiSummaryProps) {
+    const { isExpanded, toggle } = useAiSummaryToggle()
+    const toggleLabel = isExpanded ? 'Hide full AI summary' : 'View full AI summary'
+    const hasSummary = !!summary
+
+    const placeholder = (
+        <Text size="sm" c="dimmed" data-testid="ai-summary-empty">
+            No AI summary available yet.
+        </Text>
+    )
+    const content = (
+        <>
+            <AiSummaryBody isVisible={isExpanded} summary={summary ?? ''} />
+            <UnstyledButton onClick={toggle} data-testid="ai-summary-toggle">
+                <Text size="sm" c="blue.6" td="underline">
+                    {toggleLabel}
+                </Text>
+            </UnstyledButton>
+        </>
+    )
+
+    return (
+        <Stack gap="xs" data-testid="ai-summary">
+            <Text fw={700}>AI Summary: Analysis of all files</Text>
+            <Text fw={600} size="sm">
+                Overview
+            </Text>
+            {hasSummary ? content : placeholder}
+        </Stack>
+    )
+}
+
+function useStudyCodeViewer(files: CodeFile[]) {
+    const [activeFileName, setActiveFileName] = useState<string | null>(files[0]?.name ?? null)
+    const [isExpanded, setIsExpanded] = useState(true)
+    const activeFile = files.find((f) => f.name === activeFileName) ?? files[0] ?? null
+    return {
+        activeFile,
+        selectFile: setActiveFileName,
+        isExpanded,
+        toggleExpanded: () => setIsExpanded((v) => !v),
+    }
+}
+
+function FileTab({ file, isActive, onClick }: { file: CodeFile; isActive: boolean; onClick: () => void }) {
+    const display = truncateFileName(file.name)
+    return (
+        <UnstyledButton
+            onClick={onClick}
+            data-testid="study-code-file-tab"
+            data-active={isActive ? 'true' : 'false'}
+            title={file.name}
+            px="md"
+            py="xs"
+            style={{
+                borderBottom: isActive ? '2px solid var(--mantine-color-blue-6)' : '2px solid transparent',
+                fontWeight: isActive ? 700 : 400,
+                whiteSpace: 'nowrap',
+            }}
+        >
+            <Text size="sm" component="span">
+                {display}
+            </Text>
+        </UnstyledButton>
+    )
+}
+
+function FileTabsRow({
+    visible,
+    activeFileName,
+    onSelect,
+    hiddenCount,
+}: {
+    visible: CodeFile[]
+    activeFileName: string | null
+    onSelect: (name: string) => void
+    hiddenCount: number
+}) {
+    const tabs = visible.map((file) => (
+        <FileTab
+            key={file.name}
+            file={file}
+            isActive={file.name === activeFileName}
+            onClick={() => onSelect(file.name)}
+        />
+    ))
+    const overflow =
+        hiddenCount > 0 ? (
+            <Text size="sm" c="charcoal.7" data-testid="study-code-files-overflow" style={{ whiteSpace: 'nowrap' }}>
+                +{hiddenCount} more files
+            </Text>
+        ) : null
+
+    return (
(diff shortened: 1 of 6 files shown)


src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
```diff
@@ -0,0 +1,134 @@
...
+        <Stack gap="xs" data-testid="submitted-code-datasets">
+            <Text size="sm">Dataset(s) associated with the study</Text>
+            <Group gap="xs">{names.length === 0 ? empty : pills}</Group>
+        </Stack>
+    )
+}
+
+function ScanStatusIcon({ scan }: { scan: LatestCodeScanForStudy | null }) {
+    if (!scan) return null
+    const passed = scan.status === 'SCAN-COMPLETE'
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx]: this collapses every non-`SCAN-COMPLETE` status into the red X. `ScanStatus` actually has four values — `SCAN-COMPLETE` / `SCAN-FAILED` / `SCAN-PENDING` / `SCAN-RUNNING`. A still-running or pending scan will render as "failed" to reviewers, which would be misleading (and would surface anytime a reviewer opens a freshly-submitted study before the scan finishes). Could we render a third in-progress state (spinner / dimmed icon) for `SCAN-PENDING` and `SCAN-RUNNING`, and reserve the red X for `SCAN-FAILED`?

src/server/db/queries.ts
```diff
@@ -397,6 +397,32 @@ export type StudyReviewWithMeta = {
...
+    status: ScanStatus
+    results: string | null
+    createdAt: Date
+}
+
+// The codeScan table is keyed by orgCodeEnv, not by studyJob — a scan represents
+// the security baseline of the environment the code runs in. We anchor on the
+// study's (orgId, language) → most recent non-testing orgCodeEnv, then its
+// latest scan row.
+export async function latestCodeScanForStudy(studyId: string): Promise<LatestCodeScanForStudy | null> {
```
nathanstitt: These are NOT the code scans you should be looking at.  These are for scans on the base image for the code env, not for job code, which is why you're having trouble joining it to a study job.

Instead we want to always pass a job id in so we make sure we return the correct ones.  Something like:

```
  export async function jobScanResultForJob(jobId: string): Promise<JobScanResult> {
      const [latestStatus, logFile] = await Promise.all([
          Action.db
              .selectFrom('jobStatusChange')
              .select(['status', 'createdAt'])
              .where('studyJobId', '=', jobId)
              .where('status', 'in', ['CODE-SCANNED', 'JOB-ERRORED'])
              .orderBy('createdAt', 'desc')
              .orderBy('id', 'desc')
              .limit(1)
              .executeTakeFirst(),
          Action.db
              .selectFrom('studyJobFile')
              .select(['id', 'name', 'path'])
              .where('studyJobId', '=', jobId)
              .where('fileType', '=', 'ENCRYPTED-SECURITY-SCAN-LOG')
              .orderBy('createdAt', 'desc')
              .limit(1)
              .executeTakeFirst(),
      ])

      return {
          status: latestStatus?.status ?? 'IN-PROGRESS',
          scannedAt: latestStatus?.createdAt ?? null,
          logFile: logFile ?? null,
      }
  }

```

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
```diff
@@ -0,0 +1,134 @@
...
+    job: Pick<LatestJobForStudy, 'id' | 'files'>
+    review: StudyReviewWithMeta | null
+    codeScan: LatestCodeScanForStudy | null
+}
+
+// Data fetching lives in the parent (CodeReviewRedesignView) so this component
+// stays a plain sync function. Nested async server components don't render
+// under testing-library / happy-dom — the parent's await is what tests rely on.
+export function SubmittedCodeSection({ orgSlug, study, job, review, codeScan }: SubmittedCodeSectionProps) {
+    const datasetNames = (study.orgDataSources ?? []).map((ds) => ds.name)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx]: `study.orgDataSources` is selected via `jsonArrayFrom`, which always emits an array (possibly empty). The `?? []` here is dead defense and slightly obscures that the field is reliably non-nullable. Safe to drop.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx
```diff
@@ -0,0 +1,370 @@
...
+        datasets: overrides.datasets ?? null,
+    })
+    const study = actionResult(await getStudyAction({ studyId: dbStudy.id }))
+    const job = await latestJobForStudy(study.id)
+    ;(useParams as Mock).mockReturnValue({ orgSlug: ORG_SLUG, studyId: study.id })
+    return { orgId: org.id, study, jobId: job.id }
+}
+
+async function renderSection(fixture: Fixture) {
+    const job = await latestJobForStudy(fixture.study.id)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx]: `renderSection` refetches `latestJobForStudy(fixture.study.id)` on every call even though `setupBaseFixture` already stored `fixture.jobId`.  let's use that to make sure we never show conflicting jobs

nathanstitt (changes requested):
looks pretty good but you're reading the wrong code scans 😁

nathanstitt:
@Rod0352 also I know the sample query I wrote is returning encrypted logs, that's fine for now I'm working on a followup to use plain text ones

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx]: **Good call, fixed. Three states now**

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx]: Yep, dropped.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx]: switched fixture to hold the full job and reuse it

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
```diff
@@ -0,0 +1,193 @@
...
+    // content rendering lands separately. The tab + expand behavior around it is
+    // fully wired up.
+    return (
+        <Box
+            bg="charcoal.0"
+            p="md"
+            data-testid="study-code-body"
+            style={{ borderRadius: 'var(--mantine-radius-sm)', fontFamily: 'monospace' }}
+        >
+            <Text size="sm">{activeFile.name}</Text>
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx]: Heads-up: the body just renders the active file name, not the file contents. The inline comment above is clear about it being a placeholder, but the section title is "Submitted code" and the PR title is "add Submitted code section" — a reviewer landing here will see `main.R` inside a styled monospace box and reasonably expect that to *be* the code. Worth either a visible "Code preview coming soon" label in the placeholder, or splitting the file-rendering follow-up into a tracked ticket linked from this comment so it doesn't get lost.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
```diff
@@ -0,0 +1,193 @@
...
+    return (
+        <Stack gap="sm" data-testid="study-code-viewer">
+            <FileTabsRow
+                visible={visible}
+                activeFileName={activeFile?.name ?? null}
+                onSelect={selectFile}
+                hiddenCount={hiddenCount}
+            />
+            <StudyCodeBody isVisible={isExpanded} activeFile={activeFile} />
+            <UnstyledButton onClick={toggleExpanded} data-testid="study-code-toggle">
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx]: Small UX nit: the toggle is rendered unconditionally, so when there are zero files the user can click "Hide full study code" and toggle an empty state in and out. We could gate this on `files.length > 0` (or `activeFile != null`) so the empty state stays static — keeps the action honest about what it can act on.

src/server/db/queries.ts
```diff
@@ -397,6 +397,44 @@ export type StudyReviewWithMeta = {
...
+            .where('status', 'in', ['CODE-SCANNED', 'JOB-ERRORED'])
+            .orderBy('createdAt', 'desc')
+            .orderBy('id', 'desc')
+            .limit(1)
+            .executeTakeFirst(),
+        Action.db
+            .selectFrom('studyJobFile')
+            .select(['id', 'name', 'path'])
+            .where('studyJobId', '=', studyJobId)
+            .where('fileType', '=', 'ENCRYPTED-SECURITY-SCAN-LOG')
```
nathanstitt: Worth thinking about the status/logFile pairing semantics. The two sub-queries are independent: `latestStatus` is whichever of CODE-SCANNED / JOB-ERRORED happened most recently, but `logFile` is just the most recent ENCRYPTED-SECURITY-SCAN-LOG regardless of when the latest status row was inserted. A job that scanned cleanly, then errored later for unrelated reasons, would render a red "fail" icon next to the previously-passing scan log filename. If that's intentional ("show any scan log we have, paired with the most recent terminal status"), worth calling out in the doc comment above so future readers don't read the pairing as causal. If not, we'd probably want the log fetch to filter on `createdAt <= latestStatus.createdAt` or only fetch a log when status is CODE-SCANNED.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
```diff
@@ -0,0 +1,145 @@
...
+            <CheckCircle
+                size={20}
+                weight="fill"
+                color="var(--mantine-color-green-6)"
+                data-icon="pass"
+                aria-hidden="true"
+            />
+        )
+    }
+    if (status === 'JOB-ERRORED') {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx]: this has nothing to do with job run status, why are we looking at JOB-ERRORED ?  

we should be checking the contents of the scan. to see if it contains 'OK', if it does then it's succeeded, otherwise it's failed.  we may adjust that logic later but it's good for a first pass.

nathanstitt (changes requested):
seems like there's some confusion here over scan status - there is nothing tools we have that gives us a clearcut success/failure, and accordingly we do not have a job status for it. the best we can do is look for the word 'OK'

Rod0352:
> seems like there's some confusion here over scan status - there is nothing tools we have that gives us a clearcut success/failure, and accordingly we do not have a job status for it. the best we can do is look for the word 'OK'

sorry, misunderstood a couple things. going to add the "preview coming soon" label and gated the toggle so it doesn't show when there are no files. On the scan status, to check for 'OK' I need the plaintext log from your #707, should I wait for that or strip the status icon for now and do it as a follow-up.

nathanstitt:
I don't have one handy but plenty on qa etc, you can literally just check the scan text for 'OK' though, ie: `contents.includes('OK')`

Rod0352:
> I don't have one handy but plenty on qa etc, you can literally just check the scan text for 'OK' though, ie: `contents.includes('OK')`

makes sense, much cleaner. pushed.

PR #709, opened 2026-05-16
OTTER-519: Show decision date on proposal headers

stetradis:
## Summary

Approved, rejected, and change-requested study views were showing the proposal **submission** date in the header timestamp instead of the date the decision was made. This PR centralizes timestamp resolution in a shared helper and wires it through ProposalRequest so post-decision views display the correct date.

## Changes

- Add decisionTimestampForProposalHeader in src/lib/studies.ts to pick the right date by study status: 
        APPROVED → study.approvedAt
        REJECTED → study.rejectedAt
        CHANGE-REQUESTED → latest NEEDS-CLARIFICATION entry createdAt
        Otherwise → study.submittedAt
- Refactor ProposalRequest to accept optional entries and call the helper internally, replacing the timestampDate prop.
- Pass entries from PostFeedbackView and ProposalSubmitted so reviewer and researcher post-decision views get the correct timestamp.
- Set submittedAt on non-draft studies in insertTestStudyJobData so test fixtures match production expectations.

### DIFF
--- a/src/lib/studies.test.ts
+++ b/src/lib/studies.test.ts
@@ -1,0 +1,100 @@
+import { describe, expect, it } from 'vitest'
+import type { ProposalFeedbackEntry, SelectedStudy } from '@/server/actions/study.actions'
+import { decisionTimestampForProposalHeader } from './studies'
+
+const submittedAt = new Date('2025-04-16T10:00:00Z')
+const approvedAt = new Date('2026-04-20T10:00:00Z')
+const rejectedAt = new Date('2026-05-01T10:00:00Z')
+const clarificationAt = new Date('2026-04-18T10:00:00Z')
+const olderClarificationAt = new Date('2026-04-10T10:00:00Z')
+
+const study = (
+    overrides: Partial<Pick<SelectedStudy, 'status' | 'approvedAt' | 'rejectedAt' | 'submittedAt'>>,
+): SelectedStudy =>
+    ({
+        status: 'PENDING-REVIEW',
+        approvedAt: null,
+        rejectedAt: null,
+        submittedAt,
+        ...overrides,
+    }) as SelectedStudy
+
+const entry = (overrides: Partial<Pick<ProposalFeedbackEntry, 'decision' | 'createdAt'>>): ProposalFeedbackEntry =>
+    ({
+        id: 'entry-1',
+        decision: 'APPROVE',
+        createdAt: new Date('2026-04-16T10:00:00Z'),
+        ...overrides,
+    }) as ProposalFeedbackEntry
+
+describe('decisionTimestampForProposalHeader', () => {
+    it.each([
+        {
+            name: 'APPROVED uses approvedAt',
+            study: study({ status: 'APPROVED', approvedAt }),
+            entries: [] as ProposalFeedbackEntry[],
+            expected: approvedAt,
+        },
+        {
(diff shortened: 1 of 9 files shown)


src/lib/studies.ts
```diff
@@ -12,3 +13,19 @@ export function deriveStudyVersion(entries: { version: number }[]): number {
...
+    }
+    if (study.status === 'REJECTED' && study.rejectedAt) {
+        return study.rejectedAt
+    }
+    if (study.status === 'CHANGE-REQUESTED' && entries.length > 0) {
+        // entries are ordered by createdAt descending
+        const latestClarification = entries.find((e) => e.decision === 'NEEDS-CLARIFICATION')
+        if (latestClarification) return latestClarification.createdAt
+    }
+    return study.submittedAt!
```
nathanstitt: The `submittedAt!` non-null assertion lies to the type system — `submittedAt: Timestamp | null` on the schema, and `unit.helpers.tsx:289` in this same PR explicitly seeds DRAFT studies with `submittedAt: null`

src/lib/studies.ts
```diff
@@ -1,4 +1,5 @@
 import type { StudyJobStatus } from '@/database/types'
+import { ProposalFeedbackEntry, SelectedStudy } from '@/server/actions/study.actions'
```
nathanstitt: this should be `import type { ProposalFeedbackEntry, SelectedStudy } from …`. 

`study.actions.ts` is `'use server'`, and `proposal-initial-request.tsx` is a client component. we should never import server code in this case, using `type` makes it explicit what we're doing

src/lib/studies.test.ts
```diff
@@ -0,0 +1,100 @@
...
+            study: study({ status: 'CHANGE-REQUESTED' }),
+            entries: [],
+            expected: submittedAt,
+        },
+        {
+            name: 'CHANGE-REQUESTED with no clarification entry uses submittedAt',
+            study: study({ status: 'CHANGE-REQUESTED' }),
+            entries: [entry({ decision: 'APPROVE', createdAt: new Date('2026-04-16T10:00:00Z') })],
+            expected: submittedAt,
+        },
```
nathanstitt, line 73: We're testing the fallbacks for APPROVED/REJECTED when their decision date is null, but not the case where `submittedAt` itself is null. Given the `!` assertion on `studies.ts:30`, a test like `study({ status: 'DRAFT', submittedAt: null })` would document what the helper actually does in that case (and would force a decision about whether to return null vs throw, related to the other comment).

nathanstitt (approved):
looks good overall but should probably fix the `!` - I've debated disallowing that in eslint because just about any time it's used it's a code smell.

src/lib/studies.ts
stetradis: This helper is only used from `ProposalRequest`, which only renders **after** submission, so `submittedAt` should always be set here. I didn't add a null/DRAFT case for that reason.

Seeding `submittedAt` in `unit.helpers.tsx` is a separate fix: non-draft fixtures were missing it entirely; only DRAFT stays null.

Makes sense on the `!` - I'll replace it with an explicit guard

src/lib/studies.test.ts
stetradis: Added a test for `submittedAt: null` on the fallback path - it throws, which matches the guard added instead of returning null (this helper isn't used pre-submission).

PR #720, opened 2026-05-18
editor: surface real auth failure code instead of generic permission-denied

nathanstitt:
## Summary
- Hocuspocus forwards `err.reason` (not `err.message`) to the client. `AuthFailureError` set only `message`, so every editor auth failure reached the client as the literal string `"permission-denied"` and the client lost the specific code (`NO_MEMBERSHIP`, `USER_NOT_PROVISIONED`, `STUDY_NOT_EDITABLE`, etc.). Setting `.reason` on the error preserves the `CODE: message` wire format the client parser already expects.
- Enriched the server-side `auth.fail` log with the parsed `code` and the JWT subject (decoded without verification, only for diagnostics) so failures can be triaged by Clerk user id without re-walking the failing branch.
- Added a `console.error` in the editor's `onAuthError` so the resolved `code` / `message` / raw `reason` appear in the browser console when the "Editor unavailable" banner shows — previously the client only knew the failure happened, not why.

## Test plan
- [x] `services/editor/auth.test.ts` — added an assertion that `AuthFailureError.reason` carries the wire format; all 86 tests pass.
- [x] `npm run lint`
- [x] `npm run typecheck`
- [ ] Manual: trigger an auth failure in dev (e.g. open a doc for a study you don't belong to), verify the browser console shows `[collaborative-editor] auth failed { code: "NO_MEMBERSHIP", … }` and the editor-service CloudWatch log line carries `"code":"NO_MEMBERSHIP"` and the `clerkUserId`.

### DIFF
--- a/services/editor/auth.test.ts
+++ b/services/editor/auth.test.ts
@@ -517,12 +517,16 @@ describe('authenticate', () => {
         })
     })
 
-    it('AuthFailureError instances carry name and code', () => {
+    it('AuthFailureError instances carry name, code, and reason', () => {
         const err = new AuthFailureError('NO_MEMBERSHIP', 'no membership in study orgs')
         expect(err).toBeInstanceOf(Error)
         expect(err.name).toBe('AuthFailureError')
         expect(err.code).toBe('NO_MEMBERSHIP')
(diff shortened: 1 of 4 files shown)


therealmarv (approved):
nice, yes we need that

PR #723, opened 2026-05-18
OTTER-521: gate edit page to original author, fix Back route

Rod0352:
OTTER-521 follow-up: QA fixes
Addresses Malarvizhi's QA findings:

Non-author edit access — page now 404s when caller isn't the original researcher (added isStudyOwnedByCurrentUserAction). Server actions already scoped writes to the author; the page now matches so non-authors don't see a form that would silently no-op.
Back button routes to /submitted (the post-feedback page where Edit & resubmit lives) instead of /view.
Resubmission note validates on mount so the Resubmit button starts disabled until the 50-word minimum is hit.


src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx
```diff
@@ -19,6 +20,12 @@ export default async function StudyEditAndResubmitRoute(props: {
     if ('error' in study) return notFound()
     if (study.status !== 'CHANGE-REQUESTED') return notFound()
 
+    // Server-side writes are scoped to the original researcher — gate the
+    // page itself to match, so non-authors don't see a form that would
+    // silently no-op when they try to save.
+    const ownership = await isStudyOwnedByCurrentUserAction({ studyId })
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx]: We could skip the new action entirely. `getStudyAction` already selects `study.researcherId` (it's referenced at line 66 below when wiring `EditResubmitForm`), and the existing pattern for current-user checks in this codebase is `const session = await sessionFromClerk(); if (study.researcherId !== session.user.id) notFound()` — used in `review/page.tsx`, `agreements/page.tsx`, `researcher-profile/page.tsx`. That's a 2-line inline check instead of a new exported action + extra DB roundtrip + extra `'error' in` branch, and it stays consistent with how the other route gates look.

src/server/actions/study-request.ts
```diff
@@ -743,3 +743,17 @@ export const resubmitProposalAction = new Action('resubmitProposalAction', { per
...
+// study researcher before rendering the form. CASL permits other researchers
+// in the same lab to view the study, but writes are scoped to the author —
+// surfacing that as a 404 prevents non-authors from typing into a form that
+// would silently no-op on save.
+export const isStudyOwnedByCurrentUserAction = new Action('isStudyOwnedByCurrentUserAction')
+    .params(z.object({ studyId: z.string() }))
+    .middleware(async ({ params: { studyId } }) => await getInfoForStudyId(studyId))
+    .requireAbilityTo('view', 'Study')
+    .handler(async ({ db, params: { studyId }, session }) => {
+        const study = await db.selectFrom('study').select(['researcherId']).where('id', '=', studyId).executeTakeFirst()
```
nathanstitt: See comments on the page, but it's likely we do not need this action.

If we do keep it, this re-queries `study.researcherId` even though `getInfoForStudyId` in the middleware already selects it (see `server/db/queries.ts:302` — it returns `study.researcherId` alongside `orgId`, `orgSlug`, `status`, `submittedByOrgId`). Other actions in this file destructure the middleware return into the handler context — the same `researcherId` is available without the extra round trip.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/form.tsx
```diff
@@ -27,7 +27,15 @@ export const EditResubmitForm: FC<EditResubmitFormProps> = ({
...
     feedbackEntries,
 }) => {
-    const { form, saveDraft, isSaving } = useEditResubmit()
+    const { form, noteForm, saveDraft, isSaving } = useEditResubmit()
+
+    // Validate the resubmission note on mount so the Resubmit button is
+    // disabled out of the gate (initial value is empty / under the word
+    // minimum); Mantine doesn't auto-validate initial values otherwise.
+    useEffect(() => {
+        noteForm.validateField('resubmissionNote')
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/form.tsx]: Imperative validate-on-mount works, but Mantine's `useForm` initializer supports `validateInputOnChange: true` / `validateInputOnBlur: true`, which would make the initial value invalid out of the box without needing an effect + an eslint-disable. Worth checking where `noteForm` is built (the `useEditResubmit` context) — flipping one of those options there is usually cleaner than driving it from a consuming component.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/footer.tsx
```diff
@@ -37,7 +37,7 @@ export const EditResubmitFooter: FC<EditResubmitFooterProps> = ({ researcherName
             const saved = await saveDraft()
             if (!saved) return
         }
-        router.push(Routes.studyView({ orgSlug, studyId }))
+        router.push(Routes.studySubmitted({ orgSlug, studyId }))
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/footer.tsx:45]: Hardcoding the Back destination to `/submitted` works for the "Edit & resubmit" entry point, but if a user reaches this page from somewhere else (deep-linked URL, dashboard CTA), Back will send them somewhere they didn't come from. The review-side flow uses `?from=...` query params to disambiguate (see `review/page.tsx`'s `searchParams.from === 'code-review'` handling). Might be worth a similar pattern, or at least a comment here noting that `submitted` is the only legitimate entry today.

nathanstitt (changes requested):
Looks good @Rod0352 but don't think we need the action

nathanstitt (approved):

PR #724, opened 2026-05-18
OTTER-555: Update DO view for resubmitted proposals in review

stetradis:
This pull request enhances the handling and display of study proposal resubmissions in the review workflow. It updates both the UI and supporting logic to clearly indicate when a proposal is a resubmission, ensures the correct version and resubmission date are shown, and adds comprehensive test coverage for these behaviors.

Key changes include:

### UI/Component updates

* Updated `ProposalSection` to accept `priorEntries` and determine if the proposal is a resubmission, displaying the correct version number, resubmission status, and date. The section is now collapsed by default on resubmissions, and the status banner and headings reflect the resubmission state.

### Logic improvements

* Enhanced `decisionTimestampForProposalHeader` to use the latest `RESUBMISSION-NOTE` entry as the submission date for resubmitted proposals, ensuring accurate date display in the UI.

### Test coverage

* Added and updated tests to verify resubmission-specific UI behavior, including versioned headings, banner content, collapsed/expanded state, and correct date handling for resubmissions.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/proposal-review-view.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/proposal-review-view.tsx
@@ -223,7 +223,7 @@ function ProposalReviewViewContent({ orgSlug, study, priorEntries, reviewVersion
                     Review initial request
                 </Title>
 
-                <ProposalSection study={study} orgSlug={orgSlug} />
+                <ProposalSection study={study} orgSlug={orgSlug} priorEntries={priorEntries} />
                 <FeedbackAndNotesSection entries={priorEntries} />
                 <ReviewFeedbackSection
                     feedback={feedback}
--- a/src/app/[orgSlug]/study/[studyId]/review/proposal-section.test.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/proposal-section.test.tsx
@@ -1,5 +1,5 @@
 import { lexicalJson } from '@/lib/lexical'
-import { getStudyAction, type SelectedStudy } from '@/server/actions/study.actions'
+import { getStudyAction, type ProposalFeedbackEntry, type SelectedStudy } from '@/server/actions/study.actions'
 import {
     actionResult,
     fireEvent,
@@ -69,6 +69,17 @@ describe('ProposalSection', () => {
         expect(screen.getByText('Researcher')).toBeInTheDocument()
     })
 
+    it('shows resubmission copy and versioned heading when prior entries indicate a resubmission', () => {
+        const priorEntries = [{ version: 2 }] as ProposalFeedbackEntry[]
+
+        renderWithProviders(<ProposalSection study={study} orgSlug="test-org" priorEntries={priorEntries} />)
+
+        expect(screen.getByRole('heading', { name: 'Review initial request v2.0' })).toBeInTheDocument()
+        expect(screen.getByTestId('status-banner')).toHaveTextContent(
+            'has resubmitted a revised initial request requesting permission to use your data',
+        )
+    })
+
     it('renders the status banner with evaluation criteria', () => {
         renderWithProviders(<ProposalSection study={study} orgSlug="test-org" />)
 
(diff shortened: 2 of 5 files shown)


src/app/[orgSlug]/study/[studyId]/review/proposal-section.tsx
```diff
@@ -26,34 +28,47 @@ const EVALUATION_CRITERIA = [
...
+            intro={bannerIntro(labName, isResubmission)}
             criteria={EVALUATION_CRITERIA}
         />
     )
 }
 
-export function ProposalSection({ study, orgSlug, initialExpanded = true }: ProposalSectionProps) {
+export function ProposalSection({ study, orgSlug, priorEntries = [] }: ProposalSectionProps) {
     const labName = study.submittingLabName ?? study.submittedByOrgSlug
+    const studyVersion = deriveStudyVersion(priorEntries)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/proposal-section.tsx]: `ProposalReviewView` already has `reviewVersion` (from `currentReviewVersion(studyId)`, a DB MAX query) and passes both it and `priorEntries` down.  Let's use that vs re-deriving the same number locally.  While they should always agree it's better to have a single source of truth IMO.

src/app/[orgSlug]/study/[studyId]/review/proposal-section.tsx
```diff
@@ -26,34 +28,47 @@ const EVALUATION_CRITERIA = [
...
 
     return (
         <ProposalRequest
             study={study}
             orgSlug={orgSlug}
             stepLabel="STEP 1"
-            heading="Review initial request"
-            banner={<StatusBanner labName={labName} />}
-            initialExpanded={initialExpanded}
+            heading={`Review initial request ${isResubmission ? `v${studyVersion}.0` : ''}`}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/proposal-section.tsx]: Small string nit — when `isResubmission` is false, this resolves to `"Review initial request "` with a trailing space. Browsers do collapse it visually this so isn't a big deal.

```suggestion
            heading={`Review initial request${isResubmission ? ` v${studyVersion}.0` : ''}`}
```

src/app/[orgSlug]/study/[studyId]/review/proposal-section.test.tsx
```diff
@@ -69,6 +69,17 @@ describe('ProposalSection', () => {
         expect(screen.getByText('Researcher')).toBeInTheDocument()
     })
 
+    it('shows resubmission copy and versioned heading when prior entries indicate a resubmission', () => {
+        const priorEntries = [{ version: 2 }] as ProposalFeedbackEntry[]
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/proposal-section.test.tsx]: `[{ version: 2 }] as ProposalFeedbackEntry[]` casts away every other field on `ProposalFeedbackEntry`. It works because `deriveStudyVersion` only reads `.version`, but if a future change in `ProposalSection` (or `ProposalRequest` which it wraps) starts reading another field on entries, this test won't fail at typecheck — it'll fail at runtime. A small `buildEntry({ version: 2 })` helper that returns a fully-populated entry with overrides would be more refactor-safe (and there's a similar `buildEntry` pattern in the sibling `post-feedback-view.test.tsx` worth borrowing).

src/app/[orgSlug]/study/[studyId]/review/proposal-review-view.tsx
```diff
@@ -223,7 +223,7 @@ function ProposalReviewViewContent({ orgSlug, study, priorEntries, reviewVersion
                     Review initial request
                 </Title>
 
-                <ProposalSection study={study} orgSlug={orgSlug} />
+                <ProposalSection study={study} orgSlug={orgSlug} priorEntries={priorEntries} />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/proposal-review-view.tsx]: Just confirming — `ProposalSection` now derives a `studyVersion` from `priorEntries` while this component already has `reviewVersion` as a prop (line 31). They should be equal by construction (both pull from `studyProposalComment.version` MAX), but it's a coincidence we're relying on. See the comment on `proposal-section.tsx:59` — same concern, anchored here for visibility on the data flow.

nathanstitt (approved):
looks good @stetradis but we probably should fix the multiple sources of truth for version number

PR #726, opened 2026-05-19
Inject data source context into workspace CLAUDE.md

rnathuji:


src/server/coder/workspaces.ts
```diff
@@ -248,6 +249,12 @@ const initializeWorkspaceCodeFiles = async (studyId: string): Promise<void> => {
         }
         combinedContextString = combinedContextString + response.content + '\n'
     }
+
+    const orgData = await orgIdFromSlug({ db, params: { orgSlug: codeEnv.slug } })
```
nathanstitt: We already have `study.orgId` in scope because`fetchLatestCodeEnvForStudyId` joins `study` (and `org` for the slug). If we add `'study.orgId'` to that select, we can drop this extra DB query and the `orgIdFromSlug` import + the `db` import below, and just call `generateDataSourcesContextString(codeEnv.orgId)` directly.

src/server/coder/workspaces.ts
```diff
@@ -248,6 +249,12 @@ const initializeWorkspaceCodeFiles = async (studyId: string): Promise<void> => {
...
+    if (orgData !== undefined) {
```
nathanstitt: This `if (orgData !== undefined)` swallows the case where the slug doesn't resolve. Since the slug originated from a row we just joined against `org`, an undefined result here is genuinely impossible, might as well remove it

src/server/coder/workspaces.ts
```diff
@@ -257,3 +264,22 @@ const initializeWorkspaceCodeFiles = async (studyId: string): Promise<void> => {
...
+    const sources = await getDataSourcesForOrg(orgId)
+
+    const sections: string[] = []
+    for (const source of sources) {
+        const lines: string[] = [`# ${source.name}`]
+        if (source.description) lines.push(source.description)
+
+        for (const url of source.urls) {
+            if (url.url !== null) {
+                lines.push(`Documentation (${url.description || 'No description provided'}): ${url.url}`)
```
nathanstitt: This formatter is almost a line-for-line duplicate of `fetchDataDocs` in `src/server/agents/review-agent/runner.ts` which is used for to request AI review of study code.  it has same loop, same null-handling; only differences are the heading prefix (`#` vs `###`) and the URL/description order in the line.  Guessing those differences aren't really important and we could make a `docContextForStudy` helper or something like that?

src/server/db/queries.ts
```diff
@@ -223,7 +223,17 @@ export const studyInfoForStudyId = async (studyId: string) => {
...
+            'orgDataSource.name',
+            'orgDataSource.description',
+            jsonArrayFrom(
+                eb
+                    .selectFrom('orgDataSourceUrl')
+                    .select(['orgDataSourceUrl.url', 'orgDataSourceUrl.description'])
+                    .whereRef('orgDataSourceUrl.orgDataSourceId', '=', 'orgDataSource.id'),
+            ).as('urls'),
+        ])
         .where('orgDataSource.orgId', '=', orgId)
```
nathanstitt, line 238: No `orderBy` on either the outer `orgDataSource` rows or the nested `orgDataSourceUrl` rows. The new test asserts the order of multiple sources and URLs by exact string match, which works today because Postgres tends to return rows in insertion order with our v7 UUIDs — but that's not guaranteed and a future migration / VACUUM could flip it. Adding `.orderBy('orgDataSource.createdAt', 'asc')` (and the same on the nested select) would make both the output and the test deterministic.

src/server/coder/workspaces.test.ts
```diff
@@ -0,0 +1,68 @@
...
+} from '@/tests/unit.helpers'
+import { Selectable } from 'kysely'
+import { Org } from '@/database/types'
+import { generateDataSourcesContextString } from './workspaces'
+
+describe('Workspaces', async () => {
+    let org: Selectable<Org>
+
+    beforeEach(async () => {
+        const { org: createdOrg } = await mockSessionWithTestData({ isAdmin: true, orgSlug: faker.string.alpha(10) })
```
nathanstitt: Tiny nit: this test cares about `org.id` only, so `isAdmin: true` and the random slug aren't doing anything for what's being verified. Doesn't change behavior, just noise in the setup.

nathanstitt (approved):
looks good just a few suggestions, feel free to merge after checking them

src/server/coder/workspaces.ts
rnathuji: Oh -- this was just to make typescript happy as otherwise it was possibly `undefined` response from the query. But, this might go away anyways with the change on the previous comment.

src/server/coder/workspaces.ts
rnathuji: Yeah -- they are similar now, but I wanted to keep them different as I suspect we may tweak them as we begin testing, and I wanted to keep them decoupled since the target contexts are different :man_shrugging: .  I just didn't have any great reason to know what's better / worse as a starting point, but that feels like bad reasoning to make them the same :joy: .

src/server/coder/workspaces.ts
nathanstitt: oh yah you're likely correct ts will need something like this, it's fine I suppose

src/server/coder/workspaces.test.ts
```diff
@@ -0,0 +1,68 @@
+import {
```
kswanson33: FYI there are some tests of `workspace.ts` functions in [coder.test.ts](https://github.com/safeinsights/management-app/blob/main/src/server/coder.test.ts) -- do you think it's better to factor those tests out into file-specific tests, or to keep all the coder tests there?

kswanson33 (approved):

src/server/coder/workspaces.test.ts
rnathuji: I'm not seeing any direct (vs indirect) tests in `coder.test.ts` for `workspace.ts` (e.g. there's no import of functions from `./workspaces`). Am I missing something? Happy to move things over if I'm unintentionally breaking conventions :sweat_smile:

nathanstitt: there's a big `coder.test.ts` that covers all of the workspaces, it should probably be broken up a bit but does cover quite a bit

src/server/coder/workspaces.test.ts
rnathuji: Addressed in [this commit](https://github.com/safeinsights/management-app/pull/726/changes/596ac30b88df92bb79c21236244615b187bcabce).

src/server/coder/workspaces.ts
rnathuji: Addressed in [this commit](https://github.com/safeinsights/management-app/pull/726/changes/4f26149b25d35e6dc64aa4d85e293adeddd3c9a6).

src/server/db/queries.ts
rnathuji: Addressed in [this commit](https://github.com/safeinsights/management-app/pull/726/changes/1d99de5566f47e1c924afb9b3ef08761e5af92b4).

src/server/coder/workspaces.ts
rnathuji: This ended up being resolved in the fix for [this comment](https://github.com/safeinsights/management-app/pull/726#discussion_r3268593711).

src/server/coder/workspaces.test.ts
kswanson33: i think there are some direct tests, imported in a sneaky way via the `.coder` path (line [10](url)) - for example, [createUserAndWorkspace](https://github.com/safeinsights/management-app/blob/main/src/server/coder/workspaces.ts#L199). is that what you meant?

rnathuji: Yeah, I see that now :sweat_smile: . I missed that `index.ts` exporting it all out and was just looking for imports fro m`workspaces` directly.

chrisbendel (approved):

src/server/coder/workspaces.ts
rnathuji: Revisited my thoughts on this one per [this Slack thread](https://openstax.slack.com/archives/C07HE61TJGY/p1779225448091009). Corresponding changes in [this commit](https://github.com/safeinsights/management-app/pull/726/changes/4483e29f8943646c6e41da8a506473f48a96b71a).

src/server/coder/workspaces.test.ts
rnathuji: I removed this new file...but then added a new `utils.test.ts` because I created a new utils / helper file for the other comment to consolidate context string generation. As part of that, I refactored tests a bit to be more unit-y and less integration-y (probably more of a subjective improvement, but it felt right so I went with it :joy: ).

PR #727, opened 2026-05-19
OTTER-547: defer IDE starter file copy until workspace is ready

nathanstitt:
## Summary

Fixes [OTTER-547](https://openstax.atlassian.net/browse/OTTER-547).

Previously, when a researcher clicked **Launch IDE** on the study code screen we copied starter files to the shared `CODER_FILES` directory immediately, then waited several minutes for the Coder workspace to actually start. The `listWorkspaceFilesAction` poller (15s interval) would pick up the files almost immediately and flip the page from the empty/launch view to the file-review view — while the user was still staring at the "Launching IDE…" button waiting for the IDE tab to open.

This change moves the starter-file copy so it runs only once the workspace reports ready (or, in dev mode, after the simulated wait inside `getWorkspaceUrlAction`). The UI now stays on the empty/launch view until the IDE is genuinely ready.

- `createCoderWorkspace` no longer eagerly copies starter files.
- `getCoderWorkspaceUrl` copies them when `isWorkspaceReady(workspace)` first becomes true, just before returning the URL.
- Dev path mirrors the same shape: `createUserAndWorkspaceAction` no longer copies; `getWorkspaceUrlAction` (CODER_DISABLED branch) copies after the 3-second simulated wait.
- `initializeWorkspaceCodeFiles` and `initializeDevWorkspaceFiles` are now idempotent — they bail when the target directory already has files — so repeat ready-polls and re-launches of existing workspaces don't clobber user edits.

## Test plan

- [x] `npx vitest run` — `src/server/coder.test.ts`, `src/server/actions/workspaces.actions.test.ts`, `src/hooks/use-workspace-launcher.test.tsx`, `src/components/study/study-code.test.tsx` (45 + 24 unit tests pass)
- [x] `tsc --noEmit --skipLibCheck` clean
- [x] `eslint` clean on touched files
- [ ] Manual: in dev mode, click Launch IDE on a study with no files — confirm the empty/launch view stays put for the full ~3s wait, then the file list appears together with the IDE tab opening
- [ ] Manual: re-launch IDE on a study that already has user-edited files — confirm files are not clobbered

[OTTER-547]: https://openstax.atlassian.net/browse/OTTER-547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/actions/workspaces.actions.ts
+++ b/src/server/actions/workspaces.actions.ts
@@ -83,7 +83,6 @@ export const createUserAndWorkspaceAction = new Action('createUserAndWorkspaceAc
         if (!session) throw new Error('Unauthorized')
         await resetBaselineJob(db, studyId)
         if (CODER_DISABLED) {
-            await initializeDevWorkspaceFiles(studyId)
             return {
                 success: true,
                 workspace: { id: `dev-workspace-${studyId}` },
@@ -107,6 +106,7 @@ export const getWorkspaceUrlAction = new Action('getWorkspaceUrlAction', {})
         if (CODER_DISABLED) {
             // these envs do not have a 'real' coder setup
             await new Promise((resolve) => setTimeout(resolve, 3000))
+            await initializeDevWorkspaceFiles(studyId)
             return `https://coder.dev.example.com/workspace/${studyId}`
         }
         return await getCoderWorkspaceUrl(studyId, workspaceId)
--- a/src/server/coder/workspaces.ts
+++ b/src/server/coder/workspaces.ts
@@ -15,6 +15,8 @@ import { CoderWorkspace, CoderWorkspaceEvent } from './types'
 import { getCoderUser, getOrCreateCoderUser } from './users'
 import { generateWorkspaceName } from './utils'
 import { fetchLatestCodeEnvForStudyId } from '../db/queries'
+import { latestStudyJobCreatedAt } from '../db/mutations'
+import { db } from '@/database'
 import { fetchFileContents } from '../storage'
 import * as fs from 'node:fs/promises'
 import * as path from 'node:path'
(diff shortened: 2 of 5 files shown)


therealmarv (approved):
Not sure this is a blocker (I guess not): This PR assumes that coder always starts below 60s.

When that assumption is not true we have an edge case:

---

## Timing edge case: Submit can be enabled without any edits

The PR moves starter-file copying so it runs after the workspace is ready, which is great. But the baseline `studyJob` timestamp is still created at the moment the user clicks **Launch IDE**, while the starter files are backdated by a fixed **60 seconds** from the time they're copied. Those two clocks aren't tied together, so if Coder takes longer than ~60s to become ready, the untouched starter files end up with an `mtime` that's *newer* than the baseline. The UI uses exactly that comparison to decide whether the researcher has edited anything, so **Submit becomes enabled even though the researcher hasn't touched a file**.

### How it happens

1. **T = 0s** — User clicks **Launch IDE**.
   `resetBaselineJob` creates a `studyJob` with `createdAt = 12:00:00`.
2. **T = 90s** — Coder finishes provisioning and the workspace becomes ready.
   `initializeWorkspaceCodeFiles` runs and writes the starter files with `mtime = 12:01:30 − 60s = 12:00:30`.
3. UI evaluates `hasChangedSinceLastJob`:
   ```
   file.mtime (12:00:30) > jobCreatedAt (12:00:00)  →  true
   ```
   `filesChanged = true` → **Submit is enabled** on first paint, with zero edits from the user.

The condition for the bug is simply:

```
(timeWorkspaceReady − timeLaunchClicked) > 60s
```

Coder cold starts routinely exceed 60s, so this isn't a theoretical edge case, it's the common path.

### Relevant code

- [`initializeWorkspaceCodeFiles()` now runs when the workspace is ready](https://github.com/safeinsights/management-app/pull/727/files#diff-13dc2fb37f49daf42abdcde9b0ce64730040fcbbce4ad2a2863258c18f43e7b2R62)
- [`resetBaselineJob()` still runs at launch start](https://github.com/safeinsights/management-app/pull/727/files#diff-e75ddebf70d61362fe82c5c0f0e370349f13b1535737ade16c325b3d71bf9be8R84)
- [`pastDate` backdates by 60s from copy time](https://github.com/safeinsights/management-app/pull/727/files#diff-13dc2fb37f49daf42abdcde9b0ce64730040fcbbce4ad2a2863258c18f43e7b2R237)
- [`hasChangedSinceLastJob()` compares file `mtime` against the baseline](https://github.com/safeinsights/management-app/blob/19b7ee26786fb03372345a31fc3b2b3737df6bfb/src/hooks/use-ide-files.ts#L35-L38)

### Suggested fixes (either one removes the timing dependency)

1. **Backdate starter files relative to the baseline timestamp.** Look up the current `studyJob.createdAt` inside `initializeWorkspaceCodeFiles` and use `pastDate = new Date(jobCreatedAt − 1000)`. Starter files are then guaranteed to be older than the baseline, no matter how long Coder takes.
2. **Co-locate the baseline reset with the file copy.** Move `resetBaselineJob` out of `createUserAndWorkspaceAction` and into `initializeWorkspaceCodeFiles` (behind the same "directory is empty" guard), so the baseline and the starter-file `mtime`s are stamped from the same moment. The `CODER_DISABLED` path needs the equivalent move.

Option 1 is the smaller change and leaves the dev path alone, so probably the safer pick unless there's a reason to want the baseline to mean "workspace actually ready".

nathanstitt:
just made an update, ty!

therealmarv (approved):
🟢

PR #636, opened 2026-04-13
Fix reviewer redirect loop and Previous button for post-review statuses

nathanstitt:
## Summary
Fixes two related bugs for reviewers:

1. **Redirect loop on resubmission** — When code is resubmitted and a reviewer clicks "View", the URL flips between `/agreements` and `/review` forever
2. **Previous button broken (OTTER-512)** — On the code review page, the Previous button doesn't work for statuses: code approved, code errored, code rejected, result needs review, result ready, result rejected. It just refreshes the page instead of showing the proposal.

## Root cause

**Loop**: For resubmitted studies, `/agreements` redirects to `/review` (job count > 1) and `/review` redirects back to `/agreements` (code not yet reviewed on latest job).

**Previous button**: The reviewer's Previous button linked to `/agreements`, which redirected back to `/review` for post-review statuses (code already reviewed), causing a silent refresh.

## Fix

- Review page checks job count before redirecting to agreements — skips the gate for resubmissions
- Previous button on `CodeReviewView` now links to `/review?from=agreements` which renders the `ProposalReviewView` directly, avoiding the agreements redirect entirely

Resolves: https://openstax.atlassian.net/browse/OTTER-413 and https://openstax.atlassian.net/browse/OTTER-512

PR #728, opened 2026-05-19
OTTER-546: scope agreements ack to the role the page was rendered for

nathanstitt:
## Summary

Fixes [OTTER-546](https://openstax.atlassian.net/browse/OTTER-546).

QA was still reporting the Agreements page never appearing for DOs on studies in *Code needs review*, even after PR #636 wired up the gate. Investigation found that `ackAgreementsAction` was updating the ack column based on **session org membership only**, not on which Agreements page the user was actually on. When a user belonged to both the reviewer enclave and the submitting lab — common with multi-org QA accounts, and the default shape of test fixtures where `orgId === submittedByOrgId` — proceeding past the **researcher** Agreements page silently set BOTH `researcherAgreementsAckedAt` and `reviewerAgreementsAckedAt`. The reviewer gate in `review/page.tsx` then saw a non-null `reviewerAgreementsAckedAt` and skipped the Agreements page, dropping the DO directly onto the code-review view.

### Fix

- Add a required `role: 'researcher' | 'reviewer'` param to `ackAgreementsAction`.
- The action only updates the column that matches that role, and verifies the caller actually belongs to the corresponding org.
- `AgreementsPage` already knows whether it is the reviewer variant via its `isReviewer` prop, so it passes that through.

### Caveat

This fixes the behavior going forward. Studies already in the wild that had `reviewerAgreementsAckedAt` populated incorrectly by the prior behavior will still skip the gate — no migration is included because there is no reliable way to distinguish bug-set acks from legitimate ones. If QA wants those historical studies re-gated, a DBA can `UPDATE study SET reviewer_agreements_acked_at = NULL WHERE …` against the affected rows. Please flag any staging/prod studies that QA wants reset.

## Test plan

- [x] `npx vitest run src/server/actions/study.actions.test.ts src/app/[orgSlug]/study/[studyId]/agreements/page.test.tsx src/app/[orgSlug]/study/[studyId]/review/page.test.tsx` — 88 tests pass
- [x] New unit test: `does NOT set reviewerAgreementsAckedAt when role=researcher and caller would pass both org checks` — directly exercises the bug
- [x] New unit test: `fails when role=reviewer but user is not a member of the reviewer org` — ensures the role param can't be abused to ack a column the caller has no business touching
- [x] `tsc --noEmit --skipLibCheck` clean
- [x] `eslint` clean on touched files
- [ ] Manual on staging with QA's reproducible test account: researcher submits code → reviewer opens dashboard → clicks View → Agreements page is shown (not skipped). Click Proceed → land on code-review view. Re-open from dashboard → land directly on code-review view (Rule 3 still works).

[OTTER-546]: https://openstax.atlassian.net/browse/OTTER-546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/agreements/agreements-page.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/agreements/agreements-page.tsx
@@ -105,7 +105,7 @@ export function AgreementsPage({
     const resolvedProceedLabel = proceedLabel ?? (isReviewer ? 'Proceed to Step 3' : 'Proceed to Step 4')
 
     const handleProceed = async () => {
-        await ackAgreementsAction({ studyId })
+        await ackAgreementsAction({ studyId, role: isReviewer ? 'reviewer' : 'researcher' })
         router.push(proceedHref as Route)
     }
     const handlePrevious = () => router.push(previousHref as Route)
--- a/src/server/actions/study.actions.test.ts
+++ b/src/server/actions/study.actions.test.ts
@@ -516,14 +516,14 @@ describe('Study Actions', () => {
 })
 
 describe('ackAgreementsAction', () => {
-    it('sets researcherAgreementsAckedAt when called by lab member', async () => {
+    it('sets researcherAgreementsAckedAt when called with role=researcher by lab member', async () => {
         const { org: labOrg, user } = await mockSessionWithTestData({ orgType: 'lab' })
         const enclaveOrg = await insertTestOrg({ slug: 'test-enclave', type: 'enclave' })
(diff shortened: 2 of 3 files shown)


therealmarv (approved):
ups, wrong comment on wrong PR... will update this asap

nathanstitt:
I think you meant to leave this on #727 😁  probably a valid concern tho

therealmarv (approved):
now corrected: this PR is fine as it is! 🟢

PR #730, opened 2026-05-19
OTTER-471: true-parallel test

therealmarv:
Main fix for feature flag behaviour is merged through https://github.com/safeinsights/management-app/pull/710

Legacy paths are not fixed because soon gone (will also mention that in OTTER-471 card).

This branch only adds minor additions in test coverage:

- Adds a true-parallel race test for the proposal review submit action, verifying that simultaneous approve and reject calls produce exactly one winner and one terminal state.
- Adds a matching parallel race test for the code review decision submit action, verifying that concurrent approve and reject calls leave exactly one terminal status row and one review comment.


nathanstitt (approved):
Thanks!

PR #731, opened 2026-05-20
OTTER-558: Code - Edit & Resubmit Study Code

therealmarv:
see [OTTER-558](https://openstax.atlassian.net/browse/OTTER-558)

- Add a new edit-and-resubmit study code flow with an autosaving resubmission note, dirty-state aware Cancel / Save-and-exit toggle, and tooltips on the IDE button, main-file column header, and Save-and-exit.
- Add a status guard on the resubmit action so code can only be resubmitted from a valid state.
- Update the empty study code view to surface a Launch IDE option when appropriate.
- Add a database migration to store the resubmission note alongside the study.
- Share a single resubmission note section between the proposal and code resubmit flows.
- Add unit tests covering the new footer, shared note section, autosave indicator, and server action behavior.
- Gate the new flow behind a feature flag.

[OTTER-558]: https://openstax.atlassian.net/browse/OTTER-558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/contexts/edit-code-resubmit/context.tsx
```diff
@@ -0,0 +1,137 @@
...
+        onSuccess: () => setLastSavedAt(new Date()),
+        onError: reportMutationError('Unable to save resubmission note draft'),
+    })
+
+    const flushSave = useCallback(
+        async (value: string) => {
+            if (value === lastSavedValueRef.current) return true
+            // Record the attempt up front so a failed save doesn't loop the
+            // debounced autosave on every mutation-state re-render.
+            lastSavedValueRef.current = value
```
nathanstitt: Worth a careful look — this looks like a silent data-loss path on save failure. `lastSavedValueRef.current = value` is set *before* the await, and the `catch` returns `false` without rolling it back. So after a transient save failure:

1. `lastSavedValueRef.current` now equals the value that *failed* to persist.
2. The autosave `useEffect` guard (`if (currentNote === lastSavedValueRef.current) return`) then skips re-saving it, so the failed content is never retried once the user stops typing.
3. Worse, `saveDraft()` (Save and exit) calls `flushSave(pendingValueRef.current)`, which hits the line-65 early-return `return true` because the ref already matches — so the footer's `handleSaveAndExit` sees `saved === true` and navigates away, even though the note was never written.

Net effect: a failed autosave followed by 'Save and exit' reports success and loses the note. The fix is to advance `lastSavedValueRef.current` only on success, and use a separate in-flight guard (e.g. a `savingValueRef`) to prevent the re-render loop the comment is worried about.

src/server/actions/study-request.ts
```diff
@@ -755,3 +755,95 @@ export const resubmitProposalAction = new Action('resubmitProposalAction', { per
...
+        }),
+    )
+    .middleware(async ({ params: { studyId } }) => await getInfoForStudyId(studyId))
+    .requireAbilityTo('create', 'StudyJob')
+    .handler(async ({ orgSlug, params, session, db }) => {
+        const { studyId, mainFileName, fileNames, resubmissionNote } = params
+
+        const latestJob = await latestJobForStudyOrNull(studyId)
+        const latestStatus = latestJob?.statusChanges.at(0)?.status
+        const allowedStatuses = new Set(['CODE-CHANGES-REQUESTED', 'JOB-ERRORED', 'RUN-COMPLETE', 'FILES-REJECTED'])
```
nathanstitt: This allowed-status set is duplicated verbatim in `resubmit/page.tsx` (`ALLOWED_LATEST_JOB_STATUSES`). They have to stay in lockstep — if they drift, the page renders the edit UI for a status the action rejects (or hides it for one the action accepts), which is a confusing dead-end for the researcher. Worth lifting to a single exported const (somewhere both server-action and the page can import) so there's one source of truth.

src/server/actions/study-request.ts
```diff
@@ -755,3 +755,95 @@ export const resubmitProposalAction = new Action('resubmitProposalAction', { per
...
+
+        await db
+            .updateTable('studyJob')
+            .set({ resubmissionNote: JSON.parse(lexicalJson(resubmissionNote)) })
+            .where('id', '=', studyJobId)
+            .execute()
+
+        await db
+            .updateTable('study')
+            .set({ status: 'PENDING-REVIEW', codeResubmissionNoteDraft: null })
```
nathanstitt: `submitStudyCodeAction` sets `submittedAt: new Date()` when it flips a study to PENDING-REVIEW (study-request.ts:599), but the resubmit path here only sets `status` + clears the draft. Is leaving `submittedAt` at the original submission time intentional? If the review queue ordering or any 'submitted at' display reads that column, a resubmitted study would show a stale timestamp. If the intent is that `submittedAt` means *first* submission, a one-line comment here would save the next reader the same double-take.

src/server/actions/study-request.ts
```diff
@@ -755,3 +755,95 @@ export const resubmitProposalAction = new Action('resubmitProposalAction', { per
...
+            userId,
+            studyId,
+            orgSlug,
+            sanitizedMainFileName,
+            additionalFileNames,
+        )
+
+        let coderFilesPath = await getConfigValue('CODER_FILES')
+        if (!CODER_DISABLED) coderFilesPath += `/${studyId}`
+        for (const fileName of fileNames) {
```
nathanstitt, line 819: Heads up (inherited from `submitStudyCodeAction`, not introduced here, so optional): because `performsMutations: true` wraps the whole handler in `db.transaction()`, these S3 copies happen inside the DB transaction window. If a later statement rolls the transaction back, the S3 objects are already written and become orphans — S3 isn't part of the rollback. Same shape as the existing submit action, so consistent, but as this is a fresh copy of the pattern it might be a good moment to note it / consider moving the file copy after commit in both places.

src/server/actions/study.actions.ts
```diff
@@ -822,9 +848,39 @@ export const getCodeReviewFeedbackAction = new Action('getCodeReviewFeedbackActi
...
+                entryType: 'RESUBMISSION-NOTE' as const,
+                decision: null,
+                body: j.resubmissionNote as NonNullable<typeof j.resubmissionNote>,
+                criteria: null,
+                createdAt: j.createdAt,
+                authorName: j.authorName ?? '',
+                version: jobVersion.get(j.studyJobId) ?? null,
+            }))
+
+        return [...reviewerEntries, ...noteEntries].sort(
```
nathanstitt: Minor: the final ordering is `createdAt` desc only. When a reviewer decision and a resubmission note on the same round carry identical `createdAt` (possible with fast operations or fixtures created in the same tick), their relative order is unstable, and the `getCodeReviewFeedbackAction` test asserts a specific order. A tiebreaker — e.g. sort by `version` then entryType, or fall back to id — would make it deterministic.

src/app/[orgSlug]/study/[studyId]/resubmit/page.tsx
```diff
@@ -1,9 +1,19 @@
...
+import { fetchLatestCodeEnvForStudyIdOrNull, latestJobForStudyOrNull } from '@/server/db/queries'
+import { getCodeReviewFeedbackAction, getStudyAction } from '@/server/actions/study.actions'
+import { EditCodeResubmitProvider } from '@/contexts/edit-code-resubmit'
 import { ResubmitCodeProvider } from '@/contexts/resubmit-code'
 import { ResubmitStudyCodeForm } from './form'
+import { EditStudyCodeView } from './edit-study-code-view'
+import { FeatureFlagGate } from './feature-flag-gate'
+import { NotFoundOnMount } from './not-found-on-mount'
+
+const ALLOWED_LATEST_JOB_STATUSES = new Set(['CODE-CHANGES-REQUESTED', 'JOB-ERRORED', 'RUN-COMPLETE', 'FILES-REJECTED'])
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/resubmit/page.tsx]: This is the other copy of the allowed-status set (the action has the same literal at study-request.ts:794). Same note as there — extracting a shared exported const would keep the page-level gate and the server-side guard from silently diverging.

src/app/[orgSlug]/study/[studyId]/resubmit/feature-flag-gate.tsx
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/resubmit/feature-flag-gate.tsx:1]: we have somehow ended up with a lot of different flags and I'm not sure why.  With this we now have additional files as well.   lets keep it all under `StudyDetailsRedesignFeatureFlag` or pick one of the existing ones. 

The entire point of us making a flag was so we can easily grep for it and remove all instances, but somehow we have ended up with at least 5 different ones 😭

src/app/[orgSlug]/study/[studyId]/resubmit/not-found-on-mount.tsx
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/resubmit/not-found-on-mount.tsx:1]: Thanks for the comment, that explains why you made this, but if I understand the flow correctly I think we should go ahead and render the page.  my reasoning is:

* Only users who have opted in to the flag will see the link.
* If they happen to somehow visit it while they don't have the flag turned on, that's on them.  

The flag isn't intended to be a true block, just a way for us to develop without distracting existing users, so I think that bit of leakage would be fine,

src/server/actions/study.actions.ts
```diff
@@ -806,13 +807,38 @@ export const getCodeReviewFeedbackAction = new Action('getCodeReviewFeedbackActi
...
+                'studyJob.resubmissionNote',
+                'studyJob.createdAt',
+                'submission.userId as authorId',
+                'author.fullName as authorName',
+            ])
+            .where('studyJob.studyId', '=', studyId)
+            .orderBy('studyJob.createdAt', 'asc')
+            .execute()
+
+        const jobVersion = new Map(codeJobs.map((j, i) => [j.studyJobId, i + 1]))
```
nathanstitt, line 833: The version derivation, per-job version lookup, and the final sort can all move into SQL with a `row_number()` window in a CTE — which collapses the two sequential queries into one round trip and lets both row types pick up `version` from the same source instead of re-deriving it via this `Map`.

The CTE shape:

```ts
.with('versioned_jobs', (qb) =>
    qb
        .selectFrom('studyJob')
        .leftJoin('jobStatusChange as submission', (join) =>
            join.onRef('submission.studyJobId', '=', 'studyJob.id').on('submission.status', '=', 'CODE-SUBMITTED'),
        )
        .leftJoin('user as author', 'author.id', 'submission.userId')
        .where('studyJob.studyId', '=', studyId)
        .select((eb) => [
            'studyJob.id as studyJobId',
            'studyJob.resubmissionNote',
            'studyJob.createdAt',
            'submission.userId as authorId',
            'author.fullName as authorName',
            // ::int matters — row_number() is bigint, which node-postgres returns as a *string*,
            // so without the cast `version` comes back '2' not 2 and breaks numeric compares.
            sql<number>`row_number() over (partition by ${sql.ref('studyJob.studyId')} order by ${sql.ref('studyJob.createdAt')} asc)::int`.as('version'),
        ]),
)
```

Then the reviewer query `leftJoin('versioned_jobs', 'versioned_jobs.studyJobId', 'studyReviewComment.studyJobId')` and selects `versioned_jobs.version` directly — no `jobVersion` Map, no second pass to attach versions. The resubmission-note rows come straight off `versioned_jobs` (`where resubmissionNote is not null`). Both sets `UNION ALL` together with a trailing `.orderBy('createdAt', 'desc')`, so the JS `.sort()` (and its repeated `new Date(...)` parsing) goes away too.

If you do try to go all the way to the union and do it all in SQL: I tried it and got it working but was so complex that I liked your `map` in js version better.

nathanstitt (changes requested):
looks really good @therealmarv - mostly just suggestions for you to think about

PR #734, opened 2026-05-20
OTTER 537: polish the post-submission study code page and add a per-file download icon

therealmarv:
see [OTTER-537 comments](https://openstax.atlassian.net/browse/OTTER-537)

- Update the post-submission status banner copy to match the Figma design.
- Hide encrypted scan logs and other non-code artifacts from the submitted code table so only files the researcher uploaded show up.
- Remove the duplicate "Hide full study code" toggle so the show/hide control only appears once.
- Move the "View approved initial request" link to the top-right of the submitted code section header.
- Add a one-click download icon next to the eye icon for each submitted file.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/view/code-post-submission-view.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/view/code-post-submission-view.tsx
@@ -62,9 +62,6 @@ export function CodePostSubmissionView({
         ['Study code'],
     ]
 
-    const toggleLabel = expanded ? 'Hide full study code' : 'View full study code'
-    const caretRotation = expanded ? 'rotate(-90deg)' : 'rotate(0deg)'
-
     const codeFiles = job.files.filter((f) => f.fileType === 'MAIN-CODE' || f.fileType === 'SUPPLEMENTAL-CODE')
 
     return (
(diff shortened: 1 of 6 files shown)


src/app/[orgSlug]/study/[studyId]/view/code-post-submission-view.tsx
```diff
@@ -62,8 +62,7 @@ export function CodePostSubmissionView({
         ['Study code'],
     ]
 
-    const toggleLabel = expanded ? 'Hide full study code' : 'View full study code'
-    const caretRotation = expanded ? 'rotate(-90deg)' : 'rotate(0deg)'
+    const codeFiles = job.files.filter((f) => f.fileType === 'MAIN-CODE' || f.fileType === 'SUPPLEMENTAL-CODE')
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/code-post-submission-view.tsx]: We have a `filterAndOrderCodeFiles` in `src/app/[orgSlug]/study/[studyId]/review/study-code-files.ts`, could use that here to have a single source of truth.


```ts
import { filterAndOrderCodeFiles } from '@/app/[orgSlug]/study/[studyId]/review/study-code-files'
// ...
const codeFiles = filterAndOrderCodeFiles(job.files)
```

src/app/[orgSlug]/study/[studyId]/view/code-post-submission-view.tsx
```diff
@@ -86,37 +85,46 @@ export function CodePostSubmissionView({
...
-                        mt="md"
-                        display="inline-flex"
-                        style={{ alignItems: 'center', gap: 4 }}
-                        aria-expanded={expanded}
-                        data-testid="study-code-toggle"
-                    >
-                        {toggleLabel}
-                        <CaretRightIcon size={12} style={{ transition: 'transform 200ms', transform: caretRotation }} />
-                    </Anchor>
+                    {!expanded && (
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/code-post-submission-view.tsx]: CLAUDE.md has a rule against `{condition && <Component />}` for conditional rendering — preferring `isVisible` props that return `null`. This block is small enough that it's borderline (the rule's aimed at large blocks), so probably fine as-is. Flagging it in case you want to match the convention strictly with something like `<ExpandToggle isVisible={!expanded} onClick={toggle} />`.

Also worth a quick visual check: the previous version animated the caret with `transition: transform 200ms`. The new toggle drops the rotation entirely (icon unmounts on expand). Intentional per Figma, or worth keeping the rotate-in?

nathanstitt (approved):
just a few small comments, thanks @therealmarv

src/app/[orgSlug]/study/[studyId]/view/code-post-submission-view.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/view/code-post-submission-view.tsx]: Extracted ExpandToggle with an isVisible prop that returns null when hidden, to match the convention. Good catch.

On the caret rotation: that one wasn't a Figma decision, it dropped out of the refactor. The old version animated because a single anchor flipped its label and caret rotation in place, so the transform transition had a "from" and "to" on the same DOM node. The new layout splits that into two separate anchors (View above the collapsed paper, Hide inside it), and the View one unmounts on expand. CSS transitions are per-node and need a previous computed style to interpolate from, so once the node is gone and a different one mounts in its place, the transform transition simply can't fire, the new icon just renders at its starting rotation. The Mantine Collapse already animates the section open/close, which is the part you actually notice. The AC on OTTER-537 doesn't call out an animation either. Happy to put it back if you'd rather, but the only ways to get one across the swap are either to keep both anchors mounted and toggle visibility (which walks back toward the pattern the isVisible-return-null convention is trying to avoid), or to stagger a style change across frames after mount, which feels like a lot of plumbing for a small caret flicker.

PR #736, opened 2026-05-21
Migrate to PNPM

chrisbendel:


bin/build-app
```diff
@@ -4,7 +4,19 @@ set -euo pipefail
 # https://github.com/getsentry/sentry-javascript/issues/15100
 # workaround for build spewing thousands of lines of:
 # <w> [webpack.cache.PackFileCacheStrategy] Serializing big strings (135kiB) impacts deserialization performance ...
-npm run build 2>&1 | stdbuf -oL grep -v 'PackFileCacheStrategy'
+#
+# TEMPORARY: invoking `next` directly via node_modules/.bin rather than `pnpm run build`.
```
chrisbendel, line 8: highlighting this note relating https://github.com/safeinsights/iac/pull/161

package.json
```diff
@@ -130,19 +139,21 @@
         "eslint-plugin-anti-trojan-source": "^1.1.4",
         "eslint-plugin-import": "^2.32.0",
         "eslint-plugin-react": "^7.37.5",
-        "eslint-plugin-react-hooks": "^7.0.1",
+        "eslint-plugin-react-hooks": "7.0.1",
```
nathanstitt, line 142: was dropping the caret intentional?

nathanstitt (approved):
looks great, ty!

nathanstitt:
I'm going to merge this before it conflicts when #740 is merged

package.json
chrisbendel: https://openstax.atlassian.net/browse/OTTER-570

It introduces some new rules by default we should prob be following up on, it spawned this ticket as a result, just a good dev chore to do (might take some work or might be super easy)

PR #737, opened 2026-05-21
Restyle edit-and-resubmit confirmation modal to match mockup

jbwilson8:
Issue: [OTTER-568](https://openstax.atlassian.net/browse/OTTER-568)

## Summary

Restyles the "Confirm initial request submission" modal on the edit-and-resubmit
page to match the design mockup.

*Update(06/02/26):  restyled modal to matching existing modals or now. There will be a future redesign of all modals to ensure consistency across the BMA. 

[OTTER-568]: https://openstax.atlassian.net/browse/OTTER-568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/footer.tsx
```diff
@@ -105,18 +105,31 @@ export const EditResubmitFooter: FC<EditResubmitFooterProps> = ({ researcherName
             </AppModal>
 
             <AppModal
-                size="md"
+                size="lg"
+                radius={0}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/footer.tsx]: Worth a thought: `radius={0}` shows up four times in this block (modal + both buttons). If the Figma direction is that the *whole app* is moving to square corners, this probably belongs in `theme.ts` (or in `AppModal`'s defaults + a Button variant) rather than per-call. If it's truly a one-off for just this modal, no action needed — but the next person reading this will wonder which.

A short rationale comment (`// Figma calls for square corners on confirmation modals — applied here until/unless we add a wrapper variant`) would save the next reader the same question.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/footer.tsx
```diff
@@ -105,18 +105,31 @@ export const EditResubmitFooter: FC<EditResubmitFooterProps> = ({ researcherName
...
                 isOpen={isConfirmOpen}
                 onClose={() => setConfirmOpen(false)}
-                title="Resubmit initial request?"
+                title="Confirm initial request submission?"
+                styles={{
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/footer.tsx]: Small consistency thing: `AppModal`'s built-in styles use hardcoded `'0px 40px'` / `'40px'` padding (see `src/components/modal.tsx:37-46`), while this override switches to `var(--mantine-spacing-xl)` / `var(--mantine-spacing-xxl)`. Both work, but mixing tokens with hardcoded pixels in the same component is a little confusing. Either direction would be cleaner:

- **Tokens-first** (probably better long-term): refactor `AppModal`'s defaults to use the same `var(--mantine-spacing-...)` tokens, then this override stays in the same vocabulary.
- **Hardcoded** (simpler but locks in design): keep this override in pixels (`padding: '32px'` / `gap: '40px'`) to match the wrapper's convention.

Mantine resolves `xl`/`xxl` to `2rem`/`2.5rem` (from `src/theme.ts:143`), so behavior is the same either way — just a readability nudge.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/footer.tsx
```diff
@@ -105,18 +105,31 @@ export const EditResubmitFooter: FC<EditResubmitFooterProps> = ({ researcherName
...
+                    header: { padding: 'var(--mantine-spacing-xl)', gap: 'var(--mantine-spacing-xxl)' },
+                    close: { height: '28px', width: '28px', minHeight: '28px', minWidth: '28px' },
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/footer.tsx]: The close button gets bumped from the wrapper's default 16px to 28px here. If the design feedback is that 16px is too small in general (e.g. clickable-target accessibility, this size feels right across modals), that's a `AppModal` default worth raising — otherwise this modal will look visually different from every other confirmation in the app. Worth a quick check with design whether 28px is just-here or the new house default.

nathanstitt (commented):
There's a lot of custom styling going on here, let's get UX involved to determine their intent.  Should we be special casing just this modal? that doesn't sound right to me, I'm guessing they want these styles for ALL modals?

cc @dkhowala and @iccgoncalves

nathanstitt (approved):
approving inso we can update later

PR #738, opened 2026-05-21
Speed up checks?

chrisbendel:


therealmarv:
ah nice, if helpful: I've also experimented with parallel testing/checks/CI in the past:
https://github.com/safeinsights/management-app/pull/564
https://github.com/safeinsights/management-app/pull/565

I've decided at the end against it because was thinking probably good we stop early at failing unit tests. I think I got max like 2-3 minutes less with one of the closed PRs.

But in general faster checks is better I think?!

chrisbendel:
@therealmarv yea this was just an experiment to see the different times, if we want to fail as early as possible i think there's a way we can cancel all other jobs if something fails 🤔

PR #739, opened 2026-05-22
OTTER-567: Add submission confirmation modal

stetradis:
**User Interface and Experience Improvements:**

* The submit button label in `ProposalFooter` has been changed from "Submit study proposal" to "Submit initial request", and clicking it now opens a confirmation modal instead of submitting immediately. The modal asks users to confirm their submission and warns that further edits are not permitted once submitted. 

**Test Updates for Consistency:**

* All relevant tests have been updated to expect the new button label ("Submit initial request") and to interact with the confirmation modal by clicking "Yes, submit initial request" before verifying submission.


nathanstitt (approved):
#737 has an almost identical modal, once both gets merged we should refactor to share code

PR #740, opened 2026-05-23
Trivy fix: upgrade js-cookie

stetradis:
Updates the `js-cookie` dependency to version 3.0.7 across the project.


nathanstitt (approved):

nathanstitt (approved):
looks good, but is conflicting after #736 was merged

therealmarv (commented):
sorry, have not seen this fix. But I think the cleaner fix is to give Clerk libs a slight upgrade bump in version number instead of using override:
https://github.com/safeinsights/management-app/pull/743/changes/7c6b33cf7f9d380a5da14071d53c13b6cd331e2d , see https://github.com/safeinsights/management-app/pull/743

But thanks for also noticing the override not working on main!

stetradis:
Thanks @therealmarv, closing this in favor of #743 👍

PR #743, opened 2026-05-26
pnpm follow-up

therealmarv:
# PNPM follow-up

seems I've missed the review window of the (now merged) `pnpm` branch. Here are some follow up fixes.

The most important one is 1.: the migration left our dependency `overrides` in a place pnpm 11 silently ignores, so every security pin was inert. 2. clears a HIGH Trivy finding that surfaced once overrides were sorted out.

---

## Checklist

| Done |   # | File                                                                   | What changed                                                                                                                                                                                                                                                                                                                                                          |
| ---- | --: | ---------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| [x]  |   1 | [`pnpm-workspace.yaml`](pnpm-workspace.yaml), [`package.json`](package.json) | **Moved dependency `overrides` to where pnpm 11 actually reads them.** pnpm 11 ignores the npm-style top-level `overrides` field in `package.json` (and even `pnpm.overrides`), so after the migration every security pin (`minimatch`, `picomatch`, `postcss`, `fast-xml-parser`, `nodemailer`, ...) was silently inert. Moved them into `pnpm-workspace.yaml` under `overrides:` so they apply again (e.g. `minimatch` 3.1.5 → 10.2.5, `picomatch` 2.3.2 → 4.0.4). |
| [x]  |   2 | [`package.json`](package.json), [`pnpm-lock.yaml`](pnpm-lock.yaml)      | **Cleared Trivy HIGH `CVE-2026-46625` (js-cookie).** Bumped Clerk packages (`@clerk/backend`, `@clerk/nextjs`, `@clerk/types`, `@clerk/testing`) so `@clerk/shared` resolves to the patched `3.47.6`, which carries `js-cookie@3.0.7` (fix for the per-instance prototype-hijack / cookie-attribute-injection advisory). No override needed; `pnpm dedupe` collapsed the duplicate so only the patched version remains. |
| [ ]  |   3 | [`bin/build-app`](bin/build-app)                                       | Not changed. It still calls `./node_modules/.bin/next build` until the IaC/Jenkins deploy script switches fully to pnpm.                                         |
| [x]  |   4 | [`.github/workflows/claude.yml`](.github/workflows/claude.yml)         | Updated stale npm wording in the Claude workflow comments and examples.                                                                                          |
| [x]  |   5 | [`.github/dependabot.yml`](.github/dependabot.yml)                     | Reworded the dependency-update note for pnpm, while keeping the correct `npm` Dependabot ecosystem name.                                                         |
| [x]  |   6 | [`THIRD_PARTY_LICENSES.md`](THIRD_PARTY_LICENSES.md)                   | Changed the license inventory command from `npx` to native `pnpm licenses`.                                                                                                 |
| [x]  |   7 | [`.trivyignore.yaml`](.trivyignore.yaml)                               | Changed the license audit comment from `npx` to native `pnpm licenses`.                                                                                                     |
| [x]  |   8 | [`.npmrc`](.npmrc), [`services/editor/.npmrc`](services/editor/.npmrc) | Added `package-lock=false` in both places. The editor service needs its own file because npm does not inherit the root `.npmrc` when run from `services/editor`. |
| [x]  |   9 | [`.gitignore`](.gitignore)                                             | Ignored `package-lock.json`, `services/editor/package-lock.json`, and `yarn.lock`.                                                                               |
| [x]  |  10 | [`.github/workflows/checks.yml`](.github/workflows/checks.yml)         | Added `pnpm audit signatures` after the frozen pnpm install. This was removed by the `pnpm` branch.                                                              |
| [x]  |  11 | [`bin/bootstrap-from-app-template`](bin/bootstrap-from-app-template)   | Stopped deleting package-lock ignore rules. The script now preserves/adds them.                                                                                  |
| [x]  |  12 | [`THIRD_PARTY_LICENSES.md`](THIRD_PARTY_LICENSES.md)                   | Rephrased "npm packages" as "packages from the npm registry."                                                                                                   |

---

## Validation

| Check                       | Result                                                                                                                  |
| --------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
| Overrides now applied       | `pnpm-lock.yaml` gained an `overrides:` block; `minimatch` → 10.2.5 and `picomatch` → 4.0.4 are resolved (previously 3.1.5 / 2.3.2). |
| `js-cookie` resolution      | Lockfile resolves `js-cookie@3.0.7` only; the vulnerable `3.0.5` is gone (CVE-2026-46625).                              |
| `pnpm install` (full)       | Clean; lockfile and `node_modules` consistent.                                                                          |
| `pnpm audit signatures`     | Passed.                                                                                                                 |
| Prettier on changed files   | Passed.                                                                                                                 |
| `pnpm run checks`           | Typecheck and `validate-actions` passed. Lint was blocked by unrelated untracked scratch files already in the worktree. |
| Unit tests                  | Not re-run here: the project's docker unit-test harness isn't pnpm-ready yet.                                            |

---

## Still To Do Later

| File                             | When to change it                                                                                                                                        |
| -------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
| [`bin/build-app`](bin/build-app) | Once the IaC pnpm branch is merged and `ManagementAppCDStack` redeploys the Jenkins `deploy.mjs` bundle, switch this back to the normal pnpm build path. |


---

## Bump to pnpm 11.3 (`packageManager`)

Bumped `packageManager` from `pnpm@11.1.3` to `pnpm@11.3.0`.

**Reason 1**: There is no real reason to use an older pnpm version (there is no LTS version or something) and fixes won't be backported to older pnpm versions.

**Reason 2:** pnpm 11.1.3 baked a supply-chain verification pass into `pnpm install` (`Verifying lockfile against supply-chain policies (N entries)...`) that re-validates every lockfile entry against `minimumReleaseAge` / `trustPolicy` on every run, with no way to skip it. Setting `minimumReleaseAge: 0` only makes the check pass trivially; the scan still runs and logs. For a local Docker unit-test harness that is pure overhead (up to 30 seconds!), since the install is `--frozen-lockfile` and the lockfile is already trusted. pnpm 11.3 adds the `trustLockfile` setting, the first real off-switch: with it we can set `--config.trustLockfile=true` scoped to the unit-test container only, while the committed supply-chain policy in `pnpm-workspace.yaml` stays strict for real installs.


pnpm-workspace.yaml
```diff
@@ -8,3 +8,14 @@ allowBuilds:
     sharp: true
     unrs-resolver: true
 minimumReleaseAge: 2880
+overrides:
```
nathanstitt, line 11: Nice catch!

.github/workflows/claude.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 1
-          # SECURITY (OTTER-545): the claude-code-action runs `npm install`
+          # SECURITY (OTTER-545): the claude-code-action can run package-manager
```
nathanstitt: Small grammar nit on the rewritten sentence — `can run package-manager` reads strange to me, I'd say we can just use `pnpm` here since we're switching to it:

services/editor/.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
nathanstitt: Genuine question rather than a blocker: is there a real workflow that runs `npm` from inside `services/editor/`? The directory is a workspace package, so the normal install flow is `pnpm install` at root, which already honors the root `.npmrc`. If nobody runs npm in here, this file is dead weight that also faintly signals "this subdir is npm-managed," which could confuse a reader. If we do want belt-and-suspenders against an accidental `cd services/editor && npm install`, a one-line comment in the file would make the intent obvious.

package.json
```diff
@@ -58,7 +56,9 @@
         "@aws-sdk/s3-presigned-post": "^3.893.0",
         "@aws-sdk/s3-request-presigner": "^3.893.0",
         "@casl/ability": "^6.8.0",
-        "@clerk/nextjs": "^6.39.0",
+        "@clerk/backend": "^2.33.4",
```
nathanstitt, line 59: Heads-up: `services/editor/package.json` still pins `@clerk/backend: ^2.33.2` while we're bumping the root to `^2.33.4`. Caret ranges mean pnpm will pick the same latest for both today, but the lower floor in the editor service can drift in surprising ways later (e.g. if a 2.33.x is yanked or if we someday tighten ranges). Probably worth a matching bump in the same PR so the two stay in lockstep, since both ultimately verify Clerk tokens against the same backend.

nathanstitt (approved):
thanks Marvin

package.json
therealmarv: oh, good catch. I will sync the version there.

services/editor/.npmrc
therealmarv: probably right to remove this file (it's not super important). As you said it's more like a belt-and-suspenders approach

PR #744, opened 2026-05-27
OTTER-574: Enhance dashboards with last updated date and reviewer tracking

stetradis:
## Summary

- Replace the "Submitted On" column on all org dashboards with a "Last updated" column, computed from the most recent of: submission, approval, rejection, job status change, proposal-comment decision/resubmission note, or review-comment decision. Drafts fall back to `created_at`. Dashboard ordering now uses this same value.
- Make the reviewer-side "Submitted By" column consistent across scopes by always showing the researcher's name (`createdBy`), rather than the submitting lab name when scoped to org.
- Make `study.reviewerId` track the *latest* reviewer to act on a study. It's now updated on proposal approve/reject, needs-clarification, and on job-file approve/reject — not just first approval. The reviewer-org "Reviewed By" column now shows that reviewer's name (falling back to `-`) instead of the viewer's org slug.


src/server/actions/study.actions.ts
```diff
@@ -44,6 +44,31 @@ import { Action, z } from './action'
 // on any dashboard, the studyId is effectively undiscoverable. Stale editor tabs / direct URL bookmarks remain
 // a known gap.
 function fetchStudyQuery(db: DBExecutor) {
+    // Drafts have no status-change rows, so fall back to created_at
+    const lastUpdatedAtExpr = sql<Date>`COALESCE(GREATEST(
```
nathanstitt: This expression worries me on both performance and maintenance grounds.

**Performance**: It contains three correlated subqueries against `job_status_change`, `study_proposal_comment`, and `study_review_comment`, and we then use the same expression in both the SELECT clause (line 103) AND the ORDER BY (line 139). Postgres often won't memoize across those positions, so worst case we run all six subqueries per study row per page load — and the dashboards are unpaginated as far as I can tell.

**Maintenance**: We now have two parallel definitions of "what's a review event" — this expression and the backfill SQL in the new migration — and they already differ slightly (the migration doesn't look at `submitted_at`/`approved_at`/`rejected_at` or RESUBMISSION-NOTE rows).

We just added `latest_reviewer_id` as a maintained column on `study` exactly because computing it per-query was awkward. The same logic argues for a sibling `last_updated_at timestamptz` column on `study`, bumped at each of the same write sites that already touch `latest_reviewer_id` (plus the resubmission write path). The dashboard query then collapses to `SELECT study.last_updated_at`, ordering is free, and there's a single source of truth.

This can be a follow-up if we want to ship the UI now, but I'd flag it before that subquery hardens into something everyone is scared to touch.

src/database/migrations/1779300000000_add_latest_reviewer_id_to_study.ts
```diff
@@ -0,0 +1,40 @@
...
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await db.schema
+        .alterTable('study')
+        .addColumn('latest_reviewer_id', 'uuid', (col) => col.references('user.id'))
+        .execute()
+
+    // Backfill from the most recent proposal / code / results review event per study.
+    await sql`
+        UPDATE study
+        SET latest_reviewer_id = sub.user_id
```
nathanstitt: Subtle correctness gap: studies that were reviewed before this migration ran will only get `latest_reviewer_id` backfilled if they have a row in `study_proposal_comment` (with a decision), `study_review_comment` (entry_type DECISION), or `job_status_change` (FILES-APPROVED/REJECTED). But the codebase has had `study.reviewer_id` for a long time as the "current reviewer" field, and there are likely older approved/rejected studies where `reviewer_id` is set but none of those three event sources exist (e.g. early approvals that pre-date the proposal-comment table).

After the migration those rows will render `-` in the new Reviewed By column, even though they clearly have a reviewer. Worth either:
- adding a final fallback before the event union: `COALESCE(latest_reviewer_id, reviewer_id)` semantics, e.g. `UPDATE study SET latest_reviewer_id = reviewer_id WHERE latest_reviewer_id IS NULL AND reviewer_id IS NOT NULL` as a second step, or
- doing `SELECT reviewer_id, study.created_at` as a fourth branch of the UNION ALL so it gets ordered alongside the events.

Option 1 is probably the cleanest since `reviewer_id` is by definition the current reviewer at migration time.

src/server/actions/study-job.actions.ts
```diff
@@ -41,6 +41,12 @@ export const approveStudyJobFilesAction = new Action('approveStudyJobFilesAction
             })
             .executeTakeFirstOrThrow()
 
+        await db
```
nathanstitt, line 44: Now that this handler does two related writes (the existing `insertInto('jobStatusChange')` plus this new study UPDATE for `latestReviewerId`), the action probably wants `new Action('approveStudyJobFilesAction', { performsMutations: true })` at its declaration so the two writes commit atomically. Without it the `db` passed to the handler is a non-transactional connection, and a crash/disconnect between the two `await db.…execute()` calls would leave the row in a state where the job is FILES-APPROVED but `latestReviewerId` is still NULL — which would then render `-` on the dashboard for a study that was actually just approved.

Same consideration applies to `rejectStudyJobFilesAction` below.

src/database/migrations/1779300000000_add_latest_reviewer_id_to_study.ts
```diff
@@ -0,0 +1,40 @@
...
+        FROM (
+            SELECT DISTINCT ON (s.id) s.id AS study_id, review_events.user_id
+            FROM study s
+            INNER JOIN LATERAL (
+                SELECT author_id AS user_id, created_at
+                FROM study_proposal_comment
+                WHERE study_id = s.id AND decision IS NOT NULL
```
nathanstitt: Small nit: `WHERE study_id = s.id AND decision IS NOT NULL` will also match a `RESUBMISSION-NOTE` row if someone ever sets a decision on one (the table doesn't have a CHECK constraint forbidding it). By convention only `REVIEWER-FEEDBACK` rows carry a decision today, but `WHERE study_id = s.id AND entry_type = 'REVIEWER-FEEDBACK' AND decision IS NOT NULL` makes the intent explicit and survives a hypothetical schema drift.

src/server/actions/study.actions.test.ts
```diff
@@ -516,6 +516,85 @@ describe('Study Actions', () => {
             })
         })
     })
+
+    describe('fetchStudiesForOrgAction lastUpdatedAt', () => {
```
nathanstitt: Really like that this test suite covers both the DRAFT-fallback case AND the full clarification-then-resubmit timeline ❤️

nathanstitt (approved):
looks good, up to you if we wanna add the last_updated_at/last_reviewed_at column

src/database/migrations/1779300000000_add_latest_reviewer_id_to_study.ts
stetradis: @nathanstitt I think you may have reviewed the previous version of this PR that still had the migration file 😬, good news is I simplified this and don't need it at all, I just expanded the areas that update `reviewerId` instead. I'll confirm that all cases are covered

src/server/actions/study.actions.ts
stetradis: Thanks for the review @nathanstitt 

I agree with the direction here, so I implemented `last_updated_at` as a maintained column on study rather than keeping the subquery in the dashboard query. It's bumped at the same write sites as `reviewerId`, backfilled in the migration, and the query now just selects/orders by the column.

PR #745, opened 2026-05-27
OTTER-566: render DO code review page resubmission variant

Rod0352:
Summary
When a DO opens the code review page for a study whose code has been submitted more than once, swap in the resubmission variant: timestamp reads "Resubmitted on {date}", heading becomes "Review study code v{N}.0", banner copy switches "submitted" to "resubmitted", the submitted code viewer starts collapsed, and a Feedback and notes section showing the prior round's reviewer decision and the current round's resubmission note renders above the code evaluation form.

Resubmission is detected by passing getCodeReviewFeedbackAction entries into CodeReviewRedesignView — on the round currently under review the action only returns entries when at least one prior round exists. The displayed version is the max non-null entry.version. The page-level fetch falls back to [] on error so a feedback fetch failure degrades to the first-submission view. Code evaluation, code review feedback, and decision selector already start empty / unselected, and the page already sits behind CodeReviewFeatureFlag, so no new flag — same approach OTTER-555 took on the proposal side.
Adds an optional codeInitiallyExpanded prop to SubmittedCodeSection. Reworks code-review-redesign-view.test.tsx into first-submission and resubmission describe blocks covering the timestamp swap, banner copy, versioned heading, Feedback and notes contents, DOM ordering, and toggle collapsed-by-default.


src/app/[orgSlug]/study/[studyId]/review/code-review-redesign-view.tsx
```diff
@@ -62,6 +105,9 @@ export async function CodeReviewRedesignView({ orgSlug, study }: CodeReviewRedes
     const proposalHref = `${Routes.studyReview({ orgSlug, studyId: study.id })}?from=code-review`
     const latestJobStatus = job.statusChanges.at(0)?.status ?? null
 
+    const isResubmission = entries.length > 0
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/code-review-redesign-view.tsx]: `isResubmission` and `version` are derived from two independent signals here: presence of any entry vs. `max(entry.version)`. They can in principle disagree — e.g. if `entries` held only a v1 reviewer DECISION with no v2 note, we'd render the "resubmitted" banner + "Resubmitted on" while the heading stayed "Review study code" (no `v2.0`).

In practice today's invariants seem to keep them in lockstep (a code resubmission *requires* a note via `resubmissionNoteParam`, so a real v2 always yields a `version: 2` entry; and `claimInitialCodeReviewJob`'s PENDING-REVIEW guard blocks re-review of v1). But that's a non-obvious cross-file invariant to lean on. Worth considering a single source of truth — `proposal-section.tsx` does `const isResubmission = reviewVersion > 1`, so `const isResubmission = version > 1` here would keep the two in sync by construction and mirror the proposal side.

nathanstitt (approved):
lgtm, thank you!

PR #746, opened 2026-05-27
OTTER-572: resume drafts on Step 2 when Step 2 fields exist

Rod0352:
Summary

Today the "Edit" link on a draft always re-routes the researcher to Step 1 (the data-org + language picker) regardless of how far they had gotten. This PR redirects the researcher back to whichever step they last edited — Step 2 if any of the Step-2-only columns have been written, Step 1 otherwise.
Detection is a pure helper, draftHasStep2Progress(study) in src/lib/studies.ts. It returns true when piUserId is set, when datasets is a non-empty array, or when any of researchQuestions, projectSummary, impact, or additionalNotes is non-null. Step 1 only writes orgSlug / language / title / piName / document paths, so a write to any of those columns is a strong signal that the researcher reached Step 2.

src/app/[orgSlug]/study/[studyId]/edit/page.tsx widens its SELECT to include the Step-2 columns and calls redirect(Routes.studyProposal({ orgSlug, studyId })) when draftHasStep2Progress(study) is true; fresh drafts still render the existing Step 1 <StudyProposal> form. The DRAFT-status guard and not-found alert are unchanged.

Adds unit tests for the helper (each Step-2 field individually triggers true; empty array does not) and for the route (fresh draft → Step 1, non-DRAFT → not-found alert, each Step-2 field individually → redirect to /[orgSlug]/study/[studyId]/proposal).

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/edit/page.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/edit/page.tsx
@@ -5,8 +5,12 @@ import { Routes } from '@/lib/routes'
 import { draftHasStep2Progress } from '@/lib/studies'
 import { StudyProposal } from '../../request/proposal'
 
-export default async function StudyEditPage(props: { params: Promise<{ studyId: string; orgSlug: string }> }) {
+export default async function StudyEditPage(props: {
+    params: Promise<{ studyId: string; orgSlug: string }>
+    searchParams: Promise<{ from?: string }>
+}) {
     const params = await props.params
+    const searchParams = await props.searchParams
     const { studyId, orgSlug } = params
 
     // TODO: validate that member from clerk session matches memberId from url
@@ -43,8 +47,10 @@ export default async function StudyEditPage(props: { params: Promise<{ studyId: 
     }
 
     // OTTER-572: drafts that already reached Step 2 reopen on Step 2 instead of
-    // always sending the researcher back to the Step 1 data-org picker.
-    if (draftHasStep2Progress(study)) {
+    // always sending the researcher back to the Step 1 data-org picker. Step 2's
+    // "Previous" button must remain a working escape hatch, so it navigates here
+    // with ?from=step2 to explicitly request the Step 1 form.
+    if (searchParams.from !== 'step2' && draftHasStep2Progress(study)) {
         redirect(Routes.studyProposal({ orgSlug, studyId }))
     }
 
--- a/src/app/[orgSlug]/study/[studyId]/edit/page.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/edit/page.tsx
@@ -1,10 +1,13 @@
+import { redirect } from 'next/navigation'
 import { db } from '@/database'
 import { AlertNotFound } from '@/components/errors'
+import { Routes } from '@/lib/routes'
+import { draftHasStep2Progress } from '@/lib/studies'
 import { StudyProposal } from '../../request/proposal'
 
-export default async function StudyEditPage(props: { params: Promise<{ studyId: string }> }) {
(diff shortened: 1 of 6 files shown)


src/app/[orgSlug]/study/[studyId]/edit/page.tsx
```diff
@@ -33,6 +42,12 @@ export default async function StudyEditPage(props: { params: Promise<{ studyId:
         )
     }
 
+    // OTTER-572: drafts that already reached Step 2 reopen on Step 2 instead of
+    // always sending the researcher back to the Step 1 data-org picker.
+    if (draftHasStep2Progress(study)) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit/page.tsx]: I think this introduces an infinite-loop / dead-button situation with Step 2's "Previous" button. In `src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx`:

```ts
const handlePrevious = async () => {
    const saved = await saveDraft()
    if (saved) router.push(Routes.studyEdit({ orgSlug, studyId }))
}
```

So when a researcher is on Step 2 and clicks "Previous" to go back to Step 1, the Edit page will now see Step-2 progress (which `saveDraft` just wrote, or which was already there from a prior session) and immediately redirect them back to `/proposal` — the same page they were trying to leave. The URL bar flickers `/edit` → `/proposal` and the user is stuck on Step 2 with no way to revisit the Step 1 fields they wanted to fix.

A couple of escape hatches worth considering:

- Have Step 2's "Previous" pass an explicit override like `?step=1` (or `?from=step2`) and skip this redirect when the param is present. Survives hard refresh, makes the intent visible in the URL, and the Edit page just needs `if (draftHasStep2Progress(study) && !params.step === '1') redirect(...)`.
- Or route "Previous" to a separate `/edit/step-1` URL that's exempt from the gate.

Worth covering with a test once the fix lands — `it('renders Step 1 when arriving from Step 2 Previous')` would lock down the back-navigation contract.

nathanstitt (commented):
looks good but probalby need to fix the button looping issue if it's real, but seems like it'd be an issue to me

src/app/[orgSlug]/study/[studyId]/edit/page.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/edit/page.tsx]: youre right, thanks :) i'll adjust that

src/lib/studies.ts
```diff
@@ -14,6 +14,30 @@ export function deriveStudyVersion(entries: { version: number }[]): number {
...
+    impact: Json | null
+    additionalNotes: Json | null
+}
+
+// Step 1 (data org + language + docs) saves `orgSlug`, `language`, `title`,
+// `piName`, and document paths. Step 2 is the first time any of the columns
+// below are written, so any one being non-empty means the researcher has
+// reached Step 2. Used to route a "resume draft" entry to the step where
+// they last left off instead of always landing on Step 1.
+export function draftHasStep2Progress(study: DraftStep2Fields): boolean {
```
nathanstitt, line 31: Nice that this is a pure, well-tested helper. One thing worth a sentence in the doc comment: the "Step 1 never writes these columns" invariant is enforced by the *form* schemas (`studyProposalFormSchema` has no `piUserId`/`datasets`/lexical fields), not by the DB or the action layer — `onSaveDraftStudyAction` actually accepts and persists `piUserId` via `draftStudyApiSchema`. So today the assumption holds because the Step 1 form never supplies those, but a future caller that passes `piUserId` into the draft-create action on a Step-1 draft would make this return a false positive and bounce a fresh draft to Step 2. Calling out that the guarantee comes from the form, not the column, would help the next reader.

nathanstitt (approved):
looks good, ty!

PR #747, opened 2026-05-27
upgrade to vitest 4.1.x as minimum

therealmarv:
## What's this about

If you've looked at the unit test job in CI lately, you've probably noticed it got really noisy. The [`test` job on the Checks workflow](https://github.com/safeinsights/management-app/actions/runs/26520079462/job/78112042756) spews a wall of warnings before the tests even start running:

```
Importing from "vitest/coverage" is deprecated since Vitest 4.1. Please use "vitest/node" instead.
Warning: A vi.mock("next/router") call in tests/vitest.setup.ts is not at the top level of the module. Although it appears nested, it will be hoisted and executed before any tests run. Move it to the top level to reflect its actual execution order. This will become an error in a future version.
Warning: A vi.mock("next/server") call ... (and one for every other vi.mock, ~12 in total)
The plugin "vite-tsconfig-paths" is detected. Vite now supports tsconfig paths resolution natively via the resolve.tsconfigPaths option. You can remove the plugin and set resolve.tsconfigPaths: true in your Vite config instead.
```

The tests themselves still passed, but the output was hard to read and one of those warnings literally says *"this will become an error in a future version"*, so it's worth dealing with now rather than waiting for it to break.

## Why it started happening

The root cause was in `package.json`:

```json
"vitest": "*"
```

A wildcard version. Whenever the lockfile gets regenerated (which the pnpm migration did), it floats to whatever the latest Vitest is. That's how we silently jumped onto **Vitest 4.1** (and **Vite 8** alongside it, since Vitest pulls Vite in), which shipped in May 2026 ("Test Tags, Native Node.js Execution and AI Agent Reporter") and surfaced three things in our setup:

1. **The `vi.mock` hoisting warning.** Vitest transforms any file containing `vi.mock` so it can hoist those calls to the very top and rewrite static imports into dynamic ones, guaranteeing mocks register before the modules they replace are imported. Our `tests/vitest.setup.ts` had all of its `vi.mock` calls nested inside a `beforeAll(...)`. Vitest was hoisting them out anyway, so the placement was misleading, and 4.1 now warns about it (and plans to make it a hard error in a later major).

2. **The `vitest/coverage` deprecation.** In 4.1 the coverage base class moved to `vitest/node` and `vitest/coverage` became a deprecated shim. This one isn't ours: it comes from the `vitest-monocart-coverage` package (latest 4.0.2 still imports the old path), and it only appears when coverage is enabled, which is why it shows up in CI but not in a plain local run.

3. **The `vite-tsconfig-paths` plugin warning.** Vite 8 added native tsconfig `paths` resolution via `resolve.tsconfigPaths`, so the `vite-tsconfig-paths` plugin we used in `vitest.config.ts` is now redundant and Vite nudges you to drop it.

This is also the answer to "why don't I see this locally?" The wildcard meant local dev images could be on an older cached Vitest, and the coverage warning is coverage-only.

## What this PR does

- **Pins Vitest to `>=4.1.6`** instead of `*`. The wildcard had no floor at all, so the lockfile could resolve to anything (including versions older than the top-level `vi.mock` requirement). `>=4.1.6` guarantees we never drop below 4.1.6 while still accepting future minor and major releases. The resolved version is unchanged (4.1.6), so this is really just the lockfile spec catching up to reality. Note: because this also accepts new majors, a future lockfile regen could move us onto Vitest 5+; CI installing with a frozen lockfile keeps us on 4.1.6 until that bump is done deliberately.
- **Moves all `vi.mock` calls in `tests/vitest.setup.ts` to the module top level**, which is what Vitest wanted. Because the mock factories need values that aren't safe to touch at hoist time (the deferred-callback tracking and the fake `headers` map, which depend on a real import), those are now built with `vi.hoisted(...)` and the import-dependent piece is wired in via a normal top-level call afterwards. Behavior is identical to before, just expressed in the way 4.1's hoisting model expects.
- **Replaces the `vite-tsconfig-paths` plugin with Vite 8's native `resolve.tsconfigPaths: true`** and removes the plugin dependency. Our `tsconfig.json` only uses plain `compilerOptions.paths` (`@/*`, `@/tests/*`), which the native boolean option covers fully (no `projects`/`root`/`loose`/`baseUrl`-only features that the native option lacks). To make sure we can actually rely on this, I also added a direct **`"vite": ">=8.0.0"`** dependency: Vitest's own peer range is `^6 || ^7 || ^8`, so without an explicit floor the lockfile could legally resolve Vite back down to 6 or 7, where `resolve.tsconfigPaths` doesn't exist.

## What this PR deliberately does NOT do

The `vitest/coverage` deprecation is left as-is. It's entirely upstream in `vitest-monocart-coverage`, it's harmless (the shim still works and Vitest documents no removal version), and the only ways to silence it from our side are a fragile pnpm patch or a fork. I've reported it upstream instead: [cenfun/vitest-monocart-coverage#17](https://github.com/cenfun/vitest-monocart-coverage/issues/17). We'll pick up the fix for free once they publish a release that imports from `vitest/node`.

## Verification

Ran the full unit suite in Docker against Vitest 4.1.6 / Vite 8.0.13 (same versions CI uses):

- **151 test files / 1368 tests pass**, exit 0, same counts as before, none skipped
- the `vi.mock` "not at the top level" warnings are gone (12 → 0)
- the `vite-tsconfig-paths` plugin warning is gone, with **zero `@/` import-resolution errors** (the `@/*` and `@/tests/*` aliases are imported across ~all test files, so a path-resolution regression couldn't pass silently)
- `tsc --noEmit`, ESLint, and Prettier are clean on the changed files

## Files changed

- `package.json` — pin `vitest` to `>=4.1.6`, drop `vite-tsconfig-paths`, add `vite >=8.0.0`
- `pnpm-lock.yaml` — spec updates + plugin removal (resolved versions unchanged)
- `tests/vitest.setup.ts` — hoist `vi.mock` calls to top level
- `vitest.config.ts` — use native `resolve.tsconfigPaths`, drop the plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,7 +156,7 @@
         "typescript": "^5.9.3",
         "typescript-eslint": "^8.56.1",
         "vite-tsconfig-paths": "^6.1.1",
-        "vitest": "^4.1.6",
+        "vitest": ">=4.1.6",
         "vitest-monocart-coverage": "^4.0.1"
     }
 }
--- a/package.json
+++ b/package.json
@@ -156,7 +156,7 @@
         "typescript": "^5.9.3",
         "typescript-eslint": "^8.56.1",
         "vite-tsconfig-paths": "^6.1.1",
-        "vitest": "*",
+        "vitest": "^4.1.6",
         "vitest-monocart-coverage": "^4.0.1"
     }
 }
--- a/package.json
+++ b/package.json
@@ -155,7 +155,7 @@
         "tsx": "4.21.0",
         "typescript": "^5.9.3",
         "typescript-eslint": "^8.56.1",
-        "vite-tsconfig-paths": "^6.1.1",
+        "vite": ">=8.0.0",
         "vitest": ">=4.1.6",
         "vitest-monocart-coverage": "^4.0.1"
     }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -351,9 +351,9 @@ importers:
       typescript-eslint:
         specifier: ^8.56.1
         version: 8.59.4(eslint@9.39.4(jiti@2.7.0))(typescript@5.9.3)
-      vite-tsconfig-paths:
-        specifier: ^6.1.1
-        version: 6.1.1(typescript@5.9.3)(vite@8.0.13(@types/node@25.9.1)(esbuild@0.27.7)(jiti@2.7.0)(terser@5.48.0)(tsx@4.21.0)(yaml@2.9.0))
+      vite:
+        specifier: '>=8.0.0'
+        version: 8.0.13(@types/node@25.9.1)(esbuild@0.27.7)(jiti@2.7.0)(terser@5.48.0)(tsx@4.21.0)(yaml@2.9.0)
       vitest:
         specifier: '>=4.1.6'
         version: 4.1.6(@opentelemetry/api@1.9.1)(@types/node@25.9.1)(@vitest/coverage-istanbul@4.1.6)(@vitest/coverage-v8@4.1.6)(happy-dom@20.9.0)(vite@8.0.13(@types/node@25.9.1)(esbuild@0.27.7)(jiti@2.7.0)(terser@5.48.0)(tsx@4.21.0)(yaml@2.9.0))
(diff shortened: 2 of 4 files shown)


nathanstitt (approved):
lgtm, thanks Marvin

PR #748, opened 2026-05-27
Remove OpenStax feature flags; ship opt-in content

nathanstitt:
All OpenStax feature flags except the main flag have been removed. The opt-in
content each one gated now ships unconditionally, and the legacy code paths
are deleted.

## Flags removed

| Flag | Effect |
|---|---|
| `usePostCodeSubmissionFeatureFlag` | `code-upload` always redirects to `/view` after submit |
| `ProposalReviewFeatureFlag` | edit-and-resubmit page no longer gated; `feature-flag-gate.tsx` deleted |
| `StudyDetailsRedesignFeatureFlag` | `StudyDetailsRedesignView` rendered directly on view/review/resubmit pages |
| `useCodeReviewCollaborationFeatureFlag` | code-review collaboration always enabled |
| `CodeReviewFeatureFlag` | `CodeReviewRedesignView` rendered directly |
| `PostSubmissionFeatureFlag` | `PostFeedbackView` rendered directly; `SubmittedView` simplified to a pass-through (now inlined into `submitted/page.tsx`) |
| `PostCodeSubmissionFeatureFlag` | `CodePostSubmissionView` / `CodePostDecisionView` rendered directly |
| `usePostSubmissionFeatureFlag` | `useStudyHref` always routes post-submission statuses to `/submitted` (see behavior change below) |
| `useProposalCollaborationFeatureFlag` | proposal collaboration always enabled |

## ⚠️ User-visible behavior change for non-openstax orgs: `useStudyHref`

`useStudyHref` previously gated its `POST_SUBMISSION_STATUSES` → `/submitted`
branch behind `isPostSubmissionFlow` (openstax-only). With the flag removed,
that branch now applies to **every org**: APPROVED / REJECTED /
CHANGE-REQUESTED studies without job activity are routed to `/submitted`
across all orgs.

This is the intended effect of dropping the flag, and `ProposalSubmitted`
still surfaces the "Proceed to step 3" CTA back to `/agreements` so APPROVED
researchers aren't stranded. Calling it out explicitly because it changes a
visible navigation path for orgs that never opted into the flag.

## Legacy components deleted

- `app/.../review/code-review-view.tsx` (+ test)
- `app/.../resubmit/old-resubmit-code-view.tsx`
- `app/.../resubmit/form.tsx`
- `app/.../edit-and-resubmit/feature-flag-gate.tsx`
- `app/.../submitted/submitted-proposal-preview.tsx`
- `app/.../submitted/submitted-view.tsx` (was a pass-through; `submitted/page.tsx` now renders `ProposalSubmitted` directly)
- `app/.../review/proposal-review-buttons.tsx` (+ test) — its approve/reject branch was unreachable after the flag removal; the only call site (`ProposalReviewFromAgreementsView`) now embeds the single "Proceed to Step 2" button inline
- `app/.../review/study-results-with-review.tsx` — orphaned; no importers remained after the redesign rollup
- `hooks/use-legacy-proposal-review-mutation.ts` — no remaining caller after `ProposalReviewButtons` was deleted
- `contexts/resubmit-code/` (whole directory)

## Renames

- `LegacyProposalReviewView` → `ProposalReviewFromAgreementsView` — the "Legacy" prefix was misleading now that this is the canonical view for the `?from=agreements && codeSubmitted` branch (not a flag-fallback)

## Preserved

`src/components/openstax-feature-flag.tsx` is kept with renamed `Unused*`
exports so the next feature flag has a starting scaffold.

## Tests

- All page-test assertions that read `page?.type` against a flag wrapper were rewritten to assert the now-direct component (`CodeReviewRedesignView`, `StudyDetailsRedesignView`, `PostFeedbackView`, etc.).
- `useStudyHref` test updated — APPROVED / CHANGE-REQUESTED without job activity now route to `/submitted` (previously `/agreements` / `/view`).
- E2E spec comments referencing removed flags cleaned up; `enableSpyMode` dropped from both collab specs (with collaboration unconditional, spy mode no longer reflects the shipped flow).

## Verification

`pnpm typecheck` ✓ &nbsp; `pnpm lint` ✓ &nbsp; **1351/1351 unit tests pass**


src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.tsx
```diff
@@ -75,15 +73,4 @@ export function LegacyProposalReviewView({ orgSlug, study, agreementsHref }: Leg
             <ProposalReviewButtons study={study} orgSlug={orgSlug} agreementsHref={agreementsHref} />
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.tsx]: **High** ⚠️ Now that the feature-flag fallback is removed, `LegacyProposalReviewView` is only reached from `review/page.tsx`, and that call site always passes `agreementsHref`. So `ProposalReviewButtons` always takes the "Proceed to Step 2" branch in production, and its approve/reject branch plus `useLegacyProposalReviewMutation` look unreachable now.

Since the whole point of this PR is deleting legacy paths, I'd finish the job: drop the dead approve/reject branch in [`proposal-review-buttons.tsx`](https://github.com/safeinsights/management-app/blob/0fa4e751cd2fb1e87d95249c0f0bfa413c6d410e/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/review/proposal-review-buttons.tsx) and the now-unused [`use-legacy-proposal-review-mutation.ts`](https://github.com/safeinsights/management-app/blob/0fa4e751cd2fb1e87d95249c0f0bfa413c6d410e/src/hooks/use-legacy-proposal-review-mutation.ts) hook (neither is in this diff). This is the biggest leftover, and the main reason I'm requesting changes.

src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.tsx
```diff
@@ -22,7 +20,7 @@ type LegacyProposalReviewViewProps = {
 export function LegacyProposalReviewView({ orgSlug, study, agreementsHref }: LegacyProposalReviewViewProps) {
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.tsx]: **Medium** (naming) This is now the canonical view for the `?from=agreements && codeSubmitted` branch, not a "legacy" one. The `Legacy` prefix is misleading and invites a future "delete the legacy view" cleanup that would actually break the agreements navigation flow. Consider renaming to something like `ProposalReviewFromAgreementsView`.

src/hooks/use-study-href.ts
```diff
@@ -10,17 +10,14 @@ export function useStudyHref(
...
-    isPostSubmissionFlow = false,
 ) {
     // APPROVED with only a baseline job (no code submitted yet) → continue uploading
     const hasCodeSubmitted = jobStatuses?.some((s) => s === 'CODE-SUBMITTED')
     if (status === 'APPROVED' && hasJobActivity && !hasCodeSubmitted) return Routes.studyCode(studyParams)
 
     if (hasJobActivity) return Routes.studyView(studyParams)
 
-    if (isPostSubmissionFlow && POST_SUBMISSION_STATUSES.includes(status)) return Routes.studySubmitted(studyParams)
+    if (POST_SUBMISSION_STATUSES.includes(status)) return Routes.studySubmitted(studyParams)
```
therealmarv, line 20: **High** (heads-up, not a hard blocker) This used to be gated behind `isPostSubmissionFlow` (OpenStax only). Making it unconditional means APPROVED / REJECTED / CHANGE-REQUESTED studies without job activity now route to `/submitted` for *every* org. That's the intended effect of removing the flag, and `ProposalSubmitted` still offers the "Proceed to step 3" CTA back to `/agreements` so nobody is stranded. But it's a user-visible change for all orgs that the PR description currently tucks into a test-update bullet. Worth calling out explicitly in the description.

src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.test.tsx
```diff
@@ -152,36 +134,13 @@ describe('LegacyProposalReviewView', () => {
         expect(screen.queryByRole('button', { name: 'Approve request' })).not.toBeInTheDocument()
     })
 
-    describe('agreementsHref bypass', () => {
+    it('renders "Proceed to Step 2" when agreementsHref is provided', () => {
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.test.tsx]: **Medium** Tied to the dead-path note on the view itself: the no-`agreementsHref` path is no longer reachable in production, yet the suite still renders the view directly. The retained tests read as coverage on effectively dead code rather than on the shipped agreements flow. Worth pruning/retargeting alongside the cleanup. Same applies to `proposal-review-buttons.test.tsx` (not in this diff).

src/app/[orgSlug]/study/[studyId]/submitted/submitted-view.tsx
```diff
@@ -20,43 +12,14 @@ interface SubmittedViewProps {
...
-            </Paper>
-            <Stack gap="sm" align="flex-end">
-                <Button component={Link} href={Routes.orgDashboard({ orgSlug })} size="md">
-                    Go to dashboard
-                </Button>
-            </Stack>
-        </Stack>
-    )
-
     return (
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/submitted/submitted-view.tsx]: **Medium** `SubmittedView` is now a pure pass-through, forwarding identical props to `ProposalSubmitted` with no added behavior. Either inline `ProposalSubmitted` directly in `submitted/page.tsx` and drop this file, or keep it with a one-line comment explaining why the layer stays. As-is it's just indirection.

tests/proposal-review-collaboration.spec.ts
```diff
@@ -123,10 +123,6 @@ test('a reviewer in two tabs collaborates live on a proposal review; one tab sub
                 url: `/openstax/study/${studyId}/review`,
             })
             await waitForOpenstaxOrgInClerkMetadata(ctxA.page)
```
therealmarv, line 117: **Low** With the collaboration flag removed, the collab UI is always on, so enabling spy mode here no longer reflects how the feature ships. It just adds a step the real flow doesn't have. Dropping `enableSpyMode` from this spec would exercise the true no-spy-mode path. (Same in `code-review-collaboration.spec.ts`.)

tests/code-review-collaboration.spec.ts
```diff
@@ -113,26 +113,22 @@ async function uploadCodeAsResearcher(page: Page, studyTitle: string): Promise<s
...
-// at the bottom of every page (no localStorage), so each fresh browser
-// context must enable it once. The element has opacity:0 but
-// `pointer-events: auto`, so a forced click reliably fires the handler.
-// Spy mode adds `.spy-mode` to <body>; we assert that to confirm state
-// flipped before relying on it gating downstream rendering.
+// Toggle spy mode by clicking the `𝜋` symbol fixed at the bottom of every
+// page. The element has opacity:0 but `pointer-events: auto`, so a forced
+// click reliably fires the handler. Spy mode adds `.spy-mode` to <body>;
+// we assert that to confirm state flipped.
 async function enableSpyMode(page: Page): Promise<void> {
```
therealmarv: **Low** Same as the proposal collab spec: collaboration is unconditional now, so gating the assertions behind `enableSpyMode` exercises a path users won't hit. The comment was updated but the spy-mode dependency remains. Dropping it would prove the real flow works.

src/contexts/proposal/context.tsx
```diff
@@ -61,14 +57,13 @@ export function ProposalProvider({ children, studyId, draftData }: ProposalProvi
...
-    const sharedSocket = useYjsWebsocket()
-    const websocketProvider: HocuspocusProviderWebsocket | null = isCollaborationEnabled ? sharedSocket : null
+    const websocketProvider = useYjsWebsocket()
 
     const yjsForm = useYjsFormMap({
         studyId,
         form,
         websocketProvider,
-        enabled: isCollaborationEnabled,
+        enabled: true,
```
therealmarv: **Low** `enabled` is now hard-coded `true` at the only caller, so the parameter on `useYjsFormMap` is effectively dead. Either drop it and let the hook guard on `websocketProvider` alone, or leave it. Note the sibling `useCodeReviewEvaluationMap` *does* still vary `enabled` (`initiallyEditable` in `code-review-client.tsx`), so only this one is dead.

therealmarv (changes requested):
Reviewed the branch end to end. The flag removal itself is thorough: every removed flag, legacy component, and flag-related mock/conditional render is genuinely gone, and typecheck/lint/tests pass. What's left is cleanup the removal didn't quite finish, plus one behavior change worth documenting. I left inline notes on the relevant lines. A few findings touch files this PR didn't change, so I'm summarizing those here with links.

### Findings not on lines in this diff

**Medium · Coverage gap for CODE-APPROVED / CODE-REJECTED empty-entries fallback**
[`view/page.test.tsx`](https://github.com/safeinsights/management-app/blob/0fa4e751cd2fb1e87d95249c0f0bfa413c6d410e/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/view/page.test.tsx) dropped the `it.each(['CODE-APPROVED', 'CODE-REJECTED'])('renders CodeOnlyView directly…')` block. The production fallback at [`view/page.tsx#L50-L52`](https://github.com/safeinsights/management-app/blob/0fa4e751cd2fb1e87d95249c0f0bfa413c6d410e/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/view/page.tsx#L50-L52) (entries empty or error, falls back to `CodeOnlyView`) still exists but is no longer covered for those statuses. One thin test on the empty-entries path would restore the coverage.

**Medium · `StudyResultsWithReview` rendering no longer asserted**
The deleted `code-review-view.test.tsx` asserted "renders StudyResultsWithReview". The surviving [`code-review-redesign-view.test.tsx`](https://github.com/safeinsights/management-app/blob/0fa4e751cd2fb1e87d95249c0f0bfa413c6d410e/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/review/code-review-redesign-view.test.tsx) covers breadcrumbs, headings, banners, and criteria, but doesn't explicitly assert `StudyResultsWithReview` within the redesign view. If that assertion was meaningful, worth porting it forward.

**Low · Stale feature-flag comments in hooks**
- [`use-code-review-evaluation-map.ts#L39`](https://github.com/safeinsights/management-app/blob/0fa4e751cd2fb1e87d95249c0f0bfa413c6d410e/src/hooks/use-code-review-evaluation-map.ts#L39) still says it "gates the bridge behind feature-flag + editable status". Only the editable-status part remains.
- [`use-yjs-form-map.ts#L64`](https://github.com/safeinsights/management-app/blob/0fa4e751cd2fb1e87d95249c0f0bfa413c6d410e/src/hooks/use-yjs-form-map.ts#L64) references a "feature-flag fallback path" (the old non-collaborative `ProposalTextField`) that was deleted from `proposal/form.tsx`, so the comment is now incorrect.

### Recommendation

Requesting changes, softly. Nothing here is a correctness bug and the feature works as intended. The reason it isn't a plain "comment": for a PR whose entire purpose is deleting dead flag code, the leftover dead approve/reject branch plus the unused `useLegacyProposalReviewMutation` hook (inline #1), and the tests still covering that dead path (inline #3), cut against the PR's own goal. I'd like those cleaned up before merge. The `useStudyHref` behavior change is not a blocker, just please surface it in the PR description. Everything else (naming, the pass-through layer, spy-mode in the collab specs, stale comments) is fine as fast-follow.

src/app/[orgSlug]/study/[studyId]/review/page.test.tsx
```diff
@@ -66,7 +56,7 @@ describe('StudyReviewPage', () => {
         expect(mockRedirect).toHaveBeenCalledWith(expect.stringContaining('/agreements'))
     })
 
-    it('renders CodeReviewFeatureFlag for enclave with code submitted when coming from agreements', async () => {
+    it('renders CodeReviewRedesignView for enclave with code submitted when coming from agreements', async () => {
```
Comment (chrisbendel) [inline src/app/[orgSlug]/study/[studyId]/review/page.test.tsx]: do we want to still call these components "Redesign"?

src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.tsx]: Done in 3664a132 + 8c03da63. `LegacyProposalReviewView` is renamed to `ProposalReviewFromAgreementsView` and now embeds the single 'Proceed to Step 2' button inline; `proposal-review-buttons.tsx`, its test, and `use-legacy-proposal-review-mutation.ts` are deleted (the server actions they wrapped are still reachable from `study-review-buttons.tsx`).

src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.tsx]: Renamed to `ProposalReviewFromAgreementsView` in 3664a132 — agree the `Legacy` prefix was a footgun for the next person.

src/hooks/use-study-href.ts
nathanstitt: Good call — surfaced this explicitly in the PR description under a dedicated '⚠️ User-visible behavior change' section so it isn't buried in a test-update bullet.

src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.test.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/review/legacy-proposal-review-view.test.tsx]: Done in 3664a132. The renamed test file (`proposal-review-from-agreements-view.test.tsx`) only exercises the agreements flow now — the no-`agreementsHref` tests are gone, and `proposal-review-buttons.test.tsx` was deleted alongside the component in 8c03da63.

src/app/[orgSlug]/study/[studyId]/submitted/submitted-view.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/submitted/submitted-view.tsx]: Inlined in 6c69c625 — `submitted/page.tsx` renders `ProposalSubmitted` directly and `submitted-view.tsx` is gone.

tests/proposal-review-collaboration.spec.ts
nathanstitt: Dropped from both collab specs in 17e305f0. The `enableSpyMode` helper is gone and both contexts now go straight to the editor, exercising the shipped no-flag flow.

tests/code-review-collaboration.spec.ts
nathanstitt: Same change as the proposal spec — `enableSpyMode` removed in 17e305f0 so both contexts exercise the shipped no-flag path.

src/contexts/proposal/context.tsx
nathanstitt: Dropped the `enabled` parameter from `useYjsFormMap` entirely in f1631262. The hook now guards solely on `websocketProvider`, which is what the disabled-flag path was relying on anyway. Left `useCodeReviewEvaluationMap` alone since `initiallyEditable` still varies there.

src/app/[orgSlug]/study/[studyId]/review/page.test.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/review/page.test.tsx]: Fair point — with the non-redesigned variants gone there's nothing left to disambiguate against. I'd prefer to do that rename as a fast-follow rather than in this PR though: `CodeReviewRedesignView` and `StudyDetailsRedesignView` touch a lot of test imports and call sites, and folding it in here would muddy the diff's '`remove dead flag code`' purpose. Happy to open a follow-up ticket if you want — let me know if you'd rather block on it.

nathanstitt:
Thanks for the thorough review @therealmarv — pushed cleanup in 3664a132..b8394210 covering all eight inline notes plus the orphaned `study-results-with-review.tsx` (more on that below). Two of the non-inline summary findings worth a quick reply:

**Coverage gap for CODE-APPROVED / CODE-REJECTED empty-entries fallback** — `view/page.test.tsx` already covers this path at the `'falls back to CodeOnlyView when latest status is a decision but no review feedback exists'` test (line 334 on the commit you reviewed) using `CODE-APPROVED`. `CODE-REJECTED` shares the same branch (`isReviewDecision(latestJobStatus) && entries.length === 0 → CodeOnlyView`) so I read this as already covered. Happy to parametrize that test over both statuses if you'd like the explicit `CODE-REJECTED` case spelled out — let me know.

**`StudyResultsWithReview` rendering no longer asserted** — checked this and `StudyResultsWithReview` is fully orphaned: `grep -rn StudyResultsWithReview src/ tests/` returns only the definition site, no importers anywhere. Rather than port the assertion forward I deleted the component in 8c03da63 — the redesign-rollup PRs left it stranded.

The `useStudyHref` behavior change you flagged is now called out under its own ⚠️ section in the PR description. Ready for another look when you have a moment.

src/app/[orgSlug]/study/[studyId]/review/page.test.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/review/page.test.tsx]: Actually on second thought I like this, I'll rename. ty!

Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/review/page.test.tsx]: Did this one too in b647d79a — but split the rename by audience since there were actually *two* `StudyDetailsRedesignView` components (one in `review/` for the reviewer page, one in `view/` for the researcher page). Flattening both to plain `StudyDetails` would have left two same-named components disambiguated only by file path, which makes greps mixed. Renamed them to `StudyDetailsReviewer` and `StudyDetailsResearcher` instead — keeps the audience signal explicit at call sites. Also dropped the now-redundant `(DO)` / `(RL)` suffixes from the describe blocks.

therealmarv (approved):
nice, looks very good for me now!

PR #750, opened 2026-05-28
OTTER-497: allow any submitting-lab member to resubmit a change-requested proposal

therealmarv:
see comments [[1]](https://openstax.atlassian.net/browse/OTTER-497?focusedCommentId=42575) and [[2]](https://openstax.atlassian.net/browse/OTTER-497?focusedCommentId=42579) from OTTER-497:

## Issue

When a Data Org sends a proposal back to a Research Lab with changes requested, anyone in that lab can open and edit the proposal together. But when it comes time to send it back, only the person who originally created the proposal can click "Resubmit." Everyone else in the lab gets an access error. That's inconsistent with how the first submission works, where any lab member can submit.

## Proposed Fix

Make "Resubmit" behave exactly like the original submit: any member of the same Research Lab can resubmit a change-requested proposal, not just the original creator. The original creator stays listed as the study owner (no change there), we're only removing the restriction on who's allowed to press the Resubmit button.

## Implemented Fix

- Allow any member of the submitting lab to edit and resubmit a change-requested proposal, not just the original researcher who created it.
- Keep the original creator recorded as the proposal owner even when a teammate makes the edits or resubmission.
- Restrict editing and resubmitting to members of the submitting lab, blocking users from other labs.
- Prevent a second resubmission once any lab member has already resubmitted, attributing the resubmission note to whoever actually submitted it.
- Add tests covering same-lab teammate access, outside-lab rejection, and the one-resubmission-only safeguard.


src/server/actions/study-request.ts
```diff
@@ -695,13 +711,20 @@ export const resubmitProposalAction = new Action('resubmitProposalAction', { per
...
+
         const study = await db
             .selectFrom('study')
             .select(['id', 'status'])
             .where('id', '=', studyId)
-            .where('researcherId', '=', userId)
+            .where('submittedByOrgId', 'in', labScope)
             .executeTakeFirst()
 
         if (!study) throw new Error('Study not found or access denied')
```
nathanstitt: While you're in here — this `throw new Error('Study not found or access denied')` and the `throw new Error('Cannot resubmit study: …')` two lines below now sit between two new `ActionFailure` throws (line 689 in the sibling action, and line 757 below). `Error` vs `ActionFailure` propagate differently to clients: `ActionFailure` becomes a structured `{ error: { submission: '…' } }` in the action response, while a plain `Error` bubbles up as an unhandled exception and the client sees a generic failure. The two cases here are *legitimately user-facing* (wrong-lab access, wrong-status race), so promoting both to `ActionFailure` would give callers something to display. Not a blocker, but worth doing while the file is open.

src/server/actions/study-request.ts
```diff
@@ -715,16 +738,24 @@ export const resubmitProposalAction = new Action('resubmitProposalAction', { per
...
                 ...updateValues,
                 status: 'PENDING-REVIEW',
             })
             .where('id', '=', studyId)
-            .where('researcherId', '=', userId)
-            .execute()
+            .where('status', '=', 'CHANGE-REQUESTED')
+            .where('submittedByOrgId', 'in', labScope)
+            .returning(['id'])
+            .executeTakeFirst()
```
nathanstitt, line 762: Worth noting in the comment above (or as a quick separate one) that the SELECT at line 723-728 isn't strictly required for safety — the UPDATE here would 0-row on the same conditions and the `if (!claimed)` would catch it. The SELECT pulls its weight by producing a more specific error message at line 730/732 that distinguishes "not your lab" from "already submitted (race)". Calling that out keeps a future reader from trying to "simplify" by deleting the SELECT and losing the diagnostic split.

src/server/actions/study-request.ts
```diff
@@ -646,31 +646,47 @@ const resubmissionNoteParam = z
...
 })
     .params(z.object({ studyId: z.string(), studyInfo: draftStudyApiSchema }))
     .middleware(async ({ params: { studyId } }) => await getInfoForStudyId(studyId))
     .requireAbilityTo('update', 'Study')
     .handler(async ({ db, params: { studyId, studyInfo }, session }) => {
-        const userId = session.user.id
+        const userLabOrgIds = Object.values(session.orgs)
+            .filter((org) => org.type === 'lab')
+            .map((org) => org.id)
+        const labScope = userLabOrgIds.length > 0 ? userLabOrgIds : ['']
```
nathanstitt, line 664: Tiny consistency thing: the existing `onUpdateDraftStudyAction` (line 209-211) and `finalizeStudySubmissionAction` (line 326-328) inline the `userLabOrgIds.length > 0 ? userLabOrgIds : ['']` boilerplate at the `.where(...)` call site. Here (and at line 721) you've extracted it into a `labScope` local. 

Both are fine, but since we're doing this in 3 places now, maybe we need to extract a `getLabIds()` helper or something?

nathanstitt (approved):
looks good just a few suggestions, thanks Marvin

therealmarv:
I will merge this branch now, the failing e2e tests are from main and not introduced by this branch.

PR #751, opened 2026-05-28
OTTER-521 follow-up

Rod0352:
## OTTER-521 follow-up — QA fixes

Addresses the open QA comment on the ticket:

- **Other researchers in the lab got an error page on Edit & resubmit** — the page gate only let the researcher who originally submitted the study through. Relaxed it to any researcher in the submitting lab, matching how the existing draft-edit flow scopes access. The server actions that backed save and resubmit had the same `researcherId = userId` filter, so co-author saves were silently no-op'ing too; swapped both to the same lab-scoped guard.

- **Auto-save missing for the resubmission note** — mirrored the code-resubmit pattern: new `proposal_resubmission_note_draft` column on study, plus a debounced autosave wired through the edit-resubmit context. The "All changes saved" label flows into the existing autosave-status prop on the note section. Draft clears on successful resubmit.

- **Save as draft didn't save the resubmission note** — `saveDraft()` now flushes the note alongside the proposal fields, and the button's disabled-state also tracks `noteForm.isDirty()` so a note-only edit still triggers the save.

### Tests
DB-backed coverage for co-author save-draft + resubmit, cross-lab rejection, and draft-clear-on-resubmit. Unit coverage for the autosave retry-on-failure path and reload hydration.

### Migration
`ALTER TABLE study ADD COLUMN proposal_resubmission_note_draft text;` — nullable, no backfill.


src/server/actions/study-request.ts
```diff
@@ -715,15 +723,17 @@ export const resubmitProposalAction = new Action('resubmitProposalAction', { per
...
             .updateTable('study')
             .set({
                 ...updateValues,
                 status: 'PENDING-REVIEW',
+                proposalResubmissionNoteDraft: null,
             })
             .where('id', '=', studyId)
-            .where('researcherId', '=', userId)
+            .where('submittedByOrgId', 'in', userLabOrgIds.length > 0 ? userLabOrgIds : [''])
             .execute()
```
nathanstitt: This UPDATE has no `status = 'CHANGE-REQUESTED'` guard, so the transition isn't idempotent against a concurrent resubmitter. The SELECT at 708-713 only verifies status before the work starts — two co-authors both clicking Resubmit can each pass that check, then both UPDATE here (the second flips PENDING-REVIEW→PENDING-REVIEW, a no-op), then both INSERT a `RESUBMISSION-NOTE` row at 739-749 with bumped versions from `nextVersionForStudyComment({ increment: true })`. End state: two notes, two version bumps, both "successful."

Adding `.where('status', '=', 'CHANGE-REQUESTED')` on this UPDATE + checking the affected-rows count would make the second resubmitter fail cleanly (no comment row inserted, no version bump). PR #750 does exactly this on its corresponding handler — worth syncing the two PRs on this point so whichever lands first doesn't regress when the other merges.

src/server/actions/study-request.ts
```diff
@@ -771,6 +781,32 @@ export const saveCodeResubmissionNoteDraftAction = new Action('saveCodeResubmiss
...
+            .filter((org) => org.type === 'lab')
+            .map((org) => org.id)
+
+        await db
+            .updateTable('study')
+            .set({ proposalResubmissionNoteDraft: note })
+            .where('id', '=', studyId)
+            .where('status', '=', 'CHANGE-REQUESTED')
+            .where('submittedByOrgId', 'in', userLabOrgIds.length > 0 ? userLabOrgIds : [''])
+            .execute()
```
nathanstitt: Same silent-no-op pattern as `onUpdateClarifiedProposalAction`: cross-lab caller flows through the UPDATE filter, 0 rows match, and the handler returns `{ studyId, savedAt }` indicating success. The client then displays "All changes saved" while the note was never persisted to the correct row. Worth checking the affected-rows count and either throwing `ActionFailure` on 0 rows or short-circuiting earlier with a lab-membership check on `study.submittedByOrgId`.

src/server/actions/resubmit-proposal.test.ts
```diff
@@ -209,3 +213,181 @@ describe('onUpdateClarifiedProposalAction', () => {
...
+        // Switch session to a user in lab-cross-author-B
+        await mockSessionWithTestData({ orgSlug: 'lab-cross-author-B', orgType: 'lab' })
+
+        const result = await onUpdateClarifiedProposalAction({
+            studyId: study.id,
+            studyInfo: { title: 'Cross-lab hijack' },
+        })
+        // No CASL/ability error today — the action silently no-ops the UPDATE
+        // when the lab guard doesn't match. Assert the data didn't change.
+        expect('error' in result).toBe(false)
```
nathanstitt: This assertion locks in the silent-no-op behavior (`expect('error' in result).toBe(false)` for a cross-lab attempt) as the contract. I'd argue the inverse: a cross-lab attempt should be an error response (or thrown failure), and this test should assert `'error' in result === true`. The current behavior means the client-side caller has no way to distinguish "saved" from "silently dropped because you're in the wrong lab" — see the matching note on `study-request.ts:676`. Easy fix if we promote the action to reject 0-row UPDATEs.

src/contexts/edit-resubmit/context.tsx
```diff
@@ -54,11 +61,83 @@ export function EditResubmitProvider({ children, studyId, draftData }: EditResub
 
     const noteForm = useForm<ResubmitNoteValue>({
         validate: zodResolver(resubmitNoteSchema),
-        initialValues: initialResubmitNoteValue,
+        initialValues: { ...initialResubmitNoteValue, resubmissionNote: initialNote },
         validateInputOnChange: true,
     })
 
-    const { saveDraft, isSaving } = useResubmitSaveDraft({ studyId, form })
+    const { saveDraft: savePropsalDraft, isSaving } = useResubmitSaveDraft({ studyId, form })
```
nathanstitt: Tiny typo — `savePropsalDraft` should be `saveProposalDraft`. Repeated at line 137 (`.then(([proposalOk, noteOk] …)`) and 139 (in the `useCallback` deps array).

src/contexts/edit-resubmit/context.tsx
```diff
@@ -54,11 +61,83 @@ export function EditResubmitProvider({ children, studyId, draftData }: EditResub
...
+                        inFlightNoteSaveRef.current = null
+                    }
+                })
+            inFlightNoteSaveRef.current = savePromise
+
+            try {
+                return await savePromise
+            } catch {
+                return false
+            }
```
nathanstitt: Minor: the `try { return await savePromise } catch { return false }` here is dead because `savePromise` already chains `.catch(() => false)` at line 104. The chain resolves to `false` on failure rather than rejecting, so the outer `catch` block is unreachable. Either drop the try/catch around `await savePromise` or drop the inner `.catch(() => false)` — keeping both is harmless but suggests the failure handling is more defensive than it actually is.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx
```diff
@@ -20,11 +20,17 @@ export default async function StudyEditAndResubmitRoute(props: {
     if ('error' in study) return notFound()
     if (study.status !== 'CHANGE-REQUESTED') return notFound()
 
-    // Server-side writes are scoped to the original researcher — gate the
-    // page itself to match, so non-authors don't see a form that would
-    // silently no-op when they try to save.
+    // OTTER-521 follow-up: any researcher in the submitting lab can edit &
+    // resubmit (per QA — Malarvizhi/Marvin's comments on 5/13/26). Mirrors
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/page.tsx]: Stylistic — comments referencing named individuals and specific dates ("per QA — Malarvizhi/Marvin's comments on 5/13/26") tend to rot and I don't find them especially useful, that's what `git blame` is for

src/contexts/edit-resubmit/context.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+            <span data-testid="note-last-saved">{noteLastSavedAt ? noteLastSavedAt.toISOString() : 'never'}</span>
+        </>
+    )
+}
+
+describe('EditResubmitProvider — proposal resubmission note autosave', () => {
+    // Mirrors OTTER-558's behavior for the code resubmit flow. Without the retry
+    // guard a failed save would mark the value as "saved" and the next click
+    // would be a no-op, silently dropping the researcher's note.
+    it('retries the same note after a save failure instead of marking it saved', async () => {
```
nathanstitt, line 42: Nice — this is exactly the bug class that bites silent-autosave flows: a transient failure marks the value as "saved" (because the optimistic update beat the network), the next user click is a no-op (because `pendingValue === lastSavedValue`), and the researcher's note is silently dropped. Forcing one failure + one success via `mockResolvedValueOnce` and asserting two distinct save calls happened catches that. Same shape as the OTTER-558 test the comment references; worth the parallel coverage here.

nathanstitt (commented):
looking good, just a few suggestions on clarifying the security checks

nathanstitt (approved):
looks great, thank you Rodney

PR #753, opened 2026-05-30
OTTER-579: Route approved studies to code after agreement acknowledgment

stetradis:
**Root cause**: The `useStudyHref` hook (which determines where the dashboard "View" button navigates) didn't account for whether agreements had already been acknowledged. When status === 'APPROVED' with no job activity, it always routed to /agreements (or /submitted with the post-submission feature flag) — even if the researcher had already clicked "Proceed to step 4" on the agreements page.

Changes made:

`src/components/dashboard/studies-table/types.ts` — Added `researcherAgreementsAckedAt: Date | null` to the `StudyRow` type (the data was already being fetched from DB, just not exposed to the UI)

`src/hooks/use-study-href.ts` — Added an `agreementsAcked` parameter. When status === 'APPROVED', no job activity, but agreements are already acknowledged, it now routes directly to /code (step 4) instead of /agreements or /submitted

`src/components/dashboard/studies-table/study-action-link.tsx` — Passes `!!study.researcherAgreementsAckedAt` to `useStudyHref`

`src/hooks/use-study-href.test.ts` — Added tests for both the standard and post-submission flows confirming that acked agreements route to /code


nathanstitt (approved):
Looks great @stetradis thank you!

PR #755, opened 2026-05-30
OTTER-575: extract reusable footer and dynamically update year 

stetradis:
Introduced a new AppFooter component that renders the application footer with dynamic year and consistent styling.


src/components/layout/app-footer.tsx
```diff
@@ -0,0 +1,13 @@
+'use client'
+
+import { AppShellFooter, Text } from '@mantine/core'
+
+export function AppFooter() {
+    return (
+        <AppShellFooter p="md" bg="purple.9" bd="none">
+            <Text ta="left" c="white" fz="sm">
+                © {new Date().getFullYear()} SafeInsights, Rice University
```
nathanstitt: Two small things on this line:

did you mean for the copy to change?  the old footer read `© 2025 - SafeInsights, Rice University` (with the ` - ` separator); this drops it to `© {year} SafeInsights, Rice University`. Probably intentional, just flagging since it's a user-visible text change riding along with the refactor.

src/components/layout/app-footer.tsx
```diff
@@ -0,0 +1,13 @@
+'use client'
```
nathanstitt: Minor: `AppFooter` has no client-only behavior of its own (no hooks, no handlers) and every caller — `AppShell`, `FocusedLayoutShell`, `not-found`, `error-boundary` — is already a client component, so this `'use client'` is redundant. Harmless to leave, but it could come off if we wanted the footer to stay server-renderable for reuse elsewhere.

nathanstitt (approved):
Looks good, just a few things to double check you intended.

src/components/layout/app-footer.tsx
stetradis: Good catch, that was an accidental. Added it back!

PR #756, opened 2026-06-01
Update UI to indicate that a file hasn't been shared

jbwilson8:
## Summary

Issue: [OTTER-511](https://openstax.atlassian.net/browse/OTTER-511)

### Problem
When a Data Organization (DO) reviewer approves a study job but chooses **not** to share some files with the researcher (RL), the UI gave no indication those files were withheld — the deselected files simply **disappeared** from the DO's file table on the post-approval view. The DO had no way to confirm at a glance which files were excluded.

### What changed
On the DO's review panel, files that were **not** shared now render a **red ✗ icon** (with a "Not shared with researcher" tooltip) instead of vanishing. Shared files keep their green check; the selection checkboxes during active review are unchanged.

- **`file-type-helpers.ts`** — add `APPROVED_TO_ENCRYPTED` reverse map + `encryptedTypeForApproved()` helper so an approved row can resolve back to the encrypted source type its metadata is keyed under.
- **`use-encrypted-files-panel.ts`** — add a `'not-shared'` row state. Once a job is approved (`FILES-APPROVED` status), the builder diffs the encrypted-file metadata (all original paths, already fetched) against the approved set: paths with no approved counterpart become `not-shared` rows (no View/Download). Also covers a whole log type being withheld (previously fell through to the lock icon).
- **`encrypted-files-panel.tsx`** — render a red `XCircleIcon` for the `not-shared` state.

### Who sees this
**DO only.** `EncryptedFilesPanel` is rendered exclusively on enclave `/review/*` pages. The researcher view (`job-results.tsx`) fetches only `APPROVED-*` files and never mounts this panel, so the RL is never shown which files were withheld — as required by the ticket.

[OTTER-511]: https://openstax.atlassian.net/browse/OTTER-511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OTTER-511]: https://openstax.atlassian.net/browse/OTTER-511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/hooks/use-encrypted-files-panel.ts
```diff
@@ -159,12 +163,14 @@ export function useEncryptedFilesPanel({ job, onFilesApproved }: Options) {
...
 
+            const encryptedType = isEncrypted ? f.fileType : encryptedTypeForApproved(f.fileType)
             const approvedFilesForType = approvedFilesByType.get(approvedType) ?? []
             const decryptedFilesForType = decryptedFilesByType.get(approvedType) ?? []
-            const metaList = metadataByFileType.get(f.fileType) ?? []
+            const metaList = metadataByFileType.get(encryptedType) ?? []
             const label = logLabel(f.fileType)
 
-            if (approvedFilesForType.length > 0) {
+            if (isJobApproved) {
```
nathanstitt: One thing worth thinking about with the new `isJobApproved` gate. Before, this branch only ran when there were actually approved files for the type (`approvedFilesForType.length > 0`). Now it runs as soon as the job hits `FILES-APPROVED` — which means the two fallbacks below it (the locked-rows branch and the final locked-placeholder branch) can never run for an approved job.

That's the behavior we want once everything has loaded. The catch is that the encrypted-files fetch is async. While it's still in flight — or if it failed and got swallowed by the Sentry catch up top — `metaList` is empty. In that moment an approved job shows its approved rows but no "not shared" rows and no placeholder, so the withheld files briefly disappear again. That's the exact thing this PR is fixing, just shrunk down to the loading/error window.

It sorts itself out on the next render, so this probably isn't worth blocking on. If we wanted it fully airtight we could fall back to a locked-style placeholder when `isJobApproved && metaList.length === 0`.

nathanstitt (approved):
Seems good,  but think we might want to still check file length, see what you think

PR #758, opened 2026-06-01
OTTER-517: My Dashboard - Inconsistencies with dual roles

therealmarv:
see [OTTER-517](https://openstax.atlassian.net/browse/OTTER-517)

**Why this happens:** The Reviewer/Researcher toggle is rendered inside `StudiesTable`'s header, but `StudiesTable` had early `return`s for the empty (and error) states that fired *before* the header was built. So whenever the selected role had zero rows, the component returned just the empty message and the whole header, including the toggle, disappeared. The effective condition for showing the toggle became "dual-role **and** the selected tab has rows" instead of just "dual-role." That produced both reported symptoms: a new dual-role user lands on the empty Researcher tab and gets stuck with no toggle (Issue 1), and switching to an empty Reviewer tab makes the toggle vanish (Issue 2). Issue 3 was simply hardcoded empty-state copy that didn't match the per-audience wording.

**Fix:** the header always renders once studies load; only the body below it reflects empty/error/table state, and the empty copy is now audience-specific.

[OTTER-517]: https://openstax.atlassian.net/browse/OTTER-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/dashboard/studies-table/empty-state.tsx
+++ b/src/components/dashboard/studies-table/empty-state.tsx
@@ -1,13 +1,23 @@
 import { Stack, Text } from '@mantine/core'
-import { Audience } from './types'
+import { Audience, Scope } from './types'
 
-export function EmptyState({ audience }: { audience: Audience }) {
-    const message =
-        audience === 'reviewer' ? 'You currently do not have any studies to review' : 'You have not started a study yet'
+// 'My dashboard' (user scope) uses participation-focused copy so dual-role users get a
+// role-specific empty state (OTTER-517). The org dashboards keep their existing copy.
+const MESSAGES: Record<Scope, Record<Audience, string>> = {
+    user: {
+        reviewer: "You haven't yet participated in reviewing a study",
+        researcher: "You haven't yet participated in a study",
+    },
+    org: {
+        reviewer: 'You have no studies to review.',
+        researcher: "You haven't started a study yet",
+    },
+}
 
+export function EmptyState({ audience, scope }: { audience: Audience; scope: Scope }) {
     return (
         <Stack align="center" gap="md">
-            <Text>{message}</Text>
+            <Text>{MESSAGES[scope][audience]}</Text>
         </Stack>
     )
 }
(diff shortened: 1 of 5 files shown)


src/components/dashboard/studies-table/index.tsx
```diff
@@ -153,20 +146,15 @@ export function StudiesTable({
...
-                            audience={audience}
-                            scope={scope}
-                            orgSlug={effectiveOrgSlug}
-                        />
-                    ))}
-                </TableTbody>
-            </Table>
+            <StudiesTableBody
+                isError={isError}
+                error={errorToString(studies)}
```
nathanstitt: Small thing noticed while the diff has this line open: on `isError`, `studies` is still the defaulted `[]` (from `data: studies = []` above), not the error object. So `errorToString([])` falls all the way through to its `'Unknown error occured'` fallback, and the alert ends up reading "Failed to load studies: Unknown error occured" — the real reason never surfaces. 

Not introduced by you but would still be great to fix by  threading the actual error through (e.g. capture the `error` object from `useQuery` and pass that), or at least drop the now-meaningless interpolation. Not a blocker.

src/components/dashboard/studies-table/empty-state.tsx
```diff
@@ -1,29 +1,23 @@
...
+// 'My dashboard' (user scope) uses participation-focused copy so dual-role users get a
+// role-specific empty state (OTTER-517). The org dashboards keep their existing copy.
+const MESSAGES: Record<Scope, Record<Audience, string>> = {
+    user: {
+        reviewer: "You haven't yet participated in reviewing a study",
+        researcher: "You haven't yet participated in a study",
+    },
+    org: {
+        reviewer: 'You have no studies to review.',
+        researcher: "You haven't started a study yet",
```
nathanstitt, line 13: huge nitpick but: now that all four messages are co-located, the punctuation is a little inconsistent; `org.reviewer` ends in a period ("...to review.") while the other three don't. Worth aligning one way or the other while they're side by side

nathanstitt (approved):
looks great, ty, just a few tiny suggestions

PR #759, opened 2026-06-02
OTTER-538 follow-up: Fix researcher "Previous" navigation on results-stage Study Details page

therealmarv:
see comments of [OTTER-538](https://openstax.atlassian.net/browse/OTTER-538)

- Fixed the researcher "Previous" button on the results-stage Study Details page so it returns to the post-code-submission page instead of the dashboard.
- Hid the "code under review" notice on that page once results exist, since the review has already concluded.
- Added a regression test confirming the reviewer "Previous" button lands on the post-code-feedback page rather than a blank screen.
- Added tests covering the corrected researcher navigation and the hidden review notice.

[OTTER-538]: https://openstax.atlassian.net/browse/OTTER-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/view/page.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/view/page.tsx
@@ -34,7 +34,8 @@ export default async function StudyReviewPage(props: {
 
     const job = await latestJobForStudyOrNull(studyId)
 
-    const dashboardHref = searchParams.returnTo === 'org' ? Routes.orgDashboard({ orgSlug }) : Routes.dashboard
+    const returnTo = searchParams.returnTo === 'org' ? 'org' : undefined
+    const dashboardHref = returnTo ? Routes.orgDashboard({ orgSlug }) : Routes.dashboard
 
     const fromAgreements = searchParams.from === 'agreements'
     const fromCodeSubmission = searchParams.from === 'code-submission'
@@ -49,8 +50,10 @@ export default async function StudyReviewPage(props: {
 
         // RL "Previous" from the results-stage Study Details page returns here with ?from=code-submission.
         // Once results exist the under-review branch no longer fires, so render the OTTER-537 post-submission
-        // page explicitly. The "code under review" banner is hidden because review has already concluded.
-        if (fromCodeSubmission) {
+        // page explicitly. Gate on the results status (not the query param alone) so a stray
+        // ?from=code-submission at a code-decision/under-review status falls through to the correct branch
+        // below. The "code under review" banner is hidden because review has already concluded.
+        if (isStudyResultsStatus(latestJobStatus) && fromCodeSubmission) {
             const reviewingOrgName = await getOrgNameFromId(study.orgId)
             return (
                 <CodePostSubmissionView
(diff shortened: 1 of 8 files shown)


src/app/[orgSlug]/study/[studyId]/view/study-details-researcher.tsx
```diff
@@ -21,7 +22,7 @@ type StudyDetailsResearcherProps = {
 }
 
 export function StudyDetailsResearcher({ orgSlug, study, job, dashboardHref }: StudyDetailsResearcherProps) {
-    const previousHref = dashboardHref ?? Routes.dashboard
+    const previousHref = Routes.studyView({ orgSlug, studyId: study.id, from: 'code-submission' })
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/study-details-researcher.tsx]: This `previousHref` drops the `returnTo=org` context. The page computes `dashboardHref` from `searchParams.returnTo === 'org'`, and this component still receives that `dashboardHref` (it's wired into the breadcrumb on line 34) — but the Previous link rebuilds the URL with only `from: 'code-submission'`. So for a researcher who entered from the org dashboard, clicking results → Previous lands them on the post-submission page whose own breadcrumb "Dashboard" link and "Go to dashboard" button now point to the *default* dashboard, not the org one. There's prior art for threading this through: `agreements/page.tsx` builds its studyView back-link as `...?from=agreements${returnToSuffix}`. Worth either extending the `studyView` route helper to carry `returnTo` or appending it the same way here. Heads-up that the updated test only asserts `stringContaining('?from=code-submission')`, so it won't catch the missing `returnTo`.

src/app/[orgSlug]/study/[studyId]/view/page.tsx
```diff
@@ -46,6 +47,23 @@ export default async function StudyReviewPage(props: {
         const latestJobStatus = job.statusChanges[0]?.status
         const isUnderReview = study.status === 'PENDING-REVIEW' && isUnderReviewStatus(latestJobStatus)
 
+        // RL "Previous" from the results-stage Study Details page returns here with ?from=code-submission.
+        // Once results exist the under-review branch no longer fires, so render the OTTER-537 post-submission
+        // page explicitly. The "code under review" banner is hidden because review has already concluded.
+        if (fromCodeSubmission) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/page.tsx]: This `fromCodeSubmission` branch sits before the `isCodeDecisionStatus` / `isUnderReview` branches below and fires for *any* `latestJobStatus` once code was submitted 

This is _probably_ fine right now, the only producer of `?from=code-submission` is `StudyDetailsResearcher`, which renders solely at results statuses, and results statuses are disjoint from code-decision statuses, so the ordering doesn't change behavior. 

However in the future if a `?from=code-submission` arrives from a `CODE-APPROVED/REJECTED/CHANGES-REQUESTED` status would now short-circuit to the post-submission view instead of `CodePostDecisionView`. Not reachable today, but since the guard is purely on the query param, a gate like `isStudyResultsStatus(latestJobStatus) && fromCodeSubmission` would make the intent explicit and keep a future caller from tripping it.

nathanstitt (approved):
Seems good but check the comment about previousHref, that might be a bug we should correct before merging

PR #764, opened 2026-06-03
Researcher encrypted results

chrisbendel:
Pairs with https://github.com/safeinsights/encryption/pull/26

# Test plan — researcher encrypted results

## Setup — keys first (both at `/account/keys`, save PEM)
- **Reviewer** (enclave user) key — needed *before* inject (script encrypts to it).
- **Researcher** (lab user) key — needed *before* results-approve (re-wrap targets it).

## Flow

1. **Researcher:** create proposal → upload code → submit. → study `PENDING-REVIEW`, job `CODE-SUBMITTED`.
2. **Reviewer:** open study → **Approve** proposal. → job `CODE-APPROVED`+`JOB-READY`, study `APPROVED`. *(Code-approval gate. Do NOT skip — script's RUN-COMPLETE assumes it.)*
3. **Grab jobId:**

```
docker compose exec postgres psql -U mgmnt -d mgmnt_dev -c \
  "select sj.id, s.title from study_job sj join study s on s.id=sj.study_id order by sj.created_at desc limit 1;"
```

4. **Inject** (= enclave runs + uploads encrypted results) → `RUN-COMPLETE`:

```
docker exec mgmnt-app pnpm exec tsx bin/inject-encrypted-results.ts <jobId>
```

5. **Reviewer:** job page → decrypt (reviewer PEM) → **Approve** results → `FILES-APPROVED` + re-wrap fires.
6. **Researcher:** results page → decrypt (researcher PEM) → results + logs visible/downloadable.

## DB check after step 5

```
docker compose exec postgres psql -U mgmnt -d mgmnt_dev -c \
  "select file_path, fingerprint from study_job_file_key order by created_at desc;"
```

Expect **2 rows** per researcher fingerprint: `results.csv` + `run.log`. Empty before approve.

## Spot checks
- Researcher sees **results AND logs** (all-or-nothing).
- Wrong PEM → clean error.
- Reject path → researcher sees rejected, no files.
- Late-joiner: key made *after* approve → no rows, can't decrypt (no backfill).


package.json
```diff
@@ -109,7 +109,7 @@
         "react-phone-number-input": "^3.4.16",
         "remeda": "^2.33.6",
         "server-only": "^0.0.1",
-        "si-encryption": "github:safeinsights/encryption#2e35d3bb56f02bfc0e6d5cd9d83c54637ed0ac20",
+        "si-encryption": "github:safeinsights/encryption#2c495dae94942895a34e64d91f83809750867f4e",
```
chrisbendel: TODO Merge and publish new encryption package first, then update this hash

src/server/results-sharing.ts
```diff
@@ -0,0 +1,63 @@
...
+    const rows = sharedFiles.flatMap((file) =>
+        file.keys.map((key) => {
+            if (!labFingerprints.has(key.fingerprint)) {
+                throw new ActionFailure({ file: `fingerprint ${key.fingerprint} is not a lab recipient` })
+            }
+            return { studyJobFileId: file.studyJobFileId, fingerprint: key.fingerprint, crypt: key.crypt }
+        }),
+    )
+
+    const approvedFileIds = sharedFiles.map((f) => f.studyJobFileId)
```
nathanstitt: We validate every `crypt` targets a real lab fingerprint, but nothing here checks that the `studyJobFileId`s actually belong to `jobId`. `approvedFileIds` is taken straight from the client payload and fed into the `studyJobFileKey` insert and the `studyJobFile` `approvedAt` update with no `studyJobId` constraint. A reviewer with approve rights on study A could pass file IDs from another job — as long as that other job's lab happens to share a fingerprint, those foreign rows get marked approved/shared. Narrow blast radius, but worth an explicit guard: fetch the file ids for `jobId` and reject (or filter) any `studyJobFileId` not in that set before inserting/updating. The comment block above documents the *fingerprint* trust boundary thoroughly — this is the one ownership check that slips through it.

src/server/storage.ts
```diff
@@ -34,9 +35,70 @@ async function storeJobFile(info: MinimalJobInfo, path: string, file: File, file
...
+    // directly as the S3 key suffix; if that trust ever weakens, sanitize for `..`/collisions.
+    const items = decomposed.map((entry) => ({
+        entry,
+        path: `${pathForStudyJob(info)}/results/${subdir}/${entry.path}`,
+    }))
+
+    // Write the bodies to S3 first, in parallel. S3 is not part of the DB transaction
+    // below, so if the transaction fails we best-effort delete what we wrote — otherwise a
+    // rollback would leave orphaned objects with no row pointing at them.
+    await Promise.all(items.map(({ entry, path }) => storeS3File(info, new Blob([entry.body]).stream(), path)))
```
nathanstitt: Nice handling of the S3-vs-transaction split — bodies written first, then best-effort `deleteS3File` via `allSettled` in the catch so a rolled-back transaction doesn't orphan objects. One residual edge: if the *process* dies between these writes and the transaction commit, the objects are orphaned with no row pointing at them and the catch never runs. Probably fine to leave to a lifecycle policy / sweeper, but a one-line note here would tell the next reader it's known rather than missed.

nathanstitt (commented):
I'm a little confused bc this is doing quite a bit of work to support sharing only some of the result  files with researchers and is moving from storing the keys from along side the files inside the zip blob to storing the them in the DB, which i'm not sure is a good idea. 

I worry that the storing the two parts on two different locations is inviting them getting out of sync.  If we had a good reason to do so that'd probably be fine, but I'm not seeing the reason to add the risk. 

I made some comments on the JIRA https://openstax.atlassian.net/browse/SHRMP-248 about this as well

chrisbendel:
@nathanstitt thanks for the review — and apologies, you hit this at the worst possible moment: right after I'd confirmed with @philschatz that we're dropping per-file approve/reject (it's all-or-nothing at the job level now), so you caught me mid-refactor. The version you reviewed (decompose-on-ingest, keys + ciphertext both pulled into the DB / per-file S3 objects) is gone. Quick summary of where it landed:

- **Prod zip format is preserved.** TOA's output and the reviewer decrypt path are byte-identical to what's in prod — reviewer keys stay in the manifest *inside* the zip, ciphertext is one untouched S3 blob. No decompose, no `iv`/`bytes` columns.
- **The only thing in the DB now is researcher keys** (`study_job_file_key`), added at approve time. They're an *additive overlay*: TOA encrypts for the reviewer/enclave keys at run time, and researchers aren't recipients yet, so the only way to grant them access without re-encrypting the immutable blob is a separate re-wrapped key.
- On your sync worry — it's narrower than for the decompose version: the zip (reviewer keys) and the DB (researcher keys) hold **disjoint** recipient sets, so they don't drift against each other. The one real seam is "replace/re-run the zip → DB researcher keys go stale."
- **Your ownership-check catch was a real bug and is fixed** — `insertSharedFileKeys` now verifies every `studyJobFileId` belongs to the job before sharing, not just the fingerprint. Good catch, thank you.

There's still a genuine fork worth your and Phil's input: do researcher keys stay in the DB (clean per-recipient revocation for Card 74 = delete a row), or go *into* the zip manifest like the reviewer keys (one source of truth, but a read-modify-write of the stored blob on every share/revoke)? I'm going to sync with Phil on whether revocation is still a committed requirement, since that's the deciding factor. Would value your take there.

src/components/layout/navbar-profile-menu.tsx
```diff
@@ -90,15 +90,15 @@ export function NavbarProfileMenu() {
 
             <Protect role={AuthRole.Reviewer}>
```
nathanstitt, line 95: Now that lab researchers hold a Results Key too (`orgNeedsKey`), this nav link is still gated behind `<Protect role={AuthRole.Reviewer}>` — so a researcher won't see "Results Key" in their profile menu. `RequireUserKey` covers the first-time *generate* redirect, but a researcher who already has a key has no menu entry to reach `/user-key` and regenerate it. Worth widening the `Protect` (or adding a Researcher-visible entry) so the rotate path is reachable for both roles.

src/server/actions/study-job.actions.ts
```diff
@@ -167,27 +176,73 @@ export const fetchEncryptedJobFilesAction = new Action('fetchEncryptedJobFilesAc
...
+        // naturally gated). Build the {file_path -> crypt} map per artifact.
+        const wrappedKeys = await db
+            .selectFrom('studyJobFileKey')
+            .select(['studyJobFileId', 'filePath', 'crypt'])
+            .where(
+                'studyJobFileId',
+                'in',
+                encryptedFiles.map((f) => f.id),
+            )
+            .where('fingerprint', '=', userKey.fingerprint)
```
nathanstitt, line 273: This seems important to solve:

Researcher access is keyed to `userKey.fingerprint`. If a researcher rotates their key (the `update` flow swaps the fingerprint), this filter stops matching the `study_job_file_key` rows wrapped against the *old* fingerprint, and they permanently lose access to already-approved results.

Is this intended?

src/server/actions/study-job.actions.ts
```diff
@@ -167,27 +176,73 @@ export const fetchEncryptedJobFilesAction = new Action('fetchEncryptedJobFilesAc
...
+        // Include submittedByOrgId so 'view StudyJob' matches lab researchers, not just enclave
+        // reviewers — researchers fetch their own re-wrapped result files here.
+        return { studyJob, orgId: studyJob.orgId, submittedByOrgId: studyJob.submittedByOrgId }
     })
     .requireAbilityTo('view', 'StudyJob')
 
-    .handler(async ({ studyJob }) => {
+    .handler(async ({ studyJob, session, db }) => {
+        const userKey = await getUserPublicKey(session.user.id)
+        if (!userKey) return []
```
nathanstitt, line 234: Small thing: `getUserPublicKey` does `executeTakeFirst()` over all of the user's rows, so if more than one `userPublicKey` row ever exists for a user the fingerprint chosen here is non-deterministic, which would silently mismatch the wrapped-key lookup below. Worth confirming there's a one-row-per-user invariant (unique constraint on `user_id`) backing this.

nathanstitt (commented):
I see we're still storing the encrypted AES key that decodes the results in the DB for researchers vs in the zip like we do for reviewers. 

It seems better to me to have a single way to store this, so either we store in the zip (reviewers), or in the db (researchers).  Either has tradeoffs:
 * zip lets us have a single blob that is self-contained and easy to pass between systems
 * db row lets us potentially revoke access to a single user by deleting the row.  However there's also other ways to revoke access, remove from org, or implement some other access rules.

I could be persuaded either way but think we should use the same mechanism for all users.

chrisbendel:
> I see we're still storing the encrypted AES key that decodes the results in the DB for researchers vs in the zip like we do for reviewers.
> 
> It seems better to me to have a single way to store this, so either we store in the zip (reviewers), or in the db (researchers). Either has tradeoffs:
> 
>     * zip lets us have a single blob that is self-contained and easy to pass between systems
> 
>     * db row lets us potentially revoke access to a single user by deleting the row.  However there's also other ways to revoke access, remove from org, or implement some other access rules.
> 
> 
> I could be persuaded either way but think we should use the same mechanism for all users.

Fair push — and you're right that one source of truth matters. The thing is, on the read path it already is one: ResultsReader takes the DB-stored wrapped keys (additionalKeys, wired here as researcherKeys) and splices them into the manifest under the recipient's fingerprint before any decrypt. So decrypt is a single manifest.files[path].keys[fingerprint] lookup for everyone — there's no second code path, the DB rows just hydrate the same manifest the zip would've carried.

Where they differ is write timing, not mechanism:

Zip manifest = recipients that existed when the job ran in the enclave (reviewers). It's the immutable run artifact — we never rewrite it.
DB table = anyone granted access after run time. Researchers today; and per the lost-key / compromised-key / new-hire / departed-member thread, this is also where renewal re-wraps will land.
The two never hold the same recipient, so there's nothing to drift between — writes are disjoint, reads are the union. The reason it's a DB row and not "append to the zip" is that the post-hoc set is mutable and grows (renewals especially), and read-modify-writing the stored blob on every new grant means rewriting a large S3 object with no row-level concurrency control, plus it mutates the artifact TOA produced. A row with a unique constraint handles concurrency and revocation cleanly and leaves the run artifact untouched.

To make that intent obvious in the schema, I'll rename study_job_file_key → study_job_file_recipient_key (and the param researcherKeys → recipientKeys) so it reads as "post-hoc recipient keys, any role," not "the researcher special-case."

Decompose (pulling reviewer keys out of the zip into the DB too, so there's literally one store) is the other way to unify — but that's the version I just backed out of, and it means rewriting the prod ingest/decrypt path TOA depends on. Happy to revisit if you think single-store is worth that.

src/server/actions/study-job.actions.ts
chrisbendel: Backed by a constraint — user_public_key has unique_user_id (migration 1742320602314), so it's one row per user. executeTakeFirst is deterministic and setUserPublicKeyAction's insert can only ever be first-time. Rotation is update-in-place on that row.

src/server/actions/study-job.actions.ts
chrisbendel: Intended for now. Rotation swaps the fingerprint and orphans the old wrapped rows — the recovery path is the renewal/re-wrap flow (a teammate re-wraps for the new key), which is a separate workstream and not in this PR. Until it lands we're not surfacing self-rotate to researchers (first-time generate still works via the key-required redirect), and I'm filing renewal as the immediate follow-up. Tracking issue: <link>.

src/server/results-sharing.ts
chrisbendel: Fixed — insertSharedFileKeys now fetches the file ids for jobId and rejects any studyJobFileId not in that set before insert/update, so the fingerprint check is no longer the only guard.

nathanstitt (approved):
Thanks for your patience on this @chrisbendel - while I still think it'd be _slightly_ better to have everything stored in the zip, I do see your reasoning for storing study-keys in the db.  So I'm fine with it for now at least, we can easily revisit later if it does turn out to be an issue, but I doubt it will.

Thanks again!

PR #765, opened 2026-06-03
Instrument Coder workspace launch path for diagnosability

nathanstitt:
## Why

A failed or hung IDE launch currently surfaces **only as an endless UI spinner** with no server-side signal — we can't tell which step broke. While investigating a prod launch where Coder reported the workspace running but the management-app UI kept spinning, the lack of breadcrumbs made the root cause impossible to pin down from logs alone (the readiness gate passed, but nothing downstream was logged).

This adds step-level breadcrumbs and loud, Sentry-reported error logging across the entire launch path so the next occurrence leaves a clear trail.

## What

**`src/server/coder/workspaces.ts`**
- **`getCoderWorkspaceUrl`** — tracks the active step (`fetch-workspace-status` / `initialize-code-files` / `generate-workspace-url`) and logs a `logger.error` naming the failing step on any throw (→ console **and Sentry**), plus milestone info logs. Every line is prefixed `[coder-launch study=… workspace=…]` for easy grepping.
- **`isWorkspaceReady` → `describeReadiness`** — now returns `{ ready, reason }`. A not-yet-ready poll logs *which* gate is holding it (build status, agent lifecycle/status, code-server health) instead of an opaque stream of nulls.
- **`initializeWorkspaceCodeFiles`** — logs the already-populated early return, the resolved code env + starter-file list, and per-file success; distinguishes **S3-fetch** vs **local-write** vs **agent-context** failures. Also made null-safe on `starterCodeFileNames`.
- **`generateWorkspaceUrl`** — logs the "Coder user not found" case.

**`src/server/coder/client.ts`**
- **`coderFetch`** — added a **15s timeout** (`AbortSignal.timeout`) so a hung Coder request becomes a logged, throwable error rather than a forever-pending poll (a prime suspect for the spinner), plus `logger.error` on non-ok responses and timeouts.

## Behavior change

The only functional change is the **15s timeout** on Coder API calls — everything else is logging. A previously-hanging request now rejects after 15s, which the client renders as an error instead of spinning indefinitely.

## Tests

- Updated existing `coderFetch` assertions for the new `AbortSignal`.
- Added `coderFetch instrumentation` coverage: non-ok → logs + throws `CoderApiError`; timeout → logs + rethrows.
- `coder.test.ts` + `workspaces.actions.test.ts`: 34 passing. `tsc` + `eslint` clean.

---

## Client-side follow-up: `use-workspace-launcher` refactor

While the server changes make a hung/failed launch *diagnosable*, the client hook that drives the launch handled errors poorly and obscured failures. Refactored `src/hooks/use-workspace-launcher.tsx` to be robust and to surface what the new server logging now reports.

**Problems addressed**
- Rolled its own state machine (`loading` / `error` / `launchComplete` / `workspaceId` `useState`s + a `setState`-in-effect latch) that mirrored React Query's own state and drifted out of sync.
- **Swallowed failures into local state with no Sentry/notification reporting** — a failed launch left the user with a generic error and no breadcrumb, undercutting the server-side instrumentation.
- Redundantly re-implemented `'error' in result` unwrapping that the `@/common` `useQuery`/`useMutation` wrappers already do (they throw `ActionFailure`).
- A blocked popup was surfaced as `error`, rendering a misleading **"Launch failed"** button even though the workspace launched fine.
- `clearError` never actually cleared poll errors.

**Changes**
- State now **derives from the creation mutation + URL poll** — no mirrored `useState`. `workspaceId` reads from `creation.data`.
- Mutation and poll failures are reported via `reportError` (**Sentry + notification**), each latched so a StrictMode double-invoke can't double-report.
- `toLaunchError` keeps a friendly fallback for opaque (non-string) action errors instead of leaking raw JSON.
- Popup-blocked shows the clickable fallback notification **only** — no longer a fake `error`.
- `clearError` fully clears poll errors via `queryClient.removeQueries`.
- Dropped the `react-hooks/set-state-in-effect` eslint-disable.

**Tests** — updated popup-blocked expectations; added coverage for error reporting (mutation + poll) and clearing a polling error. 16/16 in-file, 105/105 across `src/hooks/`. `tsc` + `eslint` + `prettier` clean.

### DIFF
--- a/src/hooks/use-workspace-launcher.test.tsx
+++ b/src/hooks/use-workspace-launcher.test.tsx
@@ -18,6 +18,11 @@ vi.mock('@/server/actions/workspaces.actions', () => ({
     getWorkspaceUrlAction: vi.fn(),
 }))
 
+vi.mock('@/components/errors', async (importOriginal) => ({
+    ...(await importOriginal<typeof import('@/components/errors')>()),
+    reportError: vi.fn(),
+}))
+
 const mockWindowOpen = vi.fn()
 Object.defineProperty(window, 'open', {
     value: mockWindowOpen,
@@ -25,6 +30,7 @@ Object.defineProperty(window, 'open', {
 })
 
 import { createUserAndWorkspaceAction, getWorkspaceUrlAction } from '@/server/actions/workspaces.actions'
+import { reportError } from '@/components/errors'
 import { notifications } from '@mantine/notifications'
 
 const studyId = faker.string.uuid()
@@ -147,7 +153,10 @@ describe('useWorkspaceLauncher', () => {
             })
         })
 
-        it('should show notification and set error when popup is blocked', async () => {
+        // A blocked popup is not a launch failure: the workspace launched fine and the user gets a
+        // clickable fallback notification, so the hook must NOT surface it as `error` (which would
+        // render a misleading "Launch failed" state in the UI).
+        it('should show fallback notification without erroring when popup is blocked', async () => {
             const workspaceUrl = 'https://workspace.example.com'
             mockWindowOpen.mockReturnValue(null) // Simulate blocked popup
             ;(createUserAndWorkspaceAction as Mock).mockResolvedValue({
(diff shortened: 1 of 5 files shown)


therealmarv (approved):
I'm okay with it from my side 🟢

PR #766, opened 2026-06-03
ci: cache SonarQube scanner and Trivy to harden infosec job

nathanstitt:
## Problem

The `infosec` CI job fails intermittently when one of its scan steps can't fetch a tool or database from upstream. Two independent download paths cause this:

- **SonarQube** — the scan step downloads the Sonar Scanner CLI from `binaries.sonarsource.com` on every run, so any transient SonarSource outage fails the job.
- **Trivy** — `trivy-action` downloads the Trivy **binary** (via `setup-trivy`) and the **vulnerability DB** on every run, so it flakes the same way when those fetches hit a transient upstream outage.

## Change

Harden both download paths so a cache hit avoids the network entirely.

### SonarQube scanner cache

Add an `actions/cache` step (same pinned SHA already used elsewhere in `checks.yml`) that caches `SONAR_USER_HOME`, redirected into the workspace via the env var so the cache actually covers the scanner install:

- **Cache hit** → the scanner binary is reused instead of re-downloaded.
- **Key** is pinned to the `sonarqube-scan-action` SHA, so bumping the action version automatically busts the cache and pulls the matching scanner.
- **`restore-keys`** provides a warm fallback from the most recent prior cache.

### Trivy binary + DB cache

`trivy-action` already enables its built-in `actions/cache` by default, but that cache only helps if the key is stable and the cache dir is shared. So on both Trivy scan steps:

- **Pin the binary `version`** (`v0.70.0`, the action's own default for this pinned SHA — behavior unchanged, now explicit) so the cache key no longer drifts run-to-run.
- **Share one `cache-dir`** (`${{ github.workspace }}/.cache/trivy`) across both steps, so the vuln DB is fetched at most once per run and reused on subsequent runs.
- A version bump naturally busts the cache.

## Security note

This respects the OTTER-545 security comment block on the job: the cache steps add no PR-controlled code execution, and `SONAR_TOKEN` stays scoped to the single scan step.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -63,12 +63,22 @@ jobs:
             - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
               with:
                   persist-credentials: false
+            # Trivy downloads its binary (via setup-trivy) and vulnerability DB on
+            # every run, so it flakes the same way the SonarQube scanner did when
+            # those fetches hit a transient upstream outage. trivy-action enables
+            # its built-in actions/cache by default; we make that cache effective by
+            # (1) pinning the binary `version` so the cache key is stable across runs
+            # instead of drifting, and (2) pointing both scan steps at one shared
+            # `cache-dir` so the DB is fetched at most once per run and reused on
+            # subsequent runs. A version bump naturally busts the cache.
             - name: Run Trivy vulnerability scanner in fs mode
               uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
               with:
                   scan-type: 'fs'
                   scan-ref: '.'
                   trivy-config: trivy.yaml
+                  version: v0.70.0
+                  cache-dir: ${{ github.workspace }}/.cache/trivy
             - name: Run Trivy license scanner
               uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
               with:
@@ -78,6 +88,8 @@ jobs:
                   severity: 'HIGH,CRITICAL'
                   exit-code: '1'
                   trivyignores: '.trivyignore.yaml'
+                  version: v0.70.0
+                  cache-dir: ${{ github.workspace }}/.cache/trivy
             # Cache the SonarScanner CLI and analysis cache (downloaded into
             # SONAR_USER_HOME by the scan step below). This makes the job resilient
             # to intermittent failures fetching the scanner from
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -63,12 +63,22 @@ jobs:
...
+            # `cache-dir` so the DB is fetched at most once per run and reused on
+            # subsequent runs. A version bump naturally busts the cache.
             - name: Run Trivy vulnerability scanner in fs mode
               uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
               with:
                   scan-type: 'fs'
                   scan-ref: '.'
                   trivy-config: trivy.yaml
+                  version: v0.70.0
+                  cache-dir: ${{ github.workspace }}/.cache/trivy
```
therealmarv, line 81: I think the comment here is wrong about the DB caching mechanism in Trivy (non-blocking).

Checked `action.yaml` at the pinned commit (`aquasecurity/trivy-action@ed142fd` = v0.36.0) and its `setup-trivy@v0.2.6` dependency, and these three inputs are actually no-ops at this SHA:

- `version` default = `v0.70.0`, `cache` default = `true`, `cache-dir` default = `${{ github.workspace }}/.cache/trivy`. So `version: v0.70.0` and `cache-dir: …/.cache/trivy` change nothing, and both steps already shared that default dir.
- The binary is already cached by default: `setup-trivy` keys it `trivy-binary-${version}-${os}-${arch}`, and since the action passes the concrete `v0.70.0` (not `latest`), it's reused across runs.
- The DB cache key is date-based, not version-based, so pinning `version` can't stabilize it:

      # Restore DB from cache
      key: cache-trivy-${{ steps.date.outputs.date }}
      restore-keys: cache-trivy-

  Source (pinned SHA): https://github.com/aquasecurity/trivy-action/blob/ed142fd0673e97e23eac54620cfb913e5ce36c25/action.yaml

Net: the only thing fetched fresh is the DB on the first run of each calendar day (by design, the DB updates ~every 6h), and these added lines don't change that, so they don't remove the Trivy flake source. Suggest either dropping them, or reframing the comment as "make defaults explicit so a future `trivy-action` SHA bump can't silently change them" (a smaller, accurate claim). The SonarQube cache change is the real fix and looks good.

therealmarv (approved):
The SonarQube fix is nice.

I think the trivy change does nothing (no-op) because it's the defaults just made explicit. See comment.

.github/workflows/checks.yml
nathanstitt: yes pretty sure this isn't the case because we've had PRs fail because download failed

PR #767, opened 2026-06-03
OTTER-598: remove legacy CodeOnlyView fallback from RL study view

therealmarv:
see [OTTER-598](https://openstax.atlassian.net/browse/OTTER-598)

- Researchers viewing a study now always see the redesigned study page; the outdated fallback screen has been removed entirely so it can never appear.
- The page that handles approved, change-requested, and rejected code now decides what to show based on the study's full history, fixing cases where the wrong or inconsistent screen appeared on reload.
- While approved code is running, the page shows a clear "running" state without the submitted code listing, instead of the old confusing status message.
- If reviewer feedback fails to load, the page now shows a brief notice and stays on the correct screen rather than dropping to the old layout.
- Rejected studies can no longer be resubmitted; only studies with requested changes can be edited and resubmitted.
- Added and updated tests to cover these states, and removed a small unused helper.

[OTTER-598]: https://openstax.atlassian.net/browse/OTTER-598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx
```diff
@@ -80,7 +87,7 @@ function deriveCodePostDecision({
 }) {
     return {
         copy: DECISION_COPY[decision],
-        timestampDate: entries[0].createdAt,
+        timestampDate: job.statusChanges.find((s) => s.status === decision)?.createdAt ?? entries[0]?.createdAt ?? null,
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx:90]: Nice — sourcing the timestamp from the matching status-change row is better than the old `entries[0].createdAt`

src/lib/code-resubmission.ts
```diff
@@ -1,8 +1,9 @@
 import type { StudyJobStatus } from '@/database/types'
 
+// CODE-REJECTED is terminal: a rejected study allows no resubmission, only a brand-new
```
nathanstitt, line 3: Removing `CODE-REJECTED` from the resubmittable set is the right call and matches OTTER-556's rejected-state copy ("No further code submissions will be accepted for this study, but you may submit a new study proposal"). I checked both callers and they degrade safely: `resubmit/page.tsx` returns `notFound()` and `study-request.ts` throws when `canResubmitStudyCode` is false — so a rejected study can no longer reach the resubmit flow. The new test locks the behavior in.

src/app/[orgSlug]/study/[studyId]/view/page.tsx
```diff
@@ -66,18 +76,37 @@ export default async function StudyReviewPage(props: {
...
             // A code decision implies the study was submitted long ago — this branch
             // should be unreachable for DRAFTs. Guard explicitly so the narrowed view
             // type holds and a corrupt row can't surface a runtime error in render.
             if (!isSubmittedStudy(study)) {
                 notFound()
             }
             const reviewingOrgName = await getOrgNameFromId(study.orgId)
+            // During the execution window the page reads as "running / results pending": keep the
+            // approved banner but drop the code listing (OTTER-598). Plain code decisions still show it.
+            const isExecuting = hasJobStatus(job.statusChanges, STUDY_CODE_RUNNING_JOB_STATUSES)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/page.tsx:109]: `isExecuting` recomputes `hasJobStatus(job.statusChanges, STUDY_CODE_RUNNING_JOB_STATUSES)` — the same predicate already folded into `decisionStatus` up on line 52. Minor, but we could compute `isExecuting` once near the top and reuse it both for the `decisionStatus` resolution and for `showStudyCode`, so the "execution window" definition lives in one spot and the two checks can't drift apart later. Not blocking.

nathanstitt (approved):
Looks great @therealmarv ty!

PR #770, opened 2026-06-04
OTTER-538 follow-up: fix do previous routing to post-code-feedback when code auto-approved

therealmarv:
see [OTTER-538 comment](https://openstax.atlassian.net/browse/OTTER-538?focusedCommentId=43317)

- Fixed the data organization reviewer's "Previous" link on the study results page so it now returns to the code feedback page instead of the initial proposal review page (the issue Malar reported on OTTER-538).
- This covers studies whose code was approved at the same time the proposal was approved, which left no separate code-review feedback on record and so previously sent the reviewer to the wrong page.
- Also fixed a similar low-probability case on this branch where the same navigation could land on a blank page when a study had no recorded feedback at all; it now falls through to the active code review page instead.
- Also handles the case where a newer baseline/IDE-init job exists after the submitted job, so the previous-link fallback still uses the submitted code job instead of skipping or looping.
- Added tests covering the corrected routing, the no-feedback fallback, and the newer baseline job case.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/page.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/page.tsx
@@ -1,6 +1,7 @@
 'use server'
 
 import { AccessDeniedAlert, AlertNotFound } from '@/components/errors'
+import type { ReviewDecision } from '@/database/types'
 import { isActionError } from '@/lib/errors'
 import { isSubmittedProposalReviewStatus } from '@/lib/proposal-review'
 import { Routes } from '@/lib/routes'
@@ -20,6 +21,12 @@ import { ProposalReviewFromAgreementsView } from './proposal-review-from-agreeme
 import { PostFeedbackView } from './post-feedback-view'
 import { ProposalReviewView } from './proposal-review-view'
 import { StudyDetailsReviewer } from './study-details-reviewer'
+
+const CODE_DECISION_TO_REVIEW_DECISION: Record<string, ReviewDecision> = {
+    'CODE-APPROVED': 'APPROVE',
+    'CODE-CHANGES-REQUESTED': 'NEEDS-CLARIFICATION',
+    'CODE-REJECTED': 'REJECT',
+}
 
 export default async function StudyReviewPage(props: {
     params: Promise<{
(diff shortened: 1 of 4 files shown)


src/app/[orgSlug]/study/[studyId]/review/page.tsx
```diff
@@ -75,15 +76,41 @@ export default async function StudyReviewPage(props: {
...
             if (codeEntries.length > 0) {
-                const job = await latestJobForStudyOrNull(studyId)
                 return <PostFeedbackView orgSlug={orgSlug} study={study} entries={codeEntries} kind="CODE" job={job} />
             }
+
+            // OTTER-538 QA: a study can reach the code-approved/results stage via proposal approval,
+            // which auto-approves the code (CODE-APPROVED) without writing a code-review comment.
+            // With no code-review rows we must still land the DO on the post-code-feedback page
+            // (approved code), not the proposal "Review initial request" fallback below.
+            const codeApprovedAt = job?.statusChanges.find((s) => s.status === 'CODE-APPROVED')?.createdAt
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/page.tsx]: This fallback only catches `CODE-APPROVED`. `isCodeDecisionStatus` (used for `decisionMade` above) also covers `CODE-CHANGES-REQUESTED` and `CODE-REJECTED` — so if a code decision other than approve ever lands here with no code-review comment (`codeEntries.length === 0`), `codeApprovedAt` is undefined and we fall through to the proposal fallback, then to active review. 

For the auto-approval path this PR targets that's correct (only APPROVE happens without a comment), so this is probably fine as-is — but worth confirming a rejected/changes-requested decision can't reach this branch comment-less. If it can, the DO would land on the wrong page again, which is the exact class of bug this PR fixes.

src/app/[orgSlug]/study/[studyId]/review/page.tsx
```diff
@@ -54,14 +54,15 @@ export default async function StudyReviewPage(props: {
...
-
         // OTTER-552: once a code-review decision has been made, opening the study (e.g. via
         // the dashboard "View" link, which carries no `from` param) must land the DO on the
         // post-feedback code page — not the active code-review/decision page. We gate on the
         // *latest* job status so a subsequent resubmission (a fresh CODE-SUBMITTED after a
         // change request) reopens active review instead of sticking on the decision page.
         const latestDecisionJob = await latestSubmittedJobForStudy(studyId)
+        const codeSubmitted =
+            studyHasJobStatus(study, 'CODE-SUBMITTED') ||
+            (latestDecisionJob?.statusChanges.some((s) => s.status === 'CODE-SUBMITTED') ?? false)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/page.tsx:72]: Nice catch widening `codeSubmitted` to also look at the job's own status history

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -208,16 +215,25 @@ function buildCrumbs({
...
+    const decision = latest?.decision ?? fallbackDecision ?? null
+    if (decision === null) {
         return null
     }
 
-    const decision = latest.decision
     const kindCopy = COPY_BY_KIND[kind]
     const decisionCopy = kindCopy.decisionCopy[decision]
     const timestampLabel = decisionCopy?.timestampLabel ?? PROPOSAL_DECISION_COPY[decision].timestampLabel
+    const timestampDate = latest?.createdAt ?? fallbackTimestamp ?? null
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Small contract nit: `decision` can come from `fallbackDecision` while `timestampDate` independently falls back to `fallbackTimestamp ?? null`. So a caller that passes `fallbackDecision` but forgets `fallbackTimestamp` renders the decision banner with no date (ProposalStepHeader just omits the timestamp line). The page always passes both together so it's fine in practice — but since they're conceptually one unit, might be worth pairing them in a single optional prop (e.g. `fallback?: { decision; timestamp }`) so the type makes the "both or neither" invariant impossible to get wrong. Non-blocking.

nathanstitt (approved):
I think this is probably fine but see the comment about only catching 'CODE-APPROVED', that might be a bug

PR #772, opened 2026-06-04
OTTER-556 follow-up: fix researcher view dead end

therealmarv:
see [OTTER-556 comments](https://openstax.atlassian.net/browse/OTTER-556?focusedCommentId=43432)

## Problem

When a data organization submits code feedback (code approved / change requested / rejected), the researcher opens the study and sees the correct code feedback page the first time. On a subsequent open or refresh, they get bounced to the post proposal submission page with no way to proceed. The study view was deciding which page to show based on the newest job on the study, and a "baseline" job (created just to mark that a workspace was opened, carrying no submitted code) could be newer than the actual reviewed submission, hiding it. A related weakness made the choice of newest job unstable when two jobs shared the same timestamp.

## Fix

Base the researcher view's routing on the latest actually-submitted job rather than the newest job of any kind, so a bare workspace-open marker can no longer hide the real submission. Separately, make the shared "newest job" lookup pick a definite job when timestamps tie, so the result is consistent across loads.

## What changed in this branch

- Made the researcher study view choose its page based on the latest submitted code, ignoring empty workspace-open jobs.
- Gave the shared latest-job lookup a stable tiebreaker so it returns the same job every time when timestamps are equal, which also benefits other pages that use it.
- Added tests confirming the code feedback page still shows for both "changes requested" and "approved" even when a newer empty workspace job exists.


src/server/db/queries.ts
```diff
@@ -86,30 +86,36 @@ export const getReviewerPublicKeyByUserId = async (userId: string) => {
...
+                        .select(['name', 'fileType', 'createdAt'])
+                        .whereRef('studyJobFile.studyJobId', '=', 'studyJob.id'),
+                ).as('files'),
+            ])
+            .where('studyJob.studyId', '=', studyId)
+            // id is a v7 (time-ordered) uuid, so it breaks createdAt ties in insertion order.
+            // Without it, two jobs sharing a createdAt (e.g. inserted in the same transaction,
+            // where now() is constant) make this LIMIT 1 non-deterministic across identical queries.
+            .orderBy('studyJob.createdAt', 'desc')
+            .orderBy('studyJob.id', 'desc')
```
nathanstitt: Heads up on a direct collision: #775 (otter-601-reuse-job-on-submit) adds this exact `.orderBy('studyJob.id', 'desc')` tiebreaker to this same `latestJobForStudyQuery`. Whichever lands second will conflict here. The change is identical and correct in both, so it's purely a merge-ordering thing — worth a quick word with whoever merges first so the second PR just drops this hunk. (Both even use the same v7-uuid rationale in the comment.)

src/app/[orgSlug]/study/[studyId]/view/page.tsx
```diff
@@ -31,17 +31,22 @@ export default async function StudyReviewPage(props: {
 
     const study = actionResult(await getStudyAction({ studyId }))
 
-    const job = await latestJobForStudyOrNull(studyId)
+    // Anchor routing on the latest *submitted* job, not the newest job row. A fresh baseline
+    // job (IDE launch / file upload, status only INITIATED) can be newer than the reviewed
+    // submission and would otherwise mask CODE-SUBMITTED / the code decision, dead-ending the
+    // researcher on the proposal/post-submission page after a refresh (OTTER-556).
+    const job = await latestSubmittedJobForStudy(studyId)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/page.tsx:38]: This `latestJobForStudyOrNull → latestSubmittedJobForStudy` swap is the reader-side defense for the same masking bug that #775 attacks from the writer side (it stops a fresh empty IDE-launch job from ever being created on top of a submission under review). These are complementary, not redundant — #775 deliberately keeps `latestSubmittedJobForStudy` precisely because a *new round* (relaunch after change-requested / results) still legitimately opens a file-less job that this routing must skip. So this swap stays load-bearing even after #775. No conflict here (#775 doesn't touch this file); just flagging the shared root cause so we don't assume one PR makes the other moot.

src/app/[orgSlug]/study/[studyId]/view/page.test.tsx
```diff
@@ -432,6 +432,60 @@ describe('StudyViewPage', () => {
         })
     })
 
+    describe('baseline-job masking (OTTER-556 refresh dead-end)', () => {
+        // A fresh baseline job (IDE launch / file upload, status only INITIATED) landing on top of
+        // the reviewed submission must not mask the code decision. Routing anchors on the latest
+        // *submitted* job, so the researcher keeps reaching the decision page instead of being
+        // bounced to the proposal / post-submission page on the next view.
+        const insertBaselineJob = async (studyId: string, createdAt: Date) => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/page.test.tsx]: This local `insertBaselineJob` helper is the same fixture #775 needs (and the OTTER-601 card calls out promoting an `insertBaselineJobOnTop` into `tests/unit.helpers.tsx`). Might be worth lifting it into shared helpers now so both PRs — and the future ones in this family — share one definition instead of re-declaring it per test file. Not blocking, just a dedup opportunity while it's fresh.

I think #775 renames the function tho

nathanstitt (approved):
looks great @therealmarv - good to see us both tackling the same issues in #775

PR #774, opened 2026-06-07
OTTER-576: deduplicate active editors by user ID

stetradis:
## Summary

Deduplicate the "also editing" list by Clerk user ID so a single person editing in multiple tabs doesn't appear as multiple collaborators, and so the current user never sees their own name in the list.

## Changes

- **`useActiveEditors`** now takes the current user's Clerk ID and dedupes awareness states by `awarenessData.userId`:
  - Skips the local Yjs client, the current Clerk user on other tabs, and any state missing a name.
  - When the same user appears more than once, keeps the focused entry if one exists.
- **`ActiveEditor`** gains a `userId` field, which is also used as the React key in `ActiveEditorsList` (replacing the unstable `name` key).
- **`CollaborativeEditor`** publishes `awarenessData: { userId }` to the collaboration plugin so peers can identify each other by Clerk ID rather than display name.
- **Provider lifecycle** reworked:
  - New `activeProvider` state + `publishProvider` callback so subscribers and the `onProviderReady` consumer stay in sync.
  - On mount, calls `provider.attach()` and re-publishes (handles React strict-mode setup/cleanup/setup without leaving subscribers stuck on `null`).
  - On unmount, calls `awareness.setLocalState(null)` and `provider.detach()` so re-entering peers get a fresh server connection with full awareness of who's already editing.

### DIFF
--- a/src/components/editable-text/collaborative-editor.tsx
+++ b/src/components/editable-text/collaborative-editor.tsx
@@ -189,7 +189,10 @@ function SaveStatus({
 
 type ActiveEditor = { userId: string; name: string; color: string; focusing: boolean }
 
-function useActiveEditors(providerRef: React.RefObject<HocuspocusProvider | null>, currentUserId: string | undefined) {
+export function useActiveEditors(
+    providerRef: React.RefObject<HocuspocusProvider | null>,
+    currentUserId: string | undefined,
+) {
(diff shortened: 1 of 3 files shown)


nathanstitt (approved):
looks great, ty!

PR #775, opened 2026-06-08
OTTER-601: reuse the round's studyJob on submit

nathanstitt:
## OTTER-601 — [Baseline (IDE-init) jobs mask the latest submitted job](https://openstax.atlassian.net/browse/OTTER-601)

Every IDE launch / file upload minted a fresh empty `studyJob` (INITIATED, no files) and every code submit minted a *second* job, so one round of work produced two job rows. The newer empty job masked the real submission anywhere "the latest job" is read — wrong pages, dead-ends, hidden controls.

A study now carries **one `studyJob` per submission round**:

- `getOrCreateCurrentRoundJob` is the single source of truth for which job in-progress work belongs to. It reuses the latest job unless that job has reached a resubmittable/terminal state (a genuine new round) or none exists.
- IDE launch (`ensureRoundJobForLaunch`) and upload (`ensureRoundJobForUpload`) reuse the round's job rather than stacking. Launch re-anchors `createdAt` only while the round is still open work, never on a live submission.
- Submit (`attachCodeToRoundJob`) fills in that same job and overwrites its code files (DB + S3) so a re-submit before review reflects exactly the files now provided. `markCodeSubmitted` records `CODE-SUBMITTED` once per round.

`latestSubmittedJobForStudy` is kept: a new round legitimately opens a file-less job that reviewer/researcher routing must skip, so the per-route swaps the card listed (C/D/E/F/G) aren't needed.

## OTTER-603 — [Code rejection also rejects the proposal page](https://openstax.atlassian.net/browse/OTTER-603)

Rejecting code now fails the job only — records `CODE-REJECTED` and leaves the study `APPROVED`, so the proposal page no longer flips to rejected.

## Test harness

Clear every test `QueryClient` after each test and disable implicit refetches, so a lingering `refetchInterval` no longer reads the per-test `CODER_FILES` env across test boundaries (fixes nondeterministic study-code failures under full-suite runs).

Full suite: 1467 tests, green across repeated runs.

### DIFF
--- a/src/server/actions/study-request.ts
+++ b/src/server/actions/study-request.ts
@@ -16,7 +16,7 @@ import {
     triggerScanForStudyJob,
 } from '@/server/aws'
 import { CODER_DISABLED, getConfigValue, SIMULATE_CODE_BUILD } from '@/server/config'
-import { nextVersionForStudyComment } from '@/server/db/mutations'
+import { getOrCreateCurrentRoundJob, nextVersionForStudyComment } from '@/server/db/mutations'
 import { getInfoForStudyId, getInfoForStudyJobId, getOrgIdFromSlug, latestJobForStudyOrNull } from '@/server/db/queries'
 import { db as database } from '@/database'
 import { deferred, onStudyReviewRequested, onStudyCodeSubmitted, onStudyCreated } from '@/server/events'
@@ -60,36 +60,42 @@ function triggerCodeScan(studyJobId: string, orgSlug: string, studyId: string) {
     }
 }
 
-async function addStudyJob(
+/**
+ * Attach the submitted code to the study's current submission round, reusing the job opened at IDE
+ * launch / file upload instead of minting a new one (OTTER-601). A new job is created only when the
+ * latest round has closed (resubmittable/terminal) — that's a genuine new submission round.
+ *
+ * The MAIN/SUPPLEMENTAL code file set is overwritten each time so a re-submit within an un-reviewed
+ * round (or after change-requested) reflects exactly the files now provided, with no leftovers from a
+ * prior attempt — in the DB and in S3.
+ */
+async function attachCodeToRoundJob(
     db: Kysely<DB>,
-    userId: string,
     studyId: string,
     orgSlug: string,
     mainCodeFileName: string,
     codeFileNames: string[],
 ) {
-    const studyJob = await db
-        .insertInto('studyJob')
-        .values({
-            studyId: studyId,
-        })
-        .returning('id')
-        .executeTakeFirstOrThrow()
+    const job = await getOrCreateCurrentRoundJob(db, studyId)
+    const studyJobId = job.id
 
-    await db
-        .insertInto('jobStatusChange')
-        .values({
-            studyJobId: studyJob.id,
-            status: 'INITIATED',
-        })
-        .executeTakeFirstOrThrow()
+    if (!job.created) {
+        // Reused round: drop any code files a prior attempt left on this job, in the DB and in S3,
+        // so a removed file does not linger in the reviewer's view.
+        await db
+            .deleteFrom('studyJobFile')
+            .where('studyJobId', '=', studyJobId)
+            .where('fileType', 'in', ['MAIN-CODE', 'SUPPLEMENTAL-CODE'])
+            .execute()
+        await deleteFolderContents(pathForStudyJobCode({ orgSlug, studyId, studyJobId }))
+    }
 
     await db
         .insertInto('studyJobFile')
         .values({
             name: mainCodeFileName,
-            path: pathForStudyJobCodeFile({ orgSlug, studyId, studyJobId: studyJob.id }, mainCodeFileName),
-            studyJobId: studyJob.id,
+            path: pathForStudyJobCodeFile({ orgSlug, studyId, studyJobId }, mainCodeFileName),
+            studyJobId,
             fileType: 'MAIN-CODE',
         })
         .executeTakeFirstOrThrow()
@@ -99,26 +105,33 @@ async function addStudyJob(
             .insertInto('studyJobFile')
             .values({
                 name: fileName,
-                path: pathForStudyJobCodeFile({ orgSlug, studyId, studyJobId: studyJob.id }, fileName),
-                studyJobId: studyJob.id,
+                path: pathForStudyJobCodeFile({ orgSlug, studyId, studyJobId }, fileName),
+                studyJobId,
                 fileType: 'SUPPLEMENTAL-CODE',
             })
             .executeTakeFirstOrThrow()
     }
 
-    const studyJobCodePath = pathForStudyJobCode({
-        orgSlug,
-        studyId,
-        studyJobId: studyJob.id,
-    })
+    // s3 signed url for client to upload
+    const urlForCodeUpload = await createSignedUploadUrl(pathForStudyJobCode({ orgSlug, studyId, studyJobId }))
 
-    // s3 signed urls for client to upload
-    const urlForCodeUpload = await createSignedUploadUrl(studyJobCodePath)
+    return { studyJobId, urlForCodeUpload }
+}
 
-    return {
-        studyJobId: studyJob.id,
-        urlForCodeUpload,
-    }
+/**
+ * Record CODE-SUBMITTED for a submission round exactly once. A re-submit within the same un-reviewed
+ * round overwrites the files but stays one submission/version, so we skip if the round's job already
+ * carries a CODE-SUBMITTED status (it may since have advanced to CODE-SCANNED, etc.).
+ */
+async function markCodeSubmitted(db: Kysely<DB>, { studyJobId, userId }: { studyJobId: string; userId: string }) {
+    const alreadySubmitted = await db
+        .selectFrom('jobStatusChange')
+        .select('id')
+        .where('studyJobId', '=', studyJobId)
+        .where('status', '=', 'CODE-SUBMITTED')
+        .executeTakeFirst()
+    if (alreadySubmitted) return
+    await db.insertInto('jobStatusChange').values({ studyJobId, userId, status: 'CODE-SUBMITTED' }).execute()
 }
 
 // Schema for creating a new draft
@@ -276,10 +289,10 @@ export const onSubmitDraftStudyAction = new Action('onSubmitDraftStudyAction', {
             throw new Error(`Cannot submit study: expected status DRAFT or APPROVED but got ${study.status}`)
         }
 
-        // Create the study job for the code files
-        const { studyJobId, urlForCodeUpload } = await addStudyJob(
+        // Attach the code to the study's current submission round. finalizeStudySubmissionAction
+        // adds the CODE-SUBMITTED status to this same job, so the two no longer diverge.
+        const { studyJobId, urlForCodeUpload } = await attachCodeToRoundJob(
             db,
-            userId,
             studyId,
             orgSlug,
             mainCodeFileName,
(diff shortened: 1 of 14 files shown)


src/server/db/mutations.ts
```diff
@@ -39,21 +42,78 @@ async function createBaselineJob(db: Kysely<DB>, studyId: string, createdAt: Dat
...
+    const latest = await db
+        .selectFrom('studyJob')
+        .select(['studyJob.id as id', 'studyJob.createdAt as createdAt'])
+        .select((eb) =>
+            eb
+                .selectFrom('jobStatusChange')
+                .select('jobStatusChange.status')
+                .whereRef('jobStatusChange.studyJobId', '=', 'studyJob.id')
+                .orderBy('jobStatusChange.createdAt', 'desc')
+                .orderBy('jobStatusChange.id', 'desc')
```
therealmarv: This `id desc` tiebreaker doesn't actually make the result deterministic. `v7uuid()` is only time-ordered to the millisecond, so two `jobStatusChange` rows inserted in the same ms can come back in either id order. That makes `latestStatus` non-deterministic, and here it drives control flow (reuse the round vs. open a new one), not just display order. This is what makes `mutations.test.ts:134` flaky (I saw it fail with `got 1, expected 2`, then pass on reruns). Could we order on something monotonic, or resolve "latest status" in a way that doesn't depend on same-ms ordering? Same determinism gap as the `distinctOn` in `study.actions.ts:55`.

src/server/db/mutations.ts
```diff
@@ -39,21 +42,78 @@ async function createBaselineJob(db: Kysely<DB>, studyId: string, createdAt: Dat
...
+                .as('latestStatus'),
+        )
+        .where('studyJob.studyId', '=', studyId)
+        .orderBy('studyJob.createdAt', 'desc')
+        .orderBy('studyJob.id', 'desc')
+        .limit(1)
+        .executeTakeFirst()
+
+    const latestStatus = (latest?.latestStatus ?? null) as StudyJobStatus | null
+    if (!latest || canResubmitStudyCode(latestStatus)) {
```
therealmarv: Opening a brand-new `INITIATED` round here whenever the latest job is resubmittable is what breaks resubmit. After `CODE-CHANGES-REQUESTED`, the researcher launches the IDE or uploads a file (both land here) → a round-2 `INITIATED` job is created → then `resubmitStudyCodeAction`'s guard at `study-request.ts:902-906` reads that raw latest job, sees `INITIATED`, and throws. Either this shouldn't open a new round until the actual submit, or the resubmit guard needs to read the latest *submitted* job. Right now launch-time and submit-time disagree about what "the current round" is.

src/server/db/mutations.ts
```diff
@@ -39,21 +42,78 @@ async function createBaselineJob(db: Kysely<DB>, studyId: string, createdAt: Dat
...
-    return createBaselineJob(db, studyId, new Date(Date.now() - 1000))
+/**
+ * IDE launch: ensure the current round has a job. When the round is still open work (no submission
+ * yet — only INITIATED), re-anchor its createdAt to now so edits made after this launch enable Submit.
+ * A round whose job already carries a submission is left untouched. Reuses the round's job rather than
+ * stacking a new one (OTTER-601).
+ */
+export async function ensureRoundJobForLaunch(db: Kysely<DB>, studyId: string): Promise<RoundJob> {
+    const job = await getOrCreateCurrentRoundJob(db, studyId)
+    if (job.created || job.latestStatus !== 'INITIATED') return job
```
therealmarv: Not re-anchoring a submitted job here is the right call (we don't want to lose the submission baseline), but it surfaces the R2/R3 gap: the submit-enable baseline is the job's `createdAt`, which was set at *launch* time, before the files were edited. So on relaunch of a submitted study, `getLastSubmissionInfoAction` returns that old `createdAt`, the files are newer, and Submit enables with no edits. The baseline that `use-ide-files.ts` compares mtimes against needs to come from a launch/session timestamp (or the latest submitted job's submit time), not this job row.

src/server/actions/study-request.ts
```diff
@@ -884,9 +912,10 @@ export const resubmitStudyCodeAction = new Action('resubmitStudyCodeAction', { p
         const sanitizedMainFileName = sanitizeFileName(mainFileName)
         const additionalFileNames = fileNames.filter((f) => f !== mainFileName).map((f) => sanitizeFileName(f))
 
-        const { studyJobId } = await addStudyJob(
+        // The canResubmitStudyCode guard above means the latest round has closed, so this opens a
```
therealmarv, line 923: The premise in this comment ("the latest round has closed, so this opens a genuinely new round") doesn't hold once the IDE has been launched or a file uploaded: `ensureRoundJobForLaunch` / `ensureRoundJobForUpload` will already have opened the new `INITIATED` round-2 job, so the guard at L902-906 sees `INITIATED` and throws before we ever get here. The guard should look at the latest *submitted* job's status, not `latestJobForStudyOrNull`.

src/server/db/mutations.test.ts
```diff
@@ -0,0 +1,185 @@
...
+
+        await ensureRoundJobForLaunch(db, study.id)
+
+        const after = await jobsForStudy(study.id)
+        expect(after).toHaveLength(1)
+        expect(after[0].id).toBe(job.id)
+        expect(after[0].createdAt.getTime()).toBe(before.createdAt.getTime())
+    })
+
+    it('opens a new round when the latest job is resubmittable', async () => {
```
therealmarv, line 169: This is the test I saw fail intermittently (`expected length 2, got 1`) under the multi-file run, passing on reruns. With `CODE-SUBMITTED` + `RUN-COMPLETE` inserted back-to-back, the two `jobStatusChange` rows can share a `createdAt` ms and the `id desc` tiebreaker in `getOrCreateCurrentRoundJob` doesn't reliably pick `RUN-COMPLETE`, so it sometimes reuses instead of opening a new round. Worth fixing the underlying ordering rather than the test, but flagging that this currently isn't deterministic.

src/server/actions/study.actions.ts
```diff
@@ -813,7 +813,20 @@ export const submitCodeReviewDecisionAction = new Action('submitCodeReviewDecisi
...
             onStudyCodeApproved({ studyId, userId })
         } else if (decision === 'reject') {
-            await performStudyCodeRejection({ db, studyId, userId })
+            // Rejecting code only fails the job, not the proposal. Record
+            // CODE-REJECTED on the job and keep the study APPROVED so the
+            // proposal page keeps showing "approved" (OTTER-603). The study
+            // was already approved (approvedAt set) before code was reviewed.
+            await db
+                .insertInto('jobStatusChange')
+                .values({ userId, status: 'CODE-REJECTED', studyJobId: claimedJob.id })
```
therealmarv, line 826: 👍 This is the right fix for OTTER-603: record `CODE-REJECTED` on the claimed job and leave `study.status = 'APPROVED'`. I confirmed the researcher code-feedback page (`view/page.tsx`) routes off the job's decision status, not `study.status`, so leaving the study APPROVED keeps the proposal page correct without breaking the rejected-code view.

therealmarv (changes requested):
Thanks @nathanstitt, the one-job-per-round model is the right direction and `getOrCreateCurrentRoundJob` reads really cleanly as a single source of truth. OTTER-603 also looks spot on. I spent most of my time in the resubmit / relaunch flows, and I think there are two real breaks plus a flaky test that we should sort out before this merges, so I'm requesting changes.

**🔴 1. Resubmit is broken once the IDE has been launched (or a file uploaded).**
`resubmitStudyCodeAction`'s guard still reads the *raw* latest job: [study-request.ts#L902-L906](https://github.com/safeinsights/management-app/blob/9b4c024a8ba95b572829d9388078e495defa00c8/src/server/actions/study-request.ts#L902-L906). But in the real resubmit flow the researcher has to put files in the workspace first, and both ways of doing that go through `getOrCreateCurrentRoundJob`, which now opens a fresh `INITIATED` round-2 job when the latest status is `CODE-CHANGES-REQUESTED`:
- launch IDE → `createUserAndWorkspaceAction` → `ensureRoundJobForLaunch`
- upload → `uploadWorkspaceFileAction` → `ensureRoundJobForUpload`

So by the time they click Resubmit, the latest status is `INITIATED`, `canResubmitStudyCode` returns false, and the action throws `Cannot resubmit study code: latest job status is INITIATED`. That's the resubmit-action + resubmit-page acceptance criteria from the card. The fix is probably to have the guard (and `getLastSubmissionInfoAction`) anchor on the latest **submitted** job rather than the raw latest. See inline note at the `attachCodeToRoundJob` call.

Worth calling out: the test that's meant to cover this ([study-request.test.ts#L855-L885](https://github.com/safeinsights/management-app/blob/9b4c024a8ba95b572829d9388078e495defa00c8/src/server/actions/study-request.test.ts#L855-L885)) passes only because it calls `resubmitStudyCodeAction` directly without launching or uploading first, which is exactly the step that creates the masking round-2 job. Can we add a launch/upload before the resubmit so the test exercises the real sequence?

**🔴 2. Submit re-enables on relaunch of a submitted study (AC violation).**
The submit-enable baseline is still the job's `createdAt`: [getLastSubmissionInfoAction](https://github.com/safeinsights/management-app/blob/9b4c024a8ba95b572829d9388078e495defa00c8/src/server/actions/workspaces.actions.ts#L142-L147) feeds [hasChangedSinceLastJob](https://github.com/safeinsights/management-app/blob/9b4c024a8ba95b572829d9388078e495defa00c8/src/hooks/use-ide-files.ts#L33-L48). The round job's `createdAt` is set at *launch* time (`T0`), the user edits a file at `T1 > T0` to enable the first submit, and `createdAt` stays at `T0` through submit. On relaunch, `ensureRoundJobForLaunch` correctly does **not** re-anchor a submitted job, so the baseline is still `T0`, the persisted files have mtime `T1 > T0`, and Submit lights up with no edits. This is the R2/R3 "split the launch/session baseline from the latest-submitted file-set comparison" item from the card. On `main` the always-fresh marker `createdAt` accidentally masked this; removing it exposes it.

**🟠 3. Flaky test (this is the one that worries me most).**
Running `./bin/docker-test -- --run 'mutations' 'study-request' 'use-code-review-mutation' 'study.actions'`, the new test `ensureRoundJobForLaunch > opens a new round when the latest job is resubmittable` failed once with `expected length 2, got 1`, then passed on 8 subsequent runs (5× file-alone, 3× the same multi-file combo). So it's genuinely intermittent, not load-only. Root cause is the latest-status subquery in `getOrCreateCurrentRoundJob` ordering by `id desc` as a tiebreaker (inline note) — `v7uuid()` isn't monotonic within a millisecond, so when two `jobStatusChange` rows land in the same ms the "latest status" is non-deterministic, and here it decides reuse-vs-open-new-round, not just display order. The PR description says "green across repeated runs", but I hit it, so I don't think the suite is actually stable yet.

**🟡 4. Two card items that don't seem to be in here yet.**
- Item J: `saveCodeResubmissionNoteDraftAction` writes the draft with no status / lab-ownership guard and no 0-row check ([study-request.ts#L843-L845](https://github.com/safeinsights/management-app/blob/9b4c024a8ba95b572829d9388078e495defa00c8/src/server/actions/study-request.ts#L843-L845)), while its proposal twin `saveProposalResubmissionNoteDraftAction` guards `status='CHANGE-REQUESTED'` AND `submittedByOrgId in userLabOrgIds`. Worth mirroring.
- Item B: the dashboard/getStudy job pick still has no `id` tiebreaker on the `distinctOn` ([study.actions.ts#L53-L58](https://github.com/safeinsights/management-app/blob/9b4c024a8ba95b572829d9388078e495defa00c8/src/server/actions/study.actions.ts#L53-L58)), so `jobStatusChanges` / `studyHasJobStatus` stays non-deterministic on same-`createdAt` jobs. You added the tiebreaker in `latestJobForStudyQuery`, just not here.

**What I really like:** OTTER-603 is clean and the `study.status` stays `APPROVED` while the code-feedback page keys off job status, so that's consistent. The `QueryClient` teardown in `vitest.setup.ts` / `unit.helpers.tsx` is a genuinely good fix and nicely commented. The `attachCodeToRoundJob` overwrite scoping to just the `/code` subfolder is correct.

tl;dr: #1 and #2 are functional bugs in the resubmit/relaunch flows that the current tests don't catch, and #3 needs to be deflaked. Once those are addressed I think this is good. Thanks again.

nathanstitt:
Thanks for the thorough pass @therealmarv — all four were on point. Status after the latest commits:

**🔴 1. Resubmit broken after launch/upload — fixed (`dbd4d4fd`).** You diagnosed this identically to the e2e failure I hit. Both guards now anchor on the latest *submitted* job: `resubmitStudyCodeAction` and `resubmit/page.tsx` use `latestSubmittedJobForStudy`, so the fresh INITIATED round job from launch/upload no longer masks the prior submission. Added the exact regression test you asked for — `resubmit succeeds after a file upload opens the new round job first` — which does `ensureRoundJobForUpload` before the resubmit. The e2e `test` job is green again.

**🔴 2. Submit re-enables on relaunch of a submitted study — fixed (`b61fe704`).** Real regression, exactly as you described: reusing the job means `createdAt` stops advancing, and on `main` the always-fresh launch job accidentally masked it. `getLastSubmissionInfoAction` now anchors the submit-enable baseline on the last submission's `CODE-SUBMITTED` time (and returns that submission's files for the set/main comparison), falling back to the round job's `createdAt` only before the first submit. Added tests for the no-submission fallback, CODE-SUBMITTED anchoring, and ignoring a newer INITIATED relaunch job.

**🟠 3. Non-deterministic round decision — fixed (`1801d174`).** Good catch, and you were right that this was real, not load-only. `created_at` defaults to `now()` (constant within a transaction), so tied statuses + non-monotonic-within-ms v7 ids made the "latest status" lookup non-deterministic. Replaced it with two order-independent `EXISTS` checks (`roundClosed` / `hasSubmission`) — sound because a resubmittable/terminal status is never followed by another status on the same job. Added a determinism test that forces two tied-`createdAt` statuses across repeated runs.

**🟡 4b. `fetchStudyQuery` distinctOn tiebreaker — fixed (`e6f95287`).** One-liner: added the `studyJob.id` tiebreaker to the `distinctOn`, mirroring `latestJobForStudyQuery`. Same determinism class as #3.

**🟡 4a. `saveCodeResubmissionNoteDraftAction` guards (item J) — deferring to its own ticket.** Confirmed real: it lacks the `status='CHANGE-REQUESTED'` + `submittedByOrgId in userLabOrgIds` + 0-row guards its proposal twin has. But it's pre-existing (this PR doesn't touch it), and it's an authz behavior change that deserves its own focused card + tests (cross-lab reject, wrong-status reject, 0-row throw) rather than riding in this reuse-job PR. I'll file it as the card's Item J.

Thanks also for the kind words on the OTTER-603 split and the QueryClient teardown. Ready for another look when you have a moment.

src/server/actions/study-request.ts
```diff
@@ -884,9 +920,10 @@ export const resubmitStudyCodeAction = new Action('resubmitStudyCodeAction', { p
...
+        // genuinely new round job (a new submission/version) rather than reusing the prior one.
```
therealmarv, line 924: this comment is misleading now.

Suggestion: remove it completely.

or

```
// Attach the resubmitted files to the current round job, creating that job only if
// launch/upload has not already opened it.
```

therealmarv (approved):
It's really nice now 🟢 Thanks. There is only one small nitpick about a misleading comment (see above).

PR #776, opened 2026-06-08
OTTER-607: scope study/job mutations to the owning lab

nathanstitt:
[OTTER-607](https://safeinsights.atlassian.net/browse/OTTER-607)

The lab-researcher rules in `permissions.ts` granted `update`/`delete` Study and `create` StudyJob unconditionally, so CASL was a no-op for those — the only thing stopping a lab member from mutating another lab's study by id was whatever ad-hoc SQL guard each handler happened to include, and some were missing it (notably the code resubmission note draft, the card's subject).

## Changes
- **permissions.ts**: scope `update`/`delete` Study and `create` StudyJob to `{ submittedByOrgId in user's lab orgs }`. `create` Study stays unconditioned — a new draft has no `submittedByOrgId` yet; the submitting lab is set in the handler from the user's own orgs.
- **saveCodeResubmissionNoteDraftAction**: add the `status = 'CHANGE-REQUESTED'` + lab SQL guard and a 0-row `ActionFailure`, mirroring its proposal twin. Previously a cross-lab member could write a draft onto another lab's study, and the client's autosave reported "saved" when nothing persisted.
- Remove the unused `onDeleteStudyJobAction` (no callers anywhere) and its orphaned `delete StudyJob` ability.

## Tests
- Cross-lab + wrong-status rejection and same-lab success for the code resubmission note draft and study delete.
- Researcher CASL scoping (own-lab allowed, other-lab denied) in `permissions.test.ts`.

All affected suites pass; lint and typecheck clean.

[OTTER-607]: https://openstax.atlassian.net/browse/OTTER-607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/permission-types.ts
+++ b/src/lib/permission-types.ts
@@ -41,7 +41,7 @@ type Abilities =
     | Ability<'OrgStudies', 'view', { orgType: 'enclave' | 'lab'; orgId?: UUID; submittedByOrgId?: UUID }>
     | Ability<'Study', 'view' | 'create', { orgId?: UUID; submittedByOrgId?: UUID }>
     | Ability<'Study', 'review' | 'approve' | 'reject' | 'update' | 'delete', { orgId?: UUID; submittedByOrgId?: UUID }>
-    | Ability<'StudyJob', 'view' | 'create' | 'delete', { orgId?: UUID; submittedByOrgId?: UUID }>
+    | Ability<'StudyJob', 'view' | 'create', { orgId?: UUID; submittedByOrgId?: UUID }>
     | Ability<'ReviewerKey', 'view' | 'update', object>
     | Ability<'Org', 'view' | 'update' | 'create' | 'delete', { orgId?: UUID; orgSlug?: string }>
     | Ability<'OrgMembers', 'view', { orgId?: UUID; orgSlug?: string }>
--- a/src/lib/permissions.test.ts
+++ b/src/lib/permissions.test.ts
@@ -41,16 +41,28 @@ test('reviewer role', () => {
 
 test('researcher role', () => {
     const { ability, session } = createAbilty({}, 'lab')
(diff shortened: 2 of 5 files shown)


therealmarv (approved):
Approving. The CASL scoping and the `saveCodeResubmissionNoteDraftAction` guard look correct, every caller of the now-scoped abilities feeds `submittedByOrgId` through its middleware, and the `onDeleteStudyJobAction` removal is clean (no callers anywhere). Nice work.

One non-blocking thing worth a follow-up, not part of this diff so leaving it as a general note:

The IDE/workspace flows still create and re-anchor jobs without the lab scoping this PR adds. `createUserAndWorkspaceAction` and `uploadWorkspaceFileAction` only `requireAbilityTo('load', 'IDE')`, and `permit('load', 'IDE')` is granted unconditionally to any lab user (permissions.ts:62 ORs with the `{ researcherId }` rule, so the unconditioned one wins). Both handlers call `ensureRoundJobForLaunch` / `ensureRoundJobForUpload` -> `getOrCreateCurrentRoundJob`, which inserts a `studyJob` + INITIATED status (and the upload path writes workspace files), with no ownership check in between. So a researcher in another lab who knows a `studyId` can still open/re-anchor an INITIATED job and drop files for someone else's study, sidestepping the `create StudyJob` scoping.

It's pre-existing and outside this change, so not a blocker here, but since OTTER-607 is "scope study/job mutations to the owning lab" it'd be good to close this door too in a follow-up, either scoping `load IDE` to `{ submittedByOrgId in usersResearcherOrgIds }` or adding an explicit lab guard in the workspace actions.

PR #780, opened 2026-06-08
Fix SWC JSX-whitespace bug: replace HTML entities in JSX text

jpslav:
## TL;DR
HTML character entities (`&apos;`, `&rsquo;`, `&#39;`, `&amp;`, …) in **JSX text** trigger a bug in SWC (our production compiler) that **drops the space between an inline element and the adjacent text**. This PR replaces those entities with literal characters, adds a lint guard so they can't come back, and adds a canary test that will tell us when the upstream SWC fix lands so we can revert the workaround.

## The symptom
On the "Have you stored your reviewer key?" modal, the production build rendered:

> **Note:**SafeInsights does not store your reviewer key…   ← no space after "Note:"

even though the source is correct:
```tsx
<b>Note:</b> SafeInsights does not store your reviewer key. If you lose your key, you won&apos;t be …
```

image

## Why this is esoteric (and why it slipped through)
The space is present in the source, and it survives **every** layer except the production compile:

| stage | keeps the space? |
|---|---|
| source on disk | ✅ (verified at the byte level) |
| Babel (what **vitest** uses to compile components) | ✅ |
| React render / Mantine SSR | ✅ |
| Sentry `reactComponentAnnotation` transform | ✅ |
| **Next SWC (`@next/swc`) — the production build** | ❌ **drops it** |

Because our unit tests compile with **Babel**, a normal render test shows the space and passes — while production (SWC) is broken. That's the trap.

## Root cause (controlled proof)
The trigger is an **HTML character entity in the adjacent text run**. Identical text, only the token varies, compiled with Next's SWC:

| token in the text run | SWC result |
|---|---|
| plain / `'` (straight) / `’` (unicode) | space preserved ✅ |
| `&apos;` | **stripped** ❌ |
| `&#39;` (numeric) | **stripped** ❌ |
| `&amp;` | **stripped** ❌ |
| `&quot;` | **stripped** ❌ |
| `&rsquo;` | **stripped** ❌ |

SWC decodes the entity **before** trimming JSX whitespace instead of after, which corrupts the adjacent-space handling. References:
- swc-project/swc#11392 (this bug) — fixed in **swc-core 1.15.11** (PR #11474)
- babel/babel#17683 (root-cause writeup: "decode after trimming, not before")
- swc-project/swc#542 (the older, *non-entity* variant — fixed long ago in v1.1.11, which is why non-entity text is fine today)

## Why this fix (and not the alternatives)
- **Not** a Next/SWC upgrade — `@next/swc` is a **version-locked vendored binary** (can't bump SWC alone), and the latest stable Next (16.2.7) doesn't include the fix yet.
- **Not** a custom build-time loader — possible (Sentry does it), but it'd need to be registered for **both** webpack and Turbopack (dev/prod parity trap) and is hidden machinery. Literal characters aren't worse source than entities anyway.

So we remove the **trigger**: literal characters in JSX text. `&apos;`/`&rsquo;` → `’` (typographic apostrophe — renders identically, and `react/no-unescaped-entities` is happy with it).

## What's in this PR
1. **19 files**: every `&apos;`/`&rsquo;` in JSX text → literal `’`. (Prettier re-flowed a few lines since `’` is shorter than `&apos;`.)
2. **ESLint guard** (`eslint.config.mjs`): `no-restricted-syntax` on `JSXText[raw=/&\w+;/]` (and numeric) — bans HTML entities in JSX text, with a message explaining the SWC bug. (Note: it matches `raw`, since typescript-eslint already entity-decodes `value`.)
3. **Regression canary** (`src/lib/swc-jsx-entity-whitespace.test.ts`): compiles a snippet with Next's SWC and asserts the **current buggy** behavior. **When `@next/swc` ships the fix, this test fails** — that's the signal to delete the workaround, the guard, and the test. A second assertion (literal ⇒ space kept) documents why the fix works.

## Verification
- `pnpm lint` (eslint + prettier) — clean, and the new guard catches any reintroduced entity repo-wide.
- The canary test passes today (bug present) and is designed to fail when SWC is fixed.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -55,6 +55,9 @@ const eslintConfig = [
             // broken in prod. Use the literal character instead (e.g. ’ ” &).
             // Remove this rule once @next/swc ships the fix; see
             // src/lib/swc-jsx-entity-whitespace.test.ts.
+            // NOTE: extend this array to add more restricted-syntax patterns — do NOT add a
+            // second 'no-restricted-syntax' key in this block, it would silently overwrite
+            // these selectors (object-key collision) and disable the entity guard with no error.
             'no-restricted-syntax': [
                 'error',
                 {
@@ -65,6 +68,8 @@ const eslintConfig = [
                         'Do not use HTML character entities (e.g. &apos;, &rsquo;, &amp;) in JSX text — they trigger an SWC whitespace bug (swc#11392). Use the literal character instead (e.g. ’ ” &).',
                 },
                 {
+                    // Numeric refs need their own selector because the named pattern above stops at
+                    // `#` (not a `\w`). `&#\w+;` covers both decimal (&#39;) and hex (&#x27;) forms.
                     selector: 'JSXText[raw=/&#\\w+;/]',
                     message:
                         'Do not use numeric HTML character references (e.g. &#39;) in JSX text — they trigger an SWC whitespace bug (swc#11392). Use the literal character instead.',
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -49,6 +49,27 @@ const eslintConfig = [
             ...reactHooksPlugin.configs.recommended.rules,
             // Disable prop-types for TypeScript projects
             'react/prop-types': 'off',
+            // Ban HTML character entities in JSX text. Our production compiler (SWC)
+            // drops the space between an inline element and adjacent JSX text when that
+            // text contains an entity (swc#11392) — invisible in unit tests (Babel),
+            // broken in prod. Use the literal character instead (e.g. ’ ” &).
+            // Remove this rule once @next/swc ships the fix; see
+            // src/lib/swc-jsx-entity-whitespace.test.ts.
+            'no-restricted-syntax': [
+                'error',
+                {
+                    // typescript-eslint exposes the source text (with entities) on `raw`;
+                    // `value` is already entity-decoded, so it must be `raw` here.
+                    selector: 'JSXText[raw=/&\\w+;/]',
+                    message:
+                        'Do not use HTML character entities (e.g. &apos;, &rsquo;, &amp;) in JSX text — they trigger an SWC whitespace bug (swc#11392). Use the literal character instead (e.g. ’ ” &).',
+                },
+                {
+                    selector: 'JSXText[raw=/&#\\w+;/]',
+                    message:
+                        'Do not use numeric HTML character references (e.g. &#39;) in JSX text — they trigger an SWC whitespace bug (swc#11392). Use the literal character instead.',
+                },
+            ],
         },
         settings: {
             react: {
--- a/src/lib/swc-jsx-entity-whitespace.test.ts
+++ b/src/lib/swc-jsx-entity-whitespace.test.ts
@@ -34,12 +34,29 @@ import { beforeAll, describe, expect, it } from 'vitest'
  */
 
 // Internal Next compiler API — no public types, but stable enough for a canary.
+// If a future Next relocates this path, require() throws here; rather than crashing
+// the whole suite at module load (which reads as "something broke" instead of
+// "investigate the canary"), capture the failure and skip with an explanatory message.
 const require = createRequire(import.meta.url)
-const swc = require('next/dist/build/swc')
+let swc: { transform: (src: string, opts: unknown) => Promise<{ code: string }>; loadBindings: () => Promise<void> }
+let swcLoadError: unknown = null
+try {
+    swc = require('next/dist/build/swc')
+} catch (err) {
+    swcLoadError = err
+    console.warn(
+        "[swc-jsx-entity-whitespace canary] Could not load 'next/dist/build/swc' — skipping. " +
+            'Next likely relocated this internal entrypoint; re-point the require above and ' +
+            `re-enable the canary. Original error: ${err instanceof Error ? err.message : String(err)}`,
+    )
+}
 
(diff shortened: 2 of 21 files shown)


eslint.config.mjs
```diff
@@ -49,6 +49,27 @@ const eslintConfig = [
...
+                'error',
+                {
+                    // typescript-eslint exposes the source text (with entities) on `raw`;
+                    // `value` is already entity-decoded, so it must be `raw` here.
+                    selector: 'JSXText[raw=/&\\w+;/]',
+                    message:
+                        'Do not use HTML character entities (e.g. &apos;, &rsquo;, &amp;) in JSX text — they trigger an SWC whitespace bug (swc#11392). Use the literal character instead (e.g. ’ ” &).',
+                },
+                {
+                    selector: 'JSXText[raw=/&#\\w+;/]',
```
nathanstitt, line 73: The two selectors are complementary — `&\w+;` matches named entities (`&apos;`, `&amp;`) but `\w` doesn't include `#`, so numeric refs need this second rule. One small note: hex refs like `&#x27;` are caught here (`#x27` is all `\w`), so coverage is good for anything anyone would actually type. No change needed — just flagging that this is "common entities," not the full HTML entity grammar.

eslint.config.mjs
```diff
@@ -49,6 +49,27 @@ const eslintConfig = [
             ...reactHooksPlugin.configs.recommended.rules,
             // Disable prop-types for TypeScript projects
             'react/prop-types': 'off',
+            // Ban HTML character entities in JSX text. Our production compiler (SWC)
+            // drops the space between an inline element and adjacent JSX text when that
+            // text contains an entity (swc#11392) — invisible in unit tests (Babel),
+            // broken in prod. Use the literal character instead (e.g. ’ ” &).
+            // Remove this rule once @next/swc ships the fix; see
+            // src/lib/swc-jsx-entity-whitespace.test.ts.
+            'no-restricted-syntax': [
```
nathanstitt, line 61: Good that this is the only `no-restricted-syntax` in the config so there's no clobber today — but that's the fragility worth a note: it's a single keyed rule, so if someone later adds a *second* restricted-syntax pattern in this same block they'll silently overwrite these two selectors (object-key collision) and the entity guard quietly stops working, with no lint error to warn them. A one-line `// extend this array; don't add a second 'no-restricted-syntax' key` could save a future debugging session.

src/lib/swc-jsx-entity-whitespace.test.ts
```diff
@@ -0,0 +1,75 @@
...
+// leading space before "SafeInsights" survived.
+const compile = async (token: string): Promise<string> => {
+    const source =
+        `const x = (<p>\n` +
+        `    <b>Note:</b> SafeInsights ${token} and the text keeps going so it\n` +
+        `    wraps onto a second line here.\n` +
+        `</p>)`
+    const { code } = await swc.transform(source, {
+        filename: 'probe.tsx',
+        jsc: { parser: { syntax: 'typescript', tsx: true }, transform: { react: { runtime: 'classic', pragma: 'h' } } },
```
nathanstitt: The probe compiles with `runtime: 'classic', pragma: 'h'`, but production builds use the automatic runtime. The whitespace trimming is in SWC's shared JSX parse/transform stage so classic-vs-automatic shouldn't change the outcome — but since the whole point of this canary is to mirror prod exactly, compiling with `runtime: 'automatic'` would be slightly more faithful and self-documenting. Minor.

src/lib/swc-jsx-entity-whitespace.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+describe('SWC JSX entity whitespace (regression canary — see file header)', () => {
+    beforeAll(async () => {
+        await swc.loadBindings()
+    })
+
+    it('canary: SWC still drops the space when the text run contains an HTML entity', async () => {
+        const code = await compile('won&apos;t store it')
+        expect(
+            code.includes(' SafeInsights'),
```
nathanstitt: `code.includes(' SafeInsights')` is a substring check over the whole compiled output, coupled to "SafeInsights" appearing verbatim. It's the load-bearing assertion that tells the next dev to delete the workaround, so a quoting/concatenation change in SWC's emit could in theory flip it. Matching the space more directly (e.g. `/Note:\s+SafeInsights/`) would make a real SWC fix harder to confuse with an output-format change.

src/lib/swc-jsx-entity-whitespace.test.ts
```diff
@@ -0,0 +1,75 @@
...
+ *   It asserts the *current, buggy* behavior. A failure means a newer
+ *   `@next/swc` has shipped the fix. At that point you can:
+ *     1. delete this test,
+ *     2. remove the `no-restricted-syntax` JSX-entity guard in eslint.config.mjs,
+ *     3. (optionally) use HTML entities in JSX text again.
+ */
+
+// Internal Next compiler API — no public types, but stable enough for a canary.
+const require = createRequire(import.meta.url)
+const swc = require('next/dist/build/swc')
```
nathanstitt: Reaching into `next/dist/build/swc` is an undocumented internal — fine for a canary, and you flagged it in the header. Resilience note: if a future Next reshuffles this path, `require` throws at module load and the whole suite errors rather than failing cleanly, which reads as "something broke" rather than "investigate the canary." Wrapping in a `try` with a skipped test + message would degrade more gracefully when the internal API moves.

nathanstitt (approved):
Thank you, this is great!  All of these comments are very minor so I'll go ahead and approve, feel free to ignore if you don't agree

PR #781, opened 2026-06-09
OTTER-546: route code review Back button to agreements page

stetradis:
## Summary

- Code review page's **Back** button now navigates to the study's agreements page (with `?from=previous`) instead of the org dashboard, matching the linear step flow Proposal → Agreements → Code Review.
- `ProposalReviewFromAgreementsView` link to agreements now also carries`?from=previous`, so reviewers who already acknowledged agreements aren't auto-redirected back to the review page.
- `CodeReviewClient` gains a required `previousHref: Route` prop, computed server-side in `code-review.tsx` via `Routes.studyAgreements({ from: 'previous'})`.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/code-review-client.test.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/code-review-client.test.tsx
@@ -1,5 +1,4 @@
 import { vi } from 'vitest'
-import type { Route } from 'next'
 import { memoryRouter } from 'next-router-mock'
 import {
     actionResult,
@@ -14,6 +13,7 @@ import {
     userEvent,
     waitFor,
 } from '@/tests/unit.helpers'
+import { Routes } from '@/lib/routes'
 import { getStudyAction, type SelectedStudy } from '@/server/actions/study.actions'
 import { latestJobForStudy } from '@/server/db/queries'
 import { CodeReviewClient } from './code-review-client'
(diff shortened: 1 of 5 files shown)


src/app/[orgSlug]/study/[studyId]/review/code-review-client.test.tsx
```diff
@@ -76,7 +78,13 @@ describe('CodeReviewClient decision selector', () => {
     it('renders all three decision options with their titles and descriptions', async () => {
         const { study, job, orgSlug } = await setupValidReviewableJob('Rice University')
         renderWithProviders(
-            <CodeReviewClient orgSlug={orgSlug} study={study} job={job} latestJobStatus="CODE-SUBMITTED" />,
+            <CodeReviewClient
+                orgSlug={orgSlug}
+                study={study}
+                job={job}
+                latestJobStatus="CODE-SUBMITTED"
+                previousHref={`/${orgSlug}/study/${study.id}/agreements?from=previous` as Route}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/code-review-client.test.tsx]: Minor: this hardcoded `` `/${orgSlug}/study/${study.id}/agreements?from=previous` as Route `` string is repeated 7 times in this file. Production builds the same value with `Routes.studyAgreements({ orgSlug, studyId, from: 'previous' })` (code-review.tsx:106) — we could reuse that helper here too, which would drop the `as Route` casts and keep the tests in lockstep if the URL shape ever changes. A single `const previousHref = ...` near the top of the describe would also cut the repetition.

src/app/[orgSlug]/study/[studyId]/review/code-review.tsx
```diff
@@ -103,6 +103,7 @@ export async function CodeReview({ orgSlug, study, entries }: CodeReviewProps) {
 
     const [review, scan] = await Promise.all([getStudyReviewForJob(job.id), jobScanResultForJob(job.id)])
     const proposalHref = `${Routes.studyReview({ orgSlug, studyId: study.id })}?from=code-review`
+    const previousHref = Routes.studyAgreements({ orgSlug, studyId: study.id, from: 'previous' })
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/code-review.tsx:106]: Nice — computing `previousHref` server-side via the typed route helper (instead of string-building in the client) keeps the `Route` type honest end-to-end. 👍

nathanstitt (approved):
looks great, only thing really is that we should be using the Route helper

PR #782, opened 2026-06-09
fix: let SI admins review studies for any org

nathanstitt:
## Problem

SI admins (who should be able to do anything, including review and approve studies) hit **"you do not have permission"** when opening a study review URL for an org they are not a member of, e.g. `/<orgSlug>/study/<id>/review`.

## Root cause — two gaps

1. **Permissions** — the `if (isSiAdmin)` block in `permissions.ts` granted only `view` actions; the review actions (`approve`/`reject`/`review` on `Study`) were missing. Those were granted solely to enclave-org *members*, so an SI admin who isn't a member of the reviewing org had no review ability.
2. **Page gates** — `review`, `agreements`, and `researcher-profile` pages gated on org **membership** (`session.orgs[orgSlug]`), which SI admins lack for arbitrary orgs — blocking them before any ability check ran.

## Fix

- **`permission-types.ts`** — added a CASL wildcard member `Ability<'all', 'manage', object>` to the `Abilities` union so `permit('manage','all')` type-checks (no `any`, no cast).
- **`permissions.ts`** — replaced the enumerated SI-admin block with `permit('manage', 'all')` (CASL's wildcard; matches every action on every subject at runtime).
- **Pages** — replaced the manual `session.orgs[orgSlug]` membership gates with ability checks (`session.can('review', toRecord('Study', { orgId }))`), matching the existing pattern in `server/auth.ts`. Researchers who land on the review URL still redirect to their study view.

## Verification

- Confirmed empirically that `('manage','all')` grants blanket access at runtime through the real `defineAbilityFor`, and that it does **not** leak to non-admins.
- Temporarily reverted the page gate to prove the new SI-admin test fails against the old membership logic (reproduces the exact "AccessDeniedAlert" symptom), then restored the fix.
- `tsc --noEmit` clean across the project; eslint + prettier clean.
- 58 tests pass across permissions / review / agreements / researcher-profile suites, including:
  - `permissions.test.ts`: SI admin cross-org review, manage/all coverage, non-leak guard.
  - `review/page.test.tsx`: SI-admin bug reproduction + non-member denial.

## Note

`permissions.ts` and route/permission changes touch authorization — reviewers should sanity-check the `manage/all` wildcard scope before merge.


jbwilson8 (approved):

therealmarv:
**Requesting one change before merge — otherwise the direction looks good.**

The permission grant (`permit('manage', 'all')`) and the page-gate conversion to ability checks are the right call. But there's a reachable path that defeats the PR's stated goal ("review studies for any org"): the action-level guard in `ackAgreementsAction` was not updated to match the new page-level authorization.

**Flow for an SI admin (a SafeInsights platform admin, granted `manage/all`) who is not a member of the reviewing enclave org, on a study with code submitted and `reviewerAgreementsAckedAt === null`:**

1. [`review/page.tsx#L161-L162`](https://github.com/safeinsights/management-app/blob/3093d83ac14d6aed809e40764312dfc5f85a7d6d/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/review/page.tsx#L161-L162) — reviewer hasn't acked, so it redirects to the agreements page.
2. [`agreements/page.tsx#L35`](https://github.com/safeinsights/management-app/blob/3093d83ac14d6aed809e40764312dfc5f85a7d6d/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/agreements/page.tsx#L35) — `isReviewer = session.can('review', ...)` is `true` (via `manage/all`), so the admin gets the reviewer agreements flow.
3. [`agreements-page.tsx#L110`](https://github.com/safeinsights/management-app/blob/3093d83ac14d6aed809e40764312dfc5f85a7d6d/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/agreements/agreements-page.tsx#L110) — proceed calls `ackAgreementsAction({ studyId, role: 'reviewer' })`.
4. [`study.actions.ts#L223-L225`](https://github.com/safeinsights/management-app/blob/3093d83ac14d6aed809e40764312dfc5f85a7d6d/src/server/actions/study.actions.ts#L223-L225) — the handler still gates on org membership:
   ```ts
   const requiredOrgId = role === 'reviewer' ? study.orgId : study.submittedByOrgId
   if (!userOrgIds.has(requiredOrgId)) {
       throw new ActionFailure({ user: `not a member of the study ${role} org` })
   }
   ```
   `session.orgs` is membership-only ([`getOrgInfoForUserId`, `queries.ts#L316`](https://github.com/safeinsights/management-app/blob/3093d83ac14d6aed809e40764312dfc5f85a7d6d/src/server/db/queries.ts#L316)). An SI admin belongs to the `safe-insights` org, not the enclave being reviewed, so `study.orgId` isn't in the set and the admin is blocked from completing review.

For proposal-stage review (no code yet) the agreements page redirects straight to review, so this only bites the code-submitted path, but that's a core part of the workflow once code exists.

**Suggested fix** — keep membership for the researcher side, allow the reviewer ack when the admin has the review ability, mirroring the page gate:
```ts
const requiredOrgId = role === 'reviewer' ? study.orgId : study.submittedByOrgId
const canReviewStudy = role === 'reviewer' && session.can('review', toRecord('Study', { orgId: study.orgId }))
if (!userOrgIds.has(requiredOrgId) && !canReviewStudy) {
    throw new ActionFailure({ user: `not a member of the study ${role} org` })
}
```
This preserves the OTTER-546 intent (the membership check still scopes dual-member test accounts on the researcher path, since the new clause only relaxes `role === 'reviewer'`).

**Test coverage note:** the existing agreements/review suites pass but don't appear to exercise the reviewer *proceed* action for a non-member admin, which is why this slipped through. Worth adding a case that calls `ackAgreementsAction({ role: 'reviewer' })` as an SI admin who isn't a member of `study.orgId`.

nathanstitt:
Good catch — fixed in 86c9d984.

`ackAgreementsAction` now mirrors the page-level ability check: the reviewer ack is allowed when the caller can `review` the study (`session.can('review', toRecord('Study', { orgId: study.orgId }))`), so an SI admin who isn't a member of the enclave can complete the code-submitted review flow. The clause is gated on `role === 'reviewer'`, so the researcher path stays membership-only and the OTTER-546 dual-member scoping is preserved.

Added a test (`sets reviewerAgreementsAckedAt when an SI admin acks as reviewer for a non-member org`) covering the exact gap. `pnpm run checks` and the ackAgreementsAction suite are green.

PR #786, opened 2026-06-09
Fix booting a fresh env in Docker

nathanstitt:
Two fixes needed to boot a fresh environment in Docker.

## 1. pnpm purge no longer prompts

Without this you occasionally get launch errors like:

```
mgmnt-app | Scope: all 2 workspace projects
mgmnt-app | [ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY] Aborted removal of modules directory due to no TTY
mgmnt-app |
mgmnt-app | If you are running pnpm in CI, set the CI environment variable to "true", or set "confirmModulesPurge" to "false".
mgmnt-app exited with code 1
```

## 2. User sync no longer references a non-existent column

On a fresh DB, the first login hit:

```
kysely:error: error: column "metadata_user_id" of relation "user" does not exist
```

The create-user path in `syncUserToDatabase` spread the full `attrs` object into `insertInto('user')`. `attrs` carries a transient `metadataUserId` field (used only for the non-prod lookup), which Kysely turned into a `metadata_user_id` column in the INSERT. Existing users hit the update branches that `.set()` explicit columns, so only the fresh-create path was broken. Fix inserts the real columns only, and adds a test covering the `metadataUserId`-set-but-no-match case.

### DIFF
--- a/bin/docker-dev-init
+++ b/bin/docker-dev-init
@@ -5,7 +5,7 @@
 
 set -e
 
-pnpm install --frozen-lockfile
+pnpm install --config.confirm-modules-purge=false --frozen-lockfile
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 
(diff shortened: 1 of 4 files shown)


chrisbendel (approved):
hit and solved this locally in my branch, thanks for getting it upstream!

PR #787, opened 2026-06-09
add docker unit test files

therealmarv:
## What this offers

A fast, self-contained Docker runner for the Vitest unit suite. `./bin/docker-test` spins up Postgres + SeaweedFS + the app container, reconciles dependencies, migrates and seeds a throwaway database, and runs the tests. Each checkout is isolated, so you can run multiple clones and their test DBs in parallel without conflicts. No local `.env`, Postgres, or S3 setup required.

## Usage (local development)

Run from the repo root:

| Command | What it does |
|---|---|
| `./bin/docker-test` | Run the whole suite |
| `./bin/docker-test -- --run PATTERN` | Run specific file(s). `PATTERN` is a **substring** match on the path, not a regex |
| `./bin/docker-test -- --run 'a' 'b'` | Multiple files: space-separated substrings (do **not** use a pipe) |
| `./bin/docker-test -- --run 'foo' -t 'REGEX'` | Filter by test name (describe/it) with regex. Pipe `\|` for OR works here (and only here) |
| `./bin/docker-test --clean` | Fresh DB + node_modules, keeps the pnpm store (fast reinstall) |
| `./bin/docker-test --down` | Stop containers, keep volumes |
| `./bin/docker-test --prune` | Delete everything for this checkout |
| `./bin/docker-test --help` | Full help |

Notes:
- **No `.env`** is needed for unit tests (Clerk is mocked and its keys are cleared; every var the tests need is baked into the compose file).
- **The DB is auto-prepared:** it is rebuilt when an applied migration was removed (rollback / branch switch), and newly added migrations are applied on each run, so `--clean` is rarely necessary.
- If a run aborts with `ERR_PNPM_OUTDATED_LOCKFILE`, your `package.json` changed without updating the lockfile: run `pnpm install` on the host, then re-run.

### Claude Code integration

Claude Code will only use this harness effectively if you tell it to (otherwise it tends to reach for host `vitest` / `pnpm test`). Two steps:

**1. Paste this block into your `CLAUDE.local.md` (personal):**

```markdown
## Running Unit Tests (dockerized)

Unit tests run inside Docker via `./bin/docker-test` (a self-contained harness:
Postgres + SeaweedFS + the app container). Do NOT run `pnpm run test` / `pnpm test`
/ `vitest` directly on the host for unit tests.

- Whole suite:      ./bin/docker-test
- Specific file(s): ./bin/docker-test -- --run PATTERN
    - PATTERN is a SUBSTRING match on the file path, NOT a regex.
    - ./bin/docker-test -- --run src/components/foo/bar.test.tsx
    - ./bin/docker-test -- --run 'bar'
    - Multiple files: space-separated substrings. Do NOT combine with a pipe
      (`--run 'a|b'` fails). Example:
      ./bin/docker-test -- --run 'positions-section' 'researcher-profile-popover'
- Filter by TEST NAME (describe/it), regex: add `-t 'REGEX'`. Pipe `|` for OR
  works here (and only here). Example:
      ./bin/docker-test -- --run 'positions-section' -t 'auto-open|hide.delete'
- Fresh DB + node_modules: ./bin/docker-test --clean (keeps the pnpm store, so the
  reinstall stays fast). Usually NOT needed: the harness auto-rebuilds the DB when an
  applied migration was removed (rollback / branch switch) and auto-applies newly
  added migrations on each run.
- No `.env` file is needed for unit tests.
- If a run aborts with `ERR_PNPM_OUTDATED_LOCKFILE`, run `pnpm install` on the host,
  then re-run.
- `./bin/docker-test --help` lists all flags.
```

**2. Allowlist the command in `.claude/settings.local.json`** so Claude runs it without a prompt each time (merge into any existing `allow` array):

```json
{
  "permissions": {
    "allow": [
      "Bash(./bin/docker-test:*)",
      "Bash(docker compose:*)"
    ]
  }
}
```

## Architecture (brief)

- `docker-compose.testunit.yml` defines `test-runner` (built from `Dockerfile.dev`), `db-unit-test` (Postgres 16), `seaweedfs-unit-test` (S3), and a one-shot bucket creator.
- `bin/docker-test` derives a per-checkout Compose project name (hash of the path) so parallel clones never collide on containers or volumes.
- Dependency install + Panda codegen live in the container **entrypoint** (not `command:`), so they run before any command, including a filtered `docker compose run` override.
- The pnpm store is co-located inside the `node_modules` volume, so pnpm hardlinks instead of copying, which keeps cold installs near-instant.
- Volumes persist between runs for speed. `--clean` wipes the DB + node_modules but keeps the store; `--prune` wipes everything.

## Benchmarks

macOS (Apple Silicon), Docker Desktop, pnpm 11.7.0, 1367 packages, 202 test files, 1840 tests.

| Mode | What | Install | Wall time |
|---|---|---|---|
| Targeted (1 file, warm) | fast dev inner-loop | 0.29s | **~6s** |
| Warm full suite | reuse all volumes | 0.28s ("Already up to date") | **44s** |
| Cold start (`--clean`) | wipe DB + node_modules, keep store | 1.5s offline (hardlink) | **63s** |
| After `--prune` | rebuild everything from network | 39.3s download | **108s** |

- Warm reinstall is effectively free (~0.28s); cold offline reinstall is ~1.5s thanks to the co-located hardlink store.
- The warm-to-cold-start delta is fresh Postgres startup + full migration replay + seed, not package downloads.


docker-compose.testunit.yml
```diff
@@ -0,0 +1,117 @@
+services:
+    test-runner:
+        build:
+            context: .
+            dockerfile: Dockerfile.dev
+        working_dir: /app
+        env_file:
+            - .env
```
nathanstitt: `env_file: - .env` will hard-fail on a fresh clone — `.env` is gitignored/untracked (only `.env.sample` is committed), and Compose errors with `env file ... not found` when the path is missing rather than skipping it. The existing `docker-compose.yml` sidesteps this by inlining `DATABASE_URL` and friends and never referencing `.env`.

maybe we make the script complain loudly if it's missing?

bin/docker-test
```diff
@@ -0,0 +1,196 @@
...
+    applied=$(docker compose -f "$COMPOSE_FILE" exec -T db-unit-test \
+        psql -U si -d si_test -t -c "SELECT name FROM kysely_migration" 2>/dev/null | tr -d ' ' | grep -v '^$' || echo "")
+
+    # Get filesystem migrations (basenames without .ts)
+    filesystem=$(ls src/database/migrations/*.ts 2>/dev/null | xargs -n1 basename | sed 's/\.ts$//')
+
+    # Check if any applied migration is missing from filesystem
+    rebuild_db=false
+    for migration in $applied; do
+        if [ -n "$migration" ] && ! echo "$filesystem" | grep -qx "$migration"; then
```
nathanstitt, line 201: Nice touch detecting the rollback case (applied migration no longer on disk → rebuild). Worth being explicit that this only catches migrations that vanished from the filesystem — the reverse and far more common case (new migration files present but not yet applied) isn't flagged here. That's fine in practice because the runner's `kysely migrate:latest` applies pending forward migrations on every run, but the usage note on line 46 ("Database is automatically rebuilt when migration mismatch is detected") reads broader than what this block does. Maybe tighten it to "rebuilt when an applied migration is missing from the codebase (e.g. after a branch switch / rollback)" so the next person doesn't expect a rebuild on forward drift.

bin/docker-test
```diff
@@ -0,0 +1,196 @@
...
+    docker compose -f "$COMPOSE_FILE" run -T --rm test-runner \
+        sh -c "pnpm install --config.store-dir=/app/node_modules/.pnpm-store --ignore-scripts --frozen-lockfile --prefer-offline --config.trustLockfile=true --config.confirmModulesPurge=false && ([ -d src/styles/generated/css ] || pnpm exec panda codegen) && pnpm exec kysely migrate:latest && pnpm exec kysely seed:run"
+fi
+
+if [ -n "$vitest_args" ]; then
+    # Exec the vitest binary directly rather than `pnpm run test:unit -- <args>`:
+    # this pnpm version drops everything after `--` instead of forwarding it to
+    # the script, so the `-- <args>` form silently ran the whole suite (no filter).
+    # `pnpm exec` passes args straight through to vitest, so file/name filters work.
+    docker compose -f "$COMPOSE_FILE" run -T --rm test-runner pnpm exec vitest $vitest_args
```
nathanstitt: `vitest_args` is built with `vitest_args="$*"` (line 78) and expanded unquoted here, so a filtered run with a spaced pattern — e.g. `bin/docker-test -- --run 'positions section'` — will word-split into two args. The single-token examples in `usage()` work, but the moment someone passes a quoted name filter it'll silently mis-parse. Capturing into an array (`vitest_args=("$@")` after the `--`, then `... pnpm exec vitest "${vitest_args[@]}"`) preserves argument boundaries. Minor since it's a dev convenience script, but cheap to get right.

nathanstitt (approved):
looks good, just a few nitpicks

therealmarv:
I'm going to close some more gaps which I've found through AI assisted self review and then merge this this week.

nathanstitt (approved):

PR #788, opened 2026-06-09
OTTER-556 follow-up: fix gaps in tests and code

therealmarv:
see [OTTER-556 comments](https://openstax.atlassian.net/browse/OTTER-556?focusedCommentId=43559)

Issue: after a reviewer requested code changes or rejected/approved code, the async scanner could still append a later CODE-SCANNED status. Because the researcher /view page trusted the latest status too directly, reopening the study could send the researcher back to the under-review/submission page instead of the decision page.

This branch fixes that by deriving the live code decision from the full submitted-job status history, so late scan results no longer mask reviewer decisions. It also adds tests for the reopen case, all code decision outcomes, lab routing, and resubmission behavior.

- Keeps researchers on the code decision page when a late scan result arrives after review feedback.
- Preserves the existing approved-and-running flow until study results are available.
- Adds coverage for reopen behavior, all code decision outcomes, and production-like lab routing.
- Adds shared decision-status coverage for late scans, tied status ordering, and resubmission cases.
- Clarifies that same-job resubmission handling is defensive historical compatibility.
- Bumps concurrently to ^10.0.3 to clear a security advisory (includes the corresponding pnpm-lock.yaml update).

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/view/page.test.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/view/page.test.tsx
@@ -2,7 +2,6 @@ import { describe, it, expect } from 'vitest'
 import * as RouterMock from 'next-router-mock'
 import {
     insertTestBaselineJob,
-    insertTestOrg,
     insertTestStudyJobData,
     insertTestStudyOnly,
     mockSessionWithTestData,
@@ -13,7 +12,7 @@ import {
     faker,
 } from '@/tests/unit.helpers'
 import { db } from '@/database'
-import type { StudyJobStatus, StudyStatus } from '@/database/types'
+import type { StudyJobStatus } from '@/database/types'
 import StudyReviewPage from './page'
 import { CodePostDecisionView } from './code-post-decision-view'
 import { CodePostSubmissionView } from './code-post-submission-view'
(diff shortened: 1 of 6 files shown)


src/lib/study-job-status.ts
```diff
@@ -60,15 +60,36 @@ export const isCodeDecisionStatus = (status: StudyJobStatus | undefined): status
...
+// For historical/defensive same-job rounds, callers pass statusChanges in newest-first DB order
+// so the first decision in the live decided history is the current round's decision.
+export const latestSubmittedJobLiveCodeDecisionStatus = (
+    statusChanges: ReadonlyArray<{ status: StudyJobStatus }>,
+): CodeDecisionStatus | null => {
+    if (!latestSubmittedJobHasLiveCodeDecision(statusChanges)) {
+        return null
+    }
+
+    const decision = statusChanges.find((s): s is { status: CodeDecisionStatus } => isCodeDecisionStatus(s.status))
```
nathanstitt, line 97: The gate (`latestSubmittedJobHasLiveCodeDecision`) is count-based precisely *because* the comment above (lines 54-59) says array order is non-deterministic — createdAt ties + non-monotonic v7 ids within a transaction. But picking *which* decision to return then falls back to `.find()`, i.e. "the first decision in array order."

In practice this is safe in the cases this is actually called with: a single decided round has exactly one decision, so position is moot; and across rounds the two decisions come from separate transactions (separate review rounds), so they don't tie on createdAt. The only spot where `.find()` order would actually matter is two *different* decision statuses coexisting on one job with a createdAt tie — which the new-job resubmission model shouldn't produce. So I think it's correct, but the function reads as self-contradictory: order-independent to decide *whether* a decision is live, order-dependent to decide *which*. A one-line note ("safe because a tie only happens within a round, which has a single decision") would close the gap for the next reader — the rest of this file is unusually careful about exactly this hazard.

src/lib/study-job-status.test.ts
```diff
@@ -52,3 +52,61 @@ describe('latestSubmittedJobHasLiveCodeDecision', () => {
...
+    it('returns null once a resubmission adds an un-decided CODE-SUBMITTED', () => {
+        // Round 1 decided (changes requested), round 2 resubmitted and awaiting review: no live decision.
+        expect(
+            latestSubmittedJobLiveCodeDecisionStatus(
+                changes('CODE-SUBMITTED', 'CODE-CHANGES-REQUESTED', 'CODE-SUBMITTED'),
+            ),
+        ).toBeNull()
+    })
+
+    it('returns a decision once a resubmission is itself decided', () => {
```
nathanstitt, line 103: This is the one case that genuinely exercises `.find()`'s position-dependence — two decisions (`CODE-APPROVED` and `CODE-CHANGES-REQUESTED`) in one history, and the assertion depends on `CODE-APPROVED` being first in the array. The "returns %s regardless of array order" test above it only ever has a single decision, so it doesn't actually prove order-independence for the multi-decision path. Could be worth a comment here noting that this case relies on newest-first ordering (and that in `page.tsx` the `hasJobStatus(['CODE-APPROVED', ...])` short-circuit would pick CODE-APPROVED anyway, so the unit-level ordering assumption never reaches the user for this particular combo).

src/app/[orgSlug]/study/[studyId]/view/page.test.tsx
```diff
@@ -202,6 +189,47 @@ describe('StudyViewPage', () => {
             .execute()
     }
 
+    const setupCrossOrgSubmittedCode = async ({
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/page.test.tsx]: `setupCrossOrgSubmittedCode` hand-rolls the study + studyJob + jobStatusChange inserts. `insertTestStudyJobData` already does most of this — is the cross-org (separate enclave + lab) split the only reason it can't be reused? If so, maybe worth a small `orgId`/`submittedByOrgId` option on the shared helper rather than a parallel insertion path that can drift from it (e.g. the `dataSources: ['all']` / `outputMimeType` defaults here vs. whatever the helper sets).

nathanstitt (approved):
I think this is fine, but would be good to use the helper as it points out below.  Not sure about the setupCrossOrgSubmittedCode but leaving in so you can evaluate

PR #789, opened 2026-06-09
Fix code review reopening after scan completes on a decided round

nathanstitt:
## Problem

A reviewer requests changes on a study's code *before* the security scan finishes. They correctly land on the post-feedback page ("You have requested changes…"). Later, once the scan completes, clicking **View** flips them to the active resubmission page ("Review study code v2.0 / has resubmitted") even though the researcher never resubmitted.

## Root cause

The source-scan CodeBuild job posts a `CODE-SUBMITTED` webhook on start (`ON_START`). When that start webhook lands *after* a reviewer has already decided the round, the spurious `CODE-SUBMITTED` gets appended to the job. That tips `latestSubmittedJobHasLiveCodeDecision` (submittedCount=2 vs decisionCount=1) back to "no live decision", so the `/review` page falls through from `PostFeedbackView` to the active `CodeReview` resubmission view.

Confirmed in the DB for the affected study: a phantom `CODE-SUBMITTED` written ~32s before `CODE-SCANNED`, after the `CODE-CHANGES-REQUESTED` row.

## Fix

Guard the `job-scan-results` webhook route so a second `CODE-SUBMITTED` is never appended when the job already has one. A job has exactly one real `CODE-SUBMITTED`, recorded at upload time before the scan is triggered — so any later echo is a duplicate and is dropped.

This fixes the bug at the layer where the bad data is written, so it holds regardless of webhook timing or scanner version, and **does not depend on an IaC deploy landing first** (per @therealmarv's review). It also hardens the "one real submission per job" invariant that `latestSubmittedJobHasLiveCodeDecision` relies on.

As defense-in-depth (not load-bearing), the scan trigger no longer emits `ON_START_PAYLOAD` (`aws.ts`). The matching IaC change to make `ON_START_PAYLOAD` optional in the scanner can now land on its own schedule rather than being a hard merge gate.

## Note

This prevents new occurrences. Studies already carrying the phantom `CODE-SUBMITTED` row are not retroactively fixed by this change.

## Tests

- `route.test.ts`: new case asserting a late `CODE-SUBMITTED` is dropped once the round has been decided.
- Updated `aws.test.ts` to drop the `ON_START_PAYLOAD` expectation.
- `aws.test.ts` and `route.test.ts` pass; typecheck + lint clean.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15,6 +15,7 @@ overrides:
   picomatch: ^4.0.4
   postcss: ^8.5.10
   uuid: ^14.0.0
+  shell-quote: ^1.8.4
 
 importers:
 
@@ -5465,8 +5466,8 @@ packages:
     resolution: {integrity: sha512-7++dFhtcx3353uBaq8DDR4NuxBetBzC7ZQOhmTQInHEd6bSrXdiEyzCvG07Z44UYdLShWUyXt5M/yhz8ekcb1A==}
     engines: {node: '>=8'}
 
-  shell-quote@1.8.3:
-    resolution: {integrity: sha512-ObmnIF4hXNg1BqhnHmgbDETF8dLPCggZWBjkQfhZpbszZnYur5DUljTcCHii5LC3J5E0yeO/1LIMyH+UvHQgyw==}
+  shell-quote@1.8.4:
+    resolution: {integrity: sha512-VsC6n6vz1ihYYyZZwX7YZSF5l5x36ca17OC+a69h94YqB7X6XLwf+5MOgynYir2SLFUbl8gIYvBo8K8RoNQ6bQ==}
     engines: {node: '>= 0.4'}
 
   si-encryption@https://codeload.github.com/safeinsights/encryption/tar.gz/2e35d3bb56f02bfc0e6d5cd9d83c54637ed0ac20:
@@ -9526,7 +9527,7 @@ snapshots:
     dependencies:
       chalk: 4.1.2
       rxjs: 7.8.2
-      shell-quote: 1.8.3
+      shell-quote: 1.8.4
       supports-color: 8.1.1
       tree-kill: 1.2.2
       yargs: 17.7.2
@@ -11959,7 +11960,7 @@ snapshots:
 
   shebang-regex@3.0.0: {}
 
-  shell-quote@1.8.3: {}
+  shell-quote@1.8.4: {}
 
   si-encryption@https://codeload.github.com/safeinsights/encryption/tar.gz/2e35d3bb56f02bfc0e6d5cd9d83c54637ed0ac20:
     dependencies:
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -19,3 +19,4 @@ overrides:
     picomatch: ^4.0.4
     postcss: ^8.5.10
     uuid: ^14.0.0
+    shell-quote: ^1.8.4
--- a/src/app/api/services/job-scan-results/route.test.ts
+++ b/src/app/api/services/job-scan-results/route.test.ts
@@ -146,6 +146,28 @@ test.skipIf(!s3Available)('stores encrypted and plaintext logs on CODE-SCANNED',
     expect(files.some((f) => f.fileType === 'SECURITY-SCAN-LOG')).toBe(true)
 })
 
+// The scanner's start webhook can echo CODE-SUBMITTED after a reviewer has already decided the
+// round. That late echo must not append a second CODE-SUBMITTED, which would reopen active review.
+test('drops a late CODE-SUBMITTED once the round has been decided', async () => {
+    const { org, user } = await mockSessionWithTestData()
+    const { jobIds } = await insertTestStudyData({ org, researcherId: user.id })
+    const jobId = jobIds[0]
+
+    await db
+        .insertInto('jobStatusChange')
+        .values([
+            { studyJobId: jobId, status: 'CODE-SUBMITTED', userId: user.id },
+            { studyJobId: jobId, status: 'CODE-CHANGES-REQUESTED', userId: user.id },
+        ])
+        .execute()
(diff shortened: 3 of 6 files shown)


therealmarv:
Reviewed this against the IaC side. It targets the correct root cause and the right layer. One blocker to flag before it merges/deploys, plus a couple of smaller notes.

## 🔴 Blocker: deploying this alone breaks every source-code scan

The deployed scanner **requires** `ON_START_PAYLOAD` and throws if it is missing:

- [`scanner.ts#L156`](https://github.com/safeinsights/iac/blob/2e0db67f8bd94bc6bc5dd9bc3ecee66ccc368c13/management-app/codebuild/scripts/scanner.ts#L156) reads it via `requireEnv('ON_START_PAYLOAD')`
- [`types.ts#L90-L96`](https://github.com/safeinsights/iac/blob/2e0db67f8bd94bc6bc5dd9bc3ecee66ccc368c13/management-app/codebuild/scripts/types.ts#L90-L96) — `requireEnv` throws `Required environment variable ... is not set` when unset
- [`scanner-buildspec.yml#L4-L6`](https://github.com/safeinsights/iac/blob/2e0db67f8bd94bc6bc5dd9bc3ecee66ccc368c13/management-app/codebuild/scanner-buildspec.yml#L4-L6) defines no default for it (only `SCAN_MODE`/`BUILD_STAGE`), so it comes solely from this app's env override
- [`scanner.ts#L179`](https://github.com/safeinsights/iac/blob/2e0db67f8bd94bc6bc5dd9bc3ecee66ccc368c13/management-app/codebuild/scripts/scanner.ts#L179) then unconditionally sends the start webhook

So once this PR stops emitting `ON_START_PAYLOAD` ([`aws.ts#L503`](https://github.com/safeinsights/management-app/blob/ff9977fbf8a6d0e6c2a21c90b3849872eeee56fc/src/server/aws.ts#L503)), the scanner throws at startup, then `main().catch` plus the buildspec `post_build` fallback both POST `ON_FAILURE_PAYLOAD` (`JOB-ERRORED`). Every submitted study's scan would report `JOB-ERRORED` instead of `CODE-SCANNED`, which is worse than the original bug. The scanner script is `s3 sync`'d at build time, so it only changes on an IaC redeploy.

**Required companion IaC change, deployed first:** make the start payload optional in [`scanner.ts#L156`](https://github.com/safeinsights/iac/blob/2e0db67f8bd94bc6bc5dd9bc3ecee66ccc368c13/management-app/codebuild/scripts/scanner.ts#L156) (`process.env.ON_START_PAYLOAD`, and skip [`scanner.ts#L179`](https://github.com/safeinsights/iac/blob/2e0db67f8bd94bc6bc5dd9bc3ecee66ccc368c13/management-app/codebuild/scripts/scanner.ts#L179) when absent), and relax `onStartPayload` in [`types.ts` `CommonEnvVars`](https://github.com/safeinsights/iac/blob/2e0db67f8bd94bc6bc5dd9bc3ecee66ccc368c13/management-app/codebuild/scripts/types.ts#L1-L16). The containerizer is unaffected: it still sends `ON_START_PAYLOAD` ([`aws.ts#L473`](https://github.com/safeinsights/management-app/blob/ff9977fbf8a6d0e6c2a21c90b3849872eeee56fc/src/server/aws.ts#L473), `JOB-PACKAGING`) and still requires it.

## 🟢 Within the app, no functional regression

In the happy path the start `CODE-SUBMITTED` was already a no-op: it arrived while the latest status was still the real `CODE-SUBMITTED`, so the route's `last.status !== body.status` guard ([`route.ts#L57-L75`](https://github.com/safeinsights/management-app/blob/ff9977fbf8a6d0e6c2a21c90b3849872eeee56fc/src/app/api/services/job-scan-results/route.ts#L57-L75)) deduped it. It only ever inserted in the race (after a decision), the bug. The real submission always records `CODE-SUBMITTED` before the scan is triggered, so dropping the echo never leaves a job without a submission status.

## 🟡 Smaller notes

- No retro-fix for studies already carrying the phantom row (noted in the PR). Existing affected studies stay wrong until the data is corrected.
- The `latestSubmittedJobHasLiveCodeDecision` count stays fragile to any future stray `CODE-SUBMITTED`. Cheap insurance that also removes the deploy-ordering risk: have the webhook drop an incoming `CODE-SUBMITTED` when the job already has one, at [`route.ts#L57-L75`](https://github.com/safeinsights/management-app/blob/ff9977fbf8a6d0e6c2a21c90b3849872eeee56fc/src/app/api/services/job-scan-results/route.ts#L57-L75). That makes this PR safe to merge before the IaC deploy and hardens the "one real submission per job" invariant.

nathanstitt:
Thanks @therealmarv — verified the blocker against the IaC source (`scanner.ts:156` `requireEnv('ON_START_PAYLOAD')` + the unconditional start webhook), and took your 🟡 suggestion as the primary fix rather than relying on deploy ordering.

The webhook route now drops an incoming `CODE-SUBMITTED` when the job already has one (the previous `last.status` check only caught an *immediately preceding* duplicate, not one sitting behind a later `CODE-CHANGES-REQUESTED` — exactly the race). So:

- This PR is **safe to merge independently of the IaC deploy** — the scanner can keep emitting its start `CODE-SUBMITTED` and the route makes it a harmless no-op.
- The `latestSubmittedJobHasLiveCodeDecision` "one real submission per job" invariant is now enforced at the write layer, not just dependent on webhook timing.
- The `aws.ts` change (dropping `ON_START_PAYLOAD`) stays as defense-in-depth / clean long-term state, but is no longer load-bearing. The companion IaC change to make `ON_START_PAYLOAD` optional can land on its own schedule.

Added a `route.test.ts` case asserting the late `CODE-SUBMITTED` is dropped once the round has been decided. On the two smaller notes: the no-retro-fix caveat still stands (existing phantom rows aren't corrected by this), and the fragility note is what the route guard now addresses.

therealmarv (approved):
nice!!!

PR #791, opened 2026-06-09
fix tests: fix flaky resubmit tests by draining deferred code scan/copy

therealmarv:
## Description

Two resubmit tests have been failing intermittently on my local machine, especially on a clean run, while passing on CI and on reruns. This fixes the timing race behind that flakiness.

- Fixes two study resubmission tests that failed intermittently due to a background timing race.
- The tests now wait for a study's background code scan to finish before recording the reviewer's decision, matching the real-world order of events.
- Adds a small reusable test helper that lets a test force pending background work to complete before continuing.

Notes:
- It is a test-only race, no product code changed. The background scan could sometimes record its status after the reviewer decision, making the app see the wrong "latest" status and the resubmit guard wrongly block.
- The fix also hardens a second sibling test that had the same latent race but happened to pass.

### DIFF
--- a/tests/vitest.setup.ts
+++ b/tests/vitest.setup.ts
@@ -51,6 +51,18 @@ mockState.setRunWithLocalStorage((cb) => {
     localStorageContext.run({ db: undefined as never }, cb)
 })
 
+// Drain the deferred callbacks scheduled so far (the `after()` work the harness collects), so a test
+// can force fire-and-forget side effects to land before continuing. Single-level by design: snapshot
+// then clear before awaiting, so callbacks scheduled *during* the drain stay queued for afterEach
+// rather than being dropped. Use when a later step depends on a deferred side effect having committed
+// (e.g. a deferred CODE-SCANNED insert must land before the test records the next status change, or
+// the time-ordered v7 ids invert and queries reading the "latest" status see the wrong row).
+export async function flushDeferred() {
+    const toRun = mockState.pendingDeferredCallbacks.slice()
+    mockState.pendingDeferredCallbacks.length = 0
+    await Promise.allSettled(toRun)
+}
+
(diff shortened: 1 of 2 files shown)


tests/vitest.setup.ts
```diff
@@ -51,6 +51,18 @@ mockState.setRunWithLocalStorage((cb) => {
     localStorageContext.run({ db: undefined as never }, cb)
 })
 
+// Drain the deferred callbacks scheduled so far (the `after()` work the harness collects), so a test
+// can force fire-and-forget side effects to land before continuing. Single-level by design: snapshot
+// then clear before awaiting, so callbacks scheduled *during* the drain stay queued for afterEach
+// rather than being dropped. Use when a later step depends on a deferred side effect having committed
+// (e.g. a deferred CODE-SCANNED insert must land before the test records the next status change, or
+// the time-ordered v7 ids invert and queries reading the "latest" status see the wrong row).
+export async function flushDeferred() {
```
nathanstitt, line 65: Nice helper, and the comment explaining the single-level snapshot-then-clear is genuinely helpful. One subtle thing worth a sentence here for future readers: this only works because the `after()` mock (`runDeferredTestCallback`) invokes the callback *synchronously* and pushes the in-flight promise — so by the time `await submitCode(...)` returns, the deferred scan promise is already in `pendingDeferredCallbacks` for us to drain. If the deferred-callback collection ever became async (e.g. queued on a microtask before pushing), `flushDeferred()` could snapshot an empty array and silently no-op. Might be worth noting that invariant so a future change to the mock doesn't quietly reintroduce the race.

nathanstitt (approved):
looks great, ty!

nathanstitt:
hah, although I may actually flip this as part of a pr I'm working on to not care about ordering of status, this might be a good test

PR #793, opened 2026-06-10
Otter 540 followup overflow redirect

Rod0352:
### Changes
- **"+N more files" now clickable** — was static text, now a dropdown listing the
  hidden code files; selecting one makes it the active file.
- **"View approved initial request" redirect** — was opening the code-review page;
  now uses `?from=initial-request` with a new `page.tsx` branch that renders the
  approved proposal (initial request) feedback view.
- **Removed legacy code-files block** — the "View the code files that you uploaded…"
  panel (Main code file + Additional files) is gone from the DO review page; not in
  Figma and duplicated the new tabbed viewer. Left intact on the researcher /view page.


src/app/[orgSlug]/study/[studyId]/review/page.tsx
```diff
@@ -66,6 +66,18 @@ export default async function StudyReviewPage(props: {
...
+        // OTTER-540: the "View approved initial request" link on the code-review page opens
+        // this route in a new tab with `from=initial-request`. It must always land on the
+        // approved *proposal* (initial request) feedback view — never the code-review page —
+        // so it sits ahead of the code-submitted/agreements branches and ignores `codeSubmitted`.
+        if (searchParams.from === 'initial-request') {
+            const proposalEntries = await getProposalFeedbackForStudyAction({ studyId })
+            if (isActionError(proposalEntries)) {
+                return <AlertNotFound title="Feedback could not be loaded" message="Please refresh and try again" />
+            }
+            return <PostFeedbackView orgSlug={orgSlug} study={study} entries={proposalEntries} />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/page.tsx]: I think this branch can render a blank page in a pretty common case. `PostFeedbackView` returns `null` when `entries` is empty and no `fallback` is passed (post-feedback-view.tsx ~L231), and the standard approve-without-written-feedback path (`performStudyProposalApproval`) writes no `studyProposalComment` row — so for an approved study with no proposal feedback, `proposalEntries` comes back `[]` and this returns an empty `PostFeedbackView`.

The "View approved initial request" link is rendered unconditionally on the submitted-code header, so any such study would open a blank tab. The existing `from=code-review` block right below (L145-151) guards against exactly this ("An empty list would blank the page… A study approved with no feedback…"). Could we mirror that here — e.g. only render when `proposalEntries.length > 0`, and otherwise fall through / show a meaningful state instead of a blank `PostFeedbackView`?

nathanstitt (commented):
seems good but need to evaluate the possible rendering of blank page

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
```diff
@@ -193,18 +194,63 @@ function FileTab({ file, isActive, onClick }: { file: CodeFile; isActive: boolea
...
+            <Menu.Target>
+                <UnstyledButton
+                    data-testid="study-code-files-overflow"
+                    px="md"
+                    py="xs"
+                    style={{ borderRadius: 0, whiteSpace: 'nowrap' }}
+                >
+                    <Group gap={4} wrap="nowrap" align="center" style={{ whiteSpace: 'nowrap' }}>
+                        <Text size="sm" c="charcoal.7" component="span">
+                            +{hidden.length} more files
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx:238]: Nice — making the overflow clickable and resolving `activeFile` against the full `files` list (line 165) means selecting a hidden file correctly drives the body, and the tests cover both the listing and the selection.

One UX nuance worth a thought (not blocking): when a reviewer picks a file from the overflow menu it becomes active, but it stays *in* the menu — it never moves into the visible tab row, and the visible tabs show no active indicator for it. The only signal that it's selected is the `data-selected` marker, which is invisible until you reopen the menu. So a reviewer reading `e.R` sees three inactive-looking tabs + "+3 more files" with no visible cue for where they are. Some viewers promote the chosen overflow item into the last visible slot to keep an active indicator on the row. Fine to leave as-is if Figma intends the menu to own the selection state, just flagging the discoverability gap.

nathanstitt (approved):
lgtm, ty!

PR #794, opened 2026-06-10
Sign full starter-code S3 key in getStarterCodeInfoAction (OTTER-369)

Rod0352:
**Bug:** Researcher "Starter code" link 404'd with S3 `NoSuchKey`.

**Cause:** `getStarterCodeInfoAction` passed the bare filename from `starterCodeFileNames` straight to `signedUrlForFile`, signing `main.R` at the bucket root. Since the `multi_starter_code` migration that column holds bare names, not full keys — the files actually live at `code-env/<org>/<envId>/starter-code/<file>`. Re-uploading never helped because the read path was requesting the wrong key.

**Fix:** Wrap the filename in `pathForStarterCode({ orgSlug, codeEnvId, fileName })` before signing, so it resolves to the same key the upload wrote to (the sibling `getStarterCodeUrlAction` already does this). Also pass `ResponseContentDisposition: 'inline'` so the file opens in-tab instead of downloading.

**Tests:** Added coverage for `getStarterCodeInfoAction` — full-key signing + the empty-starter case. Needs local Postgres: `pnpm test:unit src/server/actions/workspaces.actions.test.ts`.

### DIFF
--- a/src/server/actions/workspaces.actions.test.ts
+++ b/src/server/actions/workspaces.actions.test.ts
@@ -3,11 +3,22 @@ import {
     actionResult,
     insertTestBaselineJob,
     insertTestStudyJobData,
+    insertTestCodeEnv,
     db,
 } from '@/tests/unit.helpers'
 import { describe, expect, test, afterEach, beforeEach, vi } from 'vitest'
 import * as fs from 'node:fs/promises'
 import * as path from 'node:path'
+import { pathForStarterCode } from '@/lib/paths'
+
+// Echo the key back so a test can assert which S3 key gets signed.
+vi.mock('@/server/aws', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('@/server/aws')>()
+    return {
+        ...actual,
+        signedUrlForFile: vi.fn(async (key: string) => `https://signed.example/${key}`),
+    }
+})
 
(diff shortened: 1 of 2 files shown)


src/server/actions/workspaces.actions.test.ts
```diff
@@ -97,6 +108,39 @@ describe('Workspace Actions', () => {
...
+
+            const { getStarterCodeInfoAction } = await import('./workspaces.actions')
+            const result = actionResult(await getStarterCodeInfoAction({ studyId: study.id }))
+
+            const expectedKey = pathForStarterCode({ orgSlug: org.slug, codeEnvId: codeEnv.id, fileName: 'main.R' })
+            expect(result.starterFiles).toHaveLength(1)
+            expect(result.starterFiles[0].name).toBe('main.R')
+            expect(result.starterFiles[0].url).toContain(expectedKey)
+            // the original bug signed the bare name as the key, which resolves to the bucket root
+            expect(result.starterFiles[0].url).not.toBe('https://signed.example/main.R')
```
nathanstitt, line 129: This negative assertion is a nice touch — it pins the exact regression (signing the bare `main.R` at bucket root) rather than just the happy path, so a future refactor that reintroduces the bug fails loudly right here.

nathanstitt (approved):
nice catch! ty

PR #795, opened 2026-06-10
fix RL study-code-view

jbwilson8:
## Summary

**Issue**: [OTTER-590](https://openstax.atlassian.net/browse/OTTER-590)

When a Researcher Lead (RL) views a study whose code is in the `CODE-CHANGES-REQUESTED` state, the post-decision page rendered the code listing using `StudyCodeViewer` — the reviewer's interactive, tabbed code reader — instead of `SubmittedCodeTable`, the file table with preview/download actions used elsewhere in the researcher flow.

Root cause: `CODE-CHANGES-REQUESTED` is correctly treated as a code *decision*, so the page routes to `CodePostDecisionView`. That view hard-coded its code listing to `StudyCodeViewer`. The routing is right (the changes-requested banner, reviewer feedback section, and "Edit and resubmit" button all live in `CodePostDecisionView`) — only the code-listing widget was wrong.

This PR swaps that widget to `SubmittedCodeTable` for **all** researcher-facing decision states (approved, changes-requested, rejected), so the RL consistently sees their code as the familiar preview/download table. `StudyCodeViewer` remains in use only on the reviewer (DO) side.

## Changes

- **`code-post-decision-view.tsx`**
  - Replaced the `StudyCodeViewer` listing in `StudyCodeSection` with a collapsible `SubmittedCodeTable`, reusing the `useExpandable` + Mantine `Collapse` + toggle pattern from `code-post-submission-view.tsx`.
  - Preserved the existing UX: collapsed by default, "View / Hide submitted study code" toggle labels (`data-testid="study-code-toggle"`), and the `showStudyCode` execution-window gating.
  - Corrected the `codeFiles` prop type to the full `LatestJobForStudy['files']` row type (what `SubmittedCodeTable` requires), and removed the now-unused `StudyCodeViewer` / `CodeFile` / toggle-label imports.
- **`code-post-decision-view.test.tsx`**
  - Updated the code-listing assertions to target `submitted-code-table` (via the toggle's `aria-expanded`) instead of `study-code-viewer` / `study-code-body`.
  - The execution-window test now asserts the table and toggle are both absent when `showStudyCode={false}`.

Routing is unchanged — `page.test.tsx` still asserts `CodePostDecisionView` renders for `CODE-CHANGES-REQUESTED`.

## Before / After

| Before | After |
| --- | --- |
| Screenshot 2026-06-10 at 7 44 05 AM| Screenshot 2026-06-10 at 7 44 39 AM |

### Manual verification

As an RL, open a study whose job is `CODE-CHANGES-REQUESTED` and confirm the page shows:
- the purple "changes requested" banner,
- the reviewer feedback section,
- the "Edit and resubmit" button, and
- the **SubmittedCodeTable** (file rows with preview/download) behind the "View submitted study code" toggle.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx
@@ -1,25 +1,24 @@
 'use client'
 
-import type { FC, ReactNode } from 'react'
+import { useCallback, useState, type FC, type ReactNode } from 'react'
 import type { Route } from 'next'
-import { Box, Group, Stack, Text, Title } from '@mantine/core'
-import { CaretLeftIcon } from '@phosphor-icons/react/dist/ssr'
+import { Anchor, Box, Collapse, Group, Stack, Text, Title } from '@mantine/core'
+import { CaretLeftIcon, CaretRightIcon } from '@phosphor-icons/react/dist/ssr'
 import { AlertNotFound } from '@/components/errors'
 import { ButtonLink } from '@/components/links'
 import { PageBreadcrumbs } from '@/components/page-breadcrumbs'
 import { FeedbackAndNotesSection } from '@/components/study/feedback-and-notes'
 import { ProposalStepHeader } from '@/components/study/proposal-step-header'
-import {
-    StudyCodeViewer,
-    type StudyCodeToggleLabels,
-} from '@/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive'
-import { filterAndOrderCodeFiles, type CodeFile } from '@/app/[orgSlug]/study/[studyId]/review/study-code-files'
+import { SubmittedCodeTable } from '@/components/study/submitted-code-table'
+import { filterAndOrderCodeFiles } from '@/app/[orgSlug]/study/[studyId]/review/study-code-files'
 import { displayOrgName } from '@/lib/string'
 import { Routes } from '@/lib/routes'
 import { type Submitted } from '@/schema/study'
 import type { CodeReviewFeedbackEntry, SelectedStudy } from '@/server/actions/study.actions'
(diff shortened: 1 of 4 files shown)


src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx
```diff
@@ -154,19 +148,44 @@ const FeedbackSection: FC<{ feedbackLoadError: boolean; entries: CodeReviewFeedb
     return <FeedbackAndNotesSection entries={entries} />
 }
 
-const StudyCodeSection: FC<{ isVisible: boolean; jobId: string; codeFiles: CodeFile[] }> = ({
+function useExpandable(initial = false) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx]: `useExpandable` and the `StudyCodeToggle` + `Collapse` + `SubmittedCodeTable` block are now near-duplicates of `code-post-submission-view.tsx` (its `useExpandable` at line 33 is the same minus a `collapse` helper). Not blocking for a bugfix, but since both files live under the same `view/` dir and render the same table, extracting a shared `useExpandable` (and maybe the toggle) would keep the two from drifting — the caret-direction mismatch above is exactly the kind of drift that creeps in once they're copies.

nathanstitt (approved):
looks good, but would be good to extract a helper as suggested

src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx
Comment (jbwilson8) [reply src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx]: Moved `useExpandable` and the toggle into a shared `study-code-collapse.tsx `that both views import, so they can't drift again. That also resolves the caret mismatch — the shared toggle uses rotate(-90deg), so both pages now match. Behavior unchanged (same test-id/labels/aria-expanded).

PR #797, opened 2026-06-10
Update term Sample data term to Example Data

jbwilson8:
Issue: [OTTER-609](https://openstax.atlassian.net/browse/OTTER-609)

### Summary
Renames user-facing "Sample Data" terminology to "Example Data" in the code environment settings form to align with updated product terminology.

### Changes
- Updated the section title from **Sample Data** → **Example Data**
- Updated the file input label from **Sample Data Files** → **Example Data Files**
- Updated the file input description text in both create and edit modes:
  - Edit: "Upload new files to replace the existing example data (optional)"
  - Create: "Upload example data files for researchers (optional)"

### Files Changed
- [code-env-form.tsx](src/app/[orgSlug]/admin/settings/code-env-form.tsx)

### Notes
This is a copy/terminology-only change — no functional, schema, or behavioral changes. The underlying form field (`sampleDataPath`) is unchanged.

### Testing
- [ ] Verified the updated labels render correctly in the code environment settings form (both create and edit modes)

[OTTER-609]: https://openstax.atlassian.net/browse/OTTER-609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/admin/settings/code-env-form.tsx
+++ b/src/app/[orgSlug]/admin/settings/code-env-form.tsx
@@ -331,7 +331,7 @@ export function CodeEnvForm({ image, onCompleteAction }: CodeEnvFormProps) {
                 <Divider />
                 <Box>
                     <Title order={5} mb={4}>
-                        Sample Data
+                        Example Data
                     </Title>
                     <Text size="xs" c="dimmed" mb="sm">
                         Files available to researchers when they develop in Coder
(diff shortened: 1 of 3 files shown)


src/app/[orgSlug]/admin/settings/code-env-form.tsx
```diff
@@ -331,7 +331,7 @@ export function CodeEnvForm({ image, onCompleteAction }: CodeEnvFormProps) {
                 <Divider />
                 <Box>
                     <Title order={5} mb={4}>
-                        Sample Data
+                        Example Data
```
Comment (nathanstitt) [inline src/app/[orgSlug]/admin/settings/code-env-form.tsx:334]: Nice and tidy rename here. One consistency thing worth a look: a couple of other user-facing "sample data" strings didn't get swept up in the same pass, so the UI ends up mixing both terms:

- `code-envs.tsx:147` — the detail row label `Sample Data Path` (shown right alongside this form's values)
- `study-code-empty-view.tsx:50` — "...test it against sample data..." in the researcher empty-state copy

If OTTER-609 means to retire "Sample Data" from the UI entirely, we'd probably want those updated too. (The internal `sampleDataPath` field name is fine to leave as-is.)

nathanstitt (approved):
looks great, but Claude found a few other locations we should maybe update?

src/app/[orgSlug]/admin/settings/code-env-form.tsx
Comment (jbwilson8) [reply src/app/[orgSlug]/admin/settings/code-env-form.tsx:334]: Makes sense, thanks for pointing these references out!

PR #798, opened 2026-06-10
Add Ladle component explorer with presentational view extractions

jpslav:
## What & why

Adds **[Ladle](https://ladle.dev)** — a fast, Vite-based component explorer — so we can develop, review, and visually QA UI in isolation (design-system primitives, page layouts, modals, auth/MFA flows) without spinning up the full app, a session, or a database. Run it with `pnpm ladle`.

To make app components renderable outside Next.js, the branch also refactors ~30 components with a **container / presentational-view** split: the container keeps all hooks, session, and data logic; a new `*-view.tsx` is pure presentation and gets a `*.stories.tsx`. The split is **behavior-preserving** — containers render the same tree as before.

## How it works

**Ladle infrastructure (`/.ladle/`, dev-only):**
- A provider that mirrors `src/app/layout.tsx`'s real styling pipeline (Panda globals + Mantine layer CSS) so stories render through the *same* CSS as production.
- Shims for the Next-only APIs components import — `next/link`, `next/image`, `next/navigation`, `@sentry/nextjs`, and `@clerk/nextjs` — since Ladle runs under Vite with no Next runtime or Clerk provider. Each shim is inert (no-op navigation/telemetry/session).
- Decorators (`BrowserFrame`, `WithAppShell`) that bound `position: fixed` AppShell chrome inside a "browser" frame so it doesn't overlap Ladle's own UI.
- A `~ladle/*` path alias (tsconfig + Vite) for story imports.

**View extractions** span admin (settings, team), study (review/view), dashboard, account (sign-in, reset password, MFA/SMS), researcher profile, reviewer key, and layout (app shell, nav, profile menu). Session/data-coupled pieces are injected into views as slots, so views render standalone in Ladle while the real containers wire up the live behavior.

## Notable decisions

- **No production bundle impact:** `@ladle/react` is a `devDependency`; stories aren't reachable from the app graph; nothing in `src/` imports `~ladle/*`.
- **Drift prevention over comments:** the AppShell dimensions/colors shared by the real shell and the Ladle decorator live in one `APP_SHELL` constant (like `APP_MAIN_BG`), and a canary test (`src/app/layout.stylesheet-parity.test.ts`) fails if `layout.tsx` gains a global stylesheet the Ladle provider doesn't mirror.

## Testing

- `pnpm typecheck` ✅ and `pnpm lint` ✅.
- Unit tests pass, incl. a new `getInitials` suite (the helper is now shared by `UserAvatar` and the users table) and the stylesheet-parity canary.
- The container/view extractions are covered by the existing component tests, which exercise the same container → view render chains.
- Smoke-checked in `pnpm ladle`: AppShell-framed page stories, focused-shell auth stories, and leaf components render correctly.


therealmarv (approved):
Reviewed the Ladle setup and the container/view refactor end to end. The isolation is clean and the extractions are faithful, so approving. Three non-blocking things worth a look before merge:

**1. `workflow_dispatch` can publish non-`main` code (medium)**
[`.github/workflows/ladle-pages.yml#L20-L25`](https://github.com/safeinsights/management-app/blob/7e09bab49d8f0b5f19c253c9461c196b4d1535b4/.github/workflows/ladle-pages.yml#L20-L25)
The header comment promises Pages publishes only from `main`, but `workflow_dispatch` isn't bound by the `push` branch filter and `actions/checkout` defaults to the dispatched ref, so a manual run from any branch would deploy that branch to the public site. Not a credential hole (write access required), just an integrity gap vs. the stated model. One-line fix: gate the build job with `if: github.ref == 'refs/heads/main'`.

**2. Public Pages renders role-gated UI (medium, decision not a bug)**
[`.ladle/config.mjs#L3`](https://github.com/safeinsights/management-app/blob/7e09bab49d8f0b5f19c253c9461c196b4d1535b4/.ladle/config.mjs#L3) publishes every story, and the Clerk shim renders `Protect`/`SignedIn` children unconditionally ([`.ladle/shims/clerk.tsx#L59-L60`](https://github.com/safeinsights/management-app/blob/7e09bab49d8f0b5f19c253c9461c196b4d1535b4/.ladle/shims/clerk.tsx#L59-L60)), so the public site exposes admin/reviewer/MFA screens. No real secrets or PII in the fixtures (just `example.com`, fake UUIDs, sample passwords). Just want this to be a conscious call: document it as public UI docs, or restrict the deploy target / filter sensitive story groups.

**3. Google Fonts at runtime (super low, nitpicky)**
[`.ladle/fonts.css#L4`](https://github.com/safeinsights/management-app/blob/7e09bab49d8f0b5f19c253c9461c196b4d1535b4/.ladle/fonts.css#L4) `@import`s Open Sans from Google, so the public Pages site leaks visitor IP/UA to Google and the `ladle:build:standalone` "offline" build isn't actually offline. Self-host the woff2 or note the fallback.

**On the refactor:** spot-checked auth, study/review, admin settings, layout/nav, profile, dashboard. No prop/handler/conditional regressions. The `add-sms-mfa-view` wraps the submit handler exactly once (matches the original, not a double-wrap), `getInitials` now guards empty strings, and `submitted-code-table` guards the modal with `!!files?.length` — both harmless.

**Confirmed safe:** no production bundle impact (`@ladle/react`/`vite-plugin-singlefile` are devDeps, nothing in `src/` imports `~ladle/*`), CI permission split is correct, third-party actions are SHA-pinned, and `minimumReleaseAge` is untouched.

PR #799, opened 2026-06-10
Agent/Context file best practices from research article

chrisbendel:
[Context](https://arxiv.org/html/2602.11988v1) (haha)

I fed claude this article, asked it to apply the practices/findings from the research results to our codebase and this is what it came up with, no need to merge or integrate this but maybe can provide some small improvements the the model context (can iterate on this too)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,26 +1,17 @@
 ## Hard Rules
 
-- Strive for simplicity and clarity
 - Keep JSX minimal: No complex ternary operators, map functions, or calculations inside the return statement.
 - Move logic out: All state management, event handling, and data processing must be in custom hooks (useFeatureName) or helper functions outside the main component function.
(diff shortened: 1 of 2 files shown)


therealmarv:
I like it overall, I was worried some information is lost but it seems okay after looking through it all.

I would even remove the unit test section with usage of `pnpm` because this is user dependent (and the instructions there only work on native installation, I never installed bma directly native on my machine).

E.g. I overwrite that in my claude.local.md and unit test with docker, see docker unit test branch [here](https://github.com/safeinsights/management-app/pull/787) (need to finish it soon)

nathanstitt (approved):
we'll see how this goes, it'll be a good test.  I suspect that it'll fail to follow the conventions and have a lot of errors to fix when lint runs, but lets try it and see 👍

PR #801, opened 2026-06-11
OTTER-558 follow up: fix review copy

therealmarv:
see [OTTER-558 all comments](https://openstax.atlassian.net/browse/OTTER-558?focusedCommentId=43191)

- Fixed the "Review files" section so it shows the correct descriptive text instead of the main-file tooltip text.
- Corrected the placeholder guidance shown in the resubmission note field.
- Made the most recent feedback or note expand automatically on the code review and resubmit screens.
- Kept earlier feedback and notes collapsed unless they overflow, matching the agreed design.
- Added tests covering the expand behavior and the corrected copy.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
@@ -265,7 +265,7 @@ export function PostFeedbackView({
                     timestampLabel={timestampLabel}
                     banner={banner}
                 />
-                <FeedbackAndNotesSection entries={entries} />
+                <FeedbackAndNotesSection entries={entries} alwaysExpandLatest={isCode} />
                 <Group justify="flex-end">
                     <GoToDashboardButton />
                 </Group>
--- a/src/components/study/feedback-and-notes.test.tsx
+++ b/src/components/study/feedback-and-notes.test.tsx
@@ -1,0 +1,82 @@
+import { describe, expect, it, renderWithProviders, screen, userEvent, vi } from '@/tests/unit.helpers'
+import { lexicalJson } from '@/lib/lexical'
+import { FeedbackAndNotesSection } from './feedback-and-notes'
+
+const reviewerEntry = {
+    id: 'reviewer-v1',
+    entryType: 'REVIEWER-FEEDBACK',
+    authorName: 'Dr. Reviewer',
+    createdAt: new Date('2026-04-18T10:00:00Z'),
+    version: 1,
+    body: JSON.parse(lexicalJson('Reviewer feedback for the first round.')),
+}
+
+const noteEntry = {
+    id: 'note-v2',
+    entryType: 'RESUBMISSION-NOTE',
+    authorName: 'Dr. Researcher',
+    createdAt: new Date('2026-04-20T10:00:00Z'),
+    version: 2,
+    body: JSON.parse(lexicalJson('Addressed all the requested changes.')),
+}
+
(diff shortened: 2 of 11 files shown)


src/components/study/feedback-and-notes.tsx
```diff
@@ -123,12 +123,14 @@ function FeedbackEntry({ entry, isExpanded, onToggle }: FeedbackEntryProps) {
...
     const [expandedIds, setExpandedIds] = useState<Set<string>>(() => {
-        // Resubmission notes must be collapsed by default per spec,
-        // so only auto-expand the latest entry when it's reviewer feedback.
+        // The newest entry is expanded by default, the rest collapsed. On proposal surfaces a
+        // newest resubmission note stays collapsed (the researcher just wrote it); code surfaces
+        // opt in via alwaysExpandLatest to expand the newest entry regardless of type (OTTER-558).
         const latest = entries[0]
-        if (!latest || latest.entryType === 'RESUBMISSION-NOTE') return new Set()
+        if (!latest) return new Set()
+        if (!alwaysExpandLatest && latest.entryType === 'RESUBMISSION-NOTE') return new Set()
```
nathanstitt, line 133: Since the expanded set is computed in the `useState` initializer, it only runs on mount — if `entries[0]` or `alwaysExpandLatest` ever changed for a mounted instance, this wouldn't re-derive. On these server-rendered surfaces entries look stable per page load so this is fine in practice, just flagging it in case any of these views start swapping entries client-side.

src/components/study/feedback-and-notes.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+    entryType: 'RESUBMISSION-NOTE',
+    authorName: 'Dr. Researcher',
+    createdAt: new Date('2026-04-20T10:00:00Z'),
+    version: 2,
+    body: JSON.parse(lexicalJson('Addressed all the requested changes.')),
+}
+
+// Entries arrive newest-first from getCodeReviewFeedbackAction.
+// happy-dom has no layout, so force overflow to make the View more/less toggle render.
+const forceOverflow = () => vi.spyOn(HTMLElement.prototype, 'scrollHeight', 'get').mockReturnValue(1000)
```
nathanstitt, line 25: Nice touch documenting *why* the overflow spy is needed. Worth knowing this leans on the `lineHeight` fallback path in `FeedbackEntry`'s effect: happy-dom returns `'normal'` for lineHeight → `NaN` → it falls through to the `scrollHeight > clientHeight` branch, which the `scrollHeight=1000` mock satisfies. If that effect's measurement logic ever changes, these toggles could quietly stop rendering and the `aria-expanded` assertions would break in a non-obvious way — might be worth a one-line comment pinning that assumption.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -265,7 +265,7 @@ export function PostFeedbackView({
                     timestampLabel={timestampLabel}
                     banner={banner}
                 />
-                <FeedbackAndNotesSection entries={entries} />
+                <FeedbackAndNotesSection entries={entries} alwaysExpandLatest={isCode} />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx:268]: Good that `alwaysExpandLatest={isCode}` lines up with the `entries={isCode ? [] : ...}` guard above — code surfaces get the auto-expand, proposal surfaces keep the collapsed-note behavior. Reads consistently.

nathanstitt (approved):

PR #803, opened 2026-06-12
OTTER-573: Track proposal draft dirty state only for fields that require explicit save

stetradis:
## Summary

Ensures the proposal form's "Save as draft" button and unsaved-changes navigation prompt only react to fields that actually require an explicit save — not Lexical editor fields that auto-save to Yjs.

Previously, `form.isDirty()` treated any form change (including edits to research questions, project summary, impact, and additional notes) as unsaved draft state. Those Lexical fields persist continuously via Yjs and rehydrate on reload, so marking the form dirty after editing them incorrectly enabled "Save as draft" and triggered the leave-page prompt when there was nothing meaningful to persist to the `study` row.

This change introduces `isProposalDraftDirty()`, which checks dirty state only on draft-tracked fields: `title`, `datasets`, `piName`, and `piUserId`. The footer and `ProxyProvider` now use this helper instead of the blanket `form.isDirty()`.
Also renames the navigation modal's "Discard study" button to "Discard changes".


src/app/[orgSlug]/study/[studyId]/proposal/schema.ts
```diff
@@ -73,6 +74,15 @@ export const proposalFormSchema = z.object({
 
 export type ProposalFormValues = z.infer<typeof proposalFormSchema>
 
+// The lexical editor fields auto-save to Yjs continuously and rehydrate from it on reload,
+// so they don't trigger the "Save as draft" button or the unsaved-changes prompt.
+// These fields reach the `study` row exclusively via Save as draft / Submit
+export const DRAFT_TRACKED_FIELDS = ['title', 'datasets', 'piName', 'piUserId'] as const
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/schema.ts]: `DRAFT_TRACKED_FIELDS` mirrors `COLLAB_FIELD_KEYS` in `use-yjs-form-map.ts`, can we make a common export from lib or somewhere and share it?

src/app/[orgSlug]/study/[studyId]/proposal/footer.test.tsx
```diff
@@ -76,3 +76,50 @@ describe('ProposalFooter submit gating (OTTER-557)', () => {
         expect(submit).toBeEnabled()
     })
 })
+
+describe('ProposalFooter save-as-draft dirty tracking', () => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/footer.test.tsx:80]: Nice, clean coverage of the three meaningful cases. One gap: the tests only exercise `datasets` (enables) and `projectSummary` (stays disabled). The other tracked fields — `title`, `piName`, `piUserId` — and the negation for the other editor fields (`impact` / `additionalNotes` / `researchQuestions` staying disabled) aren't pinned, so a quiet edit to `DRAFT_TRACKED_FIELDS` wouldn't be caught. Even one parameterized `it.each` over the tracked vs. lexical field sets would lock the contract down.

nathanstitt (approved):
looks great, really only suggestion is to share the tracked fields

PR #807, opened 2026-06-14
Study fails if the code file name has parenthesis

Rod0352:
## Summary
Uploading a study's main code file with parentheses in the name (e.g. `main(1).r`) makes
the study fail at the build/packaging step — the job goes to `JOB-ERRORED` and never runs,
surfacing as `/bin/sh: 1: Syntax error: "(" unexpected` in the containerizer logs. The
file name is user-chosen at upload, so any study can hit it (any shell metacharacter does),
and the only current workaround is renaming the file first. Reproduced on Staging
(job `019d07de-9f91-7442-a239-6a75a2b798ed`).

### DIFF
--- a/src/server/aws.ts
+++ b/src/server/aws.ts
@@ -31,7 +31,7 @@ import {
     pathForStudyJobCode,
 } from '@/lib/paths'
 import type { MinimalCodeEnvInfo } from '@/lib/types'
-import { strToAscii } from '@/lib/string'
+import { shellQuote, strToAscii } from '@/lib/string'
 import { parseCsv } from '@/lib/file-content-helpers'
 import logger from '@/lib/logger'
 import { Readable } from 'stream'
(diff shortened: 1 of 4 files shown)


src/server/aws.ts
```diff
@@ -466,7 +466,11 @@ export async function buildTriggerBuildImageCommandInput(
         containerLocation: string
     },
 ) {
-    const cmd = info.cmdLine.replace('%f', info.codeEntryPointFileName)
+    // Shell-quote the filename so names with parentheses or other shell metacharacters
+    // don't break the `/bin/sh` command the containerizer runs (OTTER-477). The
+    // replacement must be a function — a string replacement would interpret `$`
+    // sequences in the filename as special replacement patterns.
+    const cmd = info.cmdLine.replace('%f', () => shellQuote(info.codeEntryPointFileName))
```
nathanstitt: Nice catch using the replacer-function form here — that's the subtle bug a string replacement would have introduced (a filename like `main$1.r` getting its `$1` interpreted as a capture-group reference), and the comment calls it out clearly.

One thing worth noting (not a blocker, and not a regression — the original `.replace('%f', ...)` had the same behavior): the string-pattern `replace` only substitutes the **first** `%f`. If a command template ever carried two `%f` tokens, the second would stay literal and break the command. The admin UI placeholder (`Rscript %f`) treats `%f` as a single token, so this is fine today — but if we want to be defensive against future multi-token templates, `replaceAll('%f', () => shellQuote(...))` covers it for free.

nathanstitt (approved):
Looks great, I do wonder tho how well the enclave will handle the `()` chars when it comes to actual running, but at least now we can get that far in the flow.

PR #813, opened 2026-06-17
OTTER-533: hack to submit code

Rod0352:
## OTTER-533: block code submission once a round is decided/has results

Going back through the flow and clicking "Proceed to Step 4" let researchers re-submit code on studies past the code phase, overwriting prior files and wiping results under review.

Root cause: study `status` stays `APPROVED` through the whole code phase, and `submitStudyCodeAction` had no job-status guard (unlike `resubmitStudyCodeAction`).

**Changes**
- `submitStudyCodeAction` rejects submissions once the latest job is past the open/under-review state (mirrors the resubmit guard). New helper `isInitialSubmissionRound` in `study-job-status.ts`.
- `code/page.tsx` redirects to study view when code is already submitted.
- `agreements/page.tsx` back-nav "Proceed" targets the study view, not the upload page.

Re-submitting an undecided round (`CODE-SUBMITTED`/`CODE-SCANNED`) still works (OTTER-601).

**Tests:** submit rejected for `CODE-APPROVED`/`RUN-COMPLETE` (study untouched), allowed for `CODE-SUBMITTED`; route + nav redirects covered.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/code/page.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/code/page.tsx
@@ -26,9 +26,8 @@ export default async function StudyCodeUploadRoute(props: { params: Promise<{ st
         redirect(Routes.studyEdit({ orgSlug, studyId }))
     }
 
-    // OTTER-533: this first-submission upload page must not be reachable once code has been submitted
-    // (e.g. via back-navigation from a later step). Re-submitting here overwrites the prior submission
-    // and wipes results under review — send the researcher to the read-only study view instead.
+    // Once code is submitted, this upload page is a dead end: re-submitting would clobber the prior
+    // submission, so send them to the study view.
     const study = await getStudyAction({ studyId })
     if (!isActionError(study) && study && studyHasJobStatus(study, 'CODE-SUBMITTED')) {
         redirect(Routes.studyView({ orgSlug, studyId }))
--- a/src/app/[orgSlug]/study/[studyId]/code/page.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/code/page.tsx
@@ -3,6 +3,9 @@
 import { Stack } from '@mantine/core'
 import { ResearcherBreadcrumbs } from '@/components/page-breadcrumbs'
 import { getDraftStudyAction } from '@/server/actions/study-request'
+import { getStudyAction } from '@/server/actions/study.actions'
+import { studyHasJobStatus } from '@/lib/studies'
+import { isActionError } from '@/lib/errors'
 import { cleanupCoderDevFiles } from '@/server/dev'
 import { redirect } from 'next/navigation'
 import { CodeUploadPage } from './code-upload'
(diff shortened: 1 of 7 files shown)


src/app/[orgSlug]/study/[studyId]/code/page.tsx
```diff
@@ -23,6 +26,13 @@ export default async function StudyCodeUploadRoute(props: { params: Promise<{ st
         redirect(Routes.studyEdit({ orgSlug, studyId }))
     }
 
+    // Once code is submitted, this upload page is a dead end: re-submitting would clobber the prior
+    // submission, so send them to the study view.
+    const study = await getStudyAction({ studyId })
+    if (!isActionError(study) && study && studyHasJobStatus(study, 'CODE-SUBMITTED')) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/code/page.tsx:32]: This page guard and the action guard use different notions of "already submitted." Here we redirect only when the latest job is `CODE-SUBMITTED`, but `submitStudyCodeAction` refuses anything outside `OPEN_SUBMISSION_JOB_STATUSES` (so also `CODE-SCANNED`, `CODE-APPROVED`, `RUN-COMPLETE`, etc.). The net effect: a researcher who navigates back to this page on a study at `CODE-APPROVED`/`RUN-COMPLETE` still sees a fully-rendered upload form, fills it out, and only hits the wall when the action throws. Could we reuse the same predicate here — e.g. drive the redirect off `!isInitialSubmissionRound(study.jobStatusChanges)` — so the page is a dead end in exactly the cases the action rejects? That keeps the two in lockstep if the status set ever shifts.

src/app/[orgSlug]/study/[studyId]/code/page.tsx
```diff
@@ -23,6 +26,13 @@ export default async function StudyCodeUploadRoute(props: { params: Promise<{ st
         redirect(Routes.studyEdit({ orgSlug, studyId }))
     }
 
+    // Once code is submitted, this upload page is a dead end: re-submitting would clobber the prior
+    // submission, so send them to the study view.
+    const study = await getStudyAction({ studyId })
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/code/page.tsx:31]: Small thing: we now call both `getDraftStudyAction` (line 19) and `getStudyAction` here, two round-trips for largely overlapping data. If `getDraftStudyAction`'s result already carries the latest job's status changes (or could), we might fold this guard into the existing fetch and save a query.

src/lib/study-job-status.ts
```diff
@@ -30,6 +30,15 @@ export const CODE_UNDER_REVIEW_JOB_STATUSES: readonly StudyJobStatus[] = ['CODE-
 export const isCodeUnderReviewStatus = (status: StudyJobStatus | undefined): boolean =>
     !!status && CODE_UNDER_REVIEW_JOB_STATUSES.includes(status)
 
+// A round is still on its first submission while every status is INITIATED or under review
+// (CODE-SUBMITTED/CODE-SCANNED); past that, the wizard submit must refuse (see submitStudyCodeAction).
+// Check every status rather than the latest: statuses written in one transaction tie on createdAt
+// and sort unreliably (see getOrCreateCurrentRoundJob).
+export const OPEN_SUBMISSION_JOB_STATUSES: readonly StudyJobStatus[] = ['INITIATED', ...CODE_UNDER_REVIEW_JOB_STATUSES]
+
+export const isInitialSubmissionRound = (statusChanges: ReadonlyArray<{ status: StudyJobStatus }>): boolean =>
```
nathanstitt, line 39: Nice — checking `.every(...)` against the whole status set (rather than just the latest change) is the right call given the createdAt-tie ordering caveat, and the comment above documenting *why* is exactly the kind of thing future readers will thank you for.

nathanstitt (changes requested):
Thanks @Rod0352 , While I think this would work, I think we're going to refactor the logic for the entire screens very soon to fix this in a more comprehensive manner.

Lets hold off on any more changes until that lands (probably tomorrow)

PR #814, opened 2026-06-17
OTTER-608: add a download icon to every code file in the review viewer

Rod0352:
### What
Re-adds the ability for DOs to download submitted code from the Code Review
screen. Each code file now has its own download icon — on every file tab and
every entry in the "+N more files" overflow menu — linking to the existing
`/dl/study-code/{jobId}/{fileName}` route.

### Implementation
Reuses the existing download idiom from `submitted-code-table-view.tsx`
(`ActionIcon` anchor + `studyCodeURL`) — no new download plumbing.

### Testing
- Added unit tests: a download link on every visible tab, and for files hidden
  in the overflow menu.
- `pnpm lint` / `prettier` clean.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
@@ -1,11 +1,12 @@
 'use client'
 
-import { Alert, Anchor, Group, Loader, Menu, Skeleton, Stack, Text, UnstyledButton } from '@mantine/core'
-import { CaretRight } from '@phosphor-icons/react/dist/ssr'
+import { ActionIcon, Alert, Anchor, Group, Loader, Menu, Skeleton, Stack, Text, UnstyledButton } from '@mantine/core'
+import { CaretRight, DownloadSimpleIcon } from '@phosphor-icons/react/dist/ssr'
 import { useState } from 'react'
 import { useQuery } from '@/common'
 import { CodeViewer } from '@/components/file-viewers'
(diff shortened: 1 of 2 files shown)


src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
```diff
@@ -286,6 +318,33 @@ function useStudyCodeFileContents(studyJobId: string, fileName: string | null) {
...
+    studyJobId: string
+    fileName: string
+    isActive?: boolean
+}) {
+    return (
+        <ActionIcon
+            component="a"
+            href={studyCodeURL(studyJobId, fileName)}
+            download={fileName}
+            onClick={(e) => e.stopPropagation()}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx:337]: The `stopPropagation` here is meant to keep a download click from switching the active file — but `Menu.Item` also closes the menu on click via Mantine's own handler (`closeOnItemClick`), which doesn't necessarily go through React bubbling that `stopPropagation` intercepts. Have we confirmed by hand that clicking the download icon in the overflow menu downloads *and* leaves the menu open ?

nathanstitt (approved):
looks good, thanks Rodney

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx:337]: Confirmed — stays open. In v8 the menu only closes through Menu.Item's React onClick

PR #818, opened 2026-06-17
OTTER-516: Add image preview to file viewers

stetradis:
## Summary

The root cause was that both reviewer + researcher code paths decoded file contents through `TextDecoder('utf-8')`, which corrupts binary data. Image files are now detected by extension and displayed in a modal using a blob URL with the correct MIME type.

## Changes
- **`src/lib/file-content-helpers.ts`** — Added `imageMimeType(path)` helper that returns the MIME type for known image extensions (png, jpg, jpeg, gif, svg, webp, bmp) or `null` for non-image files.
- **`src/components/job-results.tsx`** (researcher view) — Image files open in an `ImagePreviewModal` instead of being decoded as text and written into a new tab.
- **`src/components/encrypted-files-panel.tsx`** (reviewer view) — Image files route to `ImagePreviewModal` instead of being decoded and passed to the text-based `FilePreviewModal`.
- **`src/components/modals/image-preview-modal.tsx`** *(new)* — Modal that renders an image from a raw `ArrayBuffer` via a blob URL. Includes a download link in the title bar, matching the existing `FilePreviewModal` pattern.
- **`src/lib/file-content-helpers.test.ts`** — Tests for `imageMimeType` covering known extensions, case insensitivity, and non-image fallback.


src/components/modals/image-preview-modal.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <Group gap="md" align="baseline">
+            <span>{name}</span>
+            <DownloadBlobLink filename={name} fileContent={contents} size="sm" fw={400} />
+        </Group>
+    )
+
+    return (
+        <AppModal isOpen onClose={onClose} title={title} size="xl">
+            <Center>
+                <Image src={src} alt={name} fit="contain" mah={600} />
```
nathanstitt: **Initial empty `src`.** On first render `src=''` (line 14) before the effect runs, so `<Image src="">` mounts for a tick — Mantine treats empty src as a broken image and may flash its fallback before the blob URL lands. Could gate with `{src && <Image .../>}` or render Mantine's `<Image>` placeholder until `src` is set, to avoid the flash.

src/lib/file-content-helpers.ts
```diff
@@ -4,6 +4,21 @@ export function decodeFileContents(contents: ArrayBuffer): string {
...
+    '.jpg': 'image/jpeg',
+    '.jpeg': 'image/jpeg',
+    '.gif': 'image/gif',
+    '.svg': 'image/svg+xml',
+    '.webp': 'image/webp',
+    '.bmp': 'image/bmp',
+}
+
+export function imageMimeType(path: string): string | null {
+    const ext = path.slice(path.lastIndexOf('.')).toLowerCase()
```
nathanstitt: `path.slice(path.lastIndexOf('.'))` has a subtle edge: when there's no dot, `lastIndexOf` returns `-1` and `slice(-1)` yields the *last character* of the path rather than the whole string (e.g. `'noext'` → `'t'`). It happens to still return `null` here since no single char is in the map, and your `'noext'` test confirms that — so it's correct, just relying on a non-obvious fallthrough. A guard like `const i = path.lastIndexOf('.'); if (i < 0) return null` would make the intent explicit. (Also, a dotfile with no real extension like `.png` as a full filename would map to `image/png` — almost certainly fine for result files, just noting.)

src/components/job-results.tsx
```diff
@@ -1,32 +1,51 @@
...
 
     return (
-        <Anchor role="button" onClick={handleClick} style={{ display: 'flex', alignItems: 'center' }}>
-            View <ArrowSquareOutIcon size={16} style={{ marginLeft: 4 }} />
-        </Anchor>
+        <>
+            <Anchor role="button" onClick={handleClick} style={{ display: 'flex', alignItems: 'center' }}>
+                View <ArrowSquareOutIcon size={16} style={{ marginLeft: 4 }} />
+            </Anchor>
+            {showImageModal && mime && (
```
nathanstitt: Minor convention nit: CLAUDE.md prefers an `isVisible` prop over `{condition && <Component/>}` for gating components. Since `ImagePreviewModal` already self-manages `isOpen`, you could pass `isVisible={showImageModal}` (returning null internally) and drop the `&&`. The `mime` half of the guard is also redundant at the point of click — `handleClick` only sets `showImageModal` true when `mime` is truthy, so `showImageModal` implies `mime`. Tiny thing; the current form works.

nathanstitt (approved):
looks good, just a few tweaks I think. ty!

PR #819, opened 2026-06-19
OTTER-573: remove proxy provider

stetradis:
## Summary

Removes the `ProxyProvider` component and its three usages in the study proposal flow.

`ProxyProvider` wrapped form pages to intercept navigation when the form was dirty, prompting the user with a modal that offered to save a draft before navigating away or discard changes. With autosave now in place across these forms, the unsaved-changes guard is redundant.

## Changes

- Deleted `src/components/proxy-provider.tsx` (~200 lines)
- Removed `<ProxyProvider>` wrapper from:
  - `src/app/[orgSlug]/study/[studyId]/proposal/form.tsx` — also drops the now-unused `saveDraft`, `isSaving`, and `isProposalDraftDirty` references
  - `src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/form.tsx` — also drops the unused `form`, `saveDraft`, `isSaving` destructures from `useEditResubmit`
  - `src/app/[orgSlug]/study/request/proposal.tsx` — also removes the `onNavigateAway`/`reset()` wiring

PR #821, opened 2026-06-22
OTTER-497 follow-up: make change-requested proposal resubmit collaborative

therealmarv:
## Summary

Makes the **change-requested proposal resubmit** flow (`/study/[id]/edit-and-resubmit`) fully collaborative, bringing it to parity with the proposal-draft flow. This closes the remaining OTTER-497 gap reported in QA: two same-lab researchers editing a change-requested proposal could not see each other's edits live, and when one submitted, the others were neither redirected nor shown a toast.

## Background

The draft proposal flow (`/study/[id]/proposal`, `ProposalProvider`) shipped collaborative (Y.js + Hocuspocus), but the change-requested resubmit flow was a separate non-collaborative fork (`EditResubmitProvider`): plain Mantine form state with debounced HTTP autosave, `EditableText`, no stateless broadcast, and no kick-out. So co-editing, the submit broadcast, and the redirect/toast were all missing for `CHANGE-REQUESTED`.

Two facts kept this change small:
- The editor service already permits collaborative editing of proposal docs in `CHANGE-REQUESTED` (`services/editor/auth.ts` `isDocumentEditable`), so no server-side auth change was needed.
- The existing `proposal-submitted` stateless event validates against the `proposal-fields` doc with study status `PENDING-REVIEW`. Since `resubmitProposalAction` already transitions `CHANGE-REQUESTED -> PENDING-REVIEW`, broadcasting after it commits reuses that event verbatim. No new event type and no `services/editor` changes.

## Changes

- **Shared collaboration hook**: extracted `tabSessionId` + `useYjsWebsocket` + `useYjsFormMap` into `src/hooks/use-proposal-collaboration.ts`, now consumed by both `ProposalProvider` and `EditResubmitProvider` so the two flows can't drift apart again.
- **Collaborative resubmit editing**: `EditResubmitProvider` exposes `websocketProvider`/`yjsForm`/`tabSessionId`; the edit section uses `CollaborativeProposalTextField` and pushes scalar fields (title, datasets, PI) into the shared Y.js form map.
- **Submit broadcast + kick-out**: `useResubmitProposal` broadcasts `proposal-submitted` after the action commits; the resubmit form mounts `useSubmissionRedirectListener` + `StudyKickOutProvider` (editable status `['CHANGE-REQUESTED']`) so peers get the immediate toast/redirect plus a reconnect safety net.
- **Server action**: `resubmitProposalAction` now returns `submitterFullName` + `orgName` for the toast and deletes stale `proposal-${studyId}-%` Y.js docs after the status flip (in-tx + deferred purge), mirroring `finalizeStudySubmissionAction` so a future change-requested reopen seeds from the canonical study columns rather than stale CRDT.

## Scope / tradeoffs

- The **resubmission note** stays last-write-wins (existing debounced autosave). The proposal fields and text are live-collaborative; the note box is intentionally not yet, pending a separate decision. No editor-service work is pulled in by this PR.
- No changes to `src/lib/permissions.ts` or route definitions; existing same-lab guards already cover access.

## Tests

- New `use-resubmit-proposal.test.ts`: broadcasts `proposal-submitted` on successful resubmit, does not broadcast on validation failure, and does not broadcast when the study can no longer be resubmitted.
- Extended `resubmit-proposal.test.ts`: action returns `submitterFullName` + `orgName`, and deletes stale `proposal-*` Y.js docs on resubmit.
- `pnpm run checks` passes (typecheck + lint + action validation).

## Manual QA

Two browser contexts, same lab, study status `CHANGE-REQUESTED`:
1. User A and B open `/edit-and-resubmit`. A edits title, datasets, PI, and the proposal text fields. B sees the updates live.
2. A clicks Resubmit. B gets the toast and is redirected to the submitted page; B cannot keep editing (immediate via stateless event, and after reconnect via kick-out).
3. A different-lab user and a DO reviewer still get `notFound()`.
4. The resubmission-note box is the one field that does not sync live (expected).


src/contexts/edit-resubmit/hooks/use-resubmit-proposal.ts
```diff
@@ -1,34 +1,61 @@
...
+                    studyId,
+                    studyInfo: buildStudyInfo(form.getValues()),
+                    resubmissionNote: noteForm.values.resubmissionNote,
+                }),
+            ),
+        onSuccess: (result) => {
             form.resetDirty()
             noteForm.resetDirty()
+            const submittedByClerkId = user?.id
+            if (!submittedByClerkId) {
```
nathanstitt: why would user id ever be missing?  they must be logged in to use the app, right?

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/form.tsx
```diff
@@ -27,34 +34,50 @@ export const EditResubmitForm: FC<EditResubmitFormProps> = ({
...
-                />
-            </Stack>
-        </ProxyProvider>
+        <StudyKickOutProvider
+            studyId={studyId}
+            orgSlug={orgSlug}
+            editableStatuses={RESUBMIT_EDITABLE_STATUSES}
+            redirectTarget="studySubmitted"
+        >
+            <ProxyProvider isDirty={form.isDirty()} onSaveDraft={saveDraft} isSavingDraft={isSaving}>
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/form.tsx]: Heads-up on a likely merge conflict: PR #819 ("remove proxy provider") deletes `ProxyProvider` and `src/components/proxy-provider.tsx` outright, including this exact usage, on the grounds that autosave makes the dirty-guard redundant. This PR keeps it and nests it inside the new `StudyKickOutProvider`. Whichever lands second will conflict here — and there's a design question worth settling between the two: now that this flow is live-collaborative (fields sync to Yjs continuously), is the `ProxyProvider` "save draft before leaving?" prompt still meaningful for the proposal fields, or does it only still matter for the last-write-wins resubmission note? Might be worth coordinating with that PR's author so the two don't undo each other.

src/server/actions/study-request.ts
```diff
@@ -806,11 +807,45 @@ export const resubmitProposalAction = new Action('resubmitProposalAction', { per
...
+        // Metadata for the `proposal-submitted` stateless broadcast: peers still
+        // editing get a toast naming the submitter and the reviewing org. Mirrors
+        // finalizeStudySubmissionAction's return.
+        const submitter = await db
+            .selectFrom('user')
+            .select(['fullName'])
+            .where('id', '=', userId)
+            .executeTakeFirstOrThrow()
+
+        const reviewerOrg = await db
```
nathanstitt: We re-`selectFrom('study').innerJoin('org')` just to fetch the org name, but `study` was already loaded at the top of the handler and `orgSlug` is in scope. Not wrong (it mirrors `finalizeStudySubmissionAction`), but since `org.name` is the only new field a direct `selectFrom('org').where('slug','=',orgSlug)` — or folding it into the earlier study select — would save a round-trip. Minor.

nathanstitt (approved):
looks good, just a few notes

src/contexts/edit-resubmit/hooks/use-resubmit-proposal.ts
therealmarv: Good point, fixed. `resubmitProposalAction` now selects `clerkId` alongside `fullName` and returns `submitterClerkId`, and the hook reads that instead of `useUser()`. The confusing `if (!submittedByClerkId)` fallback (which only ever existed to narrow `user?.id`) is gone.

src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/form.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/form.tsx]: Good catch, this resolved itself. #819 merged first (2026-06-22) and this branch has since pulled `origin/main`, so `ProxyProvider` and `src/components/proxy-provider.tsx` are already gone here, no remaining references in `src`. The form now wraps with `StudyKickOutProvider` only. And agreed on the design question: with the fields live-collaborative and autosaved, the "save draft before leaving?" guard is no longer meaningful, so dropping it is the right call.

src/server/actions/study-request.ts
therealmarv: Folded it into the earlier `study` select (which already filters by `study.id`) via an `innerJoin('org', ...)`, so the separate round-trip is gone. I kept the join on `study.orgId` rather than the `slug = orgSlug` shortcut, since `orgSlug` here is the lab slug, not the reviewing org, so it would have named the wrong org in the toast.

PR #822, opened 2026-06-22
OTTER-558 follow-up: Fix Code Edit & Resubmit Cancel button

therealmarv:
# OTTER-558 follow-up: Fix the Code "Edit & Resubmit" Cancel button

## Summary

QA found four defects in the Code "Edit & Resubmit" flow ([comment 43809](https://openstax.atlassian.net/browse/OTTER-558?focusedCommentId=43809), [comment 43968](https://openstax.atlassian.net/browse/OTTER-558?focusedCommentId=43968)). After merging the study-screen state machine (#826) into this branch, three of those four are resolved upstream and only the Cancel-button fix remains here.

## Why this branch shrank

This branch originally carried four fixes. #826 introduced a study-screen state machine plus OTTER-601 job-round semantics, where a `CODE-CHANGES-REQUESTED` resubmit reuses the same job instead of opening a fresh `INITIATED` round job. That removed the shared root cause (a baseline job masking the prior decision), so on re-evaluation against current `main`:

| Original defect | Outcome after merging main |
|---|---|
| 1. Autosave crash (`saveCodeResubmissionNoteDraftAction` gated on `study.status`) | Already fixed in main: the action now gates on `latestSubmittedJobForStudy` + `canResubmitStudyCode`. Reverted to main's version. |
| 2. DO dashboard shows "Proposal · Approved" instead of the code decision | Obsolete. Round reuse means no masking job is created; the dashboard intentionally tracks the latest job. The branch's `isSubmitted` query tiebreaker contradicted that design and was reverted. |
| 3. Reopen after Save & exit lands on `/code` with feedback gone | Resolved by the state machine: a `CODE-CHANGES-REQUESTED` study routes the researcher to `/view` -> `code-feedback` (feedback + resubmit). The branch's `useStudyHref` (deleted in main) and the "Code · Draft" pill enhancement were dropped to align with main. |
| 4. Footer shows "Save and exit" instead of "Cancel" before any edit | **Fixed here.** This is the only change that remains. |

## Defect fixed

| # | Symptom | Root cause |
|---|---------|-----------|
| 4 | The footer shows "Save and exit" instead of "Cancel" even before any edit | `filesChanged` compares workspace file mtimes to the last submission and is already true on initial load |

## Changes

`use-ide-files` now tracks a `userEditedFiles` signal that flips only on a real session edit (upload, delete, or main-file pick). The resubmit footer's Cancel-vs-Save-and-exit toggle keys on this (`filesEdited` prop) instead of the mtime-based `filesChanged`, which still drives submit-enable on the initial `/code` page and is untouched.

## Test coverage

- `use-ide-files.test.ts` (real DB + real workspace, no mocking of our own modules): `userEditedFiles` is false on load and flips on real edits.
- `edit-study-code-footer.test.tsx`: Cancel shows with no session edits, Save and exit shows once edited.

## Scope notes

- Net diff over `main` is the Cancel-button fix only (`use-ide-files.ts` + test, `edit-study-code-footer.tsx` + test, `edit-study-code-view.tsx`).
- No migrations, permission-rule, or route-definition changes.

## Verification

- `pnpm run checks` passes (type checking, ESLint, Prettier, action validation).
- Unit tests:
  ```
  ./bin/docker-test -- --run 'edit-study-code-footer' 'use-ide-files'
  ```
- Manual (as a Research Lab user): open Resubmit from a results-ready / code-errored / code-change-requested study; confirm Cancel shows before any edit and becomes Save and exit after uploading a file.


src/server/actions/study.actions.ts
```diff
@@ -52,8 +52,27 @@ function fetchStudyQuery(db: DBExecutor) {
...
+                                    .selectFrom('jobStatusChange')
+                                    .select('jobStatusChange.id')
+                                    .whereRef('jobStatusChange.studyJobId', '=', 'studyJob.id')
+                                    .where('jobStatusChange.status', '!=', 'INITIATED'),
+                            )
+                            .as('isSubmitted'),
+                    )
                     .distinctOn('studyId')
                     .orderBy('studyId')
+                    .orderBy('isSubmitted', 'desc')
```
nathanstitt: The `isSubmitted desc` tiebreaker is a clean way to un-mask the decision, and the comment explains the masking scenario really well. One thing worth confirming: this ordering now applies to *every* study the dashboard returns, not just code-resubmit ones — so any study that happens to have an older submitted job plus a newer INITIATED-only job will now surface the older job's status changes instead of the newest job's. For the code-resubmit flow that's exactly the fix, but it'd be reassuring to have a non-code regression case (or a note that no other flow produces a newer INITIATED-only job alongside an older submitted one) so we know this can't shift a pill somewhere unexpected. The three new tests cover the code path and the baseline fallback nicely; a proposal-phase-with-jobs case would close the loop.

src/server/actions/study-request.ts
```diff
@@ -829,16 +829,29 @@ export const saveCodeResubmissionNoteDraftAction = new Action('saveCodeResubmiss
...
+        // OTTER-558: code resubmission runs while the study is still APPROVED — the
+        // resubmittable state lives on the JOB (CODE-CHANGES-REQUESTED / RUN-COMPLETE /
+        // FILES-APPROVED / FILES-REJECTED / JOB-ERRORED), never on study.status. Gating on
+        // study.status = 'CHANGE-REQUESTED' (the proposal-phase value) made every autosave
+        // claim 0 rows and throw "Study is not editable". Gate on the latest *submitted*
+        // job instead, mirroring resubmit/page.tsx's render gate. Use .some() (not
+        // resubmitStudyCodeAction's .at(0)): a late CODE-SCANNED row can sort ahead of the
+        // decision, and the autosave guard must be at least as permissive as the page gate
+        // or the page renders while every keystroke errors.
+        const canResubmit = (await latestSubmittedJobForStudy(studyId))?.statusChanges.some((s) =>
```
nathanstitt: Really like that the guard is explicitly reasoned to be *at least as permissive as the page gate* — the `.some()` vs `.at(0)` note about a late `CODE-SCANNED` row sorting ahead of the decision is exactly the kind of subtlety that bites silently. Since this now intentionally mirrors `resubmit/page.tsx`'s render gate (same `latestSubmittedJobForStudy` + `canResubmitStudyCode` shape), it might be worth extracting that into a single shared helper (e.g. `canResubmitStudyCodeNow(studyId)`) so the page gate and the autosave guard can't drift apart later — the whole bug class here was two seams that were *supposed* to agree but didn't. Not blocking, just guarding against the next divergence.

src/server/actions/study.actions.ts
```diff
@@ -78,6 +97,12 @@ function fetchStudyQuery(db: DBExecutor) {
                     .select(['orgDataSource.id', 'orgDataSource.name'])
                     .where(sql<boolean>`"org_data_source"."id"::text = ANY("study"."datasets")`),
             ).as('orgDataSources'),
+            // OTTER-558: a non-empty code resubmission note draft means the RL has an in-progress
+            // "code draft" (saved via Save & exit, not yet resubmitted). Surface only the boolean —
+            // reviewer dashboards receive this row too and must not see the draft note content.
+            sql<boolean>`coalesce(btrim(${eb.ref('study.codeResubmissionNoteDraft')}) <> '', false)`.as(
```
nathanstitt: Nice detail using `coalesce(btrim(...) <> '', false)` so a whitespace-only draft counts as empty and the boolean is never null — and only exposing the boolean (never the draft text) to reviewer dashboards is the right call. 👍

nathanstitt (changes requested):
I think this is good, but lets hold off until #826 is merged because I think it does part of the needed logic

src/app/[orgSlug]/study/[studyId]/resubmit/edit-study-code-footer.tsx
```diff
@@ -33,7 +35,7 @@ export const EditStudyCodeFooter: FC<EditStudyCodeFooterProps> = ({
 
     const isBusy = isSaving || isSubmitting
     const exitTarget = Routes.studyView({ orgSlug, studyId })
-    const hasChanges = noteForm.values.resubmissionNote.length > 0 || filesChanged
+    const hasChanges = noteForm.values.resubmissionNote.length > 0 || filesEdited
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/resubmit/edit-study-code-footer.tsx]: The note form is seeded from a persisted draft (`EditCodeResubmitProvider` initializes `resubmissionNote: initialNote`). 

So if the user saved a resubmission-note draft in an earlier session and reopens, `hasChanges` is true on load and the footer shows "Save and exit" before any *session* edit, the same symptom this PR is fixing, just via the note path instead of the files path. 

Is that intended (a saved-but-unsubmitted note is itself worth a Save-and-exit), or should the note also distinguish "edited this session" from "loaded from draft"?

src/hooks/use-ide-files.ts
```diff
@@ -160,6 +168,7 @@ export function useIDEFiles({ studyId, onSubmitSuccess }: UseIDEFilesOptions) {
 
     const uploadFiles = useCallback(
         (filesToUpload: File[]) => {
+            setUserEditedFiles(true)
```
nathanstitt, line 171: Small edge case: `setUserEditedFiles(true)` fires optimistically here (and in `removeFile`) before the mutation resolves. If the upload/delete server action errors, the workspace file set is unchanged but `userEditedFiles` stays `true`, so the footer would show "Save and exit" for a change that didn't actually happen. Probably acceptable (a failed upload is rare and the user did *attempt* an edit), but if we wanted it tight we could flip the flag in the mutation's `onSuccess`/`onSettled` instead. Not blocking.

nathanstitt (approved):
Looks good, ty!

PR #825, opened 2026-06-22
OTTER-613: Show full Submitted code section on the post-decision Data Partner code page

therealmarv:
### Summary
Once a study reached a code decision (code changes requested, code approved, or code rejected), the Data Partner's study code page degraded to a bare, unstructured code dump behind the "View full study code" toggle. It dropped the AI summary, the security scan log results, and the dataset listing that reviewers rely on. This PR makes the post-decision page render the same full "Submitted code" section shown during active review, with the raw code still collapsed behind its toggle.

### Problem
As a Data Partner (enclave org), opening a study with status `CODE-CHANGES-REQUESTED`, `CODE-APPROVED`, or `CODE-REJECTED` and clicking "View full study code" showed only the file tabs and code body. The AI-generated summary and the security scan log were unreachable, and the page looked broken (raw code with no surrounding structure). This is inconsistent with the active review page, which shows the complete section.

### Root cause
After a decision, the reviewer code-feedback view renders `PostFeedbackView` (kind=CODE). Its `CodeSection` rendered a bare `StudyCodeViewer` (file tabs + code only), whereas the active-review page (`code-review.tsx`) renders the full `SubmittedCodeSection` (datasets, AI summary, security scan log, and the code viewer). The post-decision path never fetched or passed the `review` and `scan` data the full section needs.

### Fix
- `review/post-feedback-view.tsx`: `CodeSection` now renders the decision banner step header plus the full `SubmittedCodeSection` as siblings (mirroring the active-review layout), instead of a stripped-down `StudyCodeViewer`. The raw code stays collapsed via `codeInitiallyExpanded={false}`; the AI summary and scan log render up front. Extracted a `SubmittedCodePanel` that returns null when the job/scan is unavailable.
- `_screens/reviewer-code-feedback-screen.tsx`: the post-decision CODE screen now fetches `getStudyReviewForJob(job.id)` and `jobScanResultForJob(job.id)` alongside the job, and passes `review`/`scan` to both `PostFeedbackView` call sites (the code-review-entries path and the no-comment fallback path), so all three decision statuses get the full section.

### Behavior after the change
For all three post-decision statuses the Data Partner now sees: the decision banner, dataset pills, the AI summary (collapsible), the security scan log result, the "View approved initial request" link, and the code viewer collapsed behind "View full study code".

### Testing
- Updated `post-feedback-view.test.tsx`: the test that asserted the old stripped-down behavior now asserts the full section renders (AI summary + security scan log + submitted-code section) with the code body still collapsed behind the toggle, and that the toggle expands it.
- `pnpm run checks` (typecheck + lint + action validation) passes.

### Notes for reviewers
The section rendering is decision-independent (only the banner copy/color differs per decision, which is covered by the existing banner tests), so the render test covers the APPROVE case as representative. The fallback path reuses the same fetched `review`/`scan` values, so it is covered by the same component test.

### Rebased onto the study-screen state machine (PR #826)
This branch was merged with `main` after PR #826 (study-screen state machine; elimination of `?from=` routing) landed. That refactor deleted the `?from=` cascade in `review/page.tsx` (now pure state-machine dispatch) and moved the post-decision CODE rendering into the new `_screens/reviewer-code-feedback-screen.tsx` adapter.

The fix was reevaluated against the new architecture and is still required: the component change in `post-feedback-view.tsx` (and its test) merged unchanged, and the `review`/`scan` fetch that previously lived in `page.tsx` was relocated into the `reviewer-code-feedback-screen.tsx` adapter, which is the new home for the post-decision CODE view. The original `page.tsx`/`page.test.tsx` edits were superseded by the refactor and dropped.


src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -138,30 +140,70 @@ function GoToDashboardButton() {
...
+    job: LatestJobForStudy | null
+    review: StudyReviewWithMeta | null
+    scan: JobScanResult | null
+}
+
+// The full "Submitted code" section (datasets, AI summary, security scan log, code viewer)
+// is the same one shown during active review. The raw code stays collapsed behind its toggle
+// (codeInitiallyExpanded={false}); the summary and scan results render up front.
+function SubmittedCodePanel({ isVisible, orgSlug, study, job, review, scan }: SubmittedCodePanelProps) {
+    if (!isVisible || !job || !scan) return null
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Worth thinking about the `!scan` half of this guard. Today it's safe: the only caller (`reviewer-code-feedback-screen.tsx`) sources `scan` from `jobScanResultForJob`, which never returns null — it falls back to `{status:'IN-PROGRESS', logFile:null}` — so in practice `scan` is non-null exactly when `job` is non-null, and this guard only really fires on the `job === null` branch. But the prop is typed `scan: JobScanResult | null`, so a *future* caller that passes a non-null `job` with a null `scan` would silently drop the entire section (datasets, AI summary, scan log, and the code viewer), which is a quieter failure than the old `{job && <StudyCodeViewer/>}` that always showed the code when a job existed. Might be worth either tightening the prop to `scan: JobScanResult` (since job-present implies scan-present here) or leaving a short comment that scan-presence is coupled to job-presence, so nobody wires up a caller that trips this silently.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -138,30 +140,70 @@ function GoToDashboardButton() {
...
+        <>
+            <ProposalStepHeader
+                stepLabel={kindCopy.stepLabel}
+                heading={kindCopy.heading}
+                studyTitle={study.title}
+                timestampDate={timestampDate}
+                timestampLabel={timestampLabel}
+                banner={banner}
+            />
+            <SubmittedCodePanel isVisible orgSlug={orgSlug} study={study} job={job} review={review} scan={scan} />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Tiny thing: `CodeSection` already does `if (!isVisible) return null` above, so passing a hardcoded `isVisible` here means `SubmittedCodePanel`'s own `!isVisible` check can never be the thing that hides it. Not wrong, just a touch of dead logic — could drop the `isVisible` prop from `SubmittedCodePanel` entirely (its `!job`/`!scan` checks carry the real work) to keep it a hair simpler. Non-blocking.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx
```diff
@@ -13,10 +13,26 @@ export async function ReviewerCodeFeedbackScreen({ study, orgSlug }: ScreenCompo
...
+                study={study}
+                entries={safeEntries}
+                kind="CODE"
+                job={job}
+                review={review}
+                scan={scan}
+            />
+        )
     }
     const fallbackStatus = job?.statusChanges.find((s) => isCodeDecisionStatus(s.status))
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx]: This re-derives the code decision by hand, but the #826 state machine already computed it: the rule that routed us to this very screen is `[reviewer-code-feedback, { when: (s) => s.codeDecision !== null }]` (reviewer-screen-rules.ts), and `StudyState.codeDecision: CodeDecisionStatus | null` is the disambiguated source of truth. Two concerns with re-deriving it here:

1. **Drift risk** — if #826's notion of "what counts as a code decision" changes (new `CODE-*` status, or the OTTER-601 round-reuse semantics), the rule table and this local `.find(isCodeDecisionStatus)` can disagree. The machine could route a study *to* this screen (`codeDecision !== null`) while this `.find()` returns nothing, leaving `fallback` undefined and the banner silently degraded — exactly the bug class the state machine was meant to remove.

2. **`.find()` picks the first match, not the latest** — `RawJob`'s own doc comment says status order isn't significant ("treats each job's statuses as a set"), yet this returns the first decision status it sees. For a job that went `CODE-CHANGES-REQUESTED` -> resubmit -> `CODE-APPROVED` in the same reused round, this could surface the stale decision rather than the current one.

The data-fetch added in this PR (`getStudyReviewForJob` + `jobScanResultForJob`) genuinely belongs here — only the *decision derivation* overlaps the machine. Could we source `codeDecision` (+ its timestamp) from the projected state instead — threading it onto `ScreenDescriptor`, or projecting the `raw` prop through the same resolve/state helpers — rather than walking `job.statusChanges`? Not new in this PR (it was relocated from `page.tsx` in the #826 rebase), but this is the PR that carries it into the state-machine world, so it's the natural place to fix it or leave a `TODO: source from codeDecision`.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx
```diff
@@ -13,10 +13,26 @@ export async function ReviewerCodeFeedbackScreen({ study, orgSlug }: ScreenCompo
         return <AlertNotFound title="Study was not found" message="No such study exists" />
     }
     const job = await latestSubmittedJobForStudy(study.id)
+    // The post-decision code page shows the full "Submitted code" section (datasets, AI summary,
+    // security scan log, code viewer), the same section as active review, so it needs the review +
+    // scan rows, not just the job (OTTER-613).
+    const [review, scan] = job
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx:19]: Broader point related to the `.find()` note below: this screen takes `ScreenComponentProps` but destructures only `{ study, orgSlug }` (line 11), dropping the `descriptor` and `raw: RawStudyState` that the #826 machine already projected. So alongside the genuinely-new data fetch here, the decision logic re-derives from scratch. If we leaned on `raw` (or threaded a `codeDecision` field onto the descriptor), the banner/decision logic would stay in lockstep with the rule table that selected this screen instead of being a parallel computation that can drift. Worth considering as the consuming side of the state machine — #826 went to the trouble of computing this once.

nathanstitt (changes requested):
seems like we should probably update to use the state decisions and remove the latest check

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx]: Fixed in f57a1b27. The screen no longer re-derives the decision: it now reads `codeDecision` from `projectStudyState(raw)` (the same projection whose `codeDecision !== null` rule routed us to this screen), so it stays in lockstep with the rule table and inherits the count-based liveness + `CODE_DECISION_PRIORITY` logic. The stale-vs-latest `.find()` problem is gone too: the banner timestamp is now looked up by the *resolved* decision status (`statusChanges.find((s) => s.status === codeDecision)`), not the first decision seen. This mirrors the researcher-side `code-decision-screen.tsx` precedent.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx:19]: Fixed in f57a1b27. The screen now destructures `raw` from `ScreenComponentProps` and runs it through `projectStudyState(raw)`, so the decision logic is the consuming side of the #826 state machine rather than a parallel re-derivation. (`descriptor` is not needed here since `codeDecision` is the only projected fact this screen consumes.)

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Addressed in f57a1b27. Kept the `!scan` check (it is the type-narrowing that lets us pass a non-null `scan` to `SubmittedCodeSection`, which requires `scan: JobScanResult`), and added a comment documenting that scan-presence is coupled to job-presence: the caller fetches both together and `jobScanResultForJob` never returns null (falls back to `{status:"IN-PROGRESS", logFile:null}`), so in practice the guard only fires on the null-job branch. That makes the coupling explicit so nobody wires up a caller that trips it silently.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Fixed in f57a1b27. Dropped the `isVisible` prop from `SubmittedCodePanel` entirely; its `!job`/`!scan` checks carry the real work, and the parent `CodeSection` already gates the whole block with `if (!isVisible) return null`.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx
```diff
@@ -1,30 +1,61 @@
...
+        )
     }
-    const fallbackStatus = job?.statusChanges.find((s) => isCodeDecisionStatus(s.status))
+    // Source the live decision from the state machine (the same projection that routed us here:
+    // reviewer-screen-rules' `codeDecision !== null`), not a hand-rolled status walk. This tracks
+    // count-based liveness and decision priority, and looking the timestamp up by the resolved
+    // decision keeps us on the current decision rather than the first one recorded on the job.
+    const { codeDecision } = projectStudyState(raw)
+    const decisionTimestamp = codeDecision
+        ? job?.statusChanges.find((s) => s.status === codeDecision)?.createdAt
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx:43]: Nice — sourcing `codeDecision` from `projectStudyState(raw)` resolves the drift/stale-`.find()` concern cleanly, and it mirrors the researcher-side `code-decision-screen.tsx` precedent.

nathanstitt (approved):
Looks good, thanks Marvin

PR #826, opened 2026-06-23
Study screen state machine; eliminate ?from= researcher routing

nathanstitt:
## Summary

Introduces a study-screen **state machine** that derives each role's view, dashboard action, and status pill from a single projected `RawStudyState`, replacing the legacy `?from=`-based routing cascade. Covers **both** the researcher and reviewer ("Data Partners"/DO) flows.

### Researcher flow

- **State machine core**: `projectStudyState` (latest-job + set-existence), `resolveScreen`, `resolveDashboardAction`, `resolvePillStatus`/`resolveRowHighlight`, plus a single-query `rawStudyStateForStudy` bundle.
- **Researcher /view**: page is now pure state-machine dispatch through a total screen registry; revisitable pages no longer redirect via `?from=`, and `from` is dropped from the route builders.
- **Job round semantics**: a round closes only on `FILES-APPROVED`/`FILES-REJECTED`; a `CODE-CHANGES-REQUESTED` resubmit reuses the same job. `CODE-SUBMITTED` is append-only per round, and submission version = `CODE-CHANGES-REQUESTED` count + 1.

### Reviewer flow

- **Reviewer rule table**: a role-keyed `REVIEWER_SCREEN_RULES` over the **unchanged** shared projection; `resolveScreen('reviewer', …)` resolves one of six `reviewer-*` screens, mapped to thin adapters over the existing reviewer views in a compiler-exhaustive registry.
- **`/review` is pure dispatch**: the ~250-line `?from=` cascade collapses to guards + a shared `renderStudyScreen` helper (also used by `/view`). All reviewer `?from=` (`initial-request`, `code-review`, `agreements`, `agreements-proceed`, `previous`) is removed, including stragglers in the redirect hooks.
- **Dedicated proposal route**: `/review/proposal` (new `studyReviewProposal` route) backs the "View approved initial request" link; it shows the decided initial request and falls through to the canonical screen when the proposal isn't decided yet.
- **Agreements gate as a screen**: a `reviewer-agreements` screen replaces the old redirect; both reviewer page entry points share one `reviewerPageGuard`.

### Misc

- Mock `deleteFolderContents` in the code-submission tests — the round-reuse cleanup branch makes a real S3 call that left `study.status` stuck at `APPROVED` in the unit env.
- Fix a pre-existing parallel-suite flake in `study-code.test.tsx` (`waitFor` the main-file `aria-pressed` flip).

## Test plan

- `pnpm run checks` (typecheck + lint + action validation) — passing
- `pnpm run test:unit` — 1589 passing, 0 failing (stable across repeated runs)


docs/plans/2026-06-22-job-round-boundary-fix-subspec.md
chrisbendel, line 1: Do we need all of the files here inside of `plans` or `docs`?

src/lib/code-resubmission.ts
```diff
@@ -10,6 +10,15 @@ export const CODE_RESUBMITTABLE_JOB_STATUSES = [
     'RUN-COMPLETE',
 ] as const satisfies readonly StudyJobStatus[]
 
+// A round CLOSES — and the next IDE launch / upload / submit opens a NEW studyJob — only after a
+// post-run results decision. Pre-run outcomes (CODE-CHANGES-REQUESTED) and a not-yet-reviewed errored
+// run (JOB-ERRORED, awaiting the reviewer's files decision) revise/continue the SAME job.
+// CODE-REJECTED is terminal. This is DISTINCT from CODE_RESUBMITTABLE_JOB_STATUSES (resubmit eligibility).
```
therealmarv, line 23: 🔴 **High — a resubmit after `RUN-COMPLETE`/`JOB-ERRORED` is silently dropped.**

`CODE_RESUBMITTABLE_JOB_STATUSES` (just above) still permits `RUN-COMPLETE` and `JOB-ERRORED`, but `ROUND_CLOSING_JOB_STATUSES` excludes them. So when a researcher resubmits from one of those states, [`getOrCreateCurrentRoundJob`](https://github.com/safeinsights/management-app/blob/69212c8dc6941ad161a87cc09deff92388a13fcb/src/server/db/mutations.ts#L92) reuses the same job (round not closed), and [`markCodeSubmitted`](https://github.com/safeinsights/management-app/blob/69212c8dc6941ad161a87cc09deff92388a13fcb/src/server/actions/study-request.ts#L145) then no-ops because `submitted(1) > requested(0)` — no new `CODE-SUBMITTED` is appended. `projectStudyState` still sees `hasResults`, so `/view` keeps rendering the terminal results screen and the resubmission is lost for both roles.

Either add these statuses to the round-closing set, or have `markCodeSubmitted` account for a post-results resubmit.

src/server/db/queries.ts
```diff
@@ -142,23 +142,28 @@ export const latestSubmittedJobForStudy = async (studyId: string): Promise<Lates
...
 }
 
-// 1 = first submission, >=2 = resubmission round.
-export const countSubmittedJobsForStudy = async (studyId: string): Promise<number> => {
+// Submission version of the latest round = (number of CODE-CHANGES-REQUESTED review rounds on the
+// latest job) + 1. First submission = v1; each same-job change-request + resubmit bumps the version.
+// CODE-SUBMITTED is append-only per round (markCodeSubmitted is round-aware), so count(CODE-SUBMITTED)
+// would also work — counting the reviewer's CODE-CHANGES-REQUESTED is equivalent and reads as
+// "rounds reviewed". FILES-REJECTED (post-run, opens a new job) does not count.
+export const codeSubmissionVersion = async (studyId: string): Promise<number> => {
```
therealmarv, line 157: 🔴 **High — the first resubmission after a `FILES-REJECTED` round hides prior reviewer feedback.**

`codeSubmissionVersion` counts `CODE-CHANGES-REQUESTED` on the *latest* job, + 1. After a `FILES-REJECTED` decision opens a fresh job, the first resubmit computes version `1`, so the under-review screen skips `getCodeReviewFeedbackAction` (gated on `version > 1`) and the researcher no longer sees the previous round's feedback. The old `countSubmittedJobsForStudy` counted across all jobs and didn't reset at a round boundary.

src/server/actions/study-request.ts
```diff
@@ -119,18 +120,30 @@ async function attachCodeToRoundJob(
...
+            eb.fn.count<number>('id').filterWhere('status', '=', 'CODE-SUBMITTED').as('submitted'),
+            eb.fn.count<number>('id').filterWhere('status', '=', 'CODE-CHANGES-REQUESTED').as('requested'),
+        ])
         .where('studyJobId', '=', studyJobId)
-        .where('status', '=', 'CODE-SUBMITTED')
-        .executeTakeFirst()
-    if (alreadySubmitted) return
+        .executeTakeFirstOrThrow()
+
+    const currentRoundAlreadySubmitted = Number(counts.submitted) > Number(counts.requested)
```
therealmarv, line 145: 🟠 **Medium — count-then-insert with no lock or unique constraint (race).**

Two concurrent submits (double-click, two tabs) can both read `submitted <= requested` and each insert a `CODE-SUBMITTED` row. Because the new projection is count-based, a duplicate can flip the live-decision calculation and leave the study stuck under review. A unique partial index or a row lock would make this idempotent.

src/app/[orgSlug]/study/[studyId]/agreements/page.tsx
```diff
@@ -54,39 +45,29 @@ export default async function StudyAgreementsRoute(props: {
...
-            ? Routes.studyView({ orgSlug: study.submittedByOrgSlug, studyId })
-            : Routes.studyCode({ orgSlug: study.submittedByOrgSlug, studyId })
-        redirect(dest)
-    }
-
-    if (study.status !== 'APPROVED' && !isDirectAccess) {
-        redirect(Routes.studyView({ orgSlug: study.submittedByOrgSlug, studyId }))
-    }
-
+    // Researcher flow: /agreements is a revisitable step. An authorized researcher can view it
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/agreements/page.tsx:65]: 🟠 **Medium — researcher `/agreements` dropped the `status === 'APPROVED'` guard.**

A non-approved researcher can now render this step directly. `PENDING-REVIEW` is still bounced downstream by `getDraftStudyAction` (it only accepts `DRAFT`/`CHANGE-REQUESTED`/`APPROVED`), but a `DRAFT`/`CHANGE-REQUESTED` study will render the upload page via Proceed. Worth confirming the code-submission actions guard status independently, so this can't become a pre-approval bypass.

src/app/[orgSlug]/study/[studyId]/review/study-details-reviewer.tsx
```diff
@@ -20,7 +20,7 @@ export async function StudyDetailsReviewer({ orgSlug, study }: StudyDetailsRevie
         return <AlertNotFound title="No submission found" message="This study has no submitted code to review." />
     }
 
-    const previousHref = Routes.studyReview({ orgSlug, studyId: study.id, from: 'code-review' })
+    const previousHref = Routes.studyReview({ orgSlug, studyId: study.id })
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/review/study-details-reviewer.tsx]: 🟡 **Medium-low — reviewer results "Previous" is a self-link.**

`previousHref` is bare `/review`, but `REVIEWER_SCREEN_RULES` routes `hasResults → reviewer-study-results` first, so `/review` resolves right back to this same screen. "Previous" goes nowhere — it likely should target the dashboard or the post-code-feedback screen.

src/app/[orgSlug]/study/[studyId]/agreements/page.tsx
```diff
@@ -54,39 +45,29 @@ export default async function StudyAgreementsRoute(props: {
                 <AgreementsPage
                     isReviewer
                     studyId={studyId}
-                    proceedHref={`${Routes.studyReview({ orgSlug, studyId })}?from=agreements-proceed`}
-                    previousHref={`${Routes.studyReview({ orgSlug, studyId })}?from=agreements`}
+                    proceedHref={Routes.studyReview({ orgSlug, studyId })}
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/agreements/page.tsx:57]: 🟡 **Medium-low — reviewer agreements Back/Previous loop.**

Both `proceedHref` and `previousHref` point at bare `/review`. For a reviewer with code awaiting a decision, `/review` re-resolves to `reviewer-code-review`, whose "Previous" sends them back here ([code-review.tsx#L106](https://github.com/safeinsights/management-app/blob/69212c8dc6941ad161a87cc09deff92388a13fcb/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/review/code-review.tsx#L106)) → loop. The state-machine screen version uses `dashboardHref` for Previous; these route-page entry points are the inconsistent ones.

src/lib/routes/definitions.ts
```diff
@@ -100,37 +100,23 @@ export const Routes = {
     studyRequest: makeRoute(({ orgSlug }) => `/${orgSlug}/study/request`, OrgParams),
 
     studyView: makeRoute(
```
therealmarv, line 102: 🟢 **Nit — leftover `from` param.** Now that `from` is dropped from `studyView`/`studyEdit`/`studyReview`, `studyAgreements` (L131) is the only route still declaring an unused `from`. Worth dropping it here too for consistency.

src/app/[orgSlug]/study/[studyId]/review/code-review.tsx
```diff
@@ -102,7 +102,7 @@ export async function CodeReview({ orgSlug, study, entries }: CodeReviewProps) {
     }
 
     const [review, scan] = await Promise.all([getStudyReviewForJob(job.id), jobScanResultForJob(job.id)])
-    const proposalHref = `${Routes.studyReview({ orgSlug, studyId: study.id })}?from=code-review`
+    const proposalHref = Routes.studyReviewProposal({ orgSlug, studyId: study.id })
     const previousHref = Routes.studyAgreements({ orgSlug, studyId: study.id, from: 'previous' })
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/review/code-review.tsx]: 🟢 **Nit — dead `from: 'previous'`.**

The agreements page no longer reads `from` (the `isDirectAccess` logic was removed), so this param is inert. The researcher equivalent in `code/page.tsx` was cleaned up but this reviewer one was missed; `review/code-review-client.test.tsx:50` still asserts it. Drop the param and update the test.

tests/study-flow.spec.ts
```diff
@@ -715,12 +874,12 @@ test('Code rejection and resubmission', async ({ page, studyFeatures }) => {
...
-        // Reviewer is auto-redirected to agreements when code has been submitted
-        await page.waitForURL(/\/agreements(\?.*)?$/, { timeout: 10000 })
-        const studyBaseUrl = page.url().replace(/\/agreements(\?.*)?$/, '')
-
-        // Navigate to code review via agreements-proceed to bypass the agreements redirect
-        await goto(page, `${studyBaseUrl}/review?from=agreements-proceed`)
+        // "View" lands on /review; with code submitted but agreements not yet acked, the state
+        // machine renders the agreements gate there. Ack via "Proceed to Step 3" to re-resolve
+        // bare /review to the code-review editor.
+        await page.waitForURL(/\/review(\?.*)?$/, { timeout: 10000 })
```
therealmarv: 🟢 **Nit — inline Playwright timeout.**

CLAUDE.md: "Do not set inline timeouts in Playwright tests… use constants from `tests/common.helpers.ts`." Use `E2E_TIMEOUT`/`E2E_TIMEOUT_LONG` here (and at the matching `waitForURL` around L976) instead of the hardcoded `10000`.

therealmarv (changes requested):
## Review — Study screen state machine (`?from=` elimination)

Reviewed against `69212c8`. This is a big and genuinely good refactor: collapsing the `?from=` routing cascade into a single `projectStudyState` projection + role-keyed screen-rule tables + a total screen registry is a real maintainability win, and separating "resubmit eligibility" from "round closes" is the right mental model. A few correctness gaps in the new round logic should be fixed before merge, plus some navigation regressions and small cleanups. Details inline.

### What's strong
- Single source of truth: one `RawStudyState` bundle → `projectStudyState`, with order-independent set/count-based liveness and a `state.shuffle.test.ts` guard.
- The total `Record<ScreenId, …>` registry gives compile-time exhaustiveness, and the Tier-1↔Tier-2 `consistency.test.ts` is an excellent drift guard.
- The reviewer-flow rework (six `reviewer-*` screens, agreements-as-a-screen) is a clean collapse of the old ~250-line cascade.
- The round-boundary fix is well-tested (CR-reuse, FILES-*/new-round, the `createdAt`-tie determinism case).

### Severity legend
🔴 high / blocker · 🟠 medium · 🟡 medium-low · 🟢 nit

### Blockers (🔴 — see inline)
- `code-resubmission.ts` — a resubmit after `RUN-COMPLETE`/`JOB-ERRORED` is silently dropped.
- `queries.ts` — `codeSubmissionVersion` resets to 1 after a `FILES-REJECTED` new round, hiding prior feedback.

### Also worth addressing (outside the diff, so noted here)
- 🟠 [`saveCodeResubmissionNoteDraftAction` still guards `status = 'CHANGE-REQUESTED'`](https://github.com/safeinsights/management-app/blob/69212c8dc6941ad161a87cc09deff92388a13fcb/src/server/actions/study-request.ts#L838). Code resubmission runs while `study.status` stays `APPROVED`, so the resubmission-note autosave will error on every keystroke (this is OTTER-558 Defect and should be probably adressed here https://github.com/safeinsights/management-app/pull/822 ). It needs the same submitted-job / `canResubmitStudyCode` guard the resubmit page renders behind.
- 🟢 The [`study-job-status.ts` comment](https://github.com/safeinsights/management-app/blob/69212c8dc6941ad161a87cc09deff92388a13fcb/src/lib/study-job-status.ts#L65) "Current resubmissions open a brand-new job" is now backwards under the round-boundary fix (a CR resubmit reuses the job); the count logic itself is still correct.
- 🟢 The [`edit-and-resubmit/footer.tsx` comment](https://github.com/safeinsights/management-app/blob/69212c8dc6941ad161a87cc09deff92388a13fcb/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/edit-and-resubmit/footer.tsx#L45) points at "review/page.tsx for prior art" on `?from=`, which no longer exists there.

### Cleanup (🟢, optional)
- [`projectStudyState` runs 3× per dashboard row](https://github.com/safeinsights/management-app/blob/69212c8dc6941ad161a87cc09deff92388a13fcb/src/components/dashboard/studies-table/study-row.tsx#L16) (`shouldHighlight`, `useStudyStatus`, `ResearcherLink`); project once and thread it down.
- Liveness logic is duplicated between `state.ts` and the still-live `latestSubmittedJobHasLiveCodeDecision`, and several `StudyState` fields (`displayStatus`, `submissionRound`, …) have no non-test consumers.

### Note on overlap with #822
This PR fixes the OTTER-558 masking symptom at the source (the round boundary). #822 tackles the same symptom on the current architecture (its dashboard query prefers the latest *submitted* job). Since this PR lands first, that workaround becomes redundant and can be dropped when #822 is reworked on top — nothing to do here, just noting it for context.

Nice work overall — the architecture is the right direction; it's mainly the new round-edge cases that need tightening.

src/lib/code-resubmission.ts
nathanstitt: Good catch — verified and fixed in a8b0b66e. The mechanism is exactly as you describe (round not closed → job reused → `markCodeSubmitted` no-ops), but the scope is narrower than "every resubmit": the UI only renders a resubmit affordance once a FILES-* decision is present, so this was only reachable by navigating directly to `/resubmit` at a bare `RUN-COMPLETE`/`JOB-ERRORED`.

Per OTTER-558 the intended resubmit-from states ("results ready", errored) are the *decided* states (FILES-APPROVED, incl. FILES-APPROVED+JOB-ERRORED) — those already close the round and work. So rather than add `RUN-COMPLETE`/`JOB-ERRORED` to the round-closing set (they're written on the same job a FILES-* decision later closes, which would break the order-independence invariant), I dropped them from `CODE_RESUBMITTABLE_JOB_STATUSES`. The page guard + action now reject a bare-results resubmit instead of silently dropping it, and no UI capability is lost.

src/server/db/queries.ts
nathanstitt: Confirmed and fixed in a8b0b66e — and OTTER-556/558 settle the ambiguity in your favor: the feedback section must show prior feedback/notes "over time" with an ever-increasing version (v1, v2, v3…), so the post-`FILES-REJECTED` reset to v1 was a real defect, not just a display quirk.

`codeSubmissionVersion` now counts round-opening events (`CODE-CHANGES-REQUESTED` + `FILES-APPROVED` + `FILES-REJECTED`) across **all** of the study's jobs, + 1, so the version is monotonic across a results-decision round boundary and the read-only screens keep surfacing prior feedback (the `> 1` gate no longer trips back to false). Counting these events rather than `CODE-SUBMITTED` also keeps the version immune to a duplicate `CODE-SUBMITTED`. Updated the `queries.test.ts` cases accordingly (added a "keeps climbing across a results-decision round boundary" test).

src/server/actions/study-request.ts
nathanstitt: Looked into this closely. The race is real (count-then-insert, no lock/constraint), but we've decided not to add a lock or unique index — the duplicate is harmless for the projection, for two reasons:

1. A concurrent duplicate only ever lands **pre-decision** (both transactions read `requested=0` before any decision exists), so the shape is `[CODE-SUBMITTED, CODE-SUBMITTED]` with `decisionCount=0`. `hasLiveDecision` is gated on `decisionCount > 0` first, so it short-circuits and the screen is identical to the single-submit case. The "stuck under review" outcome would require `submittedCount` to exceed `decisionCount` *with a decision present in the same round*, which isn't reachable from this race window.
2. The version no longer counts `CODE-SUBMITTED` (see the queries.ts change in a8b0b66e), so a duplicate can't inflate it either.

This is also consistent with OTTER-622's expected behavior ("resubmission should not wait or be affected by [scan timing]") and the principle that nothing should depend on status order/timing. Worst case from a duplicate is a redundant history row, no review-deadlock.

src/app/[orgSlug]/study/[studyId]/agreements/page.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/agreements/page.tsx:65]: Confirmed the downstream actions guard independently, so this isn't a pre-approval bypass — leaving the guard removed as intended. The removal is deliberate: `/view`'s state machine (`resolveScreen`) is now the screen authority, and `/agreements` is a revisitable step. `getDraftStudyAction` only accepts `DRAFT`/`CHANGE-REQUESTED`/`APPROVED` (bounces `PENDING-REVIEW`), and the code-submission actions (`addJobToStudyAction`/`submitStudyCodeAction`) require the CASL `create StudyJob` ability and set `PENDING-REVIEW` — none of which a non-approved proposal can escalate past. Worst case is a researcher landing on a step out of order, not a security bypass.

src/app/[orgSlug]/study/[studyId]/review/study-details-reviewer.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/review/study-details-reviewer.tsx]: Fixed in a8b0b66e. Since `hasResults` routes to `reviewer-study-results` first, bare `/review` did self-resolve back to this screen. Pointed "Previous" at the dashboard (`Routes.orgDashboard`) — the only non-looping target while results are present — and updated the test + comment. (The OTTER-538 intent of "Previous → post-code-feedback page" isn't reachable via dispatch once results exist, since results out-rank the code decision in `REVIEWER_SCREEN_RULES`.)

src/app/[orgSlug]/study/[studyId]/agreements/page.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/agreements/page.tsx:57]: Fixed in a8b0b66e. The reviewer-agreements route page now mirrors `ReviewerAgreementsScreen`: Proceed → `/review` (re-resolves to code review after ack), Previous → dashboard. That breaks the agreements ⇄ code-review loop (code-review's Previous still legitimately points back to agreements, but agreements no longer bounces straight into code-review).

src/lib/routes/definitions.ts
nathanstitt: Done in a8b0b66e — dropped `from` from the `studyAgreements` route definition (builder + schema). It's now `returnTo`-only, consistent with `studyView`.

src/app/[orgSlug]/study/[studyId]/review/code-review.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/review/code-review.tsx]: Done in a8b0b66e — removed `from: 'previous'` from `code-review.tsx` and updated `code-review-client.test.tsx:50`. Combined with dropping `from` from the route definition, the dead param is fully gone.

tests/study-flow.spec.ts
nathanstitt: Fixed in a8b0b66e — both call sites (≈L880 and ≈L976) now use `E2E_TIMEOUT` imported from `./e2e.helpers` instead of the hardcoded `10000`.

nathanstitt:
Thanks for the thorough review — addressed in a8b0b66e (pushed). Summary, with the inline threads covering specifics:

**Blockers**
- 🔴 resubmit silent-drop: dropped bare `RUN-COMPLETE`/`JOB-ERRORED` from `CODE_RESUBMITTABLE_JOB_STATUSES` (per OTTER-558 the intended resubmit-from states are the decided FILES-* ones, which already close the round).
- 🔴 version reset: `codeSubmissionVersion` now counts round-opening events across all jobs, so it's monotonic across a results-decision boundary and keeps prior feedback visible (OTTER-556/558 confirm this was the intended behavior).

**Also-noted (outside the diff)**
- 🟠 `saveCodeResubmissionNoteDraftAction`: confirmed real (autosave threw on every keystroke while `study.status` stayed APPROVED). Re-gated on the latest submitted job via `canResubmitStudyCode`, matching the resubmit page.
- 🟢 `study-job-status.ts` "open a brand-new job" comment: fixed (a CR resubmit reuses the same job now).
- 🟢 `edit-and-resubmit/footer.tsx` stale `?from=` / `review/page.tsx` prior-art pointer: fixed.

**Cleanup notes** — the `projectStudyState`-3×-per-row and the duplicated-liveness / unused-`StudyState`-fields observations are fair; leaving those for a follow-up since they're outside this review's scope. On the #822 overlap: agreed, nothing to do here.

**Race** (🟠 inline): decided not to add a lock/index — the duplicate is harmless (lands pre-decision, short-circuited by `decisionCount > 0`; version no longer counts `CODE-SUBMITTED`). Details on the thread.

docs/plans/2026-06-22-job-round-boundary-fix-subspec.md
nathanstitt: Good question 😁   We're not sure but are experimenting with keeping them to provide more context around _why_ things changed.  If they don't prove useful, we'll likely remove them all

therealmarv (approved):
Looks good for me now! 🟢

PR #827, opened 2026-06-24
fix: absorb quotes around %f so env command templates don't double-quote the entry-point filename (OTTER-477)

Rod0352:
Fix
substituteEntryPointFile(template, fileName) makes %f substitution idempotent: it consumes one matching pair of quotes hugging the token, then shell-quotes. Result is singly quoted whether the template is %f, "%f", or '%f'. buildTriggerBuildImageCommandInput now calls it instead of replaceAll('%f', …). Admin help text updated to note %f is quoted automatically.
Testing

Unit tests for the helper: bare / double-quoted / single-quoted token, embedded single quote, multiple %f, quotes elsewhere, no token.
Regression tests in aws.test.ts for double- and single-quoted templates.
Verified through a real /bin/sh: all three forms produce one clean argument (main_revised (1).R).

### DIFF
--- a/src/server/aws.ts
+++ b/src/server/aws.ts
@@ -31,7 +31,7 @@ import {
     pathForStudyJobCode,
 } from '@/lib/paths'
 import type { MinimalCodeEnvInfo } from '@/lib/types'
-import { shellQuote, strToAscii } from '@/lib/string'
+import { substituteEntryPointFile, strToAscii } from '@/lib/string'
 import { parseCsv } from '@/lib/file-content-helpers'
 import logger from '@/lib/logger'
 import { Readable } from 'stream'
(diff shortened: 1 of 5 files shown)


src/server/aws.ts
```diff
@@ -466,12 +466,11 @@ export async function buildTriggerBuildImageCommandInput(
...
 ) {
-    // Shell-quote the filename so names with parentheses or other shell metacharacters
-    // don't break the `/bin/sh` command the containerizer runs (OTTER-477). The
-    // replacement must be a function — a string replacement would interpret `$`
-    // sequences in the filename as special replacement patterns. replaceAll covers
-    // templates that reference %f more than once.
-    const cmd = info.cmdLine.replaceAll('%f', () => shellQuote(info.codeEntryPointFileName))
+    // Substitute the %f entry-point token, shell-quoting the filename so names with
+    // parentheses or other shell metacharacters don't break the `/bin/sh` command the
+    // containerizer runs (OTTER-477). The helper also absorbs quotes an admin may have
```
nathanstitt, line 471: This comment now says the helper "absorbs quotes an admin may have wrapped around %f" — but the admin-facing help text in `code-env-form.tsx` ("Use %f for the main code file name.") still doesn't tell admins that quoting is automatic. The PR description says "Admin help text updated to note %f is quoted automatically," yet `code-env-form.tsx` isn't in this diff. Did that change get dropped? Without it, an admin who reads the docs still has no signal that `"%f"` and `%f` behave identically now.

nathanstitt (approved):
looks good, just probably need to update help text

PR #828, opened 2026-06-24
OTTER-594: unify autosave UI to 'All changes saved'

Rod0352:
## Summary
Unifies the three inconsistent autosave indicators into one shared "Saving…" → 
"All changes saved" indicator, shown under every field that autosaves, for both 
RL and DO. Strips the editor's client-side save choreography (30s heartbeat, 
relative-timestamp ticking, local-vs-remote branching) now that persistence is 
handled by the editor service. Net ~150 fewer lines.

## What changed
- **New `SaveStatusIndicator`** (`src/components/save-status.tsx`) — single source 
  of the "Saving…" / "All changes saved" copy; renders nothing when idle.
- **New `useProviderSaveStatus`** (`src/lib/realtime/use-provider-save-status.ts`) — 
  derives idle/saving/saved from a Hocuspocus provider's `unsyncedChanges`; stays 
  idle until the user's first local edit so passive readers and initial loads stay quiet.
- **Collaborative editor** — removed `useSaveStatus` / `formatTimeAgo` / heartbeat; 
  renders the shared indicator. Covers RL proposal long-text fields and DO 
  proposal-review + code-review feedback.
- **Proposal (Initial Request) page** — indicator added under Study Title, 
  Dataset(s) of interest, and Principal Investigator (driven by the shared Yjs fields doc).
- **Resubmission note** (proposal + code resubmit) — swapped to the shared 
  indicator; dropped the `at {time}` timestamp.


src/lib/realtime/use-provider-save-status.ts
```diff
@@ -0,0 +1,49 @@
...
+    const hasLocalEditRef = useRef(false)
+
+    useEffect(() => {
+        if (!provider) return undefined
+
+        const onUnsyncedChanges = () => {
+            if (provider.unsyncedChanges > 0) {
+                hasLocalEditRef.current = true
+                setStatus('saving')
+            } else if (hasLocalEditRef.current) {
```
nathanstitt, line 30: One edge worth thinking about: `hasLocalEditRef` latches to `true` for the lifetime of the hook instance and never resets, while the effect re-subscribes whenever `provider` identity changes (e.g. a websocket reconnect produces a fresh `HocuspocusProvider`, which is what `activeProvider` swaps to in the editor). After such a reconnect, the very first `unsyncedChanges === 0` settle on the *new* provider will hit this `else if` branch and re-show "All changes saved" even though the user hasn't edited anything since reconnecting. Probably harmless given the copy is generic now, but if we ever want the indicator to go quiet after a reconnect, resetting `hasLocalEditRef.current = false` in the effect cleanup (or keying it to the provider) would do it. Might be worth a one-line comment either way so the latch is clearly intentional.

src/app/[orgSlug]/study/[studyId]/proposal/form.tsx
```diff
@@ -125,6 +128,7 @@ export const ProposalForm: FC<ProposalFormProps> = ({
                                     {form.errors.title && <InputError error={form.errors.title} />}
                                     <WordCounter wordCount={titleWordCount} maxWords={WORD_LIMITS.title} />
                                 </Group>
+                                <SaveStatusIndicator status={fieldsSaveStatus} />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/form.tsx]: Since all three fields share the single `fieldsSaveStatus` derived from `yjsForm.provider`, editing just the Title will flash "Saving…" → "All changes saved" simultaneously under Dataset(s) and Principal Investigator too. The PR notes these are one shared Yjs fields doc, so this is faithful to the data model — just flagging the UX: three identical indicators reacting in lockstep to an edit in any one field reads a little oddly. If that's the intended behavior, all good; if not, a single indicator for the whole section (rather than one per field) might communicate it more cleanly.

src/components/save-status.tsx
```diff
@@ -0,0 +1,26 @@
...
+    saved: 'All changes saved',
+}
+
+interface SaveStatusIndicatorProps {
+    status: SaveStatusValue
+}
+
+// Single autosave indicator shared across every surface (collaborative editor,
+// proposal fields, resubmission note). Renders nothing until there is something
+// to report so it can be dropped under any field unconditionally.
```
nathanstitt, line 17: Really like how this collapsed three bespoke indicators into one tiny presentational component with a clean idle/saving/saved contract — renders-nothing-when-idle means call sites can drop it in unconditionally. The accompanying deletion of the 30s-heartbeat / relative-timestamp choreography in the editor is a big readability win.

nathanstitt (approved):
looks good, but we should think about the 3 saving indicators flashing at once. I wonder if we can check the form dirty status to see if the fields were actually modified or not before showing?

PR #829, opened 2026-06-24
rl-study-code-refactor

jbwilson8:
### Summary

The Researcher study proposal view, in the **code change requested** state, didn't match
the design. QA flagged two issues:

> Does not match design; submitted code not broken out, proposal link missing

This PR brings the view in line with the design by breaking the submitted code into its own
card and adding the missing "View approved initial request" link. The change mirrors the
pattern already used in the sibling post-submission view.

[OTTER-590](https://openstax.atlassian.net/browse/OTTER-590)

### What changed

- **Submitted code is now broken out into its own "Submitted code" card** below the Step 4
  card. While collapsed, the Step 4 card shows a `View submitted study code ›` toggle; once
  expanded, that opener is replaced by the separate card containing the file table and a
  `Hide submitted study code` toggle.
- **Added the "View approved initial request" link** (external-link icon, opens the submitted
  proposal in a new tab) in the broken-out card's header.
- Added an optional `testId` prop to `StudyCodeToggle` so the in-card opener and the in-panel
  closer remain distinct — Mantine's `Collapse` keeps its children mounted while collapsed,
  which otherwise made the shared test id ambiguous.
- The execution-window behavior (`showStudyCode=false`) is preserved: neither the toggle nor
  the code card renders.

### Files

| File | Change |
| --- | --- |
| `code-post-decision-view.tsx` | Lifted expand state up; split the toggle/table into a step-card toggle + a broken-out `SubmittedCodePanel` with the proposal link |
| `study-code-collapse.tsx` | Added optional `testId` prop to `StudyCodeToggle` |
| `code-post-decision-view.test.tsx` | Updated the toggle test for the new structure; added a test for the proposal link's href/target |

### Screenshots

| Before | After |
| --- | --- |
| Screenshot 2026-06-24 at 9 15 16 AM| Screenshot 2026-06-24 at 8 58 01 AM|

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/view/study-code-collapse.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/view/study-code-collapse.tsx
@@ -16,9 +16,17 @@ interface StudyCodeToggleProps {
     onClick: () => void
     isVisible?: boolean
     mt?: MantineSpacing
+    /** Override the test id so multiple toggles on one page (e.g. an in-card opener and an in-panel closer) stay distinct. */
+    testId?: string
 }
 
-export const StudyCodeToggle: FC<StudyCodeToggleProps> = ({ expanded, onClick, isVisible = true, mt }) => {
+export const StudyCodeToggle: FC<StudyCodeToggleProps> = ({
+    expanded,
+    onClick,
+    isVisible = true,
+    mt,
+    testId = 'study-code-toggle',
+}) => {
     if (!isVisible) return null
     return (
         <Anchor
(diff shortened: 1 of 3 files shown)


src/app/[orgSlug]/study/[studyId]/view/study-code-collapse.tsx
```diff
@@ -16,9 +16,17 @@ interface StudyCodeToggleProps {
     onClick: () => void
     isVisible?: boolean
     mt?: MantineSpacing
+    /** Override the test id so multiple toggles on one page (e.g. an in-card opener and an in-panel closer) stay distinct. */
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/study-code-collapse.tsx:19]: Nice touch documenting *why* the `testId` override exists. One thought: the two toggles already differ by visible text — "View submitted study code" (opener) vs "Hide submitted study code" (closer) — so the tests could disambiguate via `getByRole('button', { name: ... })` on that text instead of adding a test-only prop to the production component's API. Not a blocker, but it would keep `StudyCodeToggle` free of a prop that exists purely to satisfy the test harness. If you'd rather keep the explicit `testId` for stability, totally reasonable — just flagging the role-based alternative.

nathanstitt (approved):
looks great, ty!

PR #830, opened 2026-06-25
OTTER-600: auto-refresh RL dashboards for PENDING-REVIEW proposals

jbwilson8:
### Problem

When a proposal is approved by DO (Data Organization), the Researcher's (RL) dashboard only reflects the new status after a **manual page reload**. The studies tables don't auto-refresh while a proposal is awaiting review.

[OTTER-600](https://openstax.atlassian.net/browse/OTTER-600)

### Root Cause

The 90-second auto-refresh machinery already exists (`useTimer` → `Refresher` → `StudiesTable`), but it only runs when **both** conditions hold: the `showRefresher` prop is true **and** `needsRefresh()` returns true. Two gaps excluded the researcher case:

1. **`needsRefresh()` ignored the proposal status** — it only inspected job-lifecycle changes (`jobStatusChanges`). A `PENDING-REVIEW` proposal typically has no study job yet, so that array is empty → `needsRefresh()` returned `false` → polling never started → the researcher had to reload to see the proposal flip to `APPROVED`.
2. **The lab org dashboard never enabled the refresher** — it passed `showRefresher={isEnclave}`, which is `false` for labs (the researcher view). My Dashboard already passed `showRefresher`, so it was only blocked by gap # 1.

### Changes

- **`studies-table/types.ts`** — Added `ACTIVE_PROPOSAL_STATUS = ['PENDING-REVIEW']` alongside the existing `FINAL_STATUS`, describing proposal statuses where the researcher is awaiting a DO decision.
- **`studies-table/index.tsx`** — `needsRefresh()` now also returns `true` when any study is in an active proposal status, so My Dashboard begins polling immediately.
- **`[orgSlug]/dashboard/page.tsx`** — Changed `showRefresher={isEnclave}` → `showRefresher` so both the enclave (reviewer) and lab (researcher) org dashboards auto-refresh.
- **`studies-table.test.tsx`** — Added coverage: a lone `PENDING-REVIEW` study with no jobs keeps the countdown active; an all-final study shows "Reload inactive."

The 90s interval and spy-mode countdown are reused unchanged — only `CODE-REJECTED` / `JOB-ERRORED` / `FILES-APPROVED` / `FILES-REJECTED` (final) studies stop polling. `CHANGE-REQUESTED` / `DRAFT` are intentionally excluded since the next action there is the researcher's own.

**Manual verification:** As a researcher with a `PENDING-REVIEW` proposal, open My Dashboard and the lab org dashboard, enable spy mode (π, bottom-right) to confirm the "N seconds until refresh" countdown is active. Approve the proposal as DO in another session and confirm the status flips to Approved within ~90s without a manual reload.

### Notes

- Did not touch the duplicate final-status constants (`FINAL_STATUS` in `studies-table/types.ts` vs `JOB_FINAL_STATUSES` in `src/lib/types.ts`) — out of scope.

[OTTER-600]: https://openstax.atlassian.net/browse/OTTER-600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/dashboard/page.tsx
+++ b/src/app/[orgSlug]/dashboard/page.tsx
@@ -47,7 +47,7 @@ export default async function OrgDashboardPage(props: { params: Promise<{ orgSlu
                         : undefined
                 }
                 showNewStudyButton={!isEnclave}
-                showRefresher={isEnclave}
+                showRefresher
                 paperWrapper
             />
         </Stack>
--- a/src/components/dashboard/studies-table/index.tsx
+++ b/src/components/dashboard/studies-table/index.tsx
@@ -15,6 +15,7 @@ import {
 import { StudyRow } from './study-row'
 import { StudiesTableView } from './studies-table-view'
 import {
+    ACTIVE_PROPOSAL_STATUS,
     Audience,
     FINAL_STATUS,
     REVIEWER_ACTION_STATUSES,
@@ -46,7 +47,11 @@ function filterStudiesForUser(studies: StudyRowType[], audience: Audience, userI
 }
 
 function needsRefresh(studies: StudyRowType[]): boolean {
-    return studies.some((study) => study.jobStatusChanges.some((change) => !FINAL_STATUS.includes(change.status)))
+    return studies.some(
+        (study) =>
+            ACTIVE_PROPOSAL_STATUS.includes(study.status) ||
+            study.jobStatusChanges.some((change) => !FINAL_STATUS.includes(change.status)),
+    )
 }
 
 export function StudiesTable({
--- a/src/components/dashboard/studies-table/index.tsx
+++ b/src/components/dashboard/studies-table/index.tsx
@@ -15,7 +15,7 @@ import {
 import { StudyRow } from './study-row'
 import { StudiesTableView } from './studies-table-view'
 import {
-    ACTIVE_PROPOSAL_STATUS,
+    ACTIVE_PROPOSAL_STATUSES,
     Audience,
     FINAL_STATUS,
     REVIEWER_ACTION_STATUSES,
@@ -46,10 +46,15 @@ function filterStudiesForUser(studies: StudyRowType[], audience: Audience, userI
     )
 }
 
-function needsRefresh(studies: StudyRowType[]): boolean {
+function needsRefresh(studies: StudyRowType[], audience: Audience): boolean {
+    // Two independent reasons to keep polling:
+    //  - study-level: a researcher is awaiting a DO decision (PENDING-REVIEW typically has no job
+    //    yet, so the job-level check below can't catch it). Reviewers are excluded because
+    //    PENDING-REVIEW is their own next action — re-fetching won't change until they act.
+    //  - job-level: an in-flight (non-final) job, regardless of audience or study status.
     return studies.some(
         (study) =>
-            ACTIVE_PROPOSAL_STATUS.includes(study.status) ||
+            (audience === 'researcher' && ACTIVE_PROPOSAL_STATUSES.includes(study.status)) ||
             study.jobStatusChanges.some((change) => !FINAL_STATUS.includes(change.status)),
     )
 }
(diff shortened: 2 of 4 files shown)


src/app/[orgSlug]/dashboard/page.tsx
```diff
@@ -47,7 +47,7 @@ export default async function OrgDashboardPage(props: { params: Promise<{ orgSlu
                         : undefined
                 }
                 showNewStudyButton={!isEnclave}
-                showRefresher={isEnclave}
+                showRefresher
```
Comment (nathanstitt) [inline src/app/[orgSlug]/dashboard/page.tsx:50]: This now enables the refresher unconditionally, so the **enclave** (reviewer) dashboard also begins polling on every `PENDING-REVIEW` study. But by the same reasoning the PR uses to exclude `CHANGE-REQUESTED`/`DRAFT` ("the next action there is the researcher's own"), on the enclave view `PENDING-REVIEW` is the *reviewer's* own next action — the status won't flip until they themselves act, so the 90s poll there mostly re-fetches unchanged data. The researcher-side fix only needs the lab dashboard to refresh. Worth a sanity check that we actually want the enclave dashboard polling here too — if not, `showRefresher={isEnclave || hasActiveProposal}`-style gating (or just leaving enclave as-is) would keep the blast radius to the researcher case this ticket is about.

src/components/dashboard/studies-table/index.tsx
```diff
@@ -46,7 +47,11 @@ function filterStudiesForUser(studies: StudyRowType[], audience: Audience, userI
 }
 
 function needsRefresh(studies: StudyRowType[]): boolean {
-    return studies.some((study) => study.jobStatusChanges.some((change) => !FINAL_STATUS.includes(change.status)))
+    return studies.some(
+        (study) =>
+            ACTIVE_PROPOSAL_STATUS.includes(study.status) ||
```
nathanstitt: Nice — keying off `study.status` directly is the right fix for the no-jobs case. One edge worth thinking through: this returns `true` purely on study status regardless of job state, while the second clause keys off job state regardless of study status. They don't overlap badly today, but a study that's `APPROVED`/`REJECTED`/`ARCHIVED` with an in-flight (non-final) job will still poll via the second clause, and a `PENDING-REVIEW` study always polls via the first. That seems intended, just flagging that `needsRefresh` now blends study-level and job-level lifecycles in one `some()` — a short comment on why both axes matter would help the next reader.

src/components/dashboard/studies-table/types.ts
```diff
@@ -42,6 +42,9 @@ export type StudiesTableProps = {
 // Status changes that indicate job is in a final state (no refresh needed)
 export const FINAL_STATUS: StudyJobStatus[] = ['CODE-REJECTED', 'JOB-ERRORED', 'FILES-APPROVED', 'FILES-REJECTED']
 
+// Proposal statuses where the researcher is awaiting a DO decision — keep auto-refresh polling.
+export const ACTIVE_PROPOSAL_STATUS: StudyStatus[] = ['PENDING-REVIEW']
```
nathanstitt: Tiny naming nit: this holds an array of statuses but is named singular `ACTIVE_PROPOSAL_STATUS` (matching the existing singular `FINAL_STATUS` above, so it's consistent with the file). If we ever extend it — e.g. `INITIATED`-equivalents or other awaiting-decision states — a plural `ACTIVE_PROPOSAL_STATUSES` would read better. Not blocking.

src/components/dashboard/studies-table/studies-table.test.tsx
```diff
@@ -93,6 +93,10 @@ vi.mock('@/server/actions/study.actions', () => ({
     fetchStudiesForOrgAction: vi.fn(() => mockStudies),
 }))
 
+// The action's real return type carries many more fields than these fixtures need; the default
+// factory mock above sidesteps that, but mockResolvedValueOnce is strict, so cast through it.
+type OrgStudies = Awaited<ReturnType<typeof fetchStudiesForOrgAction>>
```
nathanstitt, line 98: Good call documenting the cast. Since both new tests reuse `mockStudies[n]` and only override `status`/`jobStatusChanges`, a small shared helper (e.g. `studyWith({ status, jobStatusChanges })`) could drop the repeated `as unknown as OrgStudies` and the full `<StudiesTable .../>` block — purely optional cleanup.

nathanstitt (approved):
looks good, mostly just tiny things below that I think you can ignore if you want

PR #831, opened 2026-06-25
OTTER-628: faster e2e + unit tests (fake Clerk, prebuilt server, parallel/sharded CI)

nathanstitt:
Replaces the e2e auth stack and overhauls how the e2e suite runs: an in-app Clerk **fake** replaces the network `clerk-stub` (which was blocked by a dev-browser handshake loop), the suite runs against a **prebuilt server** instead of `next dev`, and a batch of flakiness/speed/structure fixes bring it to a stable, fully-parallel, zero-retry state. Production builds are untouched — the fake is aliased in only when `E2E_FAKE_CLERK` is set.

~38 files changed vs main (the e2e work); 12 new files under `src/lib/clerk-fake/`.

## 1. In-app Clerk fake (replaces clerk-stub)
- **`src/lib/clerk-fake/`** — fakes the server surface (`auth`/`currentUser`/`clerkClient`/`clerkMiddleware`/`createRouteMatcher`) and the client surface (`ClerkProvider` + `useUser`/`useAuth`/`useSession`/`useClerk`/`useSignIn`/`useReverification`), backed by a single `__e2e_role` cookie + fixtures that mirror the DB seed. No Clerk server, no network, no handshake.
- The real sign-in form runs against a faked `useSignIn` state machine (honors the steps: password → `needs_second_factor` → SMS `424242` → complete).
- Aliased via `next.config.ts` **`turbopack.resolveAlias`** (the app is Next 16 / Turbopack — a webpack alias is a no-op).
- **Retired clerk-stub entirely**: removed `bin/clerk-stub-test`, the stub global setup/teardown, `tests/clerk.ts`, `bin/delete-clerk-test-users.ts`, the `@clerk/testing` dep, and (from CI) the editor-service container + all `CLERK_*` secrets.

## 2. How the suite runs (the real reliability/speed win)
- **Prebuilt standalone server, not `next dev`.** `next dev`'s lazy per-route compilation made long specs exceed their timeout — it failed both in parallel *and* serially (a serial run took ~15 min and still failed). `bin/app-test` / `bin/ci-server` now `next build` + serve the standalone server (matching CI), with `CI=true` (skips the prod-only Mailgun guard, disables Coder) and `SINGLE_USER_EDITING=t` (standalone editor — no hocuspocus websocket).
- **`unset AWS_PROFILE`** in `bin/app-test` — `.env.test` set it, making the AWS SDK try expired SSO instead of the local SeaweedFS creds, which hung the upload-heavy specs ~30s.
- **`auth setup` project → `globalSetup`.** The old auth-setup *project* spun up a whole worker pool + barrier just to write three cookie files; it's now a `globalSetup` that writes the storageState JSON directly and warms routes once with a single browser.
- **`workers: '100%'`** — tests are mostly I/O-wait, so use all cores (on a 4-vCPU CI runner that's 4 workers vs Playwright's default of 2).
- **`retries: 0` + `fullyParallel` everywhere, including CI** — enforced, and documented as a hard rule in `CLAUDE.md` (a test that only passes on retry is a bug to fix at the source).

## 3. Flakiness fixed at root (not via retries)
- org-admin "add code env" modal: retry the open across React Query refetches (the refetch detached the button mid-click).
- Split the 3-phase "study creation + upload" test into three independent seeded tests so no single test risks its timeout.
- Removed a stale `+30s` per-test timeout, a dead no-op MFA click, and the `mode: 'serial'` on the MFA describe (its 3 reads are independent → parallel).
- Coverage (V8 start/stop per test) gated behind `E2E_COVERAGE` (default on) — measured to confirm it's not the bottleneck.

## 4. Results-review tests updated for main's encrypted-results rework (#764, merged in)
- `seedCodeApprovedJobReady` no longer seeds `JOB-READY` (old polling model) — its timestamp outranked the upload's `RUN-COMPLETE` and stranded the reviewer off the results screen; the results upload now owns the terminal status.
- `insertSubmittedJob` stamps seeded statuses in the *past* (was `now + i*1000`) so a later real insert (`RUN-COMPLETE` at server-now) reliably sorts after them.
- Resolved the merge in favor of main's all-or-nothing approve UI (decrypt → approve re-wraps for the researcher; no per-file checkbox).

## Review feedback addressed
- Removed `window-bridge.ts` (overlapped `global.setup.ts`); its only remaining consumer, `e2eSignOut`, now clears the cookie directly via `context.clearCookies()`.

## Verification
- Full e2e suite: **0 failures across many full runs at `retries: 0`**, fully parallel, ~16-25s, **zero Clerk / lvh.me / Coder network**.
- `pnpm run checks` green; 1654 unit tests pass.

## Notes / scope
- Multi-tab collaboration specs removed (need a live websocket, incompatible with `SINGLE_USER_EDITING`).
- "Code upload via IDE" spec removed (needs the external Coder service — out of scope).
- Pre-existing SSR `ReferenceError: window is not defined` in the invite-email path noted but not fixed (caught/non-fatal, unrelated to this work).

---

This branch also carries earlier OTTER-628 work it builds on: the **study-screen state machine** (#826), **single-user editing + seed fixes** (#824), and the **DB-seed + storageState e2e refactor**.


src/lib/clerk-fake/window-bridge.ts
chrisbendel, line 1: there is some overlap of logic here and in `global.setup.ts` - should we do away with this file?

nathanstitt: probably so, good catch

nathanstitt: Good catch — done. The bridge's only remaining consumer after the suite cleanup was `e2eSignOut`, so I removed `window-bridge.ts` entirely: `e2eSignOut` now clears the `__e2e_role` cookie directly via `context.clearCookies()`, and the sign-in form already uses the React hooks rather than `window.Clerk`. (commit 12d48364)

chrisbendel (approved):
Nice the tests are so much faster now 👍

PR #834, opened 2026-06-27
OTTER-614: Refine researcher navigation flows

stetradis:
## Summary

Fixes researcher back/forward navigation (OTTER-614) now that `/view` always resolves to the study's latest screen from state alone.
- **Adds `/view/code`** — a read-only post-decision code route (`Routes.studyViewCode`) backed by `resolveResearcherCodeScreen`. Researchers can walk back to the code step from results; the route 404s if the study hasn't reached the code stage (no forward jumps).
- **Results is no longer terminal** — the results screen's "Previous" link goes to `/view/code`; the code screen shows "Proceed to step 5" (→ plain `/view`) once results exist instead of ending at the dashboard.
- **Fixes agreements navigation** — "Proceed to Step 4" after code submission targets `/view/code` instead of `/view`, which would skip straight to results for an advanced study. "Previous" on agreements preserves `returnTo=org` on the `/submitted` link.
- **Unifies proposal feedback with `/submitted`** — the `proposal-feedback` screen renders the same `ProposalSubmitted` page (via shared `loadProposalSubmittedData`); removes the legacy `ResearcherProposalView` component.
- **Simplifies screen resolution** — drops the unused `step` parameter from `resolveScreen`; screen selection is purely `state → screen` with no URL input.
- **Fix incorrect Back route** — Cancel on draft Step 1 navigated to Step 2 instead of dashboard — handleCancel used router.back(), which followed browser history; after a Step 1 → Step 2 → Previous → Step 1 round-trip, "back" pointed at the proposal page. Replaced with an explicit router.push(Routes.dashboard).


src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx
```diff
@@ -235,6 +253,7 @@ export function CodePostDecisionView({
     dashboardHref,
     returnTo,
     latestJobStatus,
+    resultsHref,
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx:256]: Non-blocking, while we're threading `returnTo` here: just below, `proposalHref = Routes.studySubmitted({ orgSlug, studyId: study.id })` drops `returnTo`, while the sibling `previousHref = Routes.studyResearcherAgreements({ ..., returnTo })` keeps it. That `proposalHref` backs both the "Study proposal" breadcrumb and the SubmittedCodePanel link, so a researcher who walked back here under `returnTo=org` would bounce out of org scope on either. Now that `studySubmitted` accepts `returnTo`, could pass it here too for consistency.

nathanstitt (approved):
looks good,  not sure if the returnTo things is really that important

PR #835, opened 2026-06-29
OTTER-602: don't reset submit-enable when launching IDE after manual upload

Rod0352:
### Problem
On the initial proposal code submission page, a researcher who uploads files manually and then clicks the IDE option sees every file's **Last updated** flip to "Never" and the **Submit** button go disabled — staying disabled until they edit a file in the IDE. The manual upload's work is effectively discarded from the UI's point of view.

### Fix
`ensureRoundJobForLaunch` now skips the re-anchor when the workspace already holds files. `createUserAndWorkspaceAction` checks the workspace on disk (new `studyHasWorkspaceFiles` helper, mirroring `listWorkspaceFilesAction`'s filters) and passes `hasWorkspaceFiles` in. The option defaults to `false`, so the OTTER-601 behavior — re-anchoring on a genuinely empty round so IDE edits enable Submit — is unchanged.

### Tests
- `mutations.test.ts`: `ensureRoundJobForLaunch` does not re-anchor when files exist; still re-anchors on an empty round.
- `workspaces.actions.test.ts`: `createUserAndWorkspaceAction` leaves the baseline untouched when a file is present on disk, and re-anchors when none is.

### DIFF
--- a/src/server/actions/workspaces.actions.test.ts
+++ b/src/server/actions/workspaces.actions.test.ts
@@ -233,4 +233,67 @@ describe('Workspace Actions', () => {
             expect(result?.fileNames).toEqual(['main.r'])
         })
     })
+
+    // OTTER-602: launching the IDE must not reset submit-enable when files were already uploaded
+    // manually. The re-anchor that fixed OTTER-601 (advance createdAt so post-launch edits enable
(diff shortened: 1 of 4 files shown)


nathanstitt (approved):
thanks Rodney!

PR #836, opened 2026-06-29
Update code explanation prompt

chrisbendel:


rnathuji (approved):

nathanstitt (approved):
lgtm, I do wonder if we should move these into the prompts in SI admin panel eventually, I suspect we'll continue to modify them

PR #837, opened 2026-06-29
OTTER-598 follow-up: fix study-view, keep researcher on code-approved page when a job errors before reviewer files decision

therealmarv:
## Why

Follow-up to OTTER-598 (QA comment 43898). The original card's legacy `CodeOnlyView` fallback is gone (replaced by the study-screen state machine), but QA re-reported a related mismatch:

> Code errored but the data provider hasn't reviewed → the researcher still sees the **Code approved** status, yet opening **Study Details** shows the generic results/Study Details page instead of the post-code-approval page.

### Root cause

The **pill** deliberately hides `JOB-ERRORED` from researchers until a reviewer records a `FILES-*` decision (it falls back to "Code approved"). But the **researcher screen-routing rule** fired `study-results` on any `hasResults`, and `JOB-ERRORED ∈ STUDY_RESULTS_JOB_STATUSES`. So the pill and the `/view` screen disagreed: pill says "Code approved", page jumps to results.

## What changed

- **`state.ts`** — new exported predicate `isErroredResultHiddenFromResearcher(s)` = `resultsErrored && !resultsApproved && !resultsRejected`. Single source of truth.
- **`researcher-screen-rules.ts`** — the `study-results` rule now excludes the errored-hidden case, so it falls through to `code-approved` (the post-code-approval page QA expects).
- **`pill.ts`** — refactored its inline `hideErrored` to consume the same predicate, so pill and screen can't drift again.
- **`code-decision-screen.tsx`** — `showStudyCode` now also hides the code listing when the error is hidden. A packaging-stage failure (`containerizer` posts `JOB-ERRORED` with no `JOB-RUNNING` substatus, so `isExecuting` is false) would otherwise re-expose the code listing in a state we present as "approved / results pending."

Reviewer routing is intentionally untouched: reviewers should see errored results immediately.

## Tests

- `resolve.test.ts` — errored-unreviewed → `code-approved`; errored-after-files-decision → `study-results`.
- `consistency.test.ts` — added the errored-unreviewed fixture so the dashboard "View" link must resolve to a real screen.
- `page.test.tsx` — removed `JOB-ERRORED` from the OTTER-538 "renders StudyDetailsResearcher" `it.each` (the test that pinned the bug); added two scenarios: `CODE-APPROVED → JOB-RUNNING → JOB-ERRORED` and the packaging-error path `CODE-APPROVED → JOB-ERRORED` (no `JOB-RUNNING`), both asserting `CodePostDecisionView` with `showStudyCode === false`.

`pnpm run checks` passes (typecheck, lint, action validation).


src/app/[orgSlug]/study/[studyId]/_screens/code-decision-screen.tsx
```diff
@@ -16,6 +16,11 @@ export async function CodeDecisionScreen({ study, raw, orgSlug, dashboardHref, r
         state.codeDecision === 'CODE-APPROVED' || state.isExecuting ? 'CODE-APPROVED' : state.codeDecision
     if (decisionStatus === null) notFound()
 
+    // A hidden JOB-ERRORED (e.g. a packaging failure before any JOB-RUNNING substatus) is presented to
+    // the researcher as "approved / results pending"; keep the code listing hidden as during execution,
+    // so a packaging error doesn't re-expose it (OTTER-598 follow-up).
+    const hideStudyCode = state.isExecuting || isErroredResultHiddenFromResearcher(state)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/code-decision-screen.tsx]: `isErroredResultHiddenFromResearcher` is called unconditionally here, with no `role === 'researcher'` guard — unlike `pill.ts` which gates it on role. That's actually correct: reviewers route to `reviewer-study-results` for any `hasResults` (reviewer rule 1) and never reach `CodeDecisionScreen` in an errored state, so the role-agnostic call is safe. The only snag is that the helper's name reads `...FromResearcher` while this call site has no researcher context, so a future reader might wonder whether a reviewer could hit this path and wrongly get the code hidden. A one-line `// reviewers route to reviewer-study-results, so this screen is researcher-only here` would close that gap.

src/lib/study-screen/researcher-screen-rules.ts
```diff
@@ -1,11 +1,14 @@
...
 // Researcher Tier-2 rules. Order = display precedence (see spec §6). First match wins. Each entry
 // pairs the screen it routes to with the condition that selects it; the leaf view owns its own
 // back/forward buttons.
 export const RESEARCHER_SCREEN_RULES = [
-    // Results have landed: results-only Study Details.
-    ['study-results', { when: (s) => s.hasResults }],
+    // Results have landed: results-only Study Details. A bare JOB-ERRORED is excluded until a reviewer
+    // records a FILES-* decision (isErroredResultHiddenFromResearcher) — until then the researcher
+    // holds on the code-approved page below, matching the "Code approved" pill (OTTER-598, 43898).
+    ['study-results', { when: (s) => s.hasResults && !isErroredResultHiddenFromResearcher(s) }],
```
nathanstitt, line 11: One edge to sanity-check: this excludes the hidden-errored case from `study-results` and relies on the `code-approved` rule below catching it via `codeDecision === 'CODE-APPROVED' || isExecuting`. That holds for both scenarios the tests cover. But is there a state where `resultsErrored` is true (so we exclude here) yet `codeDecision` is not a live `CODE-APPROVED` and `isExecuting` is false — e.g. a resubmission that dropped the stale decision (`dropStale`) while an old `JOB-ERRORED` still sits on the job? That would fall through to `code-under-review`, which is arguably the right screen, but it'd be worth a test or a comment confirming `JOB-ERRORED` can't coexist with a non-approved live decision on the same job in practice.

src/lib/study-screen/state.ts
```diff
@@ -114,3 +114,12 @@ export function projectStudyState(raw: RawStudyState): StudyState {
         latestJobStatuses: [...jobStatuses].sort(), // stable order for deterministic output/tests
     }
 }
+
+// OTTER-598 follow-up: a JOB-ERRORED result stays hidden from the RESEARCHER until a reviewer records
+// a FILES-* decision (errored-result triage is the reviewer's). While hidden, the researcher's pill
+// reads "Code approved" (resolvePillStatus's hideErrored) and the /view screen must hold on the
+// code-approved page — NOT jump to the results/Study Details screen. Single source of truth shared by
+// the pill and RESEARCHER_SCREEN_RULES so the two can't drift (the mismatch QA re-reported in 43898).
+export const isErroredResultHiddenFromResearcher = (
```
nathanstitt, line 125: Really like that this is a single exported predicate with a thorough why-comment, and that both the pill and the screen rules consume it — the "can't drift again" property is exactly what was missing the first time around. Minor: the param type is `Pick<StudyState, 'resultsErrored' | 'resultsApproved' | 'resultsRejected'>`, which is great for testability, but every current caller passes a full `StudyState`/`StudyState`-shaped object — so the `Pick` mostly documents intent. Fine to keep; just noting it's narrower than any caller needs today.

nathanstitt (approved):
good refactor to add the `isErroredResultHiddenFromResearcher`

PR #838, opened 2026-06-29
OTTER-638: Fix duplicate ID error blocking resubmission code review

therealmarv:
## Problem

A legitimate reviewer could not submit a decision on **resubmitted** code. The system wrongly rejected it as a duplicate with *"another reviewer has already submitted a decision for this study code"*, even when it was the same reviewer and no race occurred.

**Root cause.** `study_review_comment` had a unique constraint on `(study_job_id, review_kind)`, i.e. one CODE decision per job for the job's lifetime. But a `CODE-CHANGES-REQUESTED` resubmit revises the **same job in place** (OTTER-316), so the round-2 decision collided with the round-1 decision row already on that job. The collision was caught and surfaced as the OTTER-471 race-loser message.

Replication: DP requests changes → researcher resubmits → original reviewer tries to approve → blocked.

## Fix

Scope code-review uniqueness to the review **round** instead of the job, so each round on the same job gets its own decision row while a genuine same-round two-reviewer race still collides (preserving the OTTER-471 guard).

- **Migration** `1780200000000_study_review_comment_round.ts`: add `study_review_comment.round` (`integer NOT NULL DEFAULT 1`), backfill existing CODE rows to their historical study-wide round, drop `study_review_comment_one_code_review_per_job`, add `study_review_comment_one_code_review_per_round` on `(study_job_id, review_kind, round)`. The backfill casts `status::text` to avoid Postgres's "unsafe use of new value of enum type" within the single migration transaction.
- **`codeSubmissionVersion`** gains an optional `db` executor so the round is computed inside the action's transaction (and always matches the number the UI derives).
- **`submitCodeReviewDecisionAction`** computes `round = codeSubmissionVersion(studyId, db)` and stores it on the decision. The unique-violation catch is unchanged; post-fix it fires only on a true same-round race.
- **`getCodeReviewFeedbackAction`** labels reviewer-feedback entries by the stored `round`.

`round` is the study-wide submission version (existing `codeSubmissionVersion` semantics), so it doubles as the display version rather than introducing a third numbering scheme.

## Code-review follow-ups

High-effort reviews surfaced several issues, all now fixed:

- **Resubmission-note version divergence (would hide the resubmission UI).** The reviewer page derives `isResubmission`/heading from the max entry version; with only reviewer decisions on the study-wide `round` and notes still on job-creation order, a same-job round-2 review showed all entries as v1 → rendered as a first submission and hid prior feedback. Fixed by versioning the note on the same study-wide round: new `study_job.resubmission_round` column, set by `resubmitStudyCodeAction` (= `codeSubmissionVersion` at resubmit time) and backfilled; `getCodeReviewFeedbackAction` reads it (falling back to job-ordinal only for legacy rows). No UI component changes were needed.
- **Duplicate resubmission-note entries.** The feedback query joined every `CODE-SUBMITTED` row, so a same-job resubmit (which appends a second `CODE-SUBMITTED`) fanned one job into multiple rows and emitted duplicate note entries (identical id/body/version). Fixed by selecting the note author from the job's single latest `CODE-SUBMITTED` via a `leftJoinLateral`, so each job yields exactly one row.
- **Migration rollback was not dedupe-safe.** `down()` re-added the per-job unique constraint, which fails once a job legitimately has multiple round rows. `down()` now collapses each `(job, kind)` group to its latest round before restoring the old constraint (intentionally lossy, as a rollback to one-decision-per-job must be).

Also addressed: dropped a now-dead `studyJobId` select, and updated a stale race-guard comment that referenced the old constraint shape.

## Tests

`src/server/actions/study.actions.test.ts`:

- New: full cycle (changes-requested → resubmit → approve), full cycle with round-2 reject, distinct/increasing version labels per round, a note sharing its round version with that round's decision (and appearing exactly once despite two `CODE-SUBMITTED` rows), a second decision without a resubmit still rejected as "already decided", and round numbering across separate jobs (post `FILES-APPROVED`).
- Updated: the race-loser and unique-constraint tests are now round-aware (same round blocks; different rounds are allowed).

`src/server/actions/study-request.test.ts`: the same-job resubmit test now asserts `resubmission_round` is recorded as 2.

## Verification

- `pnpm run checks` (typecheck + lint + action validation): pass.
- Full unit suite: 1719 tests across 190 files, all passing.

## Out of scope

Removing the legacy job-creation-order `jobVersion` fallback entirely once all rows carry `resubmission_round`.

### DIFF
--- a/src/server/actions/study.actions.ts
+++ b/src/server/actions/study.actions.ts
@@ -773,6 +773,11 @@ export const submitCodeReviewDecisionAction = new Action('submitCodeReviewDecisi
         // keeps the same job, so uniqueness is scoped to (studyJobId, reviewKind, round) and each
         // round gets its own decision row (OTTER-638). The round-1 decision is written before its
         // CODE-CHANGES-REQUESTED status below, so round 1 → 1 and a post-resubmit decision → 2.
+        // Load-bearing: codeSubmissionVersion counts CODE-CHANGES-REQUESTED/FILES-APPROVED/FILES-REJECTED
+        // but NOT CODE-REJECTED, which is safe only because reject is terminal (not in
+        // CODE_RESUBMITTABLE_JOB_STATUSES) so no second decision can follow it. If reject ever becomes
+        // resubmittable, codeSubmissionVersion must count it too, or the round won't advance and the
+        // next decision would collide on the round-scoped unique constraint.
         const round = await codeSubmissionVersion(studyId, db)
 
         try {
@@ -878,17 +883,18 @@ export const getCodeReviewFeedbackAction = new Action('getCodeReviewFeedbackActi
         // label even when a same-job resubmit revises in place (OTTER-316/638) and the job ordinal
         // would not advance. jobVersion (job creation order) is only a fallback for legacy rows whose
         // resubmissionRound was not backfilled. studyJob has no userId column; the author of the
-        // resubmission note is the user recorded on the job's latest CODE-SUBMITTED. A same-job
-        // resubmit appends more than one CODE-SUBMITTED, so joining the rows directly would multiply
-        // the job into duplicate codeJobs rows (and duplicate note entries) — a lateral join to the
-        // single latest submission keeps each job to one row.
+        // resubmission note is the user recorded on the job's latest CODE-SUBMITTED (left-joined so
+        // the timestamp and author are independent: a null/deleted user does not lose the submission
+        // row). A same-job resubmit appends more than one CODE-SUBMITTED, so joining the rows directly
+        // would multiply the job into duplicate codeJobs rows (and duplicate note entries); a lateral
+        // join to the single latest submission keeps each job to one row.
         const codeJobs = await db
             .selectFrom('studyJob')
             .leftJoinLateral(
                 (eb) =>
                     eb
                         .selectFrom('jobStatusChange as cs')
-                        .innerJoin('user as submitter', 'submitter.id', 'cs.userId')
+                        .leftJoin('user as submitter', 'submitter.id', 'cs.userId')
                         .select([
                             'cs.userId as authorId',
                             'submitter.fullName as authorName',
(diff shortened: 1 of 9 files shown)


src/server/actions/study.actions.ts
```diff
@@ -767,6 +769,12 @@ export const submitCodeReviewDecisionAction = new Action('submitCodeReviewDecisi
 
         const claimedJob = await claimInitialCodeReviewJob({ studyId })
 
+        // Round = study-wide submission version at decision time. A same-job resubmit (OTTER-316)
+        // keeps the same job, so uniqueness is scoped to (studyJobId, reviewKind, round) and each
+        // round gets its own decision row (OTTER-638). The round-1 decision is written before its
+        // CODE-CHANGES-REQUESTED status below, so round 1 → 1 and a post-resubmit decision → 2.
+        const round = await codeSubmissionVersion(studyId, db)
```
nathanstitt, line 781: Nice — computing `round` here (after `claimInitialCodeReviewJob` but before the `CODE-CHANGES-REQUESTED`/`CODE-REJECTED` insert below) is what makes round 1 -> 1 and the post-resubmit decision -> 2. Worth noting this leans on `codeSubmissionVersion` *not* counting `CODE-REJECTED`: since reject is terminal (no resubmit), a same-round second decision can never legitimately occur after a reject, so that's fine — but it's a load-bearing assumption that's easy to break if reject ever becomes resubmittable. Might be worth a one-liner pinning that down.

src/server/actions/study.actions.ts
```diff
@@ -929,7 +954,7 @@ export const getCodeReviewFeedbackAction = new Action('getCodeReviewFeedbackActi
                 criteria: null,
                 createdAt: j.createdAt,
```
nathanstitt: One thing worth a look: the note entry's `createdAt` is `studyJob.createdAt` (job creation time), but for a same-job resubmit the round-2 note is *written* at resubmit time — much later than job creation. In the desc-by-`createdAt` sort below, that places the round-2 resubmission note at the round-1 timeline position (older than round 1's decision), so it could render below round-1 feedback rather than above it. We now correctly *label* it v2 via `resubmissionRound`, but the timeline slot still uses the stale job-creation timestamp. Should this use the latest `CODE-SUBMITTED.createdAt` from the lateral join instead, so position matches the round it opened?

nathanstitt (approved):
looks good Marvin, but check the date below to make sure we're looking at the correct one.

PR #839, opened 2026-06-29
Add QA cleanup API to delete users and studies on non-prod

nathanstitt:
Adds two QA-only HTTP endpoints so QA can clean up after themselves on non-production environments.

## Endpoints
- `DELETE /api/qa/users/[userId]` — fully deletes a user (DB records + backing Clerk account)
- `DELETE /api/qa/studies/[studyId]` — fully deletes a study (DB records + S3 folder)

Both are hard-gated to non-production (`PROD_ENV === false`) and require a valid Clerk session token from an SI admin. Responses: `403` in prod, `401` if unauthenticated, `403` if not an SI admin, `404` if the id is missing, `200` on success.

## Details
- All logic lives in `src/server/qa-cleanup.ts` (out of the route files). `requireQaAdmin()` reuses the existing `marshalSession` path to verify `isSiAdmin`.
- Deletion is FK-safe: relations without `ON DELETE CASCADE` are removed manually. User deletion first removes studies the user owns (researcher/pi/reviewer FKs don't cascade) and clears the `RESTRICT`-constrained review comments before the user row; a missing Clerk account is logged, not fatal.
- `onDeleteStudyAction` now delegates to the shared `deleteStudyCompletely` helper (behavior unchanged).
- Deletion is by id with no test-pattern restriction.

## Tests
- New `src/server/qa-cleanup.test.ts` covers the auth guard (prod / unauthenticated / non-admin / admin) and real DB deletion of users and studies.
- `lint:fix`, `test:unit`, and `checks` all pass.

## Also in this PR
- **E2E build fix:** the E2E Clerk fake (`src/lib/clerk-fake/server.ts`) now exports `verifyToken`, which `qa-cleanup.ts` imports. Under `E2E_FAKE_CLERK` the fake is aliased in for `@clerk/nextjs/server`, so the missing export was breaking the e2e `next build`. The e2e suite never hits the `/api/qa/*` routes; the export resolves the admin fixture's claims to stay consistent with the fake's `auth()`.
- **QA seed accounts:** seeds the persistent QA login accounts (`qa-review+admin` / `+dp` / `+researcher@safeinsights.org`) in the DB seed. Refactored `TEST_USERS` to be keyed by a unique name rather than by role (so multiple users can share a role), with role→org memberships in a single `ORG_MEMBERSHIPS_BY_ROLE` map. The QA accounts already exist in Clerk; the seed only wires up SI DB access (user row + memberships + key), mirroring the existing same-role users. Idempotent, existing users untouched.


src/server/qa-cleanup.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Delete the Clerk account. A missing Clerk user is not an error — the DB record
+ * is already gone, so log and continue.
+ */
+async function deleteClerkUser(clerkId: string) {
+    try {
+        const client = await clerkClient()
+        await client.users.deleteUser(clerkId)
+    } catch (error) {
```
therealmarv, line 178: This currently treats every Clerk delete failure as non-fatal, so the route can return `200` after deleting the DB user while the backing Clerk account is still active. I think this should only ignore the known idempotent case, a Clerk 404/not-found. Other errors, auth/config/network/5xx, should be rethrown so the caller sees that the full cleanup did not complete. Clerk documents `deleteUser(userId)` as an awaited backend call returning a `Promise<User>`, with backend calls rejecting on API/network errors, so swallowing all errors makes the endpoint report a false success.

src/server/qa-cleanup.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (!study) throw new QaCleanupNotFoundError(`study ${studyId} not found`)
+
+    await deleteStudyCompletely(db, study.orgSlug, study.studyId)
+}
+
+/**
+ * Fully delete a user: any studies they own (researcher/pi/reviewer FKs do not
+ * cascade), their dependent rows, the user row, and the backing Clerk account.
+ */
+export async function deleteUserById(db: Kysely<DB>, userId: string) {
```
therealmarv, line 120: The DB side of this cleanup should be transactional. This method deletes studies and then several user-dependent rows before deleting the user; if any statement fails midway, we can leave a partially deleted user/study graph behind. Kysely's `db.transaction().execute(async (trx) => ...)` automatically commits on success and rolls back on thrown errors, so the better shape is to run the database deletes through a transaction and only let external cleanup like Clerk/S3 happen in a deliberate order around that.

therealmarv (changes requested):
Requesting changes on two cleanup correctness blockers.

The main issue is that Clerk deletion failures are swallowed wholesale, so the endpoint can return success while the Clerk user still exists. The second is that the multi-table DB deletion is not transactional, so a mid-cleanup failure can leave partial state behind. Since these endpoints are destructive and meant to provide complete cleanup, I think both should be fixed before merge.

src/server/qa-cleanup.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    const authHeader = (await headers()).get('Authorization') || ''
+    const token = authHeader.startsWith('Bearer ') ? authHeader.slice('Bearer '.length).trim() : ''
+    if (!token) {
+        return { ok: false, status: 401, message: 'Authentication required' }
+    }
+
+    let claims
+    try {
+        claims = await verifyToken(token, {})
```
therealmarv: `verifyToken(token, {})` is called with an empty options object, and I think that makes the whole admin guard fail closed in any real environment.

Here is the problem in plain terms. The token in the `Authorization: Bearer ...` header is a signed badge, and to trust it the server has to check its signature against Clerk's keys. `verifyToken` needs to be handed a key to do that. Unlike `auth()` and `clerkClient()`, the standalone `verifyToken` does **not** read `CLERK_SECRET_KEY` (or `CLERK_JWT_KEY`) from the environment on its own. It only uses the key you pass in the options object. With `{}` it has no key, cannot resolve the JWK, and throws ("Failed to resolve JWK during verification"). The `catch` on the next line then turns that into a `401`.

So on any real deployment (staging/QA with real Clerk) this line rejects **every** request with `401`, including the SI admins these endpoints are built for. It only appears to work under `E2E_FAKE_CLERK`, where the fake `verifyToken` ignores the token and returns admin claims, and the e2e suite never calls these routes anyway.

I confirmed this three ways: Clerk's official `verifyToken` docs (they say to set `secretKey` from `CLERK_SECRET_KEY` and pass it in, and every example passes a key explicitly), and by tracing the installed `@clerk/backend` 6.39.4 source (empty options hits the "no key" branch and the public wrapper throws).

Fix, passing the secret key the same way the rest of the repo reads Clerk config:

```suggestion
        claims = await verifyToken(token, { secretKey: process.env.CLERK_SECRET_KEY })
```

Why the unit tests miss it: `qa-cleanup.test.ts` mocks `verifyToken` (`vi.mock` + `mockResolvedValue`), so the tests only ever exercise a stub that returns claims. The real key-resolution path never runs, so the suite stays green while the real endpoint is non-functional.

therealmarv (changes requested):
Thanks @nathanstitt, the two blockers from the last review are both resolved: Clerk deletion now only tolerates a genuine 404/`resource_not_found` and rethrows everything else, and the multi-table DB deletes run inside a transaction with the S3/Clerk cleanup deliberately after commit. Both paths have test coverage now too. Appreciate the turnaround.

Requesting changes on one more issue I found in this pass. The SI-admin auth guard calls `verifyToken(token, {})` with no key, so it throws in any real-Clerk environment and the guard returns `401` for everyone, which means the endpoints never actually authenticate in the deployed non-prod environments they are meant for. Full explanation and a one-line fix are in the inline comment on `src/server/qa-cleanup.ts`. Once that is in and merge conflicts resolved this looks good to me.

src/server/qa-cleanup.ts
nathanstitt: Thanks @therealmarv — good catch, and confirmed. I traced it through the installed `@clerk/backend` source to be sure: the exported `verifyToken` is `withLegacyReturn(inner)` (`if (errors) throw errors[0]`), and `inner` with empty options hits the no-key branch and returns a `"Failed to resolve JWK during verification."` error — so the call throws and our `catch` turns it into a `401` for every request. (One minor note: the installed version here is `@clerk/backend` 2.33.4 rather than 6.39.4, but the code path and conclusion are identical.)

Applied your one-line fix in `src/server/qa-cleanup.ts`, passing the secret explicitly:

```js
claims = await verifyToken(token, { secretKey: process.env.CLERK_SECRET_KEY })
```

On the test blind spot you flagged (the suite mocks `verifyToken`, so it couldn't tell `{}` from `{ secretKey }`): I added a regression guard that asserts the guard passes the `secretKey` option, checked by key rather than value since `CLERK_SECRET_KEY` is unset in the unit-test env. I verified it's not a tautology — reverting the source to `{}` makes it fail, and the fix makes it pass.

Merge conflicts with `main` are also resolved (all in the launch-IDE area, unrelated to this endpoint; the ctrl-click "open in same window" feature is preserved on top of main's refactor). `checks` is green and the full unit suite passes.

therealmarv (approved):
cool, thanks!

therealmarv (approved):

PR #846, opened 2026-06-30
OTTER-593: align study code panel with Figma & standardize file review UI

stetradis:
## Summary

Updates the Study Code panel to match the latest Figma designs and standardizes
the file review UI across the resubmit/edit flow. Action buttons now live in the
panel header alongside the study title, and the "Main file" column header (with
its info tooltip) is now owned by the table itself rather than passed in per-caller.

Also fixes a bug where the panel could flip back to the empty state after the
user had already edited files.

## Changes

**Relocate action buttons (match Figma)**
- Moved the **Edit files in IDE** (`LaunchIdeButton`) and **Upload files**
  (`UploadFilesButton`) controls out of `StudyCodeReviewView` and up into the
  `StudyCodePanel` header, right-aligned next to the study title.
- Buttons only render in the review state (not while loading or in the empty state).
- Added a divider between the header and the body; tightened header spacing.

**Standardize file review UI**
- `FileReviewTable` now owns its **Main file** column header and info tooltip
  directly, instead of accepting a `mainFileColumnHeader` ReactNode prop.
- Removed the now-redundant `mainFileColumnHeader` / `ideButtonTooltip`
  prop-drilling through `StudyCodePanel` → `StudyCodeReviewView` → `FileReviewTable`.
- The IDE-button tooltip copy is defined once in `StudyCodePanel`.
- Simplified `EditStudyCodeView` accordingly (dropped its local
  `MainFileColumnHeader` and tooltip constants).

**Bug fix: prevent empty-state flicker after edits**
- `useIDEFiles` now also gates `showEmptyState` on `!userEditedFiles`, so the
  panel no longer reverts to the empty view once the user has started editing.


src/components/study/file-review-table.tsx
```diff
@@ -52,21 +51,40 @@ function MainFileStar({ fileName, isSelected, onSelect }: MainFileStarProps) {
     )
 }
 
+const MAIN_FILE_TOOLTIP =
+    "If you're creating or uploading multiple files, please select your main file (i.e., the script that runs first)"
```
nathanstitt: Nit: when this tooltip moved here from EditStudyCodeView it lost its trailing period (the original ended with `runs first).`). Worth restoring for consistency with the other tooltip copy in this PR, which still ends in a period.

nathanstitt (approved):
thanks!

PR #847, opened 2026-06-30
add CLAUDE.md from edit as well as launch

kswanson33:
- Separate context generation logic into its own file
- Invoke generation outside the check that starter code files need
- add to dev.ts


nathanstitt (approved):
love the sentinel idea, that's a clever solution to 'did the user modify the file?' that works much better than just checking mtime or something like that.

PR #849, opened 2026-07-01
OTTER-610: standardize button alignment in modals

stetradis:
Removes explicit `justify` properties from button `Group` components in confirmation modals - default behavior is left-aligned. 

This change ensures a consistent default alignment for action buttons across these dialogs, improving overall UI uniformity.

### DIFF
--- a/src/components/modals/review-confirmation-modal.tsx
+++ b/src/components/modals/review-confirmation-modal.tsx
@@ -36,7 +36,7 @@ export const ReviewConfirmationModal: FC<ReviewConfirmationModalProps> = ({
         >
             <Stack>
                 {children}
-                <Group justify="flex-end">
+                <Group>
                     <Button variant="outline" onClick={onClose} disabled={isPending}>
                         Cancel
                     </Button>
--- a/src/components/dashboard/studies-table/delete-draft-button.tsx
+++ b/src/components/dashboard/studies-table/delete-draft-button.tsx
@@ -74,7 +74,7 @@ export function DeleteDraftButton({ study }: { study: StudyRow }) {
                         Please confirm that you are wanting to delete this proposal draft. Once this draft is deleted
                         you will not be able to recover it. This action cannot be undone.
                     </Text>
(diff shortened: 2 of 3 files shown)


src/components/modals/review-confirmation-modal.tsx
```diff
@@ -36,7 +36,7 @@ export const ReviewConfirmationModal: FC<ReviewConfirmationModalProps> = ({
         >
             <Stack>
                 {children}
-                <Group justify="flex-end">
+                <Group>
```
nathanstitt, line 39: Confirmed this works as intended — Mantine's `Group` defaults to `justify: "flex-start"`, so dropping the prop left-aligns the buttons.

One thing worth a sanity check before merging, though: app-wide the *prevailing* convention for action-button rows is actually right-aligned. `justify="flex-end"` shows up ~16 times across the app (study-review-buttons, post-feedback-view, edit-study-code-footer, researcher-profile-view ×3, position-form, review-uploaded-files, proposal-submitted, …). So standardizing these three modals to left-aligned makes them *consistent with each other* but *divergent from* most other button rows in the product. Since the PR framing is "UI uniformity," it's worth confirming the intended direction with the OTTER-610 design — if the target is left-aligned modal actions specifically, all good; if the goal is app-wide uniformity, flex-end might be the standard to converge on instead. Purely a design call, not a code issue.

nathanstitt (approved):
lgtm, but maybe we should update all the other locations as well? or do it globally?

src/components/modals/review-confirmation-modal.tsx
stetradis: Great callout, I'll confirm if the alignment should also apply globally to action buttons.

stetradis: Confirmed that the left-alignment should only apply to modal buttons, all other buttons should remain right-aligned - https://openstax.atlassian.net/browse/OTTER-610?focusedCommentId=44564

PR #850, opened 2026-07-01
OTTER-462: prompt instead of erroring when already signed in

Rod0352:
## OTTER-462: redirect instead of erroring when already signed in

**Problem:** When a Clerk session is still live on `/account/signin` (the app session can expire while Clerk's stays valid), submitting hit "You're already signed in." — blocking a user who was actually authenticated.

**Fix:** Send them onward instead. The mount effect redirects rather than silently signing out, and the submit `catch` redirects on that error. Drops the `isSigningOut` band-aid.

**Tests:** unit for the load-time redirect and the submit-time redirect.

### DIFF
--- a/src/app/account/signin/page.tsx
+++ b/src/app/account/signin/page.tsx
@@ -35,7 +35,8 @@ export default function SigninPage() {
         }
     }, [searchParams, router])
 
-    if (!isLoaded || alreadySignedIn.status === 'loading') {
+    const isResolvingSession = alreadySignedIn.status === 'loading' || alreadySignedIn.status === 'redirecting'
+    if (!isLoaded || isResolvingSession) {
         return <Loader />
     }
 
--- a/src/app/account/signin/page.tsx
+++ b/src/app/account/signin/page.tsx
@@ -5,9 +5,11 @@ import { Container, Loader } from '@mantine/core'
 import { useRouter, useSearchParams } from 'next/navigation'
 import { useEffect, useState } from 'react'
 import { Routes } from '@/lib/routes'
+import { AlreadySignedIn } from './already-signed-in'
 import { type MFAState } from './logic'
 import { RequestMFA } from './mfa'
 import { SignInForm } from './sign-in-form'
+import { useAlreadySignedIn } from './use-already-signed-in'
 
 type SignInStep = 'form' | 'mfa'
 
@@ -15,6 +17,7 @@ export default function SigninPage() {
     const { isLoaded } = useSignIn()
     const searchParams = useSearchParams()
     const router = useRouter()
+    const alreadySignedIn = useAlreadySignedIn()
 
     const [step, setStep] = useState<SignInStep>('form')
     const [mfaState, setMFAState] = useState<MFAState>(false)
@@ -32,7 +35,7 @@ export default function SigninPage() {
         }
     }, [searchParams, router])
 
-    if (!isLoaded) {
+    if (!isLoaded || alreadySignedIn.status === 'loading') {
         return <Loader />
     }
 
@@ -47,10 +50,21 @@ export default function SigninPage() {
         setStep('mfa')
     }
 
+    // Show the prompt, not the form, while a session is already active.
+    const showForm = alreadySignedIn.status === 'signed-out' && step === 'form'
+    const showMFA = alreadySignedIn.status === 'signed-out' && step === 'mfa'
+
     return (
         <Container w={500}>
-            {step === 'form' && <SignInForm onComplete={setPending} mfa={false} />}
-            {step === 'mfa' && <RequestMFA mfa={mfaState} />}
+            <AlreadySignedIn
+                isVisible={alreadySignedIn.status === 'signed-in'}
+                email={alreadySignedIn.email}
+                isSwitching={alreadySignedIn.isSwitching}
+                onContinue={alreadySignedIn.continueToApp}
+                onSwitchAccount={alreadySignedIn.switchAccount}
+            />
+            {showForm && <SignInForm onComplete={setPending} mfa={false} />}
+            {showMFA && <RequestMFA mfa={mfaState} />}
         </Container>
     )
 }
--- a/src/app/account/signin/use-already-signed-in.ts
+++ b/src/app/account/signin/use-already-signed-in.ts
@@ -1,0 +1,55 @@
+'use client'
+
+import { useCallback, useEffect, useState } from 'react'
+import { useRouter, useSearchParams } from 'next/navigation'
+import { useClerk, useUser } from '@clerk/nextjs'
+import { Routes } from '@/lib/routes'
+import { safeRedirectUrl } from '@/lib/utils'
+
+export type AlreadySignedInStatus = 'loading' | 'signed-in' | 'signed-out'
+
+export interface UseAlreadySignedIn {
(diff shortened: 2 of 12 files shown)


src/app/account/signin/page.tsx
```diff
@@ -47,10 +50,21 @@ export default function SigninPage() {
...
-            {step === 'form' && <SignInForm onComplete={setPending} mfa={false} />}
-            {step === 'mfa' && <RequestMFA mfa={mfaState} />}
+            <AlreadySignedIn
+                isVisible={alreadySignedIn.status === 'signed-in'}
+                email={alreadySignedIn.email}
+                isSwitching={alreadySignedIn.isSwitching}
+                onContinue={alreadySignedIn.continueToApp}
+                onSwitchAccount={alreadySignedIn.switchAccount}
+            />
+            {showForm && <SignInForm onComplete={setPending} mfa={false} />}
```
nathanstitt, line 67: Now that the page gates the form behind `showForm` (only mounted once `status === 'signed-out'`), a chunk of the old already-signed-in handling in `sign-in-form.tsx` looks like it's been left stranded rather than removed:

- the `isSignedIn && !signedInRecently` effect that auto-calls `signOut()` (sign-in-form.tsx:60-66)
- the `isSigningOut` state + the `if (isSignedIn || isSigningOut ...) return null` guard (lines 28, 68)
- the `errorMessage === "You're already signed in."` branch in the submit catch (line 109)

The form can now only mount *after* sign-out, so `isSignedIn` should never be true while it's rendered — which makes all of the above effectively dead. The PR description says it "Drops the `isSigningOut` band-aid" and that "the submit catch redirects on that error," but `sign-in-form.tsx` isn't touched in this diff, so that cleanup doesn't seem to have landed yet. Since this is a draft: is the `sign-in-form.tsx` teardown coming in a follow-up commit, or was it meant to be part of this one? Leaving both paths in risks two components racing to reason about the same Clerk session.

src/app/account/signin/use-already-signed-in.ts
```diff
@@ -0,0 +1,55 @@
...
+export function useAlreadySignedIn(): UseAlreadySignedIn {
+    const { isLoaded, isSignedIn, user } = useUser()
+    const { signOut } = useClerk()
+    const router = useRouter()
+    const searchParams = useSearchParams()
+
+    const [status, setStatus] = useState<AlreadySignedInStatus>('loading')
+    const [isSwitching, setIsSwitching] = useState(false)
+
+    useEffect(() => {
```
nathanstitt: On the "do we even need our own session timer?" question — per Clerk's docs, **no, we should not roll our own**:

- The session **token (JWT)** is intentionally short-lived (**60s default, not configurable**), and Clerk's frontend SDK auto-refreshes it on a ~50s interval. A 60s token is *expected* to be constantly refreshed — it is not a signal that the user's session ended.
- Actual **session** lifetime is governed server-side by Clerk: **maximum lifetime (default 7 days)** and an optional **inactivity timeout (default: disabled)**, both configured in the Clerk Dashboard.

So the app should treat `useUser().isSignedIn` / a failed `getToken()` as the source of truth and let Clerk own expiry. This hook does the right thing by keying off `useUser()`. The takeaway for the *team* (not this hook): if we want sessions to time out sooner, that's a Clerk Dashboard setting (max-lifetime / inactivity timeout), **not** something to reimplement app-side. Any place that behaves as if we have an independent app-session clock is a latent source of exactly this "already signed in" mismatch.

Refs: [Session options](https://clerk.com/docs/authentication/configuration/session-options), [Session tokens](https://clerk.com/docs/guides/sessions/session-tokens).

src/app/account/signin/page.tsx
```diff
@@ -47,10 +50,21 @@ export default function SigninPage() {
         setStep('mfa')
     }
 
+    // Show the prompt, not the form, while a session is already active.
+    const showForm = alreadySignedIn.status === 'signed-out' && step === 'form'
```
nathanstitt, line 55: Design question worth surfacing: when a fully-valid Clerk session hits `/account/signin` with a `redirect_url`, we now render an interactive "Continue / Switch account" card and wait for a click. For the dominant case (user is legitimately signed in and just navigated/deep-linked to signin), an **automatic** `router.replace(redirect_url)` would be smoother — the PR description actually describes this ("the mount effect redirects rather than silently signing out"), but the code prompts instead. Is the manual card intentional (e.g. to make account-switching discoverable), or a leftover from a redirect-first draft? If auto-redirect is acceptable, the hook could `continueToApp()` on latch when `status === 'signed-in'` and a safe `redirect_url` is present, reserving the card for the no-`redirect_url` / ambiguous case.

src/app/account/signin/page.tsx
```diff
@@ -47,10 +50,21 @@ export default function SigninPage() {
...
+    const showMFA = alreadySignedIn.status === 'signed-out' && step === 'mfa'
+
     return (
         <Container w={500}>
-            {step === 'form' && <SignInForm onComplete={setPending} mfa={false} />}
-            {step === 'mfa' && <RequestMFA mfa={mfaState} />}
+            <AlreadySignedIn
```
nathanstitt, line 60: **Root cause — worth fixing upstream rather than  or in addition to handling here.**

The app session is derived entirely from Clerk. `proxy.ts` calls [`marshalSession(clerkUserId, sessionClaims)`](https://github.com/safeinsights/management-app/blob/9bf7ed00363f4b7c692e6ef34bc933d8c3200ad2/src/server/session.ts#L66-L104), which builds the session from Clerk's `sessionClaims.userMetadata`.

What actually lands a Clerk-authenticated user on `/account/signin` is [this catch in `proxy.ts`](https://github.com/safeinsights/management-app/blob/9bf7ed00363f4b7c692e6ef34bc933d8c3200ad2/src/proxy.ts#L66-L76): when `marshalSession` **throws**, we redirect to `/account/signin?error=session`. The throw is not a session expiry — it originates from the Clerk backend call [`client.users.getUser(...)` inside `syncAndUpdateUserMetadata`](https://github.com/safeinsights/management-app/blob/9bf7ed00363f4b7c692e6ef34bc933d8c3200ad2/src/server/session.ts#L18-L25) (or the metadata sync/update around it). So a transient Clerk API hiccup for a genuinely-signed-in user boots users to the signin page.

Suggested direction — fix the cause, and **don't redirect to signin at all in this case**:
- On a marshal failure for a user Clerk still considers authenticated (`clerkUserId` present), treat the stale/broken metadata as recoverable: **null out and regenerate it** (force a re-sync from the live Clerk user via `syncAndUpdateUserMetadata` / `forceUpdate`) rather than bouncing to signin.
- If regeneration itself fails transiently, fall through to `BLANK_SESSION` (authenticated-but-empty, same as the null-metadata branch already does) so the user stays signed in — never redirect an authenticated user to `/account/signin?error=session`.

That removes the state this PR is designed to catch, instead of only catching it after the fact. This card's prompt is still a reasonable safety net, but the redirect-to-signin on a marshal throw is the real bug.

nathanstitt (dismissed):
thanks for digging into this, @Rod0352 as I was reading the code I started digging in and think I may have a found the root cause, take a look and see if that makes sense to you

therealmarv (approved):
All code-review remarks seem resolved for me and this branch looks overall good for me!

PR #851, opened 2026-07-01
OTTER-641: fix stuck "code change requested" status after resubmitted code is approved

therealmarv:
see [OTTER-641](https://openstax.atlassian.net/browse/OTTER-641)

## Problem

On the researcher's "Proposed Studies" dashboard, the status stayed on "Change requested" even after the researcher resubmitted updated code and the Data Provider approved it. The researcher was effectively stuck in that status until the outputs were approved and released.

## Root cause

Resubmitting after a change request reuses the *same* study job (status rows are append-only), so an approved job accumulates the full round history on one job:

```
CODE-SUBMITTED → CODE-CHANGES-REQUESTED → CODE-SUBMITTED (resubmit) → CODE-APPROVED
```

The state layer already resolved the correct live winner as `codeDecision = 'CODE-APPROVED'` (via `CODE_DECISION_PRIORITY` plus count-based liveness). But the pill and `displayStatus` re-derived which decision to show by walking `DISPLAY_STATUS_PRIORITY`, where `CODE-CHANGES-REQUESTED` was ranked above `CODE-APPROVED`, so the stale earlier round won. Since researchers have no execution-status labels, the pill kept falling through to that stale `CODE-CHANGES-REQUESTED` all the way until `FILES-APPROVED` finally outranked it, which is exactly the "stuck until outputs approved" symptom.

## Fix

Make the status projection defer to the already-resolved live decision instead of letting array ordering pick among coexisting decisions:

- `projectStudyState` and `resolvePillStatus` now keep a code-decision status only when it equals the live `codeDecision`; all other (stale) coexisting decisions are dropped. This unifies the previous mid-resubmission handling with the approved-after-resubmit case.
- The relative ordering of the three code-decision statuses in `DISPLAY_STATUS_PRIORITY` is therefore no longer authoritative (only their position relative to the `JOB-*`/`CODE-SUBMITTED` entries matters).

To keep a single source of truth and avoid drift:

- `projectStudyState` now reuses the existing `latestSubmittedJobHasLiveCodeDecision` helper (shared with reviewer routing and dashboard highlighting) rather than reimplementing the count-based liveness rule.
- The stale-decision rule is factored into one exported helper, `isStaleCodeDecision(status, liveDecision)`, used by both `projectStudyState` and `resolvePillStatus`, so the two can't diverge.

The existing "a job carrying both `CODE-CHANGES-REQUESTED` and a terminal `CODE-REJECTED` reads Rejected" behavior is preserved, because `codeDecision` already ranks `CODE-REJECTED` above `CODE-CHANGES-REQUESTED`.

## Tests

- `state.test.ts`: resubmit then approved yields `displayStatus === 'CODE-APPROVED'`; an approved job that then starts executing follows the execution status (`JOB-READY`).
- `pill.test.ts`: researcher pill reads "Approved" after a resubmit is approved, including once execution has started (falls through to the live decision, not the superseded change request).

## Verification

- `pnpm run checks` passes (type check, lint, action validation).
- Unit tests cover the Jira reproduction path and the execution fall-through case.

[OTTER-641]: https://openstax.atlassian.net/browse/OTTER-641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/study-screen/pill.test.ts
+++ b/src/lib/study-screen/pill.test.ts
@@ -77,6 +77,30 @@ describe('resolvePillStatus', () => {
         )
         expect(label.label).toBe('Rejected')
     })
+    // OTTER-641: resubmit approved. The stale CODE-CHANGES-REQUESTED must not win over the live approval.
+    it('resubmit then approved reads Approved, not the stale Change requested', () => {
+        const label = resolvePillStatus(
+            'researcher',
+            state({
+                latestJobStatuses: ['CODE-SUBMITTED', 'CODE-CHANGES-REQUESTED', 'CODE-SUBMITTED', 'CODE-APPROVED'],
+                codeDecision: 'CODE-APPROVED',
+            }),
+        )
+        expect(label.label).toBe('Approved')
+    })
+    // Researcher has no execution-status labels, so the pill falls through JOB-READY to the live code
+    // decision; that must be the approval, not the superseded change-request from the earlier round.
+    it('approved then executing still reads Approved for the researcher (falls through to live decision)', () => {
+        const label = resolvePillStatus(
+            'researcher',
+            state({
+                latestJobStatuses: ['CODE-CHANGES-REQUESTED', 'CODE-SUBMITTED', 'CODE-APPROVED', 'JOB-READY'],
+                codeDecision: 'CODE-APPROVED',
+                isExecuting: true,
+            }),
+        )
+        expect(label.label).toBe('Approved')
+    })
 })
 
 describe('resolveRowHighlight', () => {
--- a/src/lib/study-screen/pill.test.ts
+++ b/src/lib/study-screen/pill.test.ts
@@ -88,6 +88,19 @@ describe('resolvePillStatus', () => {
         )
         expect(label.label).toBe('Approved')
     })
+    // Reviewers DO have a label for CODE-CHANGES-REQUESTED, so before the fix the stale round-1 status
+    // could win by DISPLAY_STATUS_PRIORITY. isStaleCodeDecision now runs for every role, so a reviewer
+    // viewing an approved-after-resubmit job also reads the live decision.
+    it('reviewer: resubmit then approved reads Approved, not the stale Change requested', () => {
+        const label = resolvePillStatus(
+            'reviewer',
+            state({
+                latestJobStatuses: ['CODE-SUBMITTED', 'CODE-CHANGES-REQUESTED', 'CODE-SUBMITTED', 'CODE-APPROVED'],
+                codeDecision: 'CODE-APPROVED',
+            }),
+        )
+        expect(label.label).toBe('Approved')
+    })
     // Researcher has no execution-status labels, so the pill falls through JOB-READY to the live code
     // decision; that must be the approval, not the superseded change-request from the earlier round.
     it('approved then executing still reads Approved for the researcher (falls through to live decision)', () => {
--- a/src/lib/study-screen/state.test.ts
+++ b/src/lib/study-screen/state.test.ts
@@ -82,6 +82,18 @@ describe('projectStudyState', () => {
         expect(s.displayStatus).toBe('CODE-APPROVED')
     })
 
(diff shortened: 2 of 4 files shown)


src/lib/study-screen/state.test.ts
```diff
@@ -71,6 +71,29 @@ describe('projectStudyState', () => {
...
+    // earlier round (which used to win by DISPLAY_STATUS_PRIORITY order).
+    it('same-job resubmit then approved → displayStatus is CODE-APPROVED, not the stale changes-requested', () => {
+        const approved = job(ID1, ['CODE-SUBMITTED', 'CODE-CHANGES-REQUESTED', 'CODE-SUBMITTED', 'CODE-APPROVED'])
+        const s = projectStudyState(raw({ status: 'APPROVED', jobs: [approved] }))
+        expect(s.codeDecision).toBe('CODE-APPROVED')
+        expect(s.codeAwaitingDecision).toBe(false)
+        expect(s.displayStatus).toBe('CODE-APPROVED')
+    })
+
+    it('approved job then execution starts → displayStatus follows execution, not the code decision', () => {
```
nathanstitt, line 97: Nice pair of cases here (approved-after-resubmit + execution fall-through). One symmetry gap worth considering: `pill.test.ts` asserts the resubmit-then-*rejected* shape reads Rejected (line 70), but `projectStudyState` doesn't get the equivalent `displayStatus === 'CODE-REJECTED'` assertion for that same `['CODE-SUBMITTED','CODE-CHANGES-REQUESTED','CODE-SUBMITTED','CODE-REJECTED']` shape. Since the PR body specifically calls out that the "reads Rejected" behavior is preserved, a one-line test here would lock it in at the state layer too — the logic is identical to the approved case (`isStaleCodeDecision` drops the stale CHANGES-REQUESTED), so it should be a cheap add.

src/lib/study-screen/pill.test.ts
```diff
@@ -77,6 +77,30 @@ describe('resolvePillStatus', () => {
...
+            state({
+                latestJobStatuses: ['CODE-SUBMITTED', 'CODE-CHANGES-REQUESTED', 'CODE-SUBMITTED', 'CODE-APPROVED'],
+                codeDecision: 'CODE-APPROVED',
+            }),
+        )
+        expect(label.label).toBe('Approved')
+    })
+    // Researcher has no execution-status labels, so the pill falls through JOB-READY to the live code
+    // decision; that must be the approval, not the superseded change-request from the earlier round.
+    it('approved then executing still reads Approved for the researcher (falls through to live decision)', () => {
```
nathanstitt, line 106: Both new cases are `'researcher'`. Worth noting the fix also changes the *reviewer* pill: `isStaleCodeDecision` now runs for every role, so a reviewer viewing an approved-after-resubmit job will also drop the stale CODE-CHANGES-REQUESTED (previously it could win by DISPLAY_STATUS_PRIORITY since reviewers do have a label for it). That's the correct behavior, but it's an unlabelled-in-tests change — a `'reviewer'` variant of the resubmit-then-approved case would document that reviewers now see the live decision too, not just researchers. Not blocking, just closing the coverage loop on the role that actually has code-decision labels.

nathanstitt (approved):
thank you @therealmarv

src/lib/study-screen/state.test.ts
therealmarv: Added in e0250d53. `state.test.ts` now has a `same-job resubmit then rejected → displayStatus is CODE-REJECTED, not the stale changes-requested` case, so the "reads Rejected" behavior is locked in at the state layer too, symmetric with the pill test.

src/lib/study-screen/pill.test.ts
therealmarv: Added in e0250d53. `pill.test.ts` now has a `reviewer: resubmit then approved reads Approved, not the stale Change requested` case, documenting that the stale-decision drop now applies to the reviewer role too (which has a label for CODE-CHANGES-REQUESTED and previously could show the stale round).

PR #856, opened 2026-07-02
OTTER-597: clarify breached-password error copy on signup

Rod0352:
## Summary

Signing up with a breached password showed Clerk's raw error, which doesn't
explain the problem or the fix. It now reads: "This password was found in a
database of known breached passwords and cannot be used. Please choose a
different password."

## Changes

- Renamed `CLERK_ERROR_TITLES` → `CLERK_ERROR_COPY` (`clerk-errors.tsx`); entries can now carry a `message` that replaces Clerk's wording for that error code
- Added the OTTER-597 copy for `form_password_pwned`
- Signup alert (`signup/page.tsx`) and `ClerkErrorAlert` now use the message override; codes without one keep Clerk's wording
- Server action unchanged — raw Clerk message still passes through as the fallback body

## Tests

- New `clerk-errors.test.tsx`: pwned code renders the new copy + "Compromised Password" title; unknown codes fall back to Clerk's message
- `create-account.action.test.ts` unchanged (still asserts raw passthrough)

### DIFF
--- a/src/app/account/invitation/[inviteId]/signup/page.tsx
+++ b/src/app/account/invitation/[inviteId]/signup/page.tsx
@@ -6,7 +6,7 @@ import {
     usePasswordRequirements,
 } from '@/app/account/reset-password/password-requirements'
 import { useMutation, useQuery, z, zodResolver } from '@/common'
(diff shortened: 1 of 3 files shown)


therealmarv (approved):
looks good for me!

PR #858, opened 2026-07-06
OTTER-594 follow-up: green saved indicator, per-field gating, title spacing

Rod0352:
### Changes
1. **Green checkmark** (QA issue 1) — `SaveStatusIndicator` renders `CheckCircleIcon` (fill, `green.9`) + green text for the saved state, matching the [Figma example](https://www.figma.com/design/84pQzoYf7rKqOjo4qsDOeG/Card-59-69-71---Result-step?node-id=10928-17754&m=dev) and the existing pattern in `panel.tsx`. "Saving…" stays dimmed. Applies to every surface using the shared component (proposal fields, collaborative editors, resubmission note).
2. **Per-field gating** (QA issue 2) — `useYjsFormMap` now tracks `editedKeys` from `pushField`/`pushPI`. Each field's indicator stays idle until *that* field is edited in the current session. Initial-value seeding and remote collaborator edits never mark a field as edited. Edited flags reset on provider reconnect, mirroring the existing status-latch behavior.
3. **Title spacing** (QA issue 3) — the Study Title indicator moved into the word-counter row (left slot, alongside any validation error), so it sits directly under the input like the other fields.

### Testing
- New unit tests: green checkmark rendering; seeding does not mark fields edited; only actually-written fields are marked; no-op pushes and remote updates don't mark; `pushPI` marks both PI keys.
- Updated `yjsForm` stubs in `use-resubmit-proposal.test.ts` / `use-submit-proposal.test.ts` for the new `editedKeys` field.

### DIFF
--- a/src/components/save-status.tsx
+++ b/src/components/save-status.tsx
@@ -16,7 +16,7 @@ const SavedLabel: FC = () => {
     const theme = useMantineTheme()
 
     return (
-        <Group gap={4} wrap="nowrap" data-testid="autosave-status">
+        <Group gap={8} wrap="nowrap" data-testid="autosave-status">
             <CheckCircleIcon size={16} color={theme.colors.green[9]} weight="fill" />
             <Text size="xs" c="green.9" fw={600}>
                 All changes saved
--- a/src/components/save-status.tsx
+++ b/src/components/save-status.tsx
@@ -1,11 +1,28 @@
+'use client'
+
 import { FC } from 'react'
-import { Text } from '@mantine/core'
+import { Group, Text, useMantineTheme } from '@mantine/core'
+import { CheckCircleIcon } from '@phosphor-icons/react/dist/ssr'
 
 export type SaveStatusValue = 'idle' | 'saving' | 'saved'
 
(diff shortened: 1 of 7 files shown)


src/components/save-status.tsx
```diff
@@ -1,11 +1,28 @@
...
+    <Text size="xs" c="dimmed" data-testid="autosave-status">
+        Saving…
+    </Text>
+)
+
+const SavedLabel: FC = () => {
+    const theme = useMantineTheme()
+
+    return (
+        <Group gap={4} wrap="nowrap" data-testid="autosave-status">
```
therealmarv: Non-blocking: the Figma example uses an 8px gap between the check icon and `All changes saved`; this uses `gap={4}`. Color matches the local `green.9` theme value, so I would not block on this, but `gap={8}` would be a closer visual match.

therealmarv (approved):
Approved. I reviewed the PR against the OTTER-594 follow-up comments and the linked Figma example. The branch covers the requested green saved indicator, per-field gating for untouched fields, and title indicator placement.

I left one non-blocking visual note about the icon/text gap compared with Figma.

PR #860, opened 2026-07-07
OTTER-613 follow-up: collapse the entire Submitted code section

therealmarv:
see [OTTER-613](https://openstax.atlassian.net/browse/OTTER-613?focusedCommentId=44789)

## Problem

On the Data Partner (DO) post-decision **"Review study code"** page (study status *code change requested* / *code approved* / *code rejected*), clicking **"Hide full study code"** only collapsed the raw code viewer. The "Submitted code" header, "View approved initial request" link, dataset chips, "AI Summary" panel, and "Security scan log" panel all stayed visible.

Per the Figma spec (collapsed `683-14230`, expanded `4488-5751`), the **entire** "Submitted code" card must be hidden when collapsed. The collapsed state should show only the STEP status card with a **"View full study code >"** link inside it (below the banner), followed by "Feedback and notes". Expanding reveals the full card ending in a single **"Hide full study code ^"** link.

## Root cause

The collapse boundary sat too deep: on the post-decision path only the inner code *body* toggled, while the section header, datasets, AI summary, and scan panels always rendered.

## Fix

Lift the boundary to wrap the whole card, mirroring the already-shipped researcher-side pattern (OTTER-590, `view/code-post-decision-view.tsx`):

- **`post-feedback-view.tsx`** — `CodeSection` owns the expand state via `useExpandable()`. When collapsed it renders a "View full study code" opener inside the step card (`ProposalStepHeader` `children`); the full `SubmittedCodeSection` is wrapped in a Mantine `<Collapse in={expanded} keepMounted>` so the whole card animates open/closed and its state (selected file tab, AI-summary expand state) is preserved between expansions. `onCollapse` is wired to the closer.
- **`submitted-code-interactive.tsx`** — `StudyCodeViewer` gains an optional `onCollapse` mode: when set, the code + tabs are always shown and its toggle becomes the section's single "Hide full study code" closer (distinct `study-code-toggle-collapse` testid).
- **`submitted-code-section.tsx`** — threads `onCollapse` through to the viewer.

Net effect: collapsed → only the status card + opener link; expanded → full "Submitted code" card ending in "Hide full study code"; exactly one toggle is visible in each state.

## Review follow-ups

Addressing the code-review feedback on this PR:

- **Single toggle / collapse hook.** Consolidated to one parameterized `StudyCodeToggle` (with `labels`/`testId`/`forwardRef`) and one `useExpandable` hook, both in `view/study-code-collapse.tsx`. The duplicate toggle previously defined in `submitted-code-interactive.tsx` is gone (that module now re-exports the shared one), and `post-feedback-view.tsx` imports both the toggle and `useExpandable` from the same module. The "View full study code" / "Hide full study code" labels live in a single shared `FULL_STUDY_CODE_TOGGLE_LABELS` constant.
- **Keep mounted instead of unmounting.** Using `<Collapse keepMounted>` (rather than returning `null` while collapsed) restores the expand/collapse animation and preserves in-panel state, matching the researcher-side pattern. Tradeoff: on this post-decision page the review row already exists, so `useStudyReviewPoll` short-circuits (no polling); only a one-time code-file fetch runs while mounted.
- **Accessibility.** Focus moves to the revealed section on expand and back to the opener on collapse; the collapsed panel is `aria-hidden` + `inert` via Mantine's `Collapse`, so it is removed from the accessibility tree.

## Scope

Only the post-decision reviewer path (`kind="CODE"`) is affected. The active-review page (`code-review.tsx`) is untouched (`onCollapse` absent → unchanged behavior), confirmed by its passing tests.

## Testing

- Rewrote the OTTER-613 case in `post-feedback-view.test.tsx`: collapsed ⇒ section / AI summary / scan present in the DOM but not visible, and only the "View full study code" opener is interactive; expand ⇒ full section + code body revealed with the "Hide full study code" closer and focus moved into the panel; collapse again ⇒ section hidden (stays mounted) and focus returned to the opener.
- Added an `onCollapse`-mode case to `submitted-code-section.test.tsx`.
- Full unit suite green; `pnpm run checks` passes. Merged latest `main` into the branch.

[OTTER-613]: https://openstax.atlassian.net/browse/OTTER-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -146,25 +148,29 @@ type SubmittedCodePanelProps = {
...
+// same one shown during active review. Per OTTER-613 the entire card is hidden when collapsed and
+// revealed via the "View full study code" toggle in the step card, so it only mounts when expanded;
+// its own toggle then reads "Hide full study code" and calls onCollapse to close the whole card.
+function SubmittedCodePanel({ orgSlug, study, job, review, scan, expanded, onCollapse }: SubmittedCodePanelProps) {
     // scan-presence is coupled to job-presence: the caller fetches both together and
     // jobScanResultForJob never returns null (it falls back to {status:'IN-PROGRESS', logFile:null}),
     // so scan is null exactly when job is null. The !scan check is the type-narrowing that lets us
     // pass a non-null scan to SubmittedCodeSection; in practice it only fires on the null-job branch.
     if (!job || !scan) return null
+    if (!expanded) return null
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Unmounting here differs from the researcher-side pattern this mirrors — `view/code-post-decision-view.tsx` keeps the card mounted inside a Mantine `<Collapse>`. The visible consequences: no expand/collapse animation, and every re-expand starts fresh (selected file tab and AI-summary expand state reset, and the review/code queries refetch on mount). It also buys the "no AI polling / code fetch while collapsed" win the description calls out, so it may well be the right tradeoff — just worth a sanity check that the state reset + lack of animation match the Figma expectations.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -13,6 +13,8 @@ import type { ReactNode } from 'react'
 import type { CodeReviewFeedbackEntry, ProposalFeedbackEntry, SelectedStudy } from '@/server/actions/study.actions'
 import type { JobScanResult, LatestJobForStudy, StudyReviewWithMeta } from '@/server/db/queries'
 import { SubmittedCodeSection } from './submitted-code-section'
+import { StudyCodeToggle } from './submitted-code-interactive'
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: We now have two exported components named `StudyCodeToggle`: this one (props `isExpanded`/`labels`/`testId`) and the one in `view/study-code-collapse.tsx` (props `expanded`, hardcoded labels, also with a `testId` override). This file imports one of them while pulling `useExpandable` from the other's module, which is easy to misread — the researcher-side page uses the *other* toggle for the same interaction. Might be worth a small follow-up to consolidate into a single parameterized toggle (the `study-code-collapse` one is nearly there) or at least give them distinct names.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -204,8 +215,18 @@ function CodeSection({
                 timestampDate={timestampDate}
                 timestampLabel={timestampLabel}
                 banner={banner}
+            >
+                <StudyCodeToggle isVisible={!expanded && hasSubmittedCode} isExpanded={false} onClick={toggle} />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Small a11y note: clicking this opener unmounts it, so keyboard focus drops to `<body>` and a screen-reader user never perceives the state change — the `aria-expanded={false}` element is gone before it could flip. Same on the way back: the closer unmounts along with the whole card. Moving focus to the revealed section on expand (and back to this opener on collapse) would tighten it up; seems fine as a follow-up if we'd rather keep this PR scoped to the Figma fix.

nathanstitt (changes requested):
looks, good but think we probably should refactor to have a single collapse component vs the mismatch pointed out below

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Switched to the researcher-side approach: the section is now wrapped in `<Collapse in={expanded} keepMounted>` instead of unmounting. That restores the expand/collapse animation and preserves in-panel state (selected file tab, AI-summary expand state) across expansions.

On the "no polling while collapsed" tradeoff you flagged: it's a non-issue on this post-decision page specifically. The review row already exists by this point, so `useStudyReviewPoll` seeds from `initialData` and `refetchInterval` short-circuits (`data != null` → `false`), i.e. no polling. Only a one-time code-file fetch runs while mounted. Updated the PR description to reflect this since the original text was now inaccurate.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Consolidated to a single toggle. There's now one parameterized `StudyCodeToggle` (props `expanded`/`labels`/`testId`, `forwardRef`) and one `useExpandable` hook, both in `view/study-code-collapse.tsx`. The duplicate component in `submitted-code-interactive.tsx` was removed (that module now just re-exports the shared one), and this file imports both the toggle and `useExpandable` from the same module, so the split-import mismatch is gone. The "View full study code" / "Hide full study code" labels are now a single shared `FULL_STUDY_CODE_TOGGLE_LABELS` constant.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Added focus management. `CodeSection` now moves focus to the revealed panel on expand (a `tabIndex={-1}` wrapper `<Box>`) and back to the opener on collapse. Since the section stays mounted now, the opener/closer no longer disappear out from under focus, and while collapsed the panel is `aria-hidden` + `inert` (via Mantine's `Collapse`) so it's removed from the accessibility tree. The OTTER-613 test asserts both focus transitions.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
```diff
@@ -29,8 +29,14 @@ import {
...
 import type { StudyReviewWithMeta } from '@/server/db/queries'
 import type { CodeFile } from './study-code-files'
+import {
+    FULL_STUDY_CODE_TOGGLE_LABELS,
+    StudyCodeToggle,
+    type StudyCodeToggleLabels,
+} from '@/app/[orgSlug]/study/[studyId]/view/study-code-collapse'
 
 export type { CodeFile } from './study-code-files'
+export { StudyCodeToggle, type StudyCodeToggleLabels }
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx]: Looks like nothing imports `StudyCodeToggle` / `StudyCodeToggleLabels` from this module anymore — the remaining consumers (`submitted-code-section.tsx` and its test) pull other exports, and everything else already imports the toggle straight from `view/study-code-collapse`. We could drop this re-export so the shared module stays the single import path (the type import above is still needed for `toggleLabels`).

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -168,26 +173,42 @@ type SubmittedCodePanelProps = {
...
     return (
-        <SubmittedCodeSection
-            orgSlug={orgSlug}
-            study={study}
-            job={job}
-            review={review}
-            scan={scan}
-            codeInitiallyExpanded={false}
-        />
+        <Collapse in={expanded} keepMounted>
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx:200]: Verified the keepMounted tradeoff from the earlier thread: with `keepMounted`, Mantine keeps the collapsed panel at `height: 0` with `aria-hidden` + `inert`, exactly as described. On polling — `useStudyReviewPoll` only keeps refetching when no review row exists, and in that legacy/failed-generation case the same polling already ran on main (this section used to render unconditionally), so the collapsed panel adds no background work that wasn't happening before. No action needed, just closing the loop.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -216,7 +237,22 @@ function CodeSection({
...
+    }, [collapse])
+    const onExpand = useCallback(() => {
+        toggle()
+        requestAnimationFrame(() => panelRef.current?.focus())
+    }, [toggle])
     if (!isVisible) return null
+    // Only offer the opener when there is a Submitted code panel behind it. SubmittedCodePanel
+    // returns null without a job/scan (e.g. the fallback auto-approved page), so an unconditional
+    // opener would otherwise expand to an empty card with no way back.
+    const hasSubmittedCode = Boolean(job && scan)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx:255]: Nice catch going beyond the requested changes — without this guard the opener would expand an empty card with no closer behind it (`SubmittedCodePanel` returns null without a job). The fallback-page test asserting the opener is absent covers it too.

nathanstitt (approved):
looks good but we can probably remove the dead code found

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx]: Fixed in 3e2fcf59. Dropped the `export { StudyCodeToggle, type StudyCodeToggleLabels }` re-export so `view/study-code-collapse` stays the single import path. The remaining consumers (`submitted-code-section.tsx` and its test) only pull other exports, and everything else already imports the toggle straight from the shared module. The type import above stays since `toggleLabels` still needs it.

PR #861, opened 2026-07-07
OTTER-638 follow-up: gate code resubmit eligibility on projected state

therealmarv:
see [OTTER-638](https://openstax.atlassian.net/browse/OTTER-638?focusedCommentId=44798) (QA re-report that existing studies still fail; moved here from OTTER-558. Original bug report: [OTTER-558 comment 43809](https://openstax.atlassian.net/browse/OTTER-558?focusedCommentId=43809), QA re-report in [comment 44570](https://openstax.atlassian.net/browse/OTTER-558?focusedCommentId=44570))

_Supersedes #854 (auto-closed when its branch was renamed from `OTTER-558-follow-up-resubmit-eligibility`); same fix, since refreshed against current `main`._

_Branch refreshed against current `main` (merge commit); `pnpm run checks` (typecheck + eslint/prettier + action validation) re-passes locally. CI re-runs the unit + e2e suites on the merge commit; the "3x, 0 flaky" e2e run noted below was the pre-merge verification._

## The bug

On a **Result ready** study, opening `/resubmit` and typing in the Resubmission Note threw a stack of toasts, one per keystroke: "Unable to save resubmission note draft / Submission Study is not editable or you do not have access."

The `/resubmit` page gated eligibility order-independently (`statusChanges.some(...)`), but `saveCodeResubmissionNoteDraftAction` and `resubmitStudyCodeAction` read only the top status row (`statusChanges.at(0)`). `statusChanges` is ordered `createdAt desc, id desc`, which is non-deterministic within a transaction (`created_at = now()`, v7 ids not monotonic within a millisecond), and a late `CODE-SCANNED` webhook can append after the decision. So for a Result-ready study `FILES-APPROVED` sits in history (page renders) while `CODE-SCANNED` sorts to `at(0)` (both actions reject). The final "Resubmit study code" button would fail the same way.

This survived earlier fixes because every action test seeded exactly one status row (so `.at(0)` and `.some()` were indistinguishable) and the client tests mocked the actions; a prior change fixed a different symptom (the old `study.status='CHANGE-REQUESTED'` guard) and the close was mis-attributed.

## The fix

Route the page and both actions through the study-screen projection that every `/view` and `/review` page already uses:

```
rawStudyStateForStudy(studyId) -> projectStudyState(raw) -> canResearcherResubmitCode(state)
```

`projectStudyState` is order-independent by construction (latest job by `max(id)`, statuses as a `Set`) and liveness-aware. The new predicate `canResearcherResubmitCode` maps the retired `CODE_RESUBMITTABLE_JOB_STATUSES` to projected facts:

```ts
codeDecision === 'CODE-CHANGES-REQUESTED' || resultsApproved || resultsRejected
```

This also fixes a latent over-approval the old `.some()` had: a study already resubmitted and awaiting a new decision (a stale `CODE-CHANGES-REQUESTED` with a fresh `CODE-SUBMITTED` appended) is no longer treated as resubmittable.

## Changes

- Add `canResearcherResubmitCode` in `src/lib/study-screen/eligibility.ts` (exported from `index.ts`).
- `rawStudyStateForStudy(studyId, db: DBExecutor = defaultDb)` now accepts an executor (mirroring `codeSubmissionVersion`) so mutation actions gate on their own handler transaction.
- Wire `resubmit/page.tsx`, `saveCodeResubmissionNoteDraftAction`, and `resubmitStudyCodeAction` through the predicate. Lab-scoped 0-row update guard unchanged; the redundant `latestSubmittedJobForStudy` fetch dropped in the resubmit action.
- Retire `canResubmitStudyCode` / `CODE_RESUBMITTABLE_JOB_STATUSES`; delete `src/lib/code-resubmission.ts` (+ test); move `ROUND_CLOSING_JOB_STATUSES` to `src/lib/study-job-status.ts` (import updated in `mutations.ts`).

## Tests

Covered at every layer the bug touched, all green; the e2e was run in a real browser and passed repeatedly (3x, 0 flaky at `retries: 0`) before the main refresh (see note at top).

- **Pure predicate** (`eligibility.test.ts`): `CODE-CHANGES-REQUESTED` live, stale `CODE-CHANGES-REQUESTED` (fresh `CODE-SUBMITTED` appended) not eligible, `FILES-APPROVED` with a trailing `CODE-SCANNED` (the QA scenario), `FILES-REJECTED`, bare `JOB-ERRORED` not eligible, errored-then-`FILES-REJECTED` eligible (the "code errored" entry state), plus code-approved / awaiting / no-job, the multi-round latest-job case, and the OTTER-601 masking case (a later baseline-only `INITIATED` job must not mask a prior `FILES-APPROVED` round; pins `latestJob`'s prefer-submitted filter for the eligibility path).
- **DB-backed action tests** (`study-request.test.ts`) for **both** `saveCodeResubmissionNoteDraftAction` and `resubmitStudyCodeAction`: a mixed history where a later `CODE-SCANNED` sorts first still saves/resubmits (results-ready and change-requested), stale `CODE-CHANGES-REQUESTED` rejects, plus the results-rejected, cross-lab, oversized-payload, empty-note, and no-resubmittable-status cases.
- **De-mocked integration test** (`context.integration.test.tsx`): the real provider + debounced autosave + real action + test DB. A resubmittable study persists the draft to `study.codeResubmissionNoteDraft`; a non-resubmittable one reports save failure and leaves it null. This closes the mocked-UI blind spot the original bug slipped through. (The footer test stays mocked on purpose: its confirm path runs the full real resubmit, S3 upload, job creation, status flip, events, which the DB action tests and the e2e already exercise.)
- **E2E** (`tests/study-flow.spec.ts` + `seedCodeResultsReady` in `tests/e2e.seed.ts`): a Results-ready study drives the full page -> upload -> autosave -> Resubmit flow in a real browser, asserting the "All changes saved" indicator appears (no "not editable" error toast) and the resubmit lands on `/view`. This guards the page + autosave + resubmit wiring, NOT the ordering bug itself: the seed is deterministically ordered (`FILES-APPROVED` newest), so the old `at(0)` gate would have passed here too; the ordering-triggered failure is owned by the predicate and DB-backed action tests above.
- Page order-independence test unchanged (`resubmit/page.test.tsx`).
- `pnpm run checks` passes.

## Behavior change to confirm

Eligibility is now liveness-aware, so a study already resubmitted and awaiting a new decision is no longer resubmit-eligible (it routes to `code-under-review`, matching the pill and screen resolver). This is stricter than the old `.some()` page behavior. Flagging for product sign-off.

## Follow-ups (separate ticket)

Not addressed here and not part of this fix (this PR only touches the researcher resubmit path):

- **Migrate the remaining reviewer-side `statusChanges.at(0)` state reads onto projected facts.** Same ordering-nondeterminism bug class, deferred because they are not the reported QA failure and are low practical risk (both realistic post-resubmit statuses, `CODE-SUBMITTED` and `CODE-SCANNED`, are reviewable). They must move **together**, not piecemeal: `review/code-review.tsx` editability (server render) and `editor.actions.ts` `getStudyStatusAction` reconnect poll feed the **same** `isCodeReviewEditable` predicate, so converting only one would leave the reviewer screen and its kick-out inconsistent. `study.actions.ts:379` is a separate path (the `approveStudyProposalAction` re-approval guard, not the `submitCodeReviewDecisionAction` code-review decision). Fix using the existing order-independent `latestCodeChangeIsSubmission` / `codeAwaitingDecision`.
- **Optional per-job monotonic `seq` column** on `jobStatusChange`: evaluated and deprioritized. It only makes "latest event" ordering deterministic, not the decision-vs-event semantic (eligibility wants the latest decision, so the projection is still required), and a correct per-job monotonic insert is a migration plus a concurrency-safe backfill for little gain.

[OTTER-638]: https://openstax.atlassian.net/browse/OTTER-638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/study-flow.spec.ts
```diff
@@ -676,6 +682,38 @@ test('Code change request and resubmission', async ({ browser, studyFeatures })
...
+
+    await withRole(browser, 'researcher', async (page) => {
+        await goto(page, `/openstax-lab/study/${studyId}/resubmit`)
+        await expect(page.getByRole('heading', { name: /Edit study code/i })).toBeVisible()
+
+        const fileInput = page.locator('input[type="file"]')
+        await fileInput.setInputFiles(['tests/fixtures/code-samples/main.r', 'tests/fixtures/code-samples/code.r'])
+
+        // Filling the note fires the debounced autosave against the real action. Pre-fix this threw
+        // an error toast on every keystroke; now the "All changes saved" indicator must appear.
```
nathanstitt: Small accuracy nit on this comment: with `seedCodeResultsReady`'s spaced timestamps, `FILES-APPROVED` is the newest status row, so the old `at(0)` gate would have read `FILES-APPROVED` — which *was* in `CODE_RESUBMITTABLE_JOB_STATUSES` — and passed. Pre-fix, this test would most likely have gone green too; the error-toast repro needed the scan (or another row) to sort above the decision, which the block comment above rightly leaves to the unit tests. Worth rewording so a future reader doesn't lean on this e2e as the regression guard for the ordering bug — it guards the page + autosave + resubmit wiring, not the ordering.

src/lib/study-screen/eligibility.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        expect(
+            canResubmit([job(ID, ['CODE-SUBMITTED', 'CODE-APPROVED', 'JOB-RUNNING', 'JOB-ERRORED', 'FILES-REJECTED'])]),
+        ).toBe(true)
+    })
+
+    it('false with no jobs', () => {
+        expect(canResubmit([])).toBe(false)
+    })
+
+    it('true from the latest submitted round when a prior round closed with FILES-APPROVED', () => {
```
nathanstitt, line 83: One case that might be worth pinning here: the OTTER-601 masking scenario, where a mid-resubmit file upload has already opened a fresh baseline job after the results decision — e.g. `canResubmit([job(ID, [... 'FILES-APPROVED']), job(ID2, ['INITIATED'])])` → still `true`. That behavior comes from `latestJob`'s prefer-submitted filter in `state.ts`, and right now the only thing exercising it for eligibility is the new e2e (the upload happens before the autosave); none of the study-screen unit tests touch `INITIATED` at all. A one-liner here would keep the predicate honest if that filter ever changes.

src/server/actions/study-request.ts
```diff
@@ -865,15 +861,12 @@ export const saveCodeResubmissionNoteDraftAction = new Action('saveCodeResubmiss
...
-        // on study.status = 'CHANGE-REQUESTED' here (a proposal-clarification status)
-        // would make this autosave throw on every keystroke during a code resubmit.
-        const latestJob = await latestSubmittedJobForStudy(studyId)
-        const latestStatus = latestJob?.statusChanges.at(0)?.status
-        if (!canResubmitStudyCode(latestStatus)) {
+        // Code resubmission runs while study.status stays APPROVED (the reviewer's decision lives on the
+        // job, not the study). Eligibility is the projected-state fact shared with the resubmit page and
+        // resubmitStudyCodeAction (order-independent and liveness-aware), so this autosave can't reject a
+        // state the page rendered as editable. Run on the handler executor.
+        const raw = await rawStudyStateForStudy(studyId, db)
```
nathanstitt, line 860: Non-blocking observation: this now pulls the full raw study state — including the four lexical JSON proposal columns and the complete status history of every job — on each debounced autosave, where the old gate fetched a single job's statuses. Fine at current scale, and sharing one predicate across page + actions is the right trade; just noting that if autosave traffic ever matters, a slim variant of `rawStudyStateForStudy` that skips the Step-2 fields would be the easy trim (the projection input type would need narrowing, so not for this PR).

nathanstitt (approved):
thank you Marvin, I think it looks good and the things found are pretty minor, up to you if you wanna address 'em

tests/study-flow.spec.ts
therealmarv: Good catch, reworded in 63bce966. The inline comment no longer claims this reproduces the pre-fix error toast; it now states explicitly that this seed is deterministically ordered (FILES-APPROVED newest), so the old `at(0)` gate would have passed here too, and that it guards the page + autosave + resubmit wiring, not the ordering bug (which the unit tests own).

src/lib/study-screen/eligibility.test.ts
therealmarv: Added in 63bce966: `canResubmit([job(ID, ['CODE-SUBMITTED', 'FILES-APPROVED']), job(ID2, ['INITIATED'])])` → `true`, pinning `latestJob`'s prefer-submitted filter for the eligibility path so it stays honest if that filter ever changes.

src/server/actions/study-request.ts
therealmarv: Agreed, and leaving it as-is for this PR per your note. Sharing one predicate across page + actions is the intended trade; the slim `rawStudyStateForStudy` variant (skipping the Step-2 lexical fields, with the projection input type narrowed) is the right trim if autosave traffic ever becomes a concern, but it isn't warranted at current scale.

PR #862, opened 2026-07-08
OTTER-573: Deprecate 'Save as Draft' and mirror study title

stetradis:
This pull request removes all explicit "Save as draft" buttons and related dirty tracking logic from proposal editing and resubmission flows. Instead, it relies on continuous autosaving of proposal data via Yjs (a collaborative editing framework) and introduces a server-side mechanism to mirror the proposal title into the `study` table for dashboard discoverability. The changes simplify the UI and codebase, ensuring that data is always up-to-date without requiring manual saves.

The most important changes are:

**Removal of manual draft saving and dirty tracking:**
- Removed the "Save as draft" button and all associated dirty state tracking from both the proposal and edit/resubmit footers (`proposal/footer.tsx`, `edit-and-resubmit/footer.tsx`, `proposal/schema.ts`, and related tests). Navigation and submission now rely solely on autosave, and the UI is simplified as a result. 

**Proposal autosave and title mirroring:**
- Introduced `mirrorProposalTitleToStudy` and `readTitleFromFieldsDoc` in a new `title-mirror.ts` module, which ensures the proposal title is mirrored to the `study` table whenever the collaborative document is saved. This keeps the dashboard in sync with the latest title, even though the canonical data lives in Yjs. 

**UI and navigation adjustments:**
- Updated navigation logic so that moving between steps or going "Back" does not require explicit saving; autosave ensures data persistence. The resubmit and proposal footers now only flush the note (if needed) or simply navigate back. 

**Test and code cleanup:**
- Removed tests and code related to manual draft saving and dirty tracking, reflecting the new autosave-only flow. 
- Updated and simplified footer component props and structure in the study request flow, removing unused save-draft logic. 

**New tests for title mirroring:**
- Added unit tests for the new title mirroring logic to ensure correct behavior when syncing the proposal title to the study table.


tests/study-flow.spec.ts
```diff
@@ -378,17 +370,15 @@ test('Researcher submits a proposal', async ({ browser, studyFeatures }) => {
 })
 
 // OTTER-572: a draft left on Step 2 must reopen on Step 2 from the dashboard, not the Step 1
-// data-org picker. Owns the draft-resume surface: creates a draft, reaches Step 2, saves, leaves,
-// then reopens via the dashboard "Edit" link and asserts it lands back on Step 2.
+// data-org picker. Seeds a DRAFT with Step 2 progress so the routing rule is testable in
+// both collaborative and single-user editing modes (CI runs without Hocuspocus).
 test('Researcher resumes a Step 2 draft on Step 2', async ({ browser, studyFeatures }) => {
     const studyTitle = studyFeatures.uniqueTitle('resume-step2')
+    await seedDraftStep2(studyTitle)
```
nathanstitt: Seeding the row directly makes this test pass, but I think it stops covering what OTTER-572 cares about: `insertStudy` pre-populates `piUserId`, `datasets` and I think it's possibly why the tests are passing even tho we don't have a multi-user editor in CI

src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx
```diff
@@ -40,11 +39,10 @@ export const ProposalFooter: FC<ProposalFooterProps> = ({ researcherName, resear
         submitProposal()
     }
 
-    const handlePrevious = async () => {
-        const saved = await saveDraft()
-        // /edit is revisitable — it renders Step 1 directly and no longer resume-redirects, so the
-        // back-step needs no signal.
-        if (saved) router.push(Routes.studyEdit({ orgSlug, studyId }))
+    const handlePrevious = () => {
+        // Autosave (Yjs + server-side title mirror) persists edits, so back-navigation
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/footer.tsx]: This assumption holds in collaborative mode, but in single-user mode (the `singleUserEditing` env flag — which is what CI and pr envs use we still need to save it ourselves I think?  Of course CI is passing so maybe we're not testing this properly?

src/contexts/edit-resubmit/context.test.tsx
```diff
@@ -2,22 +2,17 @@ import { type Mock, describe, expect, it, vi } from 'vitest'
...
 import { fireEvent, renderWithProviders, screen, waitFor } from '@/tests/unit.helpers'
 import { EditResubmitProvider, useEditResubmit } from './context'
-import {
-    onUpdateClarifiedProposalAction,
-    resubmitProposalAction,
-    saveProposalResubmissionNoteDraftAction,
-} from '@/server/actions/study-request'
+import { resubmitProposalAction, saveProposalResubmissionNoteDraftAction } from '@/server/actions/study-request'
 
 vi.mock('@/server/actions/study-request', () => ({
```
nathanstitt, line 7: With `use-resubmit-save-draft.ts` deleted, `onUpdateClarifiedProposalAction` in `src/server/actions/study-request.ts` is dead code, we should remove it.

nathanstitt (commented):
Looks good but think we still will need to save title ourselves in CI & PR envs, which means we might as well save it ourselves everywhere on nav, and maybe not need to do so in the editor?

stetradis:
@nathanstitt Updated to address the review comments:

- **Single-user save (title in CI/PR):** added save-on-navigate — `useSaveProposalDraft` flushes the form (title included) to the study row via `onUpdateDraftStudyAction` before Back/Previous. Runs in all modes, so it's the write path where Hocuspocus isn't active.
- **Step 2 resume test:** rewrote it to drive the real flow (Step 1 → fill Step 2 → Previous → dashboard) instead of seeding a pre-populated row, so it actually exercises the resume path in single-user CI.
- **Dead code:** removed `onUpdateClarifiedProposalAction` and the unused `seedDraftStep2` helper.

On dropping the editor mirror — I don't think we can. Nav-save only fires on the Back/Previous button, whereas the editor mirror runs from the Hocuspocus store hook, so it rides the debounced flush and disconnect. That's what keeps the dashboard title fresh when a user leaves any other way — navigating straight to the dashboard, closing the tab, sidebar navigation, browser-back. So the two are complementary: mirror is the primary path in prod (all exits), nav-save is the fallback in single-user CI/PR (button paths).

services/editor/title-mirror.ts
```diff
@@ -0,0 +1,38 @@
...
+    if (parsed.kind !== 'proposal-fields') return
+
+    const title = readTitleFromFieldsDoc(doc)
+
+    await db.query(
+        `UPDATE study
+            SET title = $2::text
+          WHERE id = $1
+            AND status IN ('DRAFT', 'CHANGE-REQUESTED')
+            AND ($2::text IS NOT NULL OR status = 'DRAFT')`,
```
nathanstitt, line 35: Nice guard — a blank title can only ever land on a DRAFT row, so the mirror can't trip the `study_title_required_when_not_draft` constraint from a mid-rename flush, and the client-side `omitBlankTitle` option mirrors the same rule on the nav-flush path. Both sides are tested.

src/contexts/study-request/hooks/use-save-draft.ts
```diff
@@ -23,8 +23,9 @@ export function useSaveDraft({ studyId, submittingOrgSlug, onStudyCreated }: Use
 
     const mutation = useMutation({
         mutationFn: async (formValues: StudyProposalFormValues) => {
+            // title is omitted: it's owned by the Step 2 editor's autosave mirror and by
```
nathanstitt, line 26: Verified this is safe: Step 1 renders no title input (the only title fields live in the Step 2 editor and the resubmit edit section), so nothing user-entered is dropped here — it purely stops a mount-time snapshot from clobbering a title that Step 2 mirrored or flushed in the meantime. Subtle catch.

nathanstitt (approved):
thank you!

PR #863, opened 2026-07-08
Speed up the Checks CI workflow: Turbopack build cache + cache/Postgres/concurrency tuning

therealmarv:
## Summary

This PR speeds up the **Checks** CI workflow. Measured across full runs on this branch, the whole workflow dropped from **~5m12s to ~4m16s, an average of ~56s (~18%) faster**, and the `e2e` job (the former long pole) from **~4m36s to ~3m23s (~26% faster)**. It's now balanced with the unit shards instead of gating the run.

All changes are correctness-neutral: no test was weakened, retries stay at 0, and every optimization either fails loudly or was reverted (see "Tried and reverted").

## What changed

1. **Turbopack persistent build cache for the CI e2e build.** This repo is on Next 16, where `next build` uses Turbopack, whose filesystem build cache is experimental and off by default. The workflow's `.next/cache` step was therefore caching an almost-empty directory (~434 KB), so every build ran cold (~77s). Enabling `experimental.turbopackFileSystemCacheForBuild` (gated behind a new `TURBOPACK_FS_CACHE` env var set only on the CI e2e job, never on the production deploy build) makes Turbopack write a real ~430 MB cache into `.next/cache`. Warm builds drop to ~52s.

2. **Next cache: restore-always, save-on-main-only** (`actions/cache/restore` + `actions/cache/save` split). The ~430 MB cache upload added ~52s to the post-job phase of every PR run. We now save only on push to `main`; PRs restore `main`'s cache via the `tpc1-<lockfileHash>-` restore-keys prefix and rebuild incrementally. This removes the single biggest chunk of PR e2e time. A `tpc1` token in the cache key allows a one-character cache bust.

3. **Faster Postgres service startup** on both `unit` and `e2e`: `--health-interval 10s` to `2s`. The runner only waits for Docker to report the container healthy, so a shorter poll cuts the idle wait after Postgres is ready. "Initialize containers" went from ~22s to ~9s. Same `pg_isready` check.

4. **Concurrency group** that cancels superseded runs on the same PR ref (frees runner capacity, faster feedback on rapid re-pushes). Push runs get a unique group per run (`github.run_id`), so main is never grouped and every main run completes and seeds the cache, that is, main behaves exactly as it did before this change.

## Benchmark results

Baseline is PR #857 (predates this branch: no Turbopack cache, no items above). "After" is the average of the last 6 successful Checks runs on this branch.

| Metric | Before (PR #857) | After (avg of last 6) | Improvement |
|---|---|---|---|
| **Full Checks workflow** | 5m12s (312s) | **~4m16s (256s)** _(range 4m08s–4m32s)_ | **~56s (~18%)** |
| **e2e job** | 4m36s (276s) | **~3m23s (203s)** _(range 3m04s–3m46s)_ | **~73s (~26%)** |

Where the e2e time went (before → after):

| e2e step | Before | After | Cause |
|---|---|---|---|
| build and run server | ~77s (cold) | ~52s (warm) | Turbopack build cache (1) |
| Post cache save | ~52s | 0s on PRs (skipped) | save-on-main split (2) |
| Initialize containers (Postgres) | ~22s | ~9s | health-interval 2s (3) |

The full e2e suite passed at `retries: 0` on every reliable run.

### Turbopack build-cache warm/cold samples (earlier, item 1 in isolation)

| Run | Cache state | build time |
|---|---|---|
| Cold (seed) | seeding | 78s |
| Warm 1-4 | restored (restore-key) | 57s / 55s / 51s / 53s |

## Is the experimental Turbopack flag safe here?

Yes, and the CI e2e build is one of the safer places for it:

- Next's docs label the FileSystem Cache "stable for development and experimental for production builds." There is no documented case of it silently producing stale or wrong output; the identical cache engine has been stable and on-by-default for `next dev` since 16.1.
- The realistic failure mode is loud: known issues (e.g. a `Z_BUF_ERROR`) are build failures fixed by a clean-cache rebuild. A broken build fails the `build and run server` step and turns the job red; it cannot produce a false-green test run, because a broken build means no server for Playwright to hit.
- Change detection is content-hash + dependency-graph based, and our cache key also hashes every `.ts/.tsx`, so any source change misses the exact key and rebuilds incrementally.
- Bust it by bumping the `tpc1` token (or clearing `.next/cache`). The flag is scoped to CI via `TURBOPACK_FS_CACHE`, so the production Jenkins deploy build is unaffected and keeps building clean.

## Tried and reverted (kept out for correctness)

- **Backgrounding the Playwright install to overlap the build (~12s).** `playwright install-deps` shells out to `sudo`/apt, which hangs intermittently (~1 in 3 runs) when detached from the step, violating the zero-flake rule. Its ~12s benefit was also lost in run-to-run noise. Reverted to a synchronous install; a code comment records why.
- **`postgres:15-alpine`** (smaller pull). Alpine uses musl libc with different default text collation than the Debian image, which can change `ORDER BY` results and cause flaky tests. Kept the Debian image; the health-interval change delivers the win risk-free.

## Cache retention note

GitHub evicts caches automatically (7-day-since-last-access and a 10 GB per-repo LRU), so nothing needs manual cleanup. The `tpc1` cache is read on every e2e run, so it stays at the recently-used end and is not an eviction candidate; stale legacy caches are evicted first.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -145,10 +145,11 @@ jobs:
                     POSTGRES_USER: si
                     POSTGRES_DB: si_test
                     POSTGRES_PASSWORD: si
-                # Poll pg_isready every 2s (not 10s): the runner just waits for Docker to
-                # report the container healthy, so a shorter interval cuts the idle wait
-                # after Postgres is actually ready. Same check, faster detection.
-                options: --health-cmd pg_isready --health-interval 2s --health-timeout 5s --health-retries 5
+                # Poll pg_isready every 2s (not 10s) so the runner detects "healthy" sooner
+                # and cuts the idle wait after Postgres is ready. Raise retries 5 -> 25 so the
+                # total startup grace stays ~50s (2s x 25 = 50s), matching the old 10s x 5;
+                # faster polling must not shrink the tolerance for a slow cold start.
+                options: --health-cmd pg_isready --health-interval 2s --health-timeout 5s --health-retries 25
         env:
             DATABASE_URL: 'postgres://si:si@localhost:5432/si_test'
             SIMULATE_CODE_BUILD: 't'
@@ -221,10 +222,11 @@ jobs:
                     POSTGRES_USER: si
                     POSTGRES_DB: si_test
                     POSTGRES_PASSWORD: si
-                # Poll pg_isready every 2s (not 10s): the runner just waits for Docker to
-                # report the container healthy, so a shorter interval cuts the idle wait
-                # after Postgres is actually ready. Same check, faster detection.
-                options: --health-cmd pg_isready --health-interval 2s --health-timeout 5s --health-retries 5
+                # Poll pg_isready every 2s (not 10s) so the runner detects "healthy" sooner
+                # and cuts the idle wait after Postgres is ready. Raise retries 5 -> 25 so the
+                # total startup grace stays ~50s (2s x 25 = 50s), matching the old 10s x 5;
+                # faster polling must not shrink the tolerance for a slow cold start.
+                options: --health-cmd pg_isready --health-interval 2s --health-timeout 5s --health-retries 25
         env:
             DATABASE_URL: 'postgres://si:si@localhost:5432/si_test'
             SIMULATE_CODE_BUILD: 't'
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -20,6 +20,13 @@ on:
             - main
     pull_request: {}
 
+# Cancel superseded runs on the same PR ref so an outdated commit's jobs free the runner
+# for the newest push. Gated to pull_request only: push-to-main runs never cancel, so they
+# always finish and seed the shared build cache + coverage baseline.
+concurrency:
+    group: checks-${{ github.ref }}
+    cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
     lint:
         timeout-minutes: 5
@@ -133,7 +140,10 @@ jobs:
                     POSTGRES_USER: si
                     POSTGRES_DB: si_test
                     POSTGRES_PASSWORD: si
-                options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+                # Poll pg_isready every 2s (not 10s): the runner just waits for Docker to
+                # report the container healthy, so a shorter interval cuts the idle wait
+                # after Postgres is actually ready. Same check, faster detection.
+                options: --health-cmd pg_isready --health-interval 2s --health-timeout 5s --health-retries 5
         env:
             DATABASE_URL: 'postgres://si:si@localhost:5432/si_test'
             SIMULATE_CODE_BUILD: 't'
@@ -206,7 +216,10 @@ jobs:
                     POSTGRES_USER: si
                     POSTGRES_DB: si_test
                     POSTGRES_PASSWORD: si
-                options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+                # Poll pg_isready every 2s (not 10s): the runner just waits for Docker to
+                # report the container healthy, so a shorter interval cuts the idle wait
+                # after Postgres is actually ready. Same check, faster detection.
+                options: --health-cmd pg_isready --health-interval 2s --health-timeout 5s --health-retries 5
         env:
             DATABASE_URL: 'postgres://si:si@localhost:5432/si_test'
             SIMULATE_CODE_BUILD: 't'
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yml
```diff
@@ -277,6 +314,17 @@ jobs:
                   path: test-results/e2e/coverage/raw
                   retention-days: 1
                   if-no-files-found: ignore
+            # Save the Next build cache ONLY on push-to-main (see the restore step above).
+            # Runs only on success, so a failed build/test never persists a poisoned cache.
+            - name: Save Next build cache (main only)
+              # Guard on both the event and the ref: `push` is only configured for `main`, but a
+              # merged `pull_request` run also reports github.ref == refs/heads/main, so the
```
nathanstitt: Small factual nit in the comment: a `pull_request` run (including a `closed`/merged one) reports `github.ref` as `refs/pull/<n>/merge`, never `refs/heads/main` — it's `pull_request_target` that reports the base branch ref. The belt-and-braces guard itself is still worth keeping (it does protect against a future `pull_request_target` or extra push trigger), so maybe just reword the justification so the next reader doesn't absorb the wrong fact about `pull_request` refs.

.github/workflows/checks.yml
```diff
@@ -277,6 +314,17 @@ jobs:
...
+              # event check keeps the save strictly on push-to-main even if triggers change later.
+              if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
nathanstitt: On a re-run of a main workflow (e.g. re-running the same commit after a flake), the exact primary key already exists, so this save will tar/compress ~430 MB and then fail to reserve the cache — a warning, not a failure, but wasted minutes on main's critical path. Adding `&& steps.next_cache.outputs.cache-hit != 'true'` to the guard would skip the doomed save (`cache-hit` from `actions/cache/restore` is only `true` on an exact primary-key match, so incremental prefix-restores still save).

next.config.ts
```diff
@@ -9,6 +9,14 @@ const isDev = Boolean(process.env.CI || process.env.NODE_ENV === 'development')
 // (flag unset) are untouched. See src/lib/clerk-fake/README intent in server.ts.
 const fakeClerk = Boolean(process.env.E2E_FAKE_CLERK)
 
+// Turbopack's persistent filesystem cache for `next build` is experimental (opt-in) in
+// Next 16, so it's gated behind TURBOPACK_FS_CACHE and only turned on for the CI e2e build
+// (see .github/workflows/checks.yml), never for the production deploy build. It writes to
+// .next/cache, which CI persists across runs to make incremental rebuilds much faster. If
+// a cached build ever looks wrong it fails loudly at build time (never a false-green test
```
nathanstitt: The "fails loudly at build time (never a false-green test run)" claim covers the *broken*-build failure mode, but the risk mode for a persistent build cache is the *stale* build: if Turbopack's invalidation ever missed a change, the server would run without that change and e2e could pass against old code — which is precisely a quiet false green. The PR description makes the case this is unlikely (content-hash change detection, no documented staleness cases), and the exact cache key hashing every source file helps, so the risk seems acceptably low — but we could soften this comment to match, and maybe note the `tpc` bump as the recovery procedure for suspected staleness, not just loud breakage. A periodic cold build (e.g. a scheduled run that skips restore-keys) would be a cheap backstop if we ever want one.

nathanstitt (approved):
seems like a nice improvement, ty!

.github/workflows/checks.yml
therealmarv: Good catch, that was factually wrong. Fixed in f1257bc0: a `pull_request` run reports `refs/pull/<n>/merge`, not `refs/heads/main`, so today the event check alone would suffice. Kept the ref check and reworded the justification to frame it as belt-and-braces for a future `pull_request_target` or extra push trigger.

.github/workflows/checks.yml
therealmarv: Agreed, applied in f1257bc0. Added `&& steps.next_cache.outputs.cache-hit != 'true'` to the save guard so a main re-run on an unchanged commit skips the ~430 MB tar-then-fail-to-reserve. Since `cache-hit` is `true` only on an exact primary-key match, incremental prefix-restores still save on main as before.

next.config.ts
therealmarv: Fair, the original comment only covered the loud broken-cache case. Reworded in f1257bc0 to distinguish the corrupt-cache failure (loud, red build) from the quieter stale-build risk (missed invalidation), note that content-hash detection plus the source-hashing cache key make staleness unlikely, and point to the tpc token bump as the recovery procedure for suspected staleness. A periodic cold build sounds like a reasonable cheap backstop if we want one later; happy to add a scheduled restore-key-skipping run as a follow-up.

PR #864, opened 2026-07-08
remove reject code option

jbwilson8:
## OTTER-650: Remove the "Reject and end study" option from the DP Code Review page

Issue: https://openstax.atlassian.net/browse/OTTER-650

### Summary

UX and Product are re-conceptualizing how studies are "ended." As an interim step,
leadership asked to remove the **"Reject and end study"** option from the Data Partner
(DP) Code Review page. This PR removes only that option (the user-facing entry point)
and deliberately leaves the surrounding reject infrastructure in place so it can be
re-wired into a new pathway later.

### What changed

- **Removed the `reject` decision option** from the code-review decision group
  (`code-review-feedback-section.tsx`). A Data Partner now sees only **Approve and run
  code** and **Request revision**.

### Intentionally left untouched (per the card)

These are preserved so the future pathway can reuse them — none are removed:

- The "Reject study code?" confirmation modal in `code-review-client.tsx`
- The `'reject'` value in the `Decision` type
- The reject branch of `submitCodeReviewDecisionAction`
- The DP/RL post-code-rejection views (`post-feedback-view.tsx`)

They are unreachable via the UI now but remain fully wired.

### Tests

- **`code-review-client.test.tsx`** — updated the decision-options test to assert the
  reject option is absent, dropped the reject case from the enable-submit `it.each`, and
  removed the "opens the destructive reject modal" test (that flow is no longer reachable
  through the UI).
- **`study-flow.spec.ts`** — the "Code rejection ends the study" e2e test no longer drives
  the removed button. It now seeds a `CODE-REJECTED` study directly via the existing
  `seedCodeRejected` helper and verifies the preserved reviewer + researcher terminal
  views, keeping coverage of the post-rejection pages.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/code-review-client.test.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/code-review-client.test.tsx
@@ -79,7 +79,7 @@ describe('CodeReviewClient decision selector', () => {
         })
     })
 
-    it('renders all three decision options with their titles and descriptions', async () => {
+    it('renders both decision options with their titles and descriptions', async () => {
         const { study, job, orgSlug, previousHref } = await setupValidReviewableJob('Rice University')
         renderWithProviders(
             <CodeReviewClient
@@ -93,11 +93,12 @@ describe('CodeReviewClient decision selector', () => {
 
         expect(screen.getByTestId('code-review-decision-approve')).toBeInTheDocument()
         expect(screen.getByTestId('code-review-decision-needs-clarification')).toBeInTheDocument()
-        expect(screen.getByTestId('code-review-decision-reject')).toBeInTheDocument()
+        // OTTER-650: the "Reject and end study" option was removed as an interim step.
+        expect(screen.queryByTestId('code-review-decision-reject')).not.toBeInTheDocument()
 
         expect(screen.getByText('Approve and run code')).toBeInTheDocument()
         expect(screen.getByText('Request revision')).toBeInTheDocument()
-        expect(screen.getByText('Reject and end study')).toBeInTheDocument()
+        expect(screen.queryByText('Reject and end study')).not.toBeInTheDocument()
 
         expect(
             screen.getByText(
@@ -109,12 +110,6 @@ describe('CodeReviewClient decision selector', () => {
                 /Return this code submission to Rice University for necessary updates, additional information, or specific changes\./,
             ),
         ).toBeInTheDocument()
-        expect(
-            screen.getByText(
-                /Permanently end this study due to major, unresolvable issues\. Share rationale with Rice University\./,
-            ),
-        ).toBeInTheDocument()
-        expect(screen.getByText('Warning: This terminates the study and cannot be undone.')).toBeInTheDocument()
     })
 
     it('renders the editable review form (not "Code review is closed") for an APPROVED study with a reviewable job', async () => {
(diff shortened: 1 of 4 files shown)


src/app/[orgSlug]/study/[studyId]/review/code-review-feedback-section.tsx
```diff
@@ -105,20 +105,6 @@ const buildDecisionOptions = (labName: string): DecisionOption[] => [
...
-                Permanently end this study due to major, unresolvable issues. Share rationale with {labName}.
-                <br />
-                <Text component="span" size="sm" c="grey.7" fw={600}>
-                    Warning: This terminates the study and cannot be undone.
-                </Text>
-            </Text>
-        ),
-        testId: 'code-review-decision-reject',
-    },
 ]
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/code-review-feedback-section.tsx:108]: Removing the `reject` option here has a side effect worth catching for the dead-code goal: it orphans a chain in `code-review-client.tsx` (which this PR doesn't touch, so it's easy to miss). `useReviewDecision` starts at `null` and only this radio group calls `onSelect`, so with `reject` gone `decision.selected` can never be `'reject'`. That makes the following unreachable in `code-review-client.tsx`:

- the `if (decision.selected === 'reject') openReject()` branch in `handleSubmit` (~line 95)
- the `rejectOpen` / `openReject` / `closeReject` disclosure (~line 67), plus its threading through the hook return and destructure
- the second `ReviewConfirmationModal` — "Reject study code?" (~lines 267-278) — since `isOpen={rejectOpen}` is now always `false`
- the `REJECTION_WARNING` import (~line 11), used only inside that dead modal

This is distinct from the infra the description says it's intentionally keeping: the server action's reject branch, the `'reject'` value in the `Decision` type, and the post-rejection views are genuinely reachable by *other* callers (proposal reject writes `CODE-REJECTED`), so preserving those reads fine. But the modal + disclosure live inside the very component whose entry point was just deleted — they can't be re-wired without also restoring this radio option, so they're not really "reusable in place." Since dead-code removal was the explicit ask: worth deleting that modal, the disclosure, the `=== 'reject'` branch, and the `REJECTION_WARNING` import, and letting git history carry the pattern. (eslint won't flag any of it — every symbol is still referenced in the return object / JSX, just never triggerable.)

src/app/[orgSlug]/study/[studyId]/review/code-review-client.test.tsx
```diff
@@ -93,11 +93,12 @@ describe('CodeReviewClient decision selector', () => {
 
         expect(screen.getByTestId('code-review-decision-approve')).toBeInTheDocument()
         expect(screen.getByTestId('code-review-decision-needs-clarification')).toBeInTheDocument()
-        expect(screen.getByTestId('code-review-decision-reject')).toBeInTheDocument()
+        // OTTER-650: the "Reject and end study" option was removed as an interim step.
+        expect(screen.queryByTestId('code-review-decision-reject')).not.toBeInTheDocument()
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/code-review-client.test.tsx:97]: If we do delete the now-unreachable reject modal + `rejectOpen` disclosure in `code-review-client.tsx` (see the note on `code-review-feedback-section.tsx`), this file still has coverage that would need a companion trim — e.g. the removed "opens the destructive reject modal" test is already gone here, good. Just flagging that the test-side cleanup in this PR is the tell that the component-side dead code was left behind: the test for the reject modal was removable precisely because the flow is unreachable, which is the same reason the modal itself is.

nathanstitt (commented):
looks good but we should also remove the dead code in spots noted below

nathanstitt (approved):
thank you!

PR #866, opened 2026-07-09
OTTER-621: derive proposal decision from approvedAt/rejectedAt

Rod0352:
### Problem
With a study in code-needs-review, "View approved initial request" opened
/review/proposal but rendered the code-review screen instead of the decided
proposal (or a blank page when no feedback rows existed). Intermittent — only
while code was awaiting review.

### Cause
Code submit/resubmit flips study.status back to PENDING-REVIEW. The proposal
page and feedback screen checked PROPOSAL_STATUS_TO_REVIEW_DECISION[study.status],
which has no PENDING-REVIEW entry, so a decided proposal looked undecided.

### Changes
- New proposalReviewDecision() in src/lib/review-decision.ts: status map first,
  then fall back to approvedAt/rejectedAt (they survive the status flip and
  mutually clear each other)
- /review/proposal page and ReviewerProposalFeedbackScreen now use it;
  undecided proposals still fall through to the editable review

### Testing
- New review-decision.test.ts (map, timestamp fallbacks, precedence)
- Regression test: PENDING-REVIEW + approvedAt renders PostFeedbackView with
  APPROVE fallback (the staging scenario from the ticket video)
- /review page + reviewer-screen-rules suites pass; lint/tsc clean

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-proposal-feedback-screen.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-proposal-feedback-screen.tsx
@@ -1,7 +1,7 @@
 import { isSubmittedStudy } from '@/schema/study'
 import { isActionError } from '@/lib/errors'
 import { AlertNotFound } from '@/components/errors'
-import { PROPOSAL_STATUS_TO_REVIEW_DECISION } from '@/lib/review-decision'
+import { proposalReviewDecision } from '@/lib/review-decision'
 import { getProposalFeedbackForStudyAction } from '@/server/actions/study.actions'
 import { PostFeedbackView } from '../review/post-feedback-view'
 import type { ScreenComponentProps } from './types'
@@ -12,7 +12,7 @@ export async function ReviewerProposalFeedbackScreen({ study, orgSlug }: ScreenC
     }
     const entries = await getProposalFeedbackForStudyAction({ studyId: study.id })
     const safeEntries = isActionError(entries) ? [] : entries
-    const decision = PROPOSAL_STATUS_TO_REVIEW_DECISION[study.status]
+    const decision = proposalReviewDecision(study)
     const fallback = decision
         ? { decision, timestamp: study.approvedAt ?? study.rejectedAt ?? study.createdAt }
         : undefined
--- a/src/lib/review-decision.test.ts
+++ b/src/lib/review-decision.test.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it } from 'vitest'
+import type { StudyStatus } from '@/database/types'
+import { proposalReviewDecision } from './review-decision'
+
+const study = (status: StudyStatus, overrides: { approvedAt?: Date | null; rejectedAt?: Date | null } = {}) => ({
+    status,
+    approvedAt: overrides.approvedAt ?? null,
+    rejectedAt: overrides.rejectedAt ?? null,
+})
+
+describe('proposalReviewDecision', () => {
+    it('maps decided statuses directly', () => {
+        expect(proposalReviewDecision(study('APPROVED'))).toBe('APPROVE')
+        expect(proposalReviewDecision(study('REJECTED'))).toBe('REJECT')
+        expect(proposalReviewDecision(study('CHANGE-REQUESTED'))).toBe('NEEDS-CLARIFICATION')
+    })
+
+    it('falls back to approvedAt when code submission reset the status', () => {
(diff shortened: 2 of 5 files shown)


src/app/[orgSlug]/study/[studyId]/_screens/reviewer-proposal-feedback-screen.tsx
```diff
@@ -12,7 +12,7 @@ export async function ReviewerProposalFeedbackScreen({ study, orgSlug }: ScreenC
     }
     const entries = await getProposalFeedbackForStudyAction({ studyId: study.id })
     const safeEntries = isActionError(entries) ? [] : entries
-    const decision = PROPOSAL_STATUS_TO_REVIEW_DECISION[study.status]
+    const decision = proposalReviewDecision(study)
     const fallback = decision
         ? { decision, timestamp: study.approvedAt ?? study.rejectedAt ?? study.createdAt }
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-proposal-feedback-screen.tsx:17]: Nice fix — I traced the whole timestamp lifecycle and the premise holds: every write of `approvedAt`/`rejectedAt` nulls the other (study.actions `performStudyProposalApproval`/`markStudyRejected`), and the code-submit/resubmit paths flip status to PENDING-REVIEW via an explicit field allowlist (`snapshotFields` in study-request) that never touches the timestamps, so `approvedAt` genuinely survives the flip.

One small structural note here: the *decision* now comes from `proposalReviewDecision(study)` (status-map-first, timestamp-fallback), but the *timestamp* is re-derived independently as `approvedAt ?? rejectedAt ?? createdAt`. Because the two exclusivity invariants hold today, these always agree. But they encode the same "which decision won" logic in two places, so a future change to the timestamp lifecycle could desync them (e.g. if the map ever returned APPROVE while only `rejectedAt` were set, the banner would show the rejection date). Might be worth having `proposalReviewDecision` return the decision *and* its effective timestamp together so there's a single source of truth — optional cleanup, not blocking.

src/lib/review-decision.ts
```diff
@@ -25,6 +25,21 @@ export const PROPOSAL_STATUS_TO_REVIEW_DECISION: Partial<Record<StudyStatus, Rev
...
+// whether the proposal was ever decided. approvedAt/rejectedAt survive those flips (approval and
+// rejection clear each other), making them the durable record once the study is code-stage.
+export function proposalReviewDecision(study: {
+    status: StudyStatus
+    approvedAt: Date | null
+    rejectedAt: Date | null
+}): ReviewDecision | undefined {
+    const byStatus = PROPOSAL_STATUS_TO_REVIEW_DECISION[study.status]
+    if (byStatus) return byStatus
+    if (study.approvedAt) return 'APPROVE'
```
nathanstitt, line 38: Worth a one-line comment on the ordering invariant this relies on: `approvedAt` before `rejectedAt` is only unambiguous because the two are mutually exclusive at the DB layer (each write nulls the other). It's true today, but it's the kind of thing that's easy to break silently — a note here ("approvedAt/rejectedAt are mutually exclusive; see performStudyProposalApproval/markStudyRejected") would flag the invariant for anyone who later adds a path that sets both.

src/lib/review-decision.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    it('falls back to approvedAt when code submission reset the status', () => {
+        expect(proposalReviewDecision(study('PENDING-REVIEW', { approvedAt: new Date() }))).toBe('APPROVE')
+    })
+
+    it('falls back to rejectedAt when the status no longer carries the decision', () => {
+        expect(proposalReviewDecision(study('PENDING-REVIEW', { rejectedAt: new Date() }))).toBe('REJECT')
+    })
+
+    it('prefers the explicit status over stale timestamps', () => {
```
nathanstitt, line 26: This case (`CHANGE-REQUESTED` + `approvedAt` set) can't actually occur in the DB — the clarification path sets `status: 'CHANGE-REQUESTED'` and clears both timestamps in the same update (study.actions ~L665). It's a fine defensive test of the precedence rule, but since it asserts on an impossible state, a future reader might infer such rows exist. Maybe rename it to something like 'status map takes precedence (defensive: timestamps ignored when a decided status is present)' so the intent reads as guarding the branch order rather than modeling a real state.

nathanstitt (approved):
I think this is fine, but the real fix is to no longer modify study status when code changes, that's the real issue we're fighting here.  Doing so will be a larger fix though and we can tackle that later

PR #867, opened 2026-07-09
(don't merge, wait for TS 7.x API layer to arrive) Upgrade production to TypeScript 6 and add TypeScript 7 to dev tooling

therealmarv:
## Summary

> **This branch is now scoped to the TypeScript work only.** It upgrades production to TypeScript 6 (6.0.3) and wires the Go-native TypeScript 7 compiler into developer tooling (the `typecheck` used by `pnpm checks` / `pre:push`, plus the editor), while the production build, `next build`, and `typescript-eslint` all continue to use TypeScript 6. TS7 never reaches the shipped product.

No application source changed. `pnpm run checks` (TS7 typecheck), `pnpm run typecheck:ts6` (full TS6), and the Ladle build pass throughout.

### Rebased: unrelated changes now come from main

This branch was originally cut before several general repo improvements landed on `main`. It has been rebased onto current `main`, and three changes that were previously carried here have been dropped because they were merged independently and are already on `main`:

- **Parallelizing eslint/prettier and content-hash caching in `pnpm checks`** (merged in #869).
- **Removing the redundant `validate-actions` CI step** (also merged in #869).
- **Removing the unused `next-swagger-doc` dependency** (merged in #869).

As a result, this PR's diff is now limited to TypeScript 6, TypeScript 7 dev tooling, and the two TypeScript-coupled dependency bumps described below. The check-speed benchmarks that previously mixed in the lint parallelization have been trimmed to the part this PR is actually responsible for: the TypeScript compiler itself.

---

### Why TypeScript 6, and why now

TypeScript 7 is the Go-native compiler rewrite. It is much faster, but it ships CLI-first and defers a stable programmatic API to 7.1. Next.js relies on TypeScript's programmatic API (`typescript/lib/typescript.js`) for `next typegen` and its build-time type integration, and `typescript-eslint` needs it too, so a direct jump of the resolved `typescript` package to 7.x is not possible yet.

TypeScript 6 is Microsoft's designed answer to exactly this transition: the last release built on the current JavaScript codebase, positioned as a bridge that is semantically aligned with 7.0. Upgrading production to 6.0.3 now means the day we can adopt TS7 fully, the jump is a no-op semantically.

Every load-bearing claim behind this plan was verified against official sources (typescriptlang.org, devblogs.microsoft.com, npm, GitHub releases, the Next.js docs) before acting:

| Claim | Verdict |
| --- | --- |
| TypeScript 6.0 is a real, official "last JS-based / bridge" release | Confirmed |
| Latest stable 6.x is 6.0.3 (plain `typescript@^6` still resolves to it) | Confirmed |
| Next.js 16 needs only TypeScript >= 5.1.0, so 6.x needs no Next change | Confirmed |
| `typescript-eslint` already supports TS6 (declared range `>=4.8.4 <6.1.0`) | Confirmed |

### Why this is low risk

- **Not a structural rewrite.** TS6 keeps the same JS compiler codebase and the same programmatic API surface, so Next.js and `typescript-eslint` keep working natively with no shims or aliasing.
- **Zero source changes required.** `next typegen` + `tsc --noEmit`, ESLint + Prettier, action validation, and the full unit suite all passed unchanged on 6.0.3.
- **Next.js untouched.** 6.x satisfies Next 16's documented minimum (TS 5.1); the build uses SWC/Turbopack and only uses TypeScript as the type-check guard.
- **The linter already supported it.** `typescript-eslint` already declared support for the 6.x range, so there was no "unsupported TypeScript version" warning.

### Dependency bumps carried by this PR

**Dropped the one unmaintained package from production.** `tsconfck` (marked unmaintained) reached the production tree via `kysely-ctl@0.20.0`. Upgrading to `kysely-ctl@0.21.0` (which switched its config loader from `tsx` to `jiti` and no longer depends on `tsconfck`) removed it from production. The only remaining reference is a dev-only transitive via `@ladle/react` -> `vite-tsconfig-paths`, which has no `tsconfck`-free release yet; it is out of the production tree and functionally fine under TS6.

**Kept the TypeScript-coupled linters current.** `typescript-eslint` 8.56.1 -> 8.63.0 and `eslint-import-resolver-typescript` 4.4.4 -> 4.4.5. ESLint itself was intentionally left on the latest 9.x rather than 10.x, because `@next/eslint-plugin-next` is not yet validated for ESLint 10; that is a separate future change.

> Note: the removal of the unused `next-swagger-doc` dependency (which cleared the last deprecated transitive packages from the production tree) landed separately in #869 and is already on `main`, so it is no longer part of this PR.

---

### How it is wired so it cannot leak into the build

- The resolved `typescript` package stays on **6.x**. Node resolves the bare specifier `typescript` by the `node_modules/typescript` directory name, not by a package's internal name, so nothing that does `require('typescript')` can resolve to 7.x. This was verified after install: `require('typescript')` returns 6.0.3, and `typescript-eslint` also resolves 6.0.3.
- TS7 is installed under a separate name, `typescript7` (a pnpm alias to `typescript@7.0.2`), and invoked only by explicit path in the scripts. That sidesteps the `tsc` bin-name collision entirely.
- `next build`, `next typegen`, and `typescript-eslint` are therefore unaffected: production type-checking is still TypeScript 6.

### What now uses TS7

| Area | Compiler | How |
| --- | --- | --- |
| `typecheck` (feeds `pnpm checks` + `pre:push`) | TS7 Go | `node ./node_modules/typescript7/bin/tsc -p tsconfig.ts7.json` |
| `typecheck:ts6` (full-coverage fallback) | TS6 | unchanged flags |
| Next.js build | TS6 | unchanged |
| `typescript-eslint` (lint) | TS6 | unchanged (needs the API TS7 lacks until 7.1) |
| Editor (VS Code) | TS7 Go LSP | `.vscode/settings.json` `js/ts.experimental.useTsgo: true` + recommends the Native Preview extension |
| Unit tests (vitest), Ladle | TS6 / esbuild / Vite | unchanged (do not use `tsc`) |

The TS7 fast gate uses `tsconfig.ts7.json`, which excludes the `*.stories.tsx` and `.ladle` files. Those files do a type-only `import type { Story } from '@ladle/react'`, which pulls `@ladle/react`'s own bundled `.tsx` into the program; TS7 is stricter than TS6 and reports 2 pre-existing type errors inside that third-party file (`--skipLibCheck` cannot skip it because it is `.tsx`, not `.d.ts`). This is dev-tool noise with no bearing on production. Those files remain fully type-checked by TS6 via `next build` (CI and production) and via `typecheck:ts6`.

---

## Benchmarks: the TypeScript compiler

This PR's contribution to check speed is the type-check step itself. Numbers are the median of repeated warm runs on a 14-core Apple Silicon machine, over the identical file set.

| Compiler | Warm |
| --- | --- |
| TS6 `tsc --noEmit` | 2.17s |
| TS7 (Go) `tsc --noEmit` | 0.51s |

That is about a **4.2x** speedup on the raw type-check. Because `pnpm checks` runs typecheck, eslint, prettier, and validate-actions concurrently, this drop feeds directly into the warm `checks` wall-clock. (The separate lint parallelization and caching that also speed up `checks` merged independently in #869 and are already on `main`.)

---

## Why this is correct

- **Route types are never stale.** `next typegen` runs fresh before every type-check (it is not cached or skipped), so the generated `.next/types` are always current for `tsc`.
- **TS6 remains the ultimate gate.** TS7 is a file-by-file port of TS6 and is semantically compatible; where it differs it is stricter, not looser. `next build` still type-checks with TS6 in CI and production, so TS6 is the authoritative full-coverage check. The TS7 fast gate is additive, faster feedback on top of that.
- **TS7 is isolated from the build** by construction (separate package name, explicit-path invocation, `require('typescript')` still resolves to 6.x), verified after install.

---

## Supply-chain note

`typescript@7.0.2` was published inside our 48-hour `minimumReleaseAge` supply-chain safeguard when this work started, so an earlier revision of this branch carried a temporary, scoped `minimumReleaseAgeExclude` for it. That version has since aged past the window, so the exclusion has been removed: `typescript7` now installs under the normal age gate with no exemption, and `pnpm-workspace.yaml` is unchanged from `main`.

---

## Validation

`pnpm run checks` (TS7 typecheck), `pnpm run typecheck:ts6` (full TS6), and the Ladle production build all pass. The production build is unchanged and remains on TypeScript 6.


nathanstitt:
I'd rather wait a month or so and update directly to TS7 once it settles, the wins are large enough that I think we can be a _little_ aggressive on jumping on the upgrade, vs making two large changes like this

therealmarv:
sure, this branch can be upgraded once the API coverage in TS 7.x lands and it's also usable in Next etc on production. Not in hurry with this one. It was anyway a fun side task. But I was surprised we had packages installed we never used like `next-swagger-doc`

therealmarv:
This can be reopened when TS 7.1 release happens.

Current status of TS 7.1 beta: https://github.com/microsoft/TypeScript/milestone/224

PR #868, opened 2026-07-09
OTTER-621: remove the code-submit study.status flip

nathanstitt:
### Problem
Code submit/resubmit flipped study.status back to PENDING-REVIEW (and stomped submittedAt with the code-submission time) even though the proposal was already approved. At least five layers of compensation existed to undo or route around it — #866 being the latest. Any new consumer reading status naively repeats OTTER-621.

### Changes
- submitStudyCodeAction / resubmitStudyCodeAction no longer touch status or submittedAt; job statuses (CODE-SUBMITTED etc.) alone drive the code phase, which screen rules, pills, dashboards, polling, and claim gates already key on
- Delete dead addJobToStudyAction; drop the unreachable APPROVED claim from finalizeStudySubmissionAction; remove the isCodeReapproval restore path (code approval belongs to submitCodeReviewDecisionAction)
- Data migration backfills rows stranded mid-code-review (PENDING-REVIEW + approved_at set -> APPROVED)
- Decision-path APPROVED writes kept one release as self-healing for rolling-deploy stragglers (commented); the approvedAt-IS-NULL claim guard and #866's timestamp fallback stay for the same reason
- Side fix (separate commit): SignOutPanel read window.location.href during SSR, 500ing the invitation route for signed-in users

### Testing
- Unit suite 1802/1802; `pnpm run checks` clean
- New round-trip test: approve -> submit -> needs-clarification -> resubmit keeps status APPROVED and submittedAt stable; migration test covers straggler vs proposal-stage rows
- Flip-asserting tests updated; UI submit tests now key on the job's CODE-SUBMITTED marker
- Local full-parallel e2e was unreliable on a saturated machine (load 24/14 cores, shuffling failure sets, 30s timeouts); all five touched study-flow specs pass in isolation. CI should confirm the full suite.

Base is #866's branch; retarget to main once it merges.


therealmarv (approved):
this is probably for @Rod0352 to review better.

Just wanted to say I see no problems with PR and it looks good.

PR #869, opened 2026-07-10
Speed up pnpm checks (parallelize + cache lint) and remove unused next-swagger-doc

therealmarv:
This is a minimal version of the TypeScript 6/7 upgrade PR ([#867](https://github.com/safeinsights/management-app/pull/867)). It takes only the safe, high-impact `pnpm checks` speedups from that branch and **does not touch any existing TypeScript version** (`typecheck` still runs the current `tsc`, and no `typescript` / `typescript-eslint` versions change here). The TS7 fast-compiler and TS6 production bump from #867 are intentionally left out.

## Headline result

Under ideal (warm, local repeat-run) conditions, `pnpm checks` goes from about **13.6s to about 3.9s, roughly a 3.5x speedup.**

| Condition | Before this branch | After this branch | Improvement |
| --- | --- | --- | --- |
| **Warm (local repeat, ideal)** | **~13.6s** | **~3.9s** | **~3.5x faster** |
| Cold (no lint cache, the CI case) | ~13.6s | ~13.5s | unchanged |

Cold runs are unchanged by design: a first, uncached ESLint pass dominates them, and caching cannot help a run that has no cache yet. The warm win is what developers feel on every local repeat run.

## What changed

1. **Parallelize and cache the lint step.** The `lint` script previously ran `eslint . && prettier --check .` sequentially, and `checks` treated that as one serial task. Now ESLint and Prettier run as two independent parallel tasks, each with content-hash caching (`--cache --cache-strategy content`), and `checks` runs typecheck, `lint:eslint`, `lint:prettier`, and `validate-actions` all concurrently. `lint:fix` gained the same cache flags.
2. **Remove the redundant CI step.** The `lint` CI job had a separate `validate actions` step that already runs inside `pnpm run checks`; it was dropped.
3. **Remove the unused `next-swagger-doc` dependency.** It was declared as a production dependency but is never imported (no `createSwaggerSpec`/`withSwagger` call, no swagger or api-doc route, nothing in `next.config`). It is a plain explicit-import library with no install-time side effects, so being installed does nothing. Removing it prunes 25 transitive packages, including the deprecated `glob@7.1.6`, `inflight`, `lodash.get`, and `lodash.isequal` that reached the tree only through it.

## Benchmarks

All numbers are `real` wall-clock from `/usr/bin/time -p` on Apple Silicon, median of repeated runs.

### `pnpm checks` end to end

| Scenario | Runs (s) | Median |
| --- | --- | --- |
| Before this branch (no caching) | 14.86 / 13.59 / 13.64 | ~13.6s |
| After, cold (lint cache cleared) | 13.48 | ~13.5s |
| After, warm (lint cache present) | 4.60 / 3.87 / 3.74 | **~3.9s** |

### Warm component breakdown (tasks run concurrently, so total is bound by the slowest)

| Task | Warm |
| --- | --- |
| typecheck (`next typegen` + `tsc`, unchanged) | 3.51s |
| lint:eslint (cached) | 1.65s |
| lint:prettier (cached) | 1.64s |
| validate-actions | 1.06s |

The warm floor is now the (deliberately unchanged) `tsc` typecheck. #867 pushes this further to about 1.9s by swapping in the TS7 Go compiler, which this PR does not do.

## Why this is correct

- **Content-hash caching cannot hide a new error.** `--cache-strategy content` keys the cache on file content plus config, not mtime, so any changed file (or a config change) is always re-checked. Verified against the installed binaries: both `eslint@9.39.4` and `prettier@3.8.3` accept `--cache-strategy <metadata|content>`.
- **CI stays a full run.** GitHub Actions runners do not persist the ESLint/Prettier cache across runs, so CI recomputes everything every time regardless of local caching.
- **Cache files are already gitignored** (`.eslintcache`; Prettier caches under `node_modules/.cache`).
- **`next-swagger-doc` removal is safe.** No source reference exists and the package has no `postinstall`/`preinstall` script, no `sideEffects`, and no Next.js plugin/loader/transform, so nothing depended on its mere presence. `pnpm run checks` passes after removal.

## Validation

`pnpm run checks` (typecheck + parallel cached lint + action validation) passes on this branch, both cold and warm.


package.json
```diff
@@ -31,13 +31,15 @@
         "create-admin-user": "tsx bin/create-admin-user.ts",
         "start": "next start",
         "deploy": "tsx bin/deploy.ts",
-        "lint": "eslint . && prettier --check --log-level warn .",
+        "lint": "concurrently 'pnpm:lint:eslint' 'pnpm:lint:prettier'",
+        "lint:eslint": "eslint --cache --cache-strategy content .",
+        "lint:prettier": "prettier --cache --cache-strategy content --check --log-level warn .",
         "validate-actions": "tsx tests/validate-actions.ts",
-        "lint:fix": "eslint --fix . && prettier --write --log-level warn .",
+        "lint:fix": "eslint --cache --cache-strategy content --fix . && prettier --cache --cache-strategy content --write --log-level warn .",
```
nathanstitt, line 38: Nice that `lint:fix` was left sequential (`eslint --fix && prettier --write`) rather than parallelized like the read-only `lint`. Running `eslint --fix` and `prettier --write` concurrently over the same files would race on writes, so the asymmetry here is the correct call — just noting it looked intentional and is worth keeping that way.

.github/workflows/checks.yml
```diff
@@ -54,8 +54,6 @@ jobs:
               run: pnpm list
             - name: pnpm run checks
               run: pnpm run checks
```
nathanstitt, line 56: Confirmed the dropped `validate actions` CI step is safe to remove: `checks` includes `pnpm:validate-actions` in its concurrently list, and I verified `concurrently@10.0.3` defaults `--success` to `all`, so if validate-actions (or any sibling task) fails, `pnpm run checks` still exits non-zero and this job goes red. No coverage or fail-signal is lost by removing the separate step.

nathanstitt (approved):

PR #871, opened 2026-07-10
OTTER-640: add more regression tests

therealmarv:
see [OTTER-640](https://openstax.atlassian.net/browse/OTTER-640)

this issue seems fixed and all my tests conclude it's not an issue anymore. This PR only increases a regression test for it (was not existing before).

- Add regression coverage ensuring researchers can view submitted code after an enclave run is in progress or errors before review.

[OTTER-640]: https://openstax.atlassian.net/browse/OTTER-640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx
@@ -29,20 +29,8 @@ const addJobStatus = async (studyId: string, status: StudyJobStatus) => {
         .execute()
 }
 
-const seedResultsStudy = async () => {
-    const { org, user } = await mockSessionWithTestData({ orgType: 'lab' })
-    const { study } = await insertTestStudyJobData({
-        org,
-        researcherId: user.id,
-        studyStatus: 'APPROVED',
-        jobStatus: 'CODE-SUBMITTED',
-    })
-    await addJobStatus(study.id, 'CODE-APPROVED')
-    await addJobStatus(study.id, 'FILES-APPROVED')
-    return { org, study }
-}
-
-const seedRunningOrErroredCodeStudy = async (statuses: StudyJobStatus[]) => {
+// A CODE-APPROVED study with the given trailing job statuses appended (execution/results substatuses).
+const seedCodeStudy = async (statuses: StudyJobStatus[]) => {
     const { org, user } = await mockSessionWithTestData({ orgType: 'lab' })
     const { study } = await insertTestStudyJobData({
         org,
@@ -54,6 +42,8 @@ const seedRunningOrErroredCodeStudy = async (statuses: StudyJobStatus[]) => {
     for (const status of statuses) await addJobStatus(study.id, status)
     return { org, study }
 }
+
+const seedResultsStudy = () => seedCodeStudy(['FILES-APPROVED'])
 
 describe('StudyViewCode (/view/code)', () => {
     it('shows the approved-code page with a "Proceed to step 5" forward for a results study', async () => {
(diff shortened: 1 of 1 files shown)


src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx
```diff
@@ -72,6 +85,26 @@ describe('StudyViewCode (/view/code)', () => {
         expect(page?.props.resultsHref).toBe(`/${org.slug}/study/${study.id}/view?returnTo=org`)
     })
 
+    // OTTER-640: normal /view hides submitted code while execution or an unreviewed error is presented
+    // as "Code approved". The read-only /view/code step must still expose it behind this collapsed control.
+    it.each([
+        ['the code is running in the enclave', ['JOB-RUNNING']],
+        ['the run errored before the reviewer recorded a files decision', ['JOB-RUNNING', 'JOB-ERRORED']],
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx]: Worth adding a bare `['JOB-ERRORED']` third case here. In this JOB-RUNNING → JOB-ERRORED shape, `isExecuting` actually stays true (the JOB-RUNNING row is append-only and a hidden error doesn't close the execution window per `projectStudyState`), so both cases exercise the guard through the `isExecuting` operand. A JOB-ERRORED with no prior JOB-RUNNING — the "packaging failure before any JOB-RUNNING substatus" scenario the CodeDecisionScreen comment documents — is the only shape where `isExecuting` is false and `hiddenErroredResult` alone drives the hide, so it would pin the second operand independently.

src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx
```diff
@@ -42,6 +42,19 @@ const seedResultsStudy = async () => {
     return { org, study }
 }
 
+const seedRunningOrErroredCodeStudy = async (statuses: StudyJobStatus[]) => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx]: This is `seedResultsStudy` with the trailing statuses parameterized — `seedResultsStudy` is exactly this helper called with `['FILES-APPROVED']`. We could fold them into one `seedCodeStudy(statuses)` and keep `seedResultsStudy` as a one-liner (or drop it), so the seeding shape lives in one place.

nathanstitt (approved):

src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx]: Good catch. Added a bare `['JOB-ERRORED']` third case (`d7dd7dcb`). Confirmed against `projectStudyState`: with no prior `JOB-RUNNING`/packaging status, none of `STUDY_CODE_RUNNING_JOB_STATUSES` is present, so `isExecuting` is false while `isErroredResultHiddenFromResearcher` is true. It still resolves to `code-approved`/`CodePostDecisionView` (codeDecision stays `CODE-APPROVED`), so this shape now pins the `hiddenErroredResult` operand independently, which the two `JOB-RUNNING` shapes (where `isExecuting` stays true) never exercised. Added a comment documenting the distinction.

src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/view/code/page.test.tsx]: Done (`d7dd7dcb`). Folded both into a single `seedCodeStudy(statuses)` helper (a CODE-APPROVED study with the given trailing statuses appended), and kept `seedResultsStudy` as a one-liner: `const seedResultsStudy = () => seedCodeStudy(['FILES-APPROVED'])`. The seeding shape now lives in one place.

PR #872, opened 2026-07-10
OTTER 651: Fix email leaking

therealmarv:
see [OTTER-651](https://openstax.atlassian.net/browse/OTTER-651)

- Fixed review notification emails that were leaking all recipient addresses in the "To" field, exposing PII to every recipient
- Changed both study proposal and study code submitted emails to place all recipients in Bcc, with the no-reply sender address as the sole "To" recipient
- Added a regression test that verifies no org member email appears in "To" and all appear in "Bcc" for mass notification emails

[OTTER-651]: https://openstax.atlassian.net/browse/OTTER-651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/mailer.ts
+++ b/src/server/mailer.ts
@@ -3,7 +3,7 @@ import { getStudyAndOrgDisplayInfo } from '@/server/db/queries'
 import dayjs from 'dayjs'
 import { APP_BASE_URL } from './config'
 import logger from '@/lib/logger'
-import { deliver } from './mailgun'
+import { deliver, SI_EMAIL } from './mailgun'
 
 // For local testing, send to a fixed 'to' email address that is authorized to
 // receive emails from Mailgun, and remove the 'vb -' prefix from the template name.
@@ -45,16 +45,17 @@ export const sendStudyProposalEmails = async (studyId: string) => {
     const study = await getStudyAndOrgDisplayInfo(studyId)
     const reviewers = await getOrgMembers(study.orgId)
     const emails = reviewers.map((r) => r.email).filter(Boolean)
-    const to = study.reviewerEmail ?? emails.join(', ')
 
-    if (!to) {
+    if (emails.length === 0) {
         logger.warn(`No recipients for study proposal email, studyId: ${studyId}`)
         return
     }
 
+    // All recipients go in Bcc so no one sees another's address (OTTER-651).
+    // Mailgun requires at least one "To", so we use the no-reply sender.
     await deliver({
-        to,
-        ...(emails.length > 0 && { bcc: emails.join(', ') }),
+        to: SI_EMAIL,
+        bcc: emails.join(', '),
         subject: 'New study proposal',
         template: 'vb - new research proposal',
         vars: {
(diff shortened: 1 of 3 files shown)


src/server/mailer.ts
```diff
@@ -45,16 +45,17 @@ export const sendStudyProposalEmails = async (studyId: string) => {
     const study = await getStudyAndOrgDisplayInfo(studyId)
     const reviewers = await getOrgMembers(study.orgId)
     const emails = reviewers.map((r) => r.email).filter(Boolean)
-    const to = study.reviewerEmail ?? emails.join(', ')
 
-    if (!to) {
+    if (emails.length === 0) {
```
nathanstitt, line 49: Dropping `study.reviewerEmail` here (and in `sendStudyCodeSubmittedEmail`) is a small behavior change beyond the Bcc fix: previously an assigned reviewer was addressed directly in To even if they somehow weren't in the org-members query, and the send happened as long as either existed. Now everything hinges on `getOrgMembers` — a reviewer removed from the org after assignment would silently stop being notified, and this early-return skips the send entirely even when a reviewer is assigned. That's probably the *desired* semantics (ex-members shouldn't get study emails), just confirming it's intentional rather than a side effect.

src/server/mailer.ts
```diff
@@ -45,16 +45,17 @@ export const sendStudyProposalEmails = async (studyId: string) => {
...
         logger.warn(`No recipients for study proposal email, studyId: ${studyId}`)
         return
     }
 
+    // All recipients go in Bcc so no one sees another's address (OTTER-651).
+    // Mailgun requires at least one "To", so we use the no-reply sender.
     await deliver({
-        to,
-        ...(emails.length > 0 && { bcc: emails.join(', ') }),
+        to: SI_EMAIL,
```
nathanstitt, line 57: Addressing the message To the no-reply sender works and is a common pattern, one consideration though: Mailgun will actually attempt delivery to no-reply@safeinsights.org on every notification. If that mailbox doesn't accept mail, each send generates a hard bounce and Mailgun adds the address to its bounce-suppression list — harmless to the Bcc recipients, but it becomes permanent noise in the bounce/suppression metrics. The Mailgun-canonical alternative for one-template-many-recipients is batch sending with `recipient-variables`, where each recipient sees only their own address in To. Fine to ship as-is; worth knowing the trade-off.

nathanstitt (commented):
lets ask Iris about the delivery to no-reply before we merge this

src/server/mailer.ts
therealmarv: Yes, intentional. Collapsing both mass-notification sends to a single recipient source (`getOrgMembers`) is deliberate: for a PII fix, org membership should be the sole source of truth for who gets these emails. An assigned reviewer who was removed from the org no longer being notified is the desired behavior, not a side effect, since ex-members shouldn't keep receiving study emails, and the empty-members early-return follows the same reasoning.

nathanstitt (approved):

PR #873, opened 2026-07-10
OTTER-644: fix ai summary loading state

therealmarv:
see [OTTER-644](https://openstax.atlassian.net/browse/OTTER-644)

- Fixed the AI summary panel showing an error immediately on complex resubmissions by anchoring the generation timeout to the latest code-submitted event instead of the original job creation date
- Added a reset on retry so the timeout backstop restarts its clock when a reviewer retries a failed AI summary, showing the spinner instead of a stale error
- Fixed the "Submitted on / Resubmitted on" header timestamp on the code review page to also use the latest submission event, so reused jobs show the correct date
- Extracted the submission-timestamp lookup into a shared helper used by both the AI summary panel and the code review header
- Added unit tests covering the resubmission loading state, the retry-reset behavior, and the shared helper's edge cases

[OTTER-644]: https://openstax.atlassian.net/browse/OTTER-644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
@@ -113,20 +113,22 @@ const AI_SUMMARY_TIMEOUT_MS = 180_000
 // generation that hangs without writing a (success or failure) row eventually
 // surfaces as an error instead of spinning forever. `since` may be a string —
 // timestamps serialize to ISO strings across the server/client boundary.
-function useElapsedSince(since: Date | string, ms: number): boolean {
-    const sinceMs = new Date(since).getTime()
-    const [elapsed, setElapsed] = useState(() => Date.now() - sinceMs >= ms)
+function useElapsedSince(since: Date | string, ms: number) {
+    const initialSinceMs = new Date(since).getTime()
+    const [startedAt, setStartedAt] = useState(initialSinceMs)
+    const [elapsed, setElapsed] = useState(() => Date.now() - initialSinceMs >= ms)
     useEffect(() => {
-        const remaining = Math.max(0, ms - (Date.now() - sinceMs))
-        if (remaining === 0) {
-            setElapsed(true)
-            return
-        }
-        setElapsed(false)
+        const remaining = Math.max(0, ms - (Date.now() - startedAt))
         const id = setTimeout(() => setElapsed(true), remaining)
         return () => clearTimeout(id)
-    }, [sinceMs, ms])
-    return elapsed
+    }, [startedAt, ms])
+    return {
+        elapsed,
+        reset: () => {
+            setStartedAt(Date.now())
+            setElapsed(false)
+        },
+    }
 }
 
 // The review row is written by a deferred background task triggered at code
@@ -257,9 +259,9 @@ export function AiSummaryCollapsible({
     const { data: review, error } = useStudyReviewPoll(studyJobId, initialReview)
     // A successful retry is a new generation request, so it needs its own
     // timeout window instead of inheriting the original submission's age.
-    const [generationStartedAt, setGenerationStartedAt] = useState<Date | string>(submittedAt)
-    const retry = useRetryStudyReview(studyJobId, () => setGenerationStartedAt(new Date()))
-    const timedOut = useElapsedSince(generationStartedAt, timeoutMs)
+    const timeout = useElapsedSince(submittedAt, timeoutMs)
+    const retry = useRetryStudyReview(studyJobId, timeout.reset)
+    const timedOut = timeout.elapsed
     const summary = review?.report?.codeExplanation ?? null
 
     const onRetry = () => retry.mutate()
--- a/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
@@ -117,13 +117,15 @@ function useElapsedSince(since: Date | string, ms: number): boolean {
     const sinceMs = new Date(since).getTime()
     const [elapsed, setElapsed] = useState(() => Date.now() - sinceMs >= ms)
     useEffect(() => {
-        if (elapsed) return
-        // Clamp to 0 rather than setting state synchronously here; the timer
-        // fires the update on the next tick, out of the effect body.
         const remaining = Math.max(0, ms - (Date.now() - sinceMs))
+        if (remaining === 0) {
+            setElapsed(true)
+            return
+        }
+        setElapsed(false)
         const id = setTimeout(() => setElapsed(true), remaining)
         return () => clearTimeout(id)
-    }, [sinceMs, ms, elapsed])
+    }, [sinceMs, ms])
     return elapsed
 }
 
@@ -223,12 +225,13 @@ function AiSummaryContent({ summary, isExpanded, onToggle }: AiSummaryContentPro
 
 // Clears the failed row server-side, re-fires generation, then resets the
 // cached review to null so the poll resumes and the UI drops back to pending.
-function useRetryStudyReview(studyJobId: string) {
+function useRetryStudyReview(studyJobId: string, onRetryStarted: () => void) {
     const queryClient = useQueryClient()
     return useMutation({
         mutationFn: () => regenerateStudyReviewAction({ studyJobId }),
         onSuccess: () => {
             queryClient.setQueryData(['study-review', studyJobId], null)
+            onRetryStarted()
         },
     })
 }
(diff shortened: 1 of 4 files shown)


src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
```diff
@@ -106,12 +106,20 @@ function SecurityScanLog({ scan }: { scan: JobScanResult }) {
...
     review: StudyReviewWithMeta | null
     scan: JobScanResult
     codeInitiallyExpanded?: boolean
 }
 
+// A complex resubmission reuses its study job, so createdAt can predate the
+// generation request by days. The latest CODE-SUBMITTED event is the only
+// timestamp that accurately anchors the summary-generation timeout (and the
+// "Submitted/Resubmitted on" header label).
+export function latestCodeSubmittedAt(job: Pick<LatestJobForStudy, 'createdAt' | 'statusChanges'>): Date | string {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx:121]: `.find` returning the *latest* CODE-SUBMITTED depends on `statusChanges` arriving newest-first — which holds for `LatestJobForStudy` (its subquery orders `createdAt desc, id desc`), but the helper is exported and nothing in the signature enforces that ordering. Worth one line in the doc comment stating the newest-first assumption, since other status-change arrays in the codebase don't necessarily share the guarantee and a differently-ordered caller would silently get the *oldest* submission back.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx
```diff
@@ -246,6 +247,69 @@ describe('SubmittedCodeSection — AI summary', () => {
...
+            { status: 'CODE-CHANGES-REQUESTED', createdAt: new Date(Date.now() - 2 * day) },
+            { status: 'CODE-SUBMITTED', createdAt: new Date() },
+        ]
+        await db
+            .insertInto('jobStatusChange')
+            .values(statuses.map((change) => ({ ...change, studyJobId: resubmissionFixture.job.id })))
+            .execute()
+
+        const job = await latestJobForStudy(resubmissionFixture.study.id)
+        expect(latestCodeSubmittedAt(job)).not.toBe(originalSubmittedAt)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx]: `.not.toBe` on two Date objects is an identity comparison, and these can never be the same object (one was constructed locally, the other came back through the query's JSON serialization) — so this assertion passes no matter what the helper returns. Comparing values pins it properly, e.g. `expect(new Date(latestCodeSubmittedAt(job)).getTime()).toBe(statuses[3].createdAt.getTime())`. Relatedly, the comment above says losing the `id desc` tiebreak would surface here, but every seeded row has a distinct `createdAt`, so only losing the `createdAt` ordering would — exercising the tiebreak needs two rows sharing a timestamp.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx
```diff
@@ -254,7 +318,7 @@ describe('SubmittedCodeSection — AI summary', () => {
             <AiSummaryCollapsible
                 studyJobId={failedFixture.job.id}
                 initialReview={initialReview}
-                submittedAt={failedFixture.job.createdAt}
+                submittedAt={new Date(Date.now() - 10 * 60_000)}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx]: Making `submittedAt` 10 minutes old here means the backstop has already elapsed at mount, so this test now passes through *either* error branch — if the `summaryFailedAt` path broke, `timedOut` alone would produce the same error state and the test would stay green. The previous fresh `job.createdAt` was load-bearing for this particular test: it guaranteed the error could only come from the failed row. (The 10-minute-old date is exactly right in the retry test below, where an already-elapsed backstop is what makes the reset observable — just not here.)

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
```diff
@@ -113,18 +113,22 @@ const AI_SUMMARY_TIMEOUT_MS = 180_000
 // generation that hangs without writing a (success or failure) row eventually
 // surfaces as an error instead of spinning forever. `since` may be a string —
 // timestamps serialize to ISO strings across the server/client boundary.
-function useElapsedSince(since: Date | string, ms: number): boolean {
-    const sinceMs = new Date(since).getTime()
-    const [elapsed, setElapsed] = useState(() => Date.now() - sinceMs >= ms)
+function useElapsedSince(since: Date | string, ms: number) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx:121]: The docstring above is now stale on two counts: the hook returns `{ elapsed, reset }` rather than a boolean, and `since` has become mount-only — with `startedAt` held in state and `elapsed` no longer un-settable from props, a changed `since` prop is ignored entirely (the old version at least re-armed the timer via the `sinceMs` dep). That is fine for the current caller, where a new submission arrives via a fresh server render, but the signature reads as if the hook tracks `since` reactively. Worth updating the comment to say the prop is read once on mount and later changes only come through `reset()`, so a future caller is not surprised.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
```diff
@@ -106,12 +106,20 @@ function SecurityScanLog({ scan }: { scan: JobScanResult }) {
...
+    return job.statusChanges.find((change) => change.status === 'CODE-SUBMITTED')?.createdAt ?? job.createdAt
+}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx:127]: this doesn't order, which means if the caller sends things that aren't already sorted it will fail badly.  suggest we:

```suggestion
  export function latestCodeSubmittedAt(job: Pick<LatestJobForStudy, 'createdAt' | 'statusChanges'>): Date | string {
      const submissions = job.statusChanges.filter((change) => change.status === 'CODE-SUBMITTED')
      if (submissions.length === 0) return job.createdAt
      return submissions.reduce((latest, change) =>
          new Date(change.createdAt).getTime() > new Date(latest.createdAt).getTime() ? change : latest,
      ).createdAt
  }

```

nathanstitt (approved):
looks good but probably should update `latestCodeSubmittedAt`

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx:127]: Done in 0493bae9. Switched `latestCodeSubmittedAt` to scan for the max `createdAt` (via `reduce`) instead of `.find`, so it no longer depends on the array being sorted. Kept the early return to `job.createdAt` when there are no CODE-SUBMITTED events, and added a helper test that feeds a non-desc array to lock in the order-independence.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx:121]: Resolved by making the helper order-independent (0493bae9) rather than documenting the newest-first assumption: it now scans all CODE-SUBMITTED events for the max `createdAt`, so a differently-ordered caller gets the newest submission regardless. Updated the doc comment to state that.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx]: Fixed in 0493bae9. Replaced the always-passing identity check with a value comparison: `expect(new Date(latestCodeSubmittedAt(job)).getTime()).toBe(statuses[3].createdAt.getTime())`. Good catch on the tiebreak point too. since every seeded row has a distinct `createdAt`, this case only exercises the `createdAt` ordering, so I reworded the comment above to claim only that (not the `id desc` tiebreak).

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx]: Fixed in 0493bae9. Anchored `submittedAt` to `new Date()` here so the backstop has not elapsed at mount, meaning the error can only come from the persisted failure row (not `timedOut`). Left the 10-minute-old date in the retry test below, where an already-elapsed backstop is exactly what makes the reset observable.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx:121]: Updated the docstring in 0493bae9: it now says the hook returns `{ elapsed, reset }`, that `since` is read once on mount to seed the start time, and that later changes must come through a fresh server render or an explicit `reset()` rather than a changed `since` prop.

PR #875, opened 2026-07-10
OTTER-649: rework security scan log display (per-tool status + download)

therealmarv:
see [OTTER-649](https://openstax.atlassian.net/browse/OTTER-649)

## What
Reworks the &#34;Security scan log&#34; panel on the reviewer Code Review page so it no longer asserts an overall pass/fail with a check/X icon in the title. Instead it shows two independent, programmatically derived statuses and a download link for the plaintext log, per the Figma designs.

- Title &#34;Security scan log&#34; with no status icon.
- Row &#34;Trivy Filesystem Scan:&#34; `No vulnerabilities found` when clean, otherwise a red warning icon + `Vulnerabilities found`.
- Row &#34;SonarQube Quality Gate:&#34; `Passed` when the gate is OK, otherwise a red warning icon + `Needs review`.
- A `Download` link, shown whenever a plaintext scan log file exists.

## Why
Per the card, we cannot say with confidence that a scan &#34;failed&#34;, only whether it passed or needs human review. The old UI derived a single status by reading the encrypted zip and checking for the substring `OK`, which is unreliable. Scan log text files are effectively always present and reviewers need to download them (ZIPs are not offered).

## How
- Status is now parsed from the plaintext `SECURITY-SCAN-LOG` text file, not the encrypted zip. Safety bias: only report PASSED on an explicit clean signal, otherwise flag for review.
  - Trivy PASSED only on the scanner&#39;s `no vulnerabilities found` line; anything else (findings, `no results`, unrecognized) is flagged.
  - SonarQube PASSED only when the quality gate reports `OK`; any other status (ERROR, WARN, NONE, TIMEOUT, UNKNOWN) or a missing SonarQube section needs review.
- New authenticated download route `GET /dl/scan-log/[jobId]` serves the plaintext `.txt`. It authorizes the job (`canViewStudyJob`) before looking up the file, so a denied requester cannot infer from a 404-vs-401 whether another org&#39;s job produced a scan log; an unknown job and an unauthorized one both return 401. The lookup returns the newest `SECURITY-SCAN-LOG` row (`getStudyJobFileOfType` orders `createdAt` desc), so on a resubmission the download matches the log the displayed statuses were parsed from. The response forces `Content-Disposition: attachment; filename="security-scan-log.txt"` so the download survives the cross-origin redirect to S3 regardless of the stored object&#39;s content type. The encrypted zip is intentionally not downloadable here.
- `JobScanResult` now carries independent `trivy` / `sonarqube` statuses plus the log file; consumers (`code-review`, `post-feedback-view`, `reviewer-code-feedback-screen`) pass it through unchanged.

The plaintext scan log format is produced by the CodeBuild scanner in the `iac` repo (`codebuild/scripts/common.ts`), which the parser targets.

## The two static rows are always shown
The two labeled rows (&#34;Trivy Filesystem Scan:&#34; and &#34;SonarQube Quality Gate:&#34;) are always rendered, matching the acceptance criteria that lists them as static elements. Their statuses are derived from the log, so when no log has been read yet each row shows a neutral &#34;Scan in progress...&#34; value (no icon, and never a fabricated pass/fail). The Download link is shown whenever a scan log file exists, independently of the parsed statuses: if a log row exists but its contents can&#39;t be read (a transient S3 read/parse failure), the statuses stay unknown (&#34;Scan in progress...&#34;) but the log remains downloadable, since the download route only needs the DB row and a signed URL. In the normal reviewer flow the log is present and readable by the time the page is viewed, so the derived statuses are shown.

## Tests
- Component: title without icon, both static labels, per-tool pass/fail text and warning icon, download link + href, and the pending (no-log) state (rows present, no icon, no download).
- `jobScanResultForJob` and the pure parsers, including the full SonarQube status vocabulary and case-insensitivity, plus an S3-backed integration test, and the unreadable-log case (log row present, contents unreadable: null statuses but still downloadable).
- Download route: `404` (authorized but no log), `401` (different-org requester), `401` (unknown job, so existence is not disclosed), and the authorized redirect to a signed URL.

Verified visually against the Figma frames across all states (all passed, Trivy-only, SonarQube-only, both failed, and the pending state).

[OTTER-649]: https://openstax.atlassian.net/browse/OTTER-649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/dl/scan-log/[jobId]/route.ts
+++ b/src/app/dl/scan-log/[jobId]/route.ts
@@ -12,16 +12,19 @@ export const GET = async (_: Request, { params }: { params: Promise<{ jobId: str
         return NextResponse.json({ error: 'no job id provided' }, { status: 400 })
     }
 
+    // Authorize before touching the file, so a denied requester can't infer from a
+    // 404-vs-401 whether another org's job produced a scan log. An unknown job and an
+    // unauthorized one both return 401 rather than disclosing which case applied.
+    const job = await jobInfoForJobId(jobId).catch(() => null)
+
+    if (!job || !(await canViewStudyJob(job))) {
+        return NextResponse.json({ error: 'permission denied' }, { status: 401 })
+    }
+
     const file = await getStudyJobFileOfType(jobId, 'SECURITY-SCAN-LOG', false)
 
     if (!file) {
         return NextResponse.json({ error: 'scan log not found' }, { status: 404 })
-    }
-
-    const job = await jobInfoForJobId(jobId)
-
-    if (!(await canViewStudyJob(job))) {
-        return NextResponse.json({ error: 'permission denied' }, { status: 401 })
     }
 
     const url = await urlForFile(file.path)
--- a/src/app/dl/scan-log/[jobId]/route.ts
+++ b/src/app/dl/scan-log/[jobId]/route.ts
@@ -27,6 +27,11 @@ export const GET = async (_: Request, { params }: { params: Promise<{ jobId: str
         return NextResponse.json({ error: 'scan log not found' }, { status: 404 })
     }
 
-    const url = await urlForFile(file.path)
+    // Pin the download: the anchor's `download` attribute doesn't survive the cross-origin
+    // redirect to S3, so force a Content-Disposition rather than relying on the stored
+    // object's content type (uploads set none today, but that could change).
+    const url = await urlForFile(file.path, {
+        ResponseContentDisposition: 'attachment; filename="security-scan-log.txt"',
+    })
     return NextResponse.redirect(url)
 }
--- a/src/app/dl/scan-log/[jobId]/route.ts
+++ b/src/app/dl/scan-log/[jobId]/route.ts
@@ -1,0 +1,29 @@
+import { NextResponse } from 'next/server'
+import { urlForFile } from '@/server/storage'
+import { getStudyJobFileOfType, jobInfoForJobId } from '@/server/db/queries'
+import { canViewStudyJob } from '@/server/auth'
+
+// Serves the plaintext SECURITY-SCAN-LOG .txt for a job. The encrypted zip is
+// intentionally not downloadable here (OTTER-649: ZIPs are not offered).
+export const GET = async (_: Request, { params }: { params: Promise<{ jobId: string }> }) => {
+    const { jobId } = await params
+
+    if (!jobId) {
+        return NextResponse.json({ error: 'no job id provided' }, { status: 400 })
+    }
+
+    const file = await getStudyJobFileOfType(jobId, 'SECURITY-SCAN-LOG', false)
+
+    if (!file) {
+        return NextResponse.json({ error: 'scan log not found' }, { status: 404 })
+    }
+
+    const job = await jobInfoForJobId(jobId)
+
+    if (!(await canViewStudyJob(job))) {
+        return NextResponse.json({ error: 'permission denied' }, { status: 401 })
+    }
+
+    const url = await urlForFile(file.path)
+    return NextResponse.redirect(url)
+}
--- a/src/server/db/queries.ts
+++ b/src/server/db/queries.ts
@@ -530,7 +530,7 @@ export function parseSonarqubeStatus(log: string): ScanToolStatus {
 
 // Reads the plaintext SECURITY-SCAN-LOG (not the encrypted zip) and derives a
 // per-tool status from its text. No log row yet, or an unreadable file, is
-// treated as "not reported" — no statuses and no download affordance.
+// treated as "not reported" (no statuses and no download affordance).
 export async function jobScanResultForJob(studyJobId: string): Promise<JobScanResult> {
     const logFile = await Action.db
         .selectFrom('studyJobFile')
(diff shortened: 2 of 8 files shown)


src/server/db/queries.ts
```diff
@@ -502,34 +502,53 @@ export type StudyReviewWithMeta = {
...
+    if (!logFile) return { trivy: null, sonarqube: null, logFile: null }
 
     try {
         const blob = await fetchFileContents(logFile.path)
         const contents = await blob.text()
-        return { status: contents.includes('OK') ? 'PASSED' : 'FAILED', logFile }
+        return { trivy: parseTrivyStatus(contents), sonarqube: parseSonarqubeStatus(contents), logFile }
     } catch {
-        return { status: 'IN-PROGRESS', logFile }
+        return { trivy: null, sonarqube: null, logFile: null }
```
nathanstitt: The old code kept `logFile` in this catch; now a transient S3 read failure both hides the Download link and shows "Scan in progress…" for a scan that actually finished. The download route doesn't depend on `fetchFileContents` succeeding — it only needs the DB row and a signed URL — so the link would still work. Worth considering returning `logFile` here with null statuses, so reviewers can still grab the log even when we couldn't parse it?

src/app/dl/scan-log/[jobId]/route.ts
```diff
@@ -0,0 +1,32 @@
...
+    // Authorize before touching the file, so a denied requester can't infer from a
+    // 404-vs-401 whether another org's job produced a scan log. An unknown job and an
+    // unauthorized one both return 401 rather than disclosing which case applied.
+    const job = await jobInfoForJobId(jobId).catch(() => null)
+
+    if (!job || !(await canViewStudyJob(job))) {
+        return NextResponse.json({ error: 'permission denied' }, { status: 401 })
+    }
+
+    const file = await getStudyJobFileOfType(jobId, 'SECURITY-SCAN-LOG', false)
```
Comment (nathanstitt) [inline src/app/dl/scan-log/[jobId]/route.ts:24]: `getStudyJobFileOfType` has no ORDER BY, while `jobScanResultForJob` deliberately picks the newest log (`orderBy('createdAt', 'desc').limit(1)`). If a job can end up with more than one SECURITY-SCAN-LOG row (the ordering over there suggests it can), this download could serve a different log than the one the displayed statuses were parsed from. Might be worth using the same newest-first lookup here.

src/app/dl/scan-log/[jobId]/route.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (!file) {
+        return NextResponse.json({ error: 'scan log not found' }, { status: 404 })
+    }
+
+    const url = await urlForFile(file.path)
```
Comment (nathanstitt) [inline src/app/dl/scan-log/[jobId]/route.ts]: The anchor's `download` attribute won't survive the cross-origin redirect to S3, so whether this actually downloads depends on what S3 sends back. Today uploads don't set a content type, so S3 serves octet-stream and the browser happens to download it — but if the stored content type ever became `text/plain`, the log would open in-tab instead. `signedUrlForFile` already accepts a `ResponseContentDisposition` override; passing `attachment; filename="security-scan-log.txt"` here would pin the Download behavior regardless.

src/app/dl/scan-log/[jobId]/route.ts
Comment (nathanstitt) [reply src/app/dl/scan-log/[jobId]/route.ts:24]: this will def happen on the case where changes are requested and user uploads new code.  we should add the orderBy

nathanstitt (approved):
looks good, but probably do need to order the a scan query

src/server/db/queries.ts
therealmarv: Good catch, fixed in 24e7b74b. The catch in `jobScanResultForJob` now returns the `logFile` row with null statuses, so a transient S3 read/parse failure no longer hides the Download link or fakes a "Scan in progress..." state for a scan that actually finished. The download route serves the file from the DB row + a signed URL and doesn't depend on this fetch. Added a unit test that inserts a log row without a stored object (so the fetch throws) and asserts the log stays downloadable with null statuses.

src/app/dl/scan-log/[jobId]/route.ts
Comment (therealmarv) [reply src/app/dl/scan-log/[jobId]/route.ts:24]: Agreed, fixed in 24e7b74b. Added `.orderBy('studyJobFile.createdAt', 'desc')` to `getStudyJobFileOfType` so it deterministically returns the newest row, matching `jobScanResultForJob`'s newest-first lookup. This keeps the downloaded log consistent with the parsed/displayed statuses on resubmission, and also makes the other callers (APPROVED-RESULT, MAIN-CODE) deterministic when a job has multiple rows of a type.

src/app/dl/scan-log/[jobId]/route.ts
Comment (therealmarv) [reply src/app/dl/scan-log/[jobId]/route.ts]: Fixed in 24e7b74b. The route now passes `ResponseContentDisposition: 'attachment; filename="security-scan-log.txt"'` through `urlForFile` → `signedUrlForFile`, so the download is pinned regardless of the stored object's content type and doesn't rely on the anchor's `download` attribute surviving the cross-origin redirect to S3.

PR #876, opened 2026-07-10
OTTER-642: make study job file storage idempotent to prevent duplicate log rows

therealmarv:

## Problem

When a study errors out (e.g. SQL failures), each log file (`security-scan-log.txt`, `logs.json`) is displayed twice in the &#34;Study Status&#34; panel, for both the reviewer (data provider) and the researcher. After release the two copies can even show inconsistent state, one approved and one still locked.

## Root cause

Each duplicated file is a genuine duplicate `study_job_file` row. The three ingest endpoints write a file row on every delivery but only deduplicate the status change, never the file:

- `results/route.ts` (Code Run Log): its idempotency guard rejects a re-post only when a `RUN-COMPLETE` status already exists. An errored job is recorded as `JOB-ERRORED`, never `RUN-COMPLETE`, so the guard never trips and a retried error post stores a second `ENCRYPTED-CODE-RUN-LOG` and appends another `JOB-ERRORED`. This is why the bug is specific to errored / SQL-failure jobs (successful jobs are protected by the `RUN-COMPLETE` guard).
- `job-scan-results/route.ts` (Security Scan Log) and `containerizer/route.ts` (Packaging Error Log): store the log unconditionally; only the status insert is deduped.

All three funnel through `storeJobFile` in `src/server/storage.ts`, which does a plain insert with a fresh v7 UUID while the S3 `path` is deterministic. So a repeat call overwrites the same S3 object but inserts a second DB row (same name/path, new id). Nothing downstream deduplicates: no unique constraint, no `DISTINCT` in the queries, no dedup in the UI hook. The re-delivery itself comes from the enclave client retrying, but the defect that turns a retry into a visible duplicate is in this codebase.

## Fix

Enforce the invariant &#34;one `study_job_file` row per `(study_job_id, path)`&#34; at the write layer.

1. **Migration** `1780400000000_dedupe_and_unique_study_job_file_path.ts`: collapse any pre-existing duplicates, then add a unique index on `(study_job_id, path)`. A unique index cannot be created while duplicates exist, so the dedupe is a required prerequisite. Recipient keys (researcher access to approved results) can be attached to any of the duplicate rows, and different approval cycles can key different duplicates, so the dedupe preserves access in two steps: pick a survivor per `(study_job_id, path)` (most recipient keys, then earliest `created_at`/`id`), copy every key from the other duplicates onto it (`INSERT ... SELECT ... ON CONFLICT (study_job_file_id, file_path, fingerprint) DO NOTHING`, so an already-granted inner file + recipient collapses to one), then delete the duplicates so their now-redundant originals are removed by the `ON DELETE CASCADE`. No key is lost regardless of which duplicate an approval happened to reference.
2. **`storeJobFile` idempotent upsert**: `onConflict((oc) =&gt; oc.columns([&#39;studyJobId&#39;, &#39;path&#39;]).doUpdateSet({ name, fileType, sourceId }))`. A retried delivery updates the existing row in place instead of inserting a duplicate. `sourceId` is refreshed alongside `name`/`fileType` so a future caller that sets it is not surprised by a stale value on a conflicting write. Centralized, so all three routes are fixed at once.
3. **Finalize the run exactly once in `results/route.ts`**: an errored job never reaches `RUN-COMPLETE`, so the run phase is treated as finalized when results were stored (`RUN-COMPLETE`) or a run-error log was stored. The error case is keyed on the `ENCRYPTED-CODE-RUN-LOG` file, which only this route writes, rather than on a bare `JOB-ERRORED` status: the scan and packaging steps also record `JOB-ERRORED` (with their own log types), so keying on the status would let one of those wrongly block a later legitimate results delivery for the same job. The finalize check + status insert run inside a `SELECT ... FOR UPDATE` transaction on the job row, so two deliveries arriving at the same instant are serialized: only the first writer stores the file, records the status, and sends the &#34;results ready&#34; email; a loser is rejected with 422. The file write runs on the transaction&#39;s own connection (a threaded executor), so the nested insert cannot deadlock against the lock the transaction holds.
4. **Serialize submits and dedupe colliding code-file paths in `attachCodeToRoundJob` (`study-request.ts`)**: the new unique index also constrains the `MAIN-CODE` / `SUPPLEMENTAL-CODE` rows written at submission time. Two uploaded filenames can resolve to the same storage path (names that sanitize identically, or a supplemental matching the main file), which under the index would fail the insert and 500 the submit, so the action collapses these to one row per path, reserving the `MAIN-CODE` row first so the entry-point file keeps its type. Concurrent submits for the same study (double-click, two tabs) are serialized by a `SELECT ... FOR UPDATE` on the parent `study` row taken before the round job is chosen: locking the study rather than the job also covers the first-submit case where no job exists yet (two transactions would otherwise each create a separate job), and makes the delete-then-insert atomic per round so two submits cannot interleave into a mixed file set. The `(study_job_id, path)` upsert is kept as defense-in-depth.

## Tests

- `results/route.test.ts`: a second log-only (errored) upload is rejected with 422 and leaves exactly one `ENCRYPTED-CODE-RUN-LOG` row and one `JOB-ERRORED` status; a prior scan/packaging `JOB-ERRORED` does not block a later results delivery; two concurrent errored uploads yield exactly one `JOB-ERRORED` row.
- `job-scan-results/route.test.ts`: delivering `CODE-SCANNED` twice yields exactly one `ENCRYPTED-SECURITY-SCAN-LOG` and one `SECURITY-SCAN-LOG` row (proves the `storeJobFile` upsert dedups).
- `study-request.test.ts`: code files that resolve to the same storage path collapse to a single `SUPPLEMENTAL-CODE` row with the `MAIN-CODE` row preserved; two concurrent submits with different file sets resolve to exactly one submit&#39;s files (never a mix); and two racing first submits create exactly one round job.
- `migrations.test.ts`: recipient keys spread across duplicate `study_job_file` rows (including a shared grant) are all preserved on the surviving row after the dedupe migration.

## Notes

- No read-side / UI dedup; the invariant is enforced at write time.
- `path` (not `file_type`) is the uniqueness key. Legitimate multi-file cases (multiple supplemental-code files) are unaffected; files that would collapse to the same path are deduped at their write site rather than rejected.
- Existing duplicate rows in an environment are collapsed by the migration; going forward no new duplicates are created.


src/database/migrations/1780300000000_dedupe_and_unique_study_job_file_path.ts
```diff
@@ -0,0 +1,34 @@
...
+    // the invariant, keeping the earliest row per (study_job_id, path). The earliest row is the one
+    // that holds recipient keys after an approval (a post-approval duplicate is always the later,
+    // key-less row), and study_job_file_recipient_key cascades on delete, so access is preserved and
+    // orphaned keys are removed with the duplicate.
+    await sql`
+        DELETE FROM study_job_file
+        WHERE id IN (
+            SELECT id FROM (
+                SELECT id, ROW_NUMBER() OVER (
+                    PARTITION BY study_job_id, path ORDER BY created_at ASC, id ASC
```
nathanstitt: The keep-earliest rule leans on "the earliest row is the one that holds recipient keys", but keys attach to whichever row id the approval client happened to reference — `buildSharedFiles` sends `file.sourceId`, and row lookups like `getStudyJobFileOfType` have no ORDER BY, so which duplicate the client saw isn't pinned down. If duplicates existed *before* an approval (e.g. the concurrent-delivery race this PR also fixes), the keyed row could be the later one, and this DELETE would cascade its keys away. Ordering the window on "has recipient keys" first would make the dedupe safe regardless of which assumption holds, e.g. `ORDER BY (SELECT count(*) FROM study_job_file_recipient_key k WHERE k.study_job_file_id = study_job_file.id) DESC, created_at ASC, id ASC`.

src/app/api/job/[jobId]/results/route.ts
```diff
@@ -44,21 +36,51 @@ export const POST = wrapApiOrgAction(async (req: Request, { params }: { params:
...
+    })
 
-    await db
-        .insertInto('jobStatusChange')
-        .values({
-            status: logs && !results ? 'JOB-ERRORED' : 'RUN-COMPLETE', // TODO: verify this is correct status,
-            studyJobId: info.studyJobId,
-        })
-        .execute()
+    if (!finalized) return new NextResponse('job already finalized', { status: 422 })
```
Comment (nathanstitt) [inline src/app/api/job/[jobId]/results/route.ts:83]: A retry caused by a lost 200 (enclave stored everything, response dropped on the network) now surfaces to the client as a 422 for the error-log case too. That matches the pre-existing contract for RUN-COMPLETE replays, so no objection — but do we know how the enclave client treats a 422 on retry? If it counts non-2xx as delivery failure it could keep retrying or alert on a delivery that actually landed. Since the endpoint is otherwise idempotent now, answering a replay with 200 would also be defensible.

src/server/storage.ts
```diff
@@ -25,20 +26,47 @@ export async function urlForStudyDocumentFile(info: MinimalStudyInfo, fileType:
...
     await storeS3File(info, file.stream(), path)
 
-    return await db
+    // Idempotent on (study_job_id, path): a retried/re-delivered ingest webhook overwrites the same
+    // S3 object, so it must update the existing row in place rather than insert a duplicate that
+    // shows up as a doubled log/result in the reviewer and researcher views (OTTER-642).
+    return await executor
         .insertInto('studyJobFile')
         .values({ path, name: file.name, studyJobId: info.studyJobId, fileType, sourceId })
+        .onConflict((oc) => oc.columns(['studyJobId', 'path']).doUpdateSet({ name: file.name, fileType }))
```
nathanstitt: `doUpdateSet` refreshes `name` and `fileType` but not `sourceId`. No production caller passes `sourceId` today, so this is dormant — but if one ever does and the row already exists, the upsert would silently keep the old row's (null) `sourceId`. Cheap to include it now and spare someone the surprise later.

src/server/actions/study-request.ts
```diff
@@ -96,28 +96,26 @@ async function attachCodeToRoundJob(
...
-            })
-            .executeTakeFirstOrThrow()
+        const filePath = pathForStudyJobCodeFile({ orgSlug, studyId, studyJobId }, fileName)
+        if (filesByPath.has(filePath)) continue
+        filesByPath.set(filePath, { name: fileName, path: filePath, studyJobId, fileType: 'SUPPLEMENTAL-CODE' })
     }
 
+    await db
+        .insertInto('studyJobFile')
+        .values([...filesByPath.values()])
```
nathanstitt, line 124: Two concurrent submits for the same study (double-click, two tabs) both reuse the same round job, and the delete-then-insert isn't serialized — the statements can interleave as A-delete, B-delete, A-insert, B-insert, at which point B's batch hits the new unique index and the submit 500s where it previously produced silent duplicates. An `onConflict((oc) => oc.columns(['studyJobId', 'path']).doUpdateSet(...))` here, mirroring `storeJobFile`, would keep that race harmless.

nathanstitt (approved):
seems correct, but does make a little nervous to overwrite previous files just because the name is the same.  However I guess we were already doing so because we were overwriting them on s3 regardless.

src/database/migrations/1780300000000_dedupe_and_unique_study_job_file_path.ts
therealmarv: Updated in the latest push (`edfd2604`). The final version goes further than rank-and-keep: it first copies every recipient key from the non-survivor duplicates onto the survivor (`INSERT ... SELECT ... ON CONFLICT (study_job_file_id, file_path, fingerprint) DO NOTHING`), then deletes the duplicates so their originals cascade away. A key held only on a later duplicate is therefore preserved rather than lost, and duplicate grants for the same inner file + recipient collapse to one. The survivor is the max-keys row (then earliest `created_at`/`id`); since step 1 only ever adds keys to it, it stays the survivor for the step-2 delete. New `migrations.test.ts` case spreads keys across two duplicate rows and asserts the survivor ends up with all distinct keys. https://github.com/safeinsights/management-app/blob/edfd260452307a860b95474d660f95493b799bae/src/database/migrations/1780300000000_dedupe_and_unique_study_job_file_path.ts#L3-L55

src/app/api/job/[jobId]/results/route.ts
Comment (therealmarv) [reply src/app/api/job/[jobId]/results/route.ts:83]: Leaving this as a 422 for now. It keeps the error-log replay consistent with the existing `RUN-COMPLETE` replay contract (both go through the same `!finalized` branch), and switching a replay to 200 is a cross-service contract decision that should be made deliberately with knowledge of how the enclave client actually treats a 422 on retry, rather than changed here as a side effect. Agreed that, since the endpoint is now idempotent, answering a replay with 200 is defensible and would be the natural follow-up if we confirm the client counts 422 as a delivery failure. I can file that as a separate ticket if you'd like it tracked.

src/server/storage.ts
therealmarv: Done, `sourceId` is now included in the `doUpdateSet` so a conflicting write refreshes it alongside `name`/`fileType` instead of silently keeping the old row's value. https://github.com/safeinsights/management-app/blob/edfd260452307a860b95474d660f95493b799bae/src/server/storage.ts#L45-L52

src/server/actions/study-request.ts
therealmarv: Update (latest push `edfd2604`): the real serialization is now a `SELECT ... FOR UPDATE` on the parent `study` row, taken before the round job is chosen, rather than relying on the upsert. Locking the job *after* selection was insufficient for the first-submit case (no job exists yet, so two transactions could each create and lock a *different* job); locking the study serializes both round creation and the delete-then-insert, so two concurrent submits can neither interleave into a mixed file set nor 500 on the unique index. The `(studyJobId, path)` upsert is kept as defense-in-depth. Covered by new tests `serializes concurrent submissions with different file sets` and `creates one round job when first submissions race`. Lock: https://github.com/safeinsights/management-app/blob/edfd260452307a860b95474d660f95493b799bae/src/server/actions/study-request.ts#L85-L90

therealmarv:
After deep(er) analysis of this PR, I also do not feel comfortable merging it. Fixing the issue at the database level feels technically straightforward, but the risk analysis does not support it yet.

**Technical risks**

- The database rule treats every file with the same study job and storage path as a duplicate. It cannot distinguish a repeated delivery from a genuinely newer file.
- The migration permanently deletes existing rows. Its rollback only removes the new index and cannot restore deleted data.
- Existing rows may have source-file references or researcher decryption keys attached to different copies. Choosing one row and deleting the others can preserve the wrong relationships.
- The database transaction does not include S3. The file in S3 can be replaced while the database keeps decryption keys created for the previous file, potentially making a released file impossible to decrypt.
- A database uniqueness rule prevents duplicate rows, but it does not prevent duplicate encryption, S3 uploads, status updates, notifications, or other side effects.

**PM and product risks which are unclear from the card:**

- OTTER-642 says scan logs, runtime-error logs, and output files should appear only once, but it does not define what makes two files the same.
- It is unclear whether an exact resend should be ignored, whether a newer file should replace the earlier file, or whether separate attempts should remain available.
- It is unclear what should happen if the Data Partner has already released a file and another version arrives later. The new version could be rejected, replace the released file, or require another review and release.
- The current PR makes these product decisions implicitly and permanently at the database level.
- Until the expected behavior is clarified, the safer approach is to preserve existing data, prevent obvious repeated deliveries before files are stored, and show users one appropriate copy without deleting the underlying records.

I will update the card and ask Iris for clarification before either reopening this PR or (most likely) creating a new one with less risks.

PR #877, opened 2026-07-11
OTTER-614: Fix proposal page showing "submitted" instead of "approved" when navigating back

stetradis:
## Summary

When a researcher navigates backward from the code review stage to the Initial Request page, the proposal now correctly displays as "approved" instead of "submitted". Code submission resets study.status to PENDING-REVIEW, but `approvedAt` survives that flip — the fix uses approvedAt/rejectedAt as durable signals to derive the true proposal status.

Consolidates the durable-decision logic into a shared `effectiveProposalStatus` helper in review-decision.ts, built on top of the existing `proposalReviewDecision` function. This ensures researcher and reviewer paths use the same source of truth and can't drift.


src/lib/studies.ts
```diff
@@ -32,18 +33,19 @@ export function draftHasStep2Progress(study: DraftStep2Fields): boolean {
 
 /** Returns the timestamp of the latest decision for the submitted proposal header. */
 export function decisionTimestampForProposalHeader(study: SelectedStudy, entries: ProposalFeedbackEntry[]): Date {
-    if (study.status === 'APPROVED' && study.approvedAt) {
+    const status = effectiveProposalStatus(study)
```
nathanstitt, line 36: Nice side benefit here: code submission overwrites `study.submittedAt` (the submit-code actions set `submittedAt: now`), so before this change the fallback at the bottom of this function labeled the *code* submission date as "Submitted on". Deriving the durable status first means the header now shows the real approval date — this fixes the wrong-date case as well as the wrong-label one.

src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx
```diff
@@ -63,23 +64,24 @@ const PROPOSAL_BANNERS: Partial<Record<StudyStatus, ProposalBannerConfig>> = {
...
     studyVersion,
 }: {
     orgName: string
-    status: StudyStatus
+    study: Pick<SelectedStudy, 'status' | 'approvedAt' | 'rejectedAt'>
     studyVersion: number
 }) {
-    const config = PROPOSAL_BANNERS[status]
+    const proposalStatus = effectiveProposalStatus(study)
+    const config = PROPOSAL_BANNERS[proposalStatus]
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx:75]: With the effective status, the APPROVED banner copy now also renders mid-code-review (the back-nav case this PR targets). At that point the researcher has already signed agreements and submitted code, so "then proceed to Step 3 - Agreements to sign the required legal documents" is guidance for a step they've completed. 

Not sure if we want to update this as well as part of this fix or if it's best left for later.

src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx
```diff
@@ -165,7 +168,8 @@ export function ProposalSubmitted({
     feedbackError,
     returnTo,
 }: ProposalSubmittedProps) {
-    const bannerConfig = PROPOSAL_BANNERS[study.status]
+    const proposalStatus = effectiveProposalStatus(study)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx:171]: Optional: `effectiveProposalStatus` is now derived in three spots in this render tree (here, `StatusBanner`, and `ProposalNavigation`). It's a cheap pure function so no real cost — and having `StatusBanner` take `study` does prevent a caller from handing it a raw status — so take it or leave it: computing once here and passing `proposalStatus` down would keep a single derivation point.

src/lib/review-decision.ts
```diff
@@ -15,15 +15,21 @@ export function toReviewDecision(decision: Decision): ReviewDecision {
...
 // (performStudyProposalApproval) record the decision on the study itself without writing feedback.
 // Mapping the study status lets the feedback view synthesize a decision so it renders the decided
 // proposal instead of a blank PostFeedbackView (which returns null when no decision exists).
-export const PROPOSAL_STATUS_TO_REVIEW_DECISION: Partial<Record<StudyStatus, ReviewDecision>> = {
-    APPROVED: 'APPROVE',
-    REJECTED: 'REJECT',
-    'CHANGE-REQUESTED': 'NEEDS-CLARIFICATION',
-}
+// Derived from REVIEW_DECISION_TO_STATUS to keep the two maps in sync.
+export const PROPOSAL_STATUS_TO_REVIEW_DECISION = Object.fromEntries(
```
nathanstitt, line 30: Deriving the inverse from the canonical map is a nice touch — one less pair to keep in sync. One caveat: `Object.fromEntries` plus the cast means TS can't verify the 1:1 property the comment above asserts, so if a second decision ever mapped to the same status the inversion would silently drop an entry. Fine as-is given the enum is small, just flagging the assumption.

nathanstitt (approved):
good fix, ty!

PR #878, opened 2026-07-13
OTTER-658: make proposal resubmission note collaborative, fix duplicate saved checkmark

Rod0352:
## Changes

- Resubmission note on edit-and-resubmit is now collaborative: bound to a round-versioned
  Yjs doc (`proposal-{studyId}-resubmission-note-v{n}`) via the shared `Editor` — live sync,
  "Also editing", provider-driven save status.
- Editor service recognizes the new doc kind: lab-member auth while CHANGE-REQUESTED,
  cold-seed from the draft column (legacy plain-text drafts normalized), cleanup via the
  existing `proposal-{studyId}-%` purge on resubmit.
- Note validation accepts Lexical JSON or plain text — code-resubmit flow unchanged;
  single-user mode keeps the debounced draft autosave.
- Removed the extra `CheckCircleIcon` next to `SaveStatusIndicator` (the duplicate checkmark).

## Test

- Unit: doc-name parsing (both parsers), editability gating, seed normalization, schema
  bounds, one-icon regression test.
- e2e: resubmit note fill switched to `fillLexicalField`.
- Manual: two browsers on the same page — live sync, "Also editing", single checkmark.


src/server/actions/study-request.ts
```diff
@@ -899,7 +901,9 @@ export const saveCodeResubmissionNoteDraftAction = new Action('saveCodeResubmiss
 export const saveProposalResubmissionNoteDraftAction = new Action('saveProposalResubmissionNoteDraftAction', {
     performsMutations: true,
 })
-    .params(z.object({ studyId: z.string().uuid(), note: z.string().max(10_000) }))
+    // 40_000: the draft is serialized Lexical JSON since OTTER-658, which
+    // inflates a 300-word note well past the plain-text 10K bound.
+    .params(z.object({ studyId: z.string().uuid(), note: z.string().max(40_000) }))
```
nathanstitt: 40K is probably fine in practice, but it's not unreachable by a legitimate note: validation caps at 300 *words*, and Lexical serializes each formatting run as its own ~100-byte node, so a 300-word note with per-word formatting or a pile of links can brush up against this. When that happens, the single-user autosave fails with an error toast on every keystroke, while resubmit itself has no byte cap — so the note would still submit fine, but drafting gets noisy. The column is `text`, so bumping to something like 100K is cheap insurance.

nathanstitt (approved):
looks good, might bump the limit as suggested tho.  Thank you!

PR #880, opened 2026-07-13
Update color of star denoting selected main  file in study proposal view

jbwilson8:
Issue: [OTTER-590](https://openstax.atlassian.net/browse/OTTER-590)

## Summary

Updates the "Main file" star on the post-submission study code view to use the **disabled grey** style instead of the active blue (indigo), matching the Figma design.

Per UX guidance:
- The **blue** star represents an *active choice* — used while the researcher is choosing which file is their main file (the upload/selection flow).
- The **grey** star represents a *submitted, locked* state — the main file has already been selected and can no longer be changed. It stays filled to show it's still selected, but the greyed styling communicates that changes are disabled.

The post-submission view (and the post-decision view, which shares the same table component) is a read-only, locked context, so it should reflect the disabled grey version.

## Changes

- `src/components/study/submitted-code-table-view.tsx`: the main-file star now renders with `--mantine-color-gray-5` instead of `--mantine-color-indigo-6`. It keeps the `fill` weight so it still reads as "selected," while supplemental files remain grey outlines. The active indigo styling remains untouched in `file-review-table.tsx` where main-file selection actually happens.

### DIFF
--- a/src/components/study/submitted-code-table-view.tsx
+++ b/src/components/study/submitted-code-table-view.tsx
@@ -18,6 +18,11 @@ const formatUpdatedAt = (date: Date | string) =>
         minute: '2-digit',
     })
 
+// Post-submission the main file is locked in, so the star renders in the disabled
+// grey style (still filled to show it's selected) rather than the active indigo used
+// while the researcher is choosing their main file.
+const STAR_COLOR = 'var(--mantine-color-gray-5)'
+
 const SubmittedCodeRow: FC<{
     file: SubmittedFile
     jobId: string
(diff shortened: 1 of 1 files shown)


src/components/study/submitted-code-table-view.tsx
```diff
@@ -25,7 +25,10 @@ const SubmittedCodeRow: FC<{
 }> = ({ file, jobId, onPreview }) => {
     const isMain = file.fileType === 'MAIN-CODE'
     const starWeight = isMain ? 'fill' : 'regular'
-    const starColor = isMain ? 'var(--mantine-color-indigo-6)' : 'var(--mantine-color-gray-5)'
+    // Post-submission the main file is locked in, so the star renders in the disabled
+    // grey style (still filled to show it's selected) rather than the active indigo used
+    // while the researcher is choosing their main file.
+    const starColor = 'var(--mantine-color-gray-5)'
```
nathanstitt: Tiny optional nit: now that `starColor` no longer depends on `isMain`, it reads a little oddly sandwiched between the two `isMain`-derived ternaries — at a glance it implies it still varies per row. We could hoist it (and this comment) to a module-level const next to `formatUpdatedAt`, or inline it on the `StarIcon` below. Totally fine as-is though.

nathanstitt (approved):
thanks!

PR #881, opened 2026-07-13
Lost Key Page Refactoring and New Key Generation Flow Refactoring

chrisbendel:
closes OTTER-654 and OTTER-655


src/app/user-key/page.tsx
```diff
@@ -1,5 +1,12 @@
+import dayjs from 'dayjs'
+import { actionResult } from '@/lib/utils'
+import { getUserPublicKeyAction } from '@/server/actions/user-keys.actions'
 import { RegenerateKey } from './regenerate-key'
 
 export default async function ManageKeysPage() {
-    return <RegenerateKey />
+    // The layout redirects to key generation when no key exists, so a key is always present here.
+    const key = actionResult(await getUserPublicKeyAction())
+    const generatedOn = dayjs(key?.updatedAt).format('MMM DD, YYYY')
```
nathanstitt: Small robustness snag: since the comment above says a key is always present here, the optional chain `key?.updatedAt` is doing double duty. If `key` were ever null, `dayjs(undefined)` doesn't throw — it silently resolves to *now*, so the page would confidently render "You generated a security key on <today>". If the invariant truly holds, `key.updatedAt` (non-optional) is both clearer and fails loudly instead of showing a plausible-but-wrong date.

src/components/layout/navbar-profile-menu.tsx
```diff
@@ -27,6 +27,8 @@ export function NavbarProfileMenu() {
 
     const menuRef = useClickOutside<HTMLDivElement>(handleClickOutside)
     const isSiAdmin = session?.user.isSiAdmin || false
+    // Any key-holding member (enclave or lab); mirrors the UserKey ability + RequireUserKey guard.
+    const needsKey = Object.values(session?.orgs || {}).some(orgNeedsKey)
```
nathanstitt: With this change, what types of orgs and users will NOT need a key?  It seems simpler to just require a key for everyone to me

src/server/actions/user-keys.actions.ts
```diff
@@ -79,10 +79,7 @@ export const updateUserPublicKeyAction = new Action('updateUserPublicKeyAction')
         await assertValidPublicKey(publicKey)
         const fingerprint = await fingerprintKeyData(publicKey)
 
-        // Rotation swaps the fingerprint, orphaning study_job_file_recipient_key rows wrapped to the
-        // old one — the researcher loses access to already-approved results. Recovery is the renewal
-        // re-wrap flow (a teammate re-wraps for the new key); until that ships, self-rotate is not
-        // surfaced to researchers. See renewal follow-up.
+        // Rotation swaps the fingerprint, orphaning outputs wrapped to the old key; the loss is confirmed in the reset modal.
```
nathanstitt, line 82: Worth a sanity check on this one: the old comment here flagged that self-rotate orphans a *researcher's* access to already-approved results with no recovery until the re-wrap/renewal flow ships — which is why the menu entry used to be reviewer-only. This PR widens that entry to lab (researcher) users and swaps the code guard for the modal warning in `regenerate-key-view.tsx` ("...those outputs will be lost"). Reasonable product call if intended — just flagging that the safety rail moved from code to copy. Is the renewal follow-up still tracked somewhere so it doesn't get lost?

src/app/account/keys/generate-keys.tsx
```diff
@@ -38,102 +26,117 @@ type GenerateKeysProps = {
...
+                    </Code>
+                </Stack>
+
+                <Stack gap="xs">
+                    <Group>
+                        <Button onClick={onCopy}>Copy key</Button>
+                        <NextButton isVisible={hasAttemptedCopy} onClick={openConfirm} />
+                    </Group>
+                    <CopySucceededIndicator isVisible={clipboard.copied} />
+                    <CopyFailedIndicator isVisible={hasAttemptedCopy && clipboard.error != null} />
```
nathanstitt: Nice touch — decoupling "Next appears" (`hasAttemptedCopy`) from "copy actually succeeded" (`clipboard.copied`/`clipboard.error`) means a blocked clipboard shows the manual-copy fallback without trapping the user, and the tests cover exactly that. One thing to be aware of: with `timeout: 2000`, `clipboard.copied` flips back to false after 2s, so the "Copied!" indicator disappears while Next stays — presumably fine, just flagging the transient.

nathanstitt (commented):
Seems good, but think we can simplify the logic to just require a key for all users

src/components/layout/navbar-profile-menu.tsx
chrisbendel: wasnt sure about the admin case here, if we can assume admins will all have keys (even if they dont belong to an org) then i agree on the simplification

@nathanstitt let me know what you think in terms of opening it up for everyone (and relaxing the requirements/checks elsewhere in the code) vs keeping it just to enclave/lab and avoiding the admin edge case 🤔

src/app/user-key/page.tsx
chrisbendel: already hit this in a prior commit 👍

src/server/actions/user-keys.actions.ts
chrisbendel: @nathanstitt this is a good question for @iccgoncalves on the product side, im assuming this was already thought of and considered a trade off based on the text the users will now see. re-wrap should be a followup issue regardless of what we choose at this point

src/lib/types.ts
```diff
@@ -44,6 +44,16 @@ export function orgNeedsKey(org: { type: OrgType }): boolean {
...
 }
 
+// Single source of truth for "this user must hold an encryption key": any member of a key-holding
+// org, plus SI admins (who can review/decrypt anything and are the intended re-key operators).
+// Mirrored by the UserKey ability (permissions.ts), the RequireUserKey guard, and the sign-in
+// redirect. SI admins already satisfy orgNeedsKey today via the enclave `safe-insights` org — the
+// explicit isSiAdmin term makes the requirement independent of that org's type.
+export function sessionNeedsKey(session: Pick<UserSession, 'user' | 'orgs'> | null | undefined): boolean {
+    if (!session) return false
+    return session.user.isSiAdmin || Object.values(session.orgs).some(orgNeedsKey)
```
nathanstitt: sorry I think I wasn't clear on earlier review: **everyone** needs a key, so this might as well just `return true`, (or just don't call it), but is fine to leave it in case _maybe_ someday we do want to adjust the logic.

src/lib/types.ts
```diff
@@ -44,6 +44,16 @@ export function orgNeedsKey(org: { type: OrgType }): boolean {
     return isEnclaveOrg(org) || isLabOrg(org)
```
nathanstitt: Minor, and it ties into Nathan's earlier 'why not just require a key for everyone' thread: `OrgType` is exactly `'enclave' | 'lab'`, so `isEnclaveOrg(org) || isLabOrg(org)` is `true` for every possible org — the branch reads as if some org type is exempt, but none is. Effectively `orgNeedsKey` already means 'is a member of any org'. If the product answer lands on 'everyone with an org needs a key' this could collapse to `Object.values(session.orgs).length > 0` (or just be inlined), which would also make the intent unambiguous for the next reader. Not blocking — just noting the current form implies a distinction the type system says can't occur.

src/app/user-key/page.tsx
```diff
@@ -1,5 +1,17 @@
...
+import { Routes } from '@/lib/routes'
+import { getUserPublicKeyAction } from '@/server/actions/user-keys.actions'
 import { RegenerateKey } from './regenerate-key'
 
 export default async function ManageKeysPage() {
-    return <RegenerateKey />
+    // The layout redirects to key generation when no key exists; guard here too so a missing key
+    // never silently formats today's date as the generated-on value.
+    const key = actionResult(await getUserPublicKeyAction())
+    if (!key) redirect(Routes.accountKeys)
```
nathanstitt, line 12: Confirming the earlier thread is resolved here — the `redirect(Routes.accountKeys)` guard plus non-optional `key.updatedAt` on the next line means a missing key can no longer silently render 'generated on <today>'. `getUserPublicKey` now also selects `updatedAt`, so this reads a real value. 👍

nathanstitt (changes requested):
hey sorry, I wasn't clear before, but since all users are either lab or dp, then everyone needs a key

src/lib/types.ts
chrisbendel: def makes sense, wasn't sure how defensive we'd want to be, but i like the idea of simplifying it to everyone

src/lib/types.ts
chrisbendel: simplified per above ^ comment

nathanstitt (approved):
Looks great, thanks @chrisbendel !

PR #883, opened 2026-07-15
OTTER-516: preview png plots in code file view modals

Rod0352:
###Problem: #818 fixed the results page, but png plots still didn't render in the view modal on the study code Review files page or the Submitted code table (QA follow-up on the ticket). Both paths decoded file bytes to utf-8 strings server-side — fs.readFile(path, 'utf-8') in readWorkspaceFileAction, blob.text() in fetchStudyJobCodeFileAction — which corrupts binary files.

##Fix: Both actions return raw bytes like the results path does, and the image-vs-text dispatch from encrypted-files-panel moves to a shared FileOrImagePreviewModal used by all three preview sites. The reviewer's inline code viewer decodes at point of use; text previews are unchanged.

##Tests: unit for the png preview on both screens, plus a byte-exact round-trip of non-utf-8 bytes through readWorkspaceFileAction.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
@@ -19,8 +19,8 @@ import { useEffect, useState } from 'react'
 import Markdown, { type Components } from 'react-markdown'
 import remarkGfm from 'remark-gfm'
 import { useMutation, useQuery, useQueryClient } from '@/common'
-import { CodeViewer } from '@/components/file-viewers'
-import { decodeFileContents } from '@/lib/file-content-helpers'
+import { CodeViewer, ImageViewer } from '@/components/file-viewers'
+import { decodeFileContents, imageMimeType } from '@/lib/file-content-helpers'
 import { highlightLanguageForFile } from '@/lib/languages'
 import { studyCodeURL } from '@/lib/paths'
 import {
@@ -509,6 +509,15 @@ function StudyCodeBody({
             </Alert>
         )
     }
+    const mime = imageMimeType(activeFile.name)
+    if (mime) {
+        return (
+            <div data-testid="study-code-body">
+                <ImageViewer name={activeFile.name} contents={data.contents} mime={mime} />
+            </div>
+        )
+    }
+
     const code = decodeFileContents(data.contents)
     return (
         <div data-testid="study-code-body">
--- a/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
@@ -20,6 +20,7 @@ import Markdown, { type Components } from 'react-markdown'
 import remarkGfm from 'remark-gfm'
 import { useMutation, useQuery, useQueryClient } from '@/common'
 import { CodeViewer } from '@/components/file-viewers'
+import { decodeFileContents } from '@/lib/file-content-helpers'
 import { highlightLanguageForFile } from '@/lib/languages'
 import { studyCodeURL } from '@/lib/paths'
 import {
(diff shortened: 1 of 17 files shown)


src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
```diff
@@ -508,9 +509,10 @@ function StudyCodeBody({
             </Alert>
         )
     }
+    const code = decodeFileContents(data.contents)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx:521]: This is the one consumer of `fetchStudyJobCodeFileAction` that still decodes unconditionally — the inline tab-strip viewer on the reviewer page, as opposed to the modal. Since the action now hands back raw bytes for *all* files (including png), a png that shows up as a tab here would get `decodeFileContents`'d into garbage and rendered as text in `CodeViewer`. `useStudyCodeViewer`/`FileTab` don't seem to filter images out of the tab list, so it looks reachable.

I get that the PR scope was explicitly the *modals* ("reviewer's inline code viewer decodes at point of use; text previews unchanged"), so this may be a deliberate follow-up — but it does mean the same OTTER-516 corruption still exists on this surface. Worth either confirming png can't land here as a tab, or noting it as a known gap so it doesn't look fixed when it isn't.

src/components/modals/file-or-image-preview-modal.tsx
```diff
@@ -0,0 +1,22 @@
...
+    contents: ArrayBuffer | null
+}
+
+// Callers hand over raw bytes so binary files like png plots survive the trip intact;
+// decoding to utf-8 happens here and only for non-image files (OTTER-516).
+export function FileOrImagePreviewModal({ file, onClose }: { file: PreviewFile | null; onClose: () => void }) {
+    if (!file) return null
+
+    const mime = imageMimeType(file.name)
+    if (mime && file.contents) {
```
nathanstitt, line 16: Nice consolidation pulling the image-vs-text dispatch into one place all three call sites share. One small thing worth a mental note: the branch keys purely on `imageMimeType(file.name)`, i.e. the extension. A binary file with a non-image extension (or a text file misnamed `.png`) routes the wrong way — the former decodes to garbage, the latter tries to render text bytes as an image. That's the same trade-off the results path already made and probably fine given code submissions are extension-conventional, just flagging that the dispatch is name-driven, not content-sniffed.

src/server/actions/workspace-files.actions.ts
```diff
@@ -42,8 +42,9 @@ export const readWorkspaceFileAction = new Action('readWorkspaceFileAction', {})
         const coderFilesPath = await getStudyFilesPath(studyId)
         const sanitized = sanitizeFileName(fileName)
         const filePath = path.join(coderFilesPath, sanitized)
-        const contents = await fs.readFile(filePath, 'utf-8')
-        return { fileName: sanitized, contents }
+        // Raw bytes, not utf-8: workspace files include binary artifacts like png plots (OTTER-516)
+        const contents = await fs.readFile(filePath)
+        return { fileName: sanitized, contents: new Uint8Array(contents).buffer }
```
nathanstitt, line 47: Good catch using `new Uint8Array(contents).buffer` rather than `contents.buffer` here — `fs.readFile` returns a `Buffer` that's a view onto a shared, pooled ArrayBuffer, so handing back `.buffer` directly could leak neighbouring bytes / the wrong length. Copying into a fresh exact-size `Uint8Array` first is the correct move, and the byte-exact round-trip test in use-ide-files locks it in.

nathanstitt (approved):
looks good but should also include the tab strip viewer as well

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx:521]: Added in aeac1a2f — StudyCodeBody now branches on imageMimeType and renders the image inline. The object-URL rendering moved out of ImagePreviewModal into a shared ImageViewer next to the other file viewers, with a test covering a png tab.

src/components/modals/file-or-image-preview-modal.tsx
Rod0352: Agreed it's name-driven — kept it consistent with the results path since submissions are extension-conventional. If misnamed files show up in practice, content sniffing (png magic bytes) can layer into imageMimeType in one place now.

PR #884, opened 2026-07-15
OTTER-646: add 1h escalation tooltip to reviewer code processing status

Rod0352:
fix: reviewer `JOB-RUNNING` ("Code processing") tooltip on the DP dashboard now reads "The code is now running against the enclave. If it stays in this status for over 1h, contact your Org Admin." (was the "you'll receive an email once results are ready" copy)

test: add hover test asserting the new tooltip in `display-study-status.test.tsx`

### DIFF
--- a/src/components/study/display-study-status.test.tsx
+++ b/src/components/study/display-study-status.test.tsx
@@ -1,6 +1,5 @@
 import { describe, expect, it } from 'vitest'
-import { renderWithProviders } from '@/tests/unit.helpers'
-import { screen } from '@testing-library/react'
+import { renderWithProviders, screen, userEvent, waitFor } from '@/tests/unit.helpers'
 import { DisplayStudyStatus } from '@/components/study/display-study-status'
 import { REVIEWER_STATUS_LABELS, RESEARCHER_STATUS_LABELS } from '@/lib/status-labels'
 import type { AllStatus } from '@/lib/types'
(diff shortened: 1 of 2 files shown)


src/components/study/display-study-status.test.tsx
```diff
@@ -29,6 +28,21 @@ describe('DisplayStudyStatus', () => {
             expect(textEl).toBeDefined()
             expect(textEl.parentElement?.textContent?.trim()).toBe(expectedText)
         })
+
+        it('shows the escalation tooltip when hovering the JOB-RUNNING pill', async () => {
```
nathanstitt, line 32: Nice call adding a real hover test rather than just asserting the value in `REVIEWER_STATUS_LABELS` — this exercises the whole `DisplayStudyStatus` → `InfoTooltip` → Mantine portal path, so it'd catch a regression where the tooltip stops rendering (not just a copy change). It's also the first tooltip-open test in the repo, which is a good pattern to have around.

src/components/study/display-study-status.test.tsx
```diff
@@ -29,6 +28,21 @@ describe('DisplayStudyStatus', () => {
...
+            const user = userEvent.setup()
+            renderWithProviders(<DisplayStudyStatus status={REVIEWER_STATUS_LABELS['JOB-RUNNING']!} />)
+
+            await user.hover(screen.getByText('Code processing'))
+
+            await waitFor(() => {
```
nathanstitt: Small thing: `getByText` already throws when the node is absent, so `expect(...).toBeDefined()` never actually does the asserting here — the real signal is `getByText` inside `waitFor` retrying until the portal mounts. We could collapse the whole `waitFor(() => expect(getByText(...)).toBeDefined())` block into one `expect(await screen.findByText('The code is now running…')).toBeVisible()` — `findByText` bundles the retry + throw, and `toBeVisible()` additionally confirms the tooltip actually opened rather than just being present in the DOM. Purely a tidy-up, current version works.

nathanstitt (approved):
looks good, thanks Rodney

PR #886, opened 2026-07-15
OTTER-652: align status banner background colors with UX spec

therealmarv:
see [OTTER-652](https://openstax.atlassian.net/browse/OTTER-652)

## Problem

Status banner background colors on the proposal and code feedback pages did not match the UX spec and risked accessibility issues. The same states were also colored differently across three separate files, which is the "inconsistency" called out in the ticket.

## Root cause

Each banner hardcoded its own Mantine palette shade, mostly the `.1` (second-lightest) shade. The spec wants the `.0` (lightest) shade, which already exists in the theme palette and matches Figma's semantic `*/Light` design tokens exactly.

Verified against the Figma nodes linked in the ticket:

| State | Figma token | Hex | Theme token |
|---|---|---|---|
| Approved (DP & RL) | `Color/Success/Light` | `#E8F8EB` | `green.0` |
| Rejected (DP & RL) | `Color/Error/Light` | `#FFE0E0` | `red.0` |
| Change requested, reviewer (DP) | `Color/Warning/Light` | `#FFF9E5` | `yellow.0` |
| Change requested, researcher (RL) | `Color/Brand/Light` | `#EAE8FC` | `purple.0` |

## Changes

- **New `src/lib/status-banner-colors.ts`**: a single `STATUS_BANNER_BG` map with semantic keys (`approved`, `rejected`, `changesRequestedReviewer`, `changesRequestedResearcher`) resolving to existing palette tokens. This reuses the theme rather than adding a parallel color system, and centralizes the reviewer-vs-researcher difference for change-requested.
- **`post-feedback-view.tsx`** (reviewer, proposal + code): swapped `green.1` / `yellow.1` / `red.1` to the shared map.
- **`code-post-decision-view.tsx`** (researcher, code): swapped `green.1` / `purple.1` / `red.1` to the shared map.
- **`proposal-submitted.tsx`** (researcher, proposal): the `<Alert>` now takes an explicit `bg` token from the map; change-requested moves from the blue family (`#ECF3FF`) to purple to match `purple.0`.
- **New `status-banner-colors.test.ts`**: a deterministic regression test asserting each semantic token resolves to the exact required hex, so a palette reorder or accidental token change is caught.

## Scope

Kept to the six banner states listed in the ticket. The small dashboard status pills (`status-labels.ts`), the "under review" banners, and the review-criteria banners are intentionally left unchanged.

## Testing

- `pnpm run checks` passes (typecheck, lint, action validation).
- Existing banner tests assert on `data-testid` presence, not colors, so they are unaffected.

[OTTER-652]: https://openstax.atlassian.net/browse/OTTER-652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx
@@ -14,6 +14,7 @@ import { ProposalHeader } from '../../request/page-header'
 import { Routes } from '@/lib/routes'
 import { Link } from '@/components/links'
 import { effectiveProposalStatus } from '@/lib/review-decision'
+import { STATUS_BANNER_BG } from '@/lib/status-banner-colors'
 
 interface ProposalSubmittedProps {
     orgSlug: string
@@ -32,6 +33,7 @@ function proposalHeading(studyVersion: number): string {
 
 type ProposalBannerConfig = {
     color: string
+    bg?: string
     message: (orgName: string) => string
     statusBadge?: string
 }
@@ -44,18 +46,21 @@ const PROPOSAL_BANNERS: Partial<Record<StudyStatus, ProposalBannerConfig>> = {
     },
     APPROVED: {
         color: 'green',
+        bg: STATUS_BANNER_BG.approved,
         statusBadge: 'Approved on',
         message: (orgName) =>
             `${displayOrgName(orgName)} has reviewed and approved your initial request. Review their feedback below, then proceed to Step 3 - Agreements to sign the required legal documents.`,
     },
     REJECTED: {
         color: 'red',
+        bg: STATUS_BANNER_BG.rejected,
         statusBadge: 'Rejected on',
         message: (orgName) =>
             `${displayOrgName(orgName)} has reviewed your initial request and is unable to support it at this time. Please review their feedback below for more details.`,
     },
     'CHANGE-REQUESTED': {
-        color: 'blue',
+        color: 'purple',
+        bg: STATUS_BANNER_BG.changesRequestedResearcher,
         statusBadge: 'Clarification requested on',
         message: (orgName) =>
             `${displayOrgName(orgName)} has reviewed your initial request and has requested clarifications. Please review their feedback below. You can revise and resubmit your request to address their questions.`,
@@ -81,7 +86,7 @@ function StatusBanner({
         : config.message(orgName)
 
     return (
-        <Alert color={config.color} mb="md" data-testid={`status-banner-${proposalStatus}`}>
+        <Alert color={config.color} bg={config.bg} mb="md" data-testid={`status-banner-${proposalStatus}`}>
             {message}
         </Alert>
     )
--- a/src/lib/status-banner-colors.test.ts
+++ b/src/lib/status-banner-colors.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, it } from 'vitest'
+import { theme } from '@/theme'
+import { STATUS_BANNER_BG } from './status-banner-colors'
+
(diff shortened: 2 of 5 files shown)


src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx
```diff
@@ -81,7 +86,7 @@ function StatusBanner({
         : config.message(orgName)
 
     return (
-        <Alert color={config.color} mb="md" data-testid={`status-banner-${proposalStatus}`}>
+        <Alert color={config.color} bg={config.bg} mb="md" data-testid={`status-banner-${proposalStatus}`}>
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx:89]: One subtle divergence worth a sanity check: this is the only one of the three banners that's a Mantine `<Alert>` rather than a plain `<Box bg={...}>`. Passing both `color` and `bg` means `color` still drives the Alert's default `light` variant border + title/text tint, and `bg` only overrides the background. So for CHANGE-REQUESTED we now get both `color: 'purple'` (border/text) *and* `bg: purple.0` (background), whereas the researcher code banner in `code-post-decision-view.tsx` gets a flat `purple.0` Box with no border tint. Not a bug — but worth confirming the purple border/text that rides along with the blue→purple swap here is intended and matches Figma, since it's an extra semantic layer the other two banners don't have.

src/lib/status-banner-colors.test.ts
```diff
@@ -0,0 +1,24 @@
+import { describe, expect, it } from 'vitest'
+import { theme } from '@/theme'
+import { STATUS_BANNER_BG } from './status-banner-colors'
+
+// Resolve a Mantine "color.shade" token (e.g. "green.0") to its concrete hex via the theme palette.
+function resolveToken(token: string): string {
+    const [color, shade] = token.split('.')
+    const tuple = theme.colors?.[color as keyof typeof theme.colors]
+    return tuple?.[Number(shade)] ?? ''
```
nathanstitt, line 9: Nice touch that the `?? ''` fallback keeps this honest — a typo'd token like `green.99` resolves to `''`, and `''.toUpperCase()` won't equal the expected hex, so the test still fails loudly rather than throwing. Solid regression guard for both an accidental token edit and a palette reorder.

nathanstitt (approved):

src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/submitted/proposal-submitted.tsx:89]: Good callout, I dug into this and it turns out the extra purple tint doesn't actually render here. In Mantine 8's default (`light`) Alert variant:

- The message body is forced to `var(--mantine-color-black)` in light scheme (`.mantineAlertMessage`), so `color` never tints the body text.
- The border is `1px solid transparent` in the light variant, so `color` doesn't produce a visible border either.
- `--alert-color` (the `purple` light-color) only surfaces on the title / icon / close button, and this Alert renders none of those.

So `color="purple"` is effectively inert here: the only thing that shows is `bg={purple.0}` (background) plus black body text. I confirmed that against the Figma RL Proposal Change node (`10434-3280`): flat `#EAE8FC` background, dark `#0d0d0d` body text, no colored border. So it matches the spec, and the researcher code banner in `code-post-decision-view.tsx` (flat `purple.0` Box, black text) renders the same visible result despite using a different element.

(`color` still matters for `PENDING-REVIEW`, which has no `bg` override and relies on it for the yellow-light background, so I left it in place.)

PR #887, opened 2026-07-15
Detect merged PR only on the main build

nathanstitt:
Merge detection ran in the Deploy stage on every branch and only matched `Merge pull request #N`.

- On a `PR-*` build Jenkins checks out `refs/pull/N/head` (the PR's own head commit, never a merge), so it never matched — while `CHANGE_ID` drove a preview deploy anyway.
- The merge/squash commit actually lands on the `main` build, and squash/rebase merges use `(#N)`, which the old regex missed.

Gate detection to the `main` build and recognize `(#N)` messages too, so `MERGED_PR_NUMBER` is set on merge and the deploy tears down the merged PR's preview.

Paired with safeinsights/iac#188, which consumes `MERGED_PR_NUMBER` to run the teardown.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,15 +54,25 @@ pipeline {
                     echo commitMsg
                     env.COMMIT_MESSAGE = commitMsg
 
-                    // try two common patterns: "Merge pull request #123" or any "#123"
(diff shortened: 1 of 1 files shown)


nathanstitt:
I think this won't be needed, we'll address on the deploy side in IaC PR

PR #891, opened 2026-07-15
OTTER-635: Increase results approval regression coverage

therealmarv:
see [OTTER-635](https://openstax.atlassian.net/browse/OTTER-635)

## Summary

- Investigation found that the originally reported stuck researcher status is no longer reproducible in the current code. Data Partner approval already records `FILES-APPROVED`, resolves the researcher dashboard to `Result ready`, and exposes the shared result artifact.
- Increase regression coverage with a connected cross-role test that follows Data Partner approval through the researcher dashboard status and researcher artifact access.
- Keep production behavior unchanged. This PR only strengthens coverage against the reported regression returning.

## Testing

- `pnpm run checks`
- `./bin/docker-test -- --run 'study-job.actions'`

No E2E tests were run.

[OTTER-635]: https://openstax.atlassian.net/browse/OTTER-635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/actions/study-job.actions.test.ts
+++ b/src/server/actions/study-job.actions.test.ts
@@ -1,5 +1,13 @@
 import { describe, expect, test, vi, type Mock } from 'vitest'
-import { db, insertTestStudyJobData, mockSessionWithTestData, actionResult } from '@/tests/unit.helpers'
+import {
+    actionResult,
+    db,
+    insertTestOrg,
+    insertTestStudyJobData,
+    insertTestUser,
+    mockClerkSession,
+    mockSessionWithTestData,
+} from '@/tests/unit.helpers'
 import {
     approveStudyJobFilesAction,
     fetchEncryptedJobFilesAction,
@@ -9,6 +17,10 @@ import {
 } from './study-job.actions'
 import { sendStudyResultsRejectedEmail } from '@/server/mailer'
 import { onStudyReviewRequested } from '@/server/events'
+import { fetchStudiesForOrgAction } from './study.actions'
+import { dashboardRawStateFromRow } from '@/components/dashboard/studies-table/dashboard-raw-state'
+import type { StudyRow } from '@/components/dashboard/studies-table/types'
+import { projectStudyState, resolvePillStatus } from '@/lib/study-screen'
 
 vi.mock('@/server/storage', () => ({
     fetchCodeManifest: vi.fn(() => ({})),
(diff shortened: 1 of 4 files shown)


src/server/actions/study-job.actions.test.ts
```diff
@@ -162,29 +220,44 @@ describe('Study Job Actions', () => {
...
+                orgSlug: lab.slug,
+                userId: researcher.id,
+                orgId: lab.id,
+                orgType: 'lab',
+            })
+            const studies = actionResult(await fetchStudiesForOrgAction({ orgSlug: lab.slug }))
+            const dashboardStudy = studies.find((candidate) => candidate.id === study.id)!
+            const state = projectStudyState(dashboardRawStateFromRow(dashboardStudy as StudyRow))
+            expect(state.resultsApproved).toBe(true)
+            expect(resolvePillStatus('researcher', state)).toMatchObject({ stage: 'Results', label: 'Ready' })
```
nathanstitt, line 253: Really like that this drives the *real* `approveStudyJobFilesAction` and then asserts the researcher's dashboard pill resolves to Results/Ready — that end-to-end path (reviewer approves → `insertSharedFileKeys` → dashboard state → researcher fetch) is exactly the regression the ticket worried about, and the pre-existing `fetchEncryptedJobFilesAction returns researcher keys` test only covered the fetch side in isolation. Nice connected coverage.

src/server/actions/study-job.actions.test.ts
```diff
@@ -162,29 +220,44 @@ describe('Study Job Actions', () => {
...
+            mockClerkSession({
+                clerkUserId: researcher.clerkId,
+                orgSlug: lab.slug,
+                userId: researcher.id,
+                orgId: lab.id,
+                orgType: 'lab',
+            })
+            const studies = actionResult(await fetchStudiesForOrgAction({ orgSlug: lab.slug }))
+            const dashboardStudy = studies.find((candidate) => candidate.id === study.id)!
+            const state = projectStudyState(dashboardRawStateFromRow(dashboardStudy as StudyRow))
```
nathanstitt, line 251: seems suspicious to me that we need a cast here, do we need it?

The `dashboardStudy as StudyRow` cast is a little load-bearing here — `dashboardRawStateFromRow` reads `jobStatusChanges`/`piUserId`/`datasets`/etc off the row, so if `fetchStudiesForOrgAction`'s shape ever drifts from `StudyRow` the cast would hide it rather than fail the type-check (the test would just silently compute a different `RawStudyState`). It lines up today since `fetchStudyQuery` selects all those fields, so more a note than a blocker — if there's a typed accessor for the dashboard row it'd be nice to lean on it instead of the cast.

nathanstitt (approved):

src/server/actions/study-job.actions.test.ts
therealmarv: Good catch. The cast mirrors the existing production boundary in `studies-table/index.tsx`, so I kept it here to make this regression test follow the real dashboard path. That said, I agree the cast is a type-safety smell in both places and could hide some query-shape drift.

I would prefer to address that separately by defining and enforcing a shared dashboard-row contract at the server-action boundary, then removing both the production and test casts. For this test-only PR, I would leave the cast as-is rather than introduce that broader refactor.

PR #892, opened 2026-07-15
OTTER-570: bump eslint-plugin-react-hooks to ^7.1.1 and fix newly-surfaced rule violations

therealmarv:
see [OTTER-570](https://openstax.atlassian.net/browse/OTTER-570)

During the npm to pnpm migration, `eslint-plugin-react-hooks` was pinned to exact `7.0.1` so `pnpm install` would not pull a newer minor with stricter rules. This unpins it, adopts the current rules, and fixes every violation they surface.

## Dependency bump

- `eslint-plugin-react-hooks`: `7.0.1` -> `^7.1.1` (devDependencies), lockfile resolves to `7.1.1`.
- `7.1.1` is the current latest and still the same major (7), so this is a semver-safe caret. `7.1.0` was a React 19.2 tooling minor and `7.1.1` was a compatibility patch (it restores `component-hook-factories` as a deprecated no-op). No CVEs and no breaking rule-behavior changes.

## Violations fixed

The bump surfaced 10 violations: 1 `react-hooks/static-components` and 9 `react-hooks/set-state-in-effect`. Real fixes were preferred over suppressions.

- `admin/settings/code-envs.tsx` (`static-components`): `DeleteCodeEnv` was declared inside another component's render body. Hoisted it to a module-level component that takes an `isVisible` prop and returns null when hidden, matching the project's conditional-visibility convention.
- `components/layout/org-admin-dashboard-link.tsx`: replaced the effect that synced the menu open state to the route with the React "adjust state during render when a prop changes" pattern (tracking the previous route value). No cascading re-render, same behavior.
- `components/timer.ts`: the disabled branch wrote `0` back into state from the effect. Now the disabled value is derived on return (`isEnabled ? timeRemaining : 0`), so no state write in the effect.
- `account/signin/use-already-signed-in.ts`: the one-time landing status (`signed-out` / `signed-in` / `redirecting`) is now resolved during render, latched by the existing `status === 'loading'` guard. The actual navigation runs in an effect that fires exactly once via a ref, so an unstable router reference cannot cause a redirect loop. `switchAccount` still forces `signed-out` explicitly.
- `hooks/use-positions-section.ts`: the auto-open of the first position is now decided during render (keyed on the persisted positions). The form-reset effect keeps its original `editingIndex !== null` guard.
- `hooks/use-personal-info-section.ts`, `hooks/use-education-section.ts`, `hooks/use-research-details-section.ts`: kept the original seeding effect that only runs while the user is not editing (`if (isEditing) return`), so a background refetch can never overwrite unsaved input. The initial edit-mode decision is coupled to seeding the Mantine form from server data and must run in the same pass, so those `setState` calls carry a justified `set-state-in-effect` suppression.

## Guarding against form data loss on refetch

Researcher-profile data is fetched with a 15-minute `refetchInterval` and refetches on window focus, so `data` can change while a section form is open for editing. The seeding effects in the three profile hooks are guarded so they never resync the form while `isEditing` is true. This protects three cases:

- unsaved edits to Mantine-managed fields,
- an interest typed but not yet committed (separate draft state),
- an interest committed through Enter or removal.

An earlier iteration guarded with `form.isDirty()` instead of `isEditing`, but Mantine's dirty check is not a reliable divergence signal here (an uncommitted draft is separate state, and committed interests go through list operations), so a value-changing refetch could still discard an edit. Guarding on `isEditing` is the robust choice and matches the pre-existing behavior.

## Justified suppressions

Five `set-state-in-effect` suppressions remain, all for genuine external-store synchronization that the effect model is meant for:

- `contexts/study-request/study-request-context.tsx`: seeding the Mantine form and document-file store from a server-provided draft.
- `hooks/use-yjs-form-map.ts`: exposing the Hocuspocus provider created and torn down by the effect.
- the three profile hooks above: the initial edit-mode decision coupled to seeding the form from server data.

No new suppressions were added beyond these; all other pre-existing `exhaustive-deps` comments are unchanged.

## Tests

Added regression tests that render the real sections/components and exercise the guarded behaviors:

- personal info: an unsaved field edit survives a refetch that changes server data.
- education: an unsaved field edit survives a refetch that changes server data.
- research details: an uncommitted interest draft survives a refetch.
- research details: a committed interest pill survives a refetch (this one failed against the `form.isDirty()` approach, which is what motivated the `isEditing` guard).
- org admin dashboard link: the submenu open state re-syncs when the route changes into and out of an admin path (covers the render-phase pattern that replaced the effect).

Each new guard test was confirmed to fail if its guard is removed, so it genuinely locks the behavior.

## Validation

- `pnpm run checks` passes (type check, lint, action validation), no warnings.
- Full unit suite passes: 210 files, 1955 tests.
- E2E was not run locally; it will run in CI.

Mantine v8 dirty-tracking behavior was confirmed against current docs via context7.

[OTTER-570]: https://openstax.atlassian.net/browse/OTTER-570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
         "eslint-plugin-anti-trojan-source": "^1.1.4",
         "eslint-plugin-import": "^2.32.0",
         "eslint-plugin-react": "^7.37.5",
-        "eslint-plugin-react-hooks": "7.0.1",
+        "eslint-plugin-react-hooks": "^7.1.1",
         "globals": "^17.4.0",
         "happy-dom": "*",
         "husky": "^9.1.7",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -321,8 +321,8 @@ importers:
         specifier: ^7.37.5
         version: 7.37.5(eslint@9.39.4(jiti@2.7.0))
       eslint-plugin-react-hooks:
-        specifier: 7.0.1
-        version: 7.0.1(eslint@9.39.4(jiti@2.7.0))
+        specifier: ^7.1.1
+        version: 7.1.1(eslint@9.39.4(jiti@2.7.0))
       globals:
         specifier: ^17.4.0
         version: 17.6.0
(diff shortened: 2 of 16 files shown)


nathanstitt (approved):
thanks!

PR #895, opened 2026-07-16
OTTER-636: show Proposal/Code draft on the first real edit

therealmarv:
see [OTTER-636](https://openstax.atlassian.net/browse/OTTER-636)

## Goal

A study should read as a **draft** the moment a real edit is made to a proposal or code, never on mere navigation or on opening/downloading a submitted file. Reuses the existing model: proposal draft = "Draft" (Proposal) pill; code draft = latest job `INITIATED` ("Draft", Code).

## Approach (revised after code review)

The proposal side is now a **display-only** signal rather than a status flip. Flipping `study.status` to `DRAFT` mid-edit collided with the whole `CHANGE-REQUESTED`-gated resubmission flow (note autosave, collaborative editor auth, kick-out, back-nav, delete rules, dashboards). Instead:

- New nullable column `study.proposalEditedAt` (migration). Stamped by `markProposalEditedAction` on the first real edit to a change-requested proposal. `study.status` stays `CHANGE-REQUESTED`, so every existing gate keeps working and the reviewer keeps seeing "Change requested" until the proposal is resubmitted.
- Projection fact `proposalDraftInProgress = CHANGE-REQUESTED && proposalEditedAt != null`; the researcher pill reads "Proposal draft". Cleared on resubmit.
- Edit detection covers every editable field in both single-user and collaborative modes and ignores hydration: title/datasets/PI fire from their user-driven change handlers; rich-text fields fire only when content actually changes; the note fires in single-user mode (where its onChange is user-driven). The stamp latch only sets after the action succeeds, so a transient failure retries.

The code side (Gap C) opens a fresh `INITIATED` round on the first real edit after a live `CODE-CHANGES-REQUESTED` (in addition to the existing round-closing `FILES-*` case), so the dashboard already reads "Code draft". Deleting/replacing a submitted code file also opens that round.

## Review response

The external review was essentially all correct; this revision addresses it:

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | Most proposal fields didn't trigger the draft | Edit detection now fires from every field's user-driven handler (title/datasets/PI/rich-text/note), independent of Yjs availability, excluding hydration. |
| 2 | Reverted DRAFT broke note-save / editor-auth / kick-out | Root cause removed: status no longer flips, so all `CHANGE-REQUESTED` gates keep working. |
| 3 | Previously-submitted proposal became deletable | Removed: no reverted-DRAFT state, so delete-draft never applies to a submitted study. |
[...]

PR #897, opened 2026-07-16
OTTER-636: draft concept, immutable proposal snapshots + revision/code round architecture

therealmarv:
see [OTTER-636](https://openstax.atlassian.net/browse/OTTER-636)

Full-architecture implementation of the draft concept, replacing the display-only approach in #895 (which will be closed). Immutable submitted-proposal snapshots + an explicit fresh-vs-revision discriminator + an immutable code-round model, so literal proposal `DRAFT` and code Draft are safe without leaking in-progress content to reviewers.

## Core model
- `study.status` is the proposal workflow state; a proposal revision literally uses `DRAFT`.
- Fresh vs revision drafts are distinguished by `study.proposal_revision_base_submission_id`, never by `submittedAt`.
- Reviewers read the latest immutable `study_proposal_submission` snapshot, never the mutable `study` row / live Yjs.
- Code rounds are immutable; absolute-latest round drives Draft display, latest submitted round drives reviewer content; at most one open code-draft round; round identity is UUIDv7 id order.
- Every mutating action uses a positive eligibility predicate.

## Phase checklist
- [x] **Phase 1** migrations (snapshot table, revision-base FK + check + indexes, backfill) + `types.ts` (validated against the real dev DB via `migrate-dev-db`; codegen diff is empty).
- [x] **Phase 3 foundation** snapshot Zod schema/serializer + central proposal/code predicates (+ unit tests).
- [x] **Phase 4** reviewer `/review`, read-only `/view`, `/submitted`, AI review agent read the latest snapshot.
- [x] **Phase 8/9 round model** fresh `INITIATED` round after any revision-entry decision (change-requested, bare `JOB-ERRORED`, FILES-approved/rejected); per-study advisory lock; UUIDv7 identity; delete opens/reuses a round; `codeRoundStatusesForStudy` selector.
- [x] **Phase 8/9 submit gate (Finding 6)** positive `canSubmitInitialCode` gate: the initial-submit path can't overwrite reviewed code or bypass the resubmission note.
- [x] **Phase 2 / 5 (snapshots)** every submission (fresh finalize + resubmit) writes an immutable snapshot version; `finalizeStudySubmissionAction` is fresh-only; `/proposal` redirects change-requested (and revision drafts) to edit-and-resubmit (Finding 3).
- [x] **Phase 6** literal `startProposalRevisionAction` (CHANGE-REQUESTED to DRAFT + base FK, idempotent, lab-scoped); the first *local* user edit flips the study via a shared `useStartProposalRevision` hook wired through `onLocalUserEdit` on the collaborative/single-user editors (Yjs hydration and remote-collaborator updates never trigger it); editor/auth (`services/editor/auth.ts`) accepts a revision DRAFT for proposal fields/text and the resubmission note; `resubmitProposalAction` accepts a revision DRAFT and clears the base on resubmit; Resubmit is blocked while the transition is pending/failed.
- [x] **Phase 10 (reviewer dashboard)** reviewers now see a revision draft (change-requested proposal being edited) as a non-actionable "Proposal Draft" and never a fresh draft; opening it renders the last immutable submitted snapshot + existing feedback read-only (routed to `reviewer-proposal-feedback`), never the mutable draft; the researcher dashboard routes a revision draft straight to edit-and-resubmit with no delete affordance. Threaded `isProposalRevisionDraft` through the study-state projection.
- [x] **Phase 7 (create)** lazy-create a fresh draft the moment a Data Partner is chosen (the first persistable Step-1 edit), not only on "Proceed to Step 2"; subsequent Step-1 edits autosave; the Data Partner selector locks once the draft exists. Duplicate-create is guarded by a synchronous id ref.
- [x] **Phase 10 auth** `load IDE` is scoped to the submitting lab (plus the author) in `src/lib/permissions.ts`, so a member of an unrelated lab can no longer open another lab's workspace. Applied with explicit sign-off; the IDE ability subject gained `submittedByOrgId`.
- [x] **Query invalidation** every transition refetches the dashboards: lazy-create + code submit/resubmit invalidate the researcher study keys, the revision flip invalidates both researcher and reviewer keys, proposal resubmit / reviewer decisions revalidate the dashboard + review paths.

## Verification
Dev stack runs under Docker; migrations applied to the real dev DB and `types.ts` regen produced no diff. Every commit is `pnpm run checks`-green and the full unit suite passes under `./bin/docker-test` (currently 219 files / 2010 tests).

## Review round (addressed)
- **Resubmit is strict (architecture §4/§7):** `resubmitProposalAction` accepts only a revision `DRAFT` (non-null base) via `canResubmitProposal`; `startProposalRevisionAction` fails visibly if a submitted study somehow has no snapshot; Resubmit stays disabled until the first-edit flip commits. The e2e seed now writes an immutable snapshot for every submitted study (mirroring finalize/backfill), which is what lets the flip succeed in tests.
- **Selection is not an edit:** the editor `onLocalUserEdit` fires only on real content mutations (`dirtyElements`/`dirtyLeaves`), never on focus/cursor moves — honoring the Jira non-trigger.
- **Revision drafts protected:** cannot be soft-deleted (submitted history preserved); `/edit` redirects them to the lab-gated `/edit-and-resubmit`.
- **Cache:** the flip invalidates both researcher and reviewer dashboard key families by prefix.
- **Snapshot dataset resolution** uses `id::text = ANY(...)` (uuid `IN` on non-uuid dataset labels 500'd reviewer pages).

## Notes
- Collaborative editor / two-session reviewer flows need manual QA (not unit-testable). In particular: the first-edit DRAFT flip across two collaborators, and a reviewer viewing a revision draft while a researcher edits it.
- The reviewer's read-only view of a revision draft reuses the existing change-requested feedback screen (the snapshot overlay presents its effective status as change-requested), so the standard "clarification requested, you'll be notified once they resubmit" banner already conveys that reviewer action is unavailable.
- Deferred (known-incomplete architecture items, not regressions): code-lifecycle correctness (concurrency claim-inside-lock, positive launch/upload/delete gates, dashboard code-draft representation, bare `JOB-ERRORED` resubmit wiring), server-side rejection of proposal decisions on revision drafts, and single-user proposal-field autosave.

[OTTER-636]: https://openstax.atlassian.net/browse/OTTER-636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/proposal-snapshot.ts
+++ b/src/lib/proposal-snapshot.ts
@@ -1,0 +1,73 @@
+import { z } from 'zod'
+
+// OTTER-636: the single validated shape of an immutable submitted-proposal snapshot. Every write to and
+// read from `study_proposal_submission.snapshot` goes through this schema + the serializer below — do
+// NOT hand-build partial snapshot objects in individual actions. It captures every reviewer-visible
+// proposal field (the funnel is getStudyAction -> ProposalRequest, plus the AI review agent), so a
+// reviewer never has to read the mutable `study` row or live Yjs docs.
+export const PROPOSAL_SNAPSHOT_SCHEMA_VERSION = 1
+
+export const proposalSnapshotSchema = z.object({
+    title: z.string().nullable(),
+    piName: z.string(),
+    piUserId: z.string().nullable(),
+    language: z.enum(['R', 'PYTHON']),
+    datasets: z.array(z.string()).nullable(),
+    dataSources: z.array(z.string()),
+    // Lexical JSON bodies are opaque here; they are validated by the editor/form layer on the way in.
+    researchQuestions: z.unknown().nullable(),
+    projectSummary: z.unknown().nullable(),
+    impact: z.unknown().nullable(),
+    additionalNotes: z.unknown().nullable(),
+    irbProtocols: z.string().nullable(),
+    descriptionDocPath: z.string().nullable(),
+    irbDocPath: z.string().nullable(),
+    agreementDocPath: z.string().nullable(),
+})
+
+export type ProposalSnapshot = z.infer<typeof proposalSnapshotSchema>
+
+// The canonical proposal columns as read off a `study` row (Kysely selects them as these types).
+export type ProposalSnapshotSource = {
+    title: string | null
+    piName: string
+    piUserId: string | null
+    language: 'R' | 'PYTHON'
+    datasets: string[] | null
+    dataSources: string[]
+    researchQuestions: unknown
+    projectSummary: unknown
+    impact: unknown
+    additionalNotes: unknown
+    irbProtocols: string | null
+    descriptionDocPath: string | null
+    irbDocPath: string | null
+    agreementDocPath: string | null
+}
+
+// Build the validated snapshot object from a study row's canonical proposal columns. Callers persist the
+// returned object as-is into `study_proposal_submission.snapshot`.
+export function serializeProposalSnapshot(source: ProposalSnapshotSource): ProposalSnapshot {
+    return proposalSnapshotSchema.parse({
+        title: source.title,
+        piName: source.piName,
+        piUserId: source.piUserId,
+        language: source.language,
+        datasets: source.datasets,
+        dataSources: source.dataSources ?? [],
+        researchQuestions: source.researchQuestions ?? null,
+        projectSummary: source.projectSummary ?? null,
+        impact: source.impact ?? null,
+        additionalNotes: source.additionalNotes ?? null,
+        irbProtocols: source.irbProtocols,
+        descriptionDocPath: source.descriptionDocPath,
+        irbDocPath: source.irbDocPath,
+        agreementDocPath: source.agreementDocPath,
+    })
+}
+
+// Validate a stored jsonb snapshot back into a typed ProposalSnapshot. Reviewer loaders use this so a
+// malformed/legacy row fails loudly rather than rendering partial data.
+export function parseProposalSnapshot(raw: unknown): ProposalSnapshot {
+    return proposalSnapshotSchema.parse(raw)
+}
--- a/src/lib/status-labels.ts
+++ b/src/lib/status-labels.ts
@@ -30,8 +30,17 @@ const COLORS = {
 
 // Proposal -> Code -> Results
 export const REVIEWER_STATUS_LABELS: Partial<Record<AllStatus, StatusLabel>> = {
-    // note: there is no 'DRAFT' label here even though that status exists on studies
-    // BECAUSE a reviewer should never see a DRAFT study
+    // OTTER-636: a reviewer never sees a FRESH draft (a brand-new proposal, filtered out of reviewer
+    // listings), but they DO see a REVISION draft: a change-requested proposal the researcher has begun
+    // editing. It shows here as a non-actionable "Proposal Draft" while they revise; reviewer actions
+    // stay unavailable until they resubmit (→ PENDING-REVIEW).
+    DRAFT: {
+        stage: 'Proposal',
+        label: 'Proposal Draft',
+        tooltip:
+            'The researcher is revising this proposal after your change request. You’ll be notified once they resubmit.',
+        colors: COLORS.draft,
+    },
 
     // Proposal
     'PENDING-REVIEW': {
--- a/src/server/actions/study-request.ts
+++ b/src/server/actions/study-request.ts
@@ -17,7 +17,12 @@ import {
 } from '@/server/aws'
 import { CODER_DISABLED, getConfigValue, SIMULATE_CODE_BUILD } from '@/server/config'
 import { getOrCreateCurrentRoundJob, nextVersionForStudyComment } from '@/server/db/mutations'
-import { codeSubmissionVersion, getInfoForStudyId, getOrgIdFromSlug } from '@/server/db/queries'
+import {
+    codeRoundStatusesForStudy,
+    codeSubmissionVersion,
+    getInfoForStudyId,
+    getOrgIdFromSlug,
+} from '@/server/db/queries'
 import { rawStudyStateForStudy } from '@/server/db/study-state-query'
 import { db as database } from '@/database'
 import { deferred, onStudyReviewRequested, onStudyCodeSubmitted, onStudyCreated } from '@/server/events'
@@ -34,7 +39,7 @@ import {
     resubmissionNoteToLexicalJson,
     resubmissionNoteWordCount,
 } from '@/app/[orgSlug]/study/[studyId]/edit-and-resubmit/schema'
-import { canResearcherResubmitCode, projectStudyState } from '@/lib/study-screen'
+import { canResearcherResubmitCode, canSubmitInitialCode, projectStudyState } from '@/lib/study-screen'
 
 const simulateJobScan = deferred(async (studyJobId: string) => {
     await sleep({ 1: 'seconds' })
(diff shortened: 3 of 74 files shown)


src/server/actions/study-request.ts
```diff
@@ -525,6 +547,17 @@ export const submitStudyCodeAction = new Action('submitStudyCodeAction', { perfo
             throw new Error('Main file not in file list')
         }
 
+        // OTTER-636 (Finding 6): the initial code-submit path carries no resubmission note, so it is
+        // valid ONLY for a genuine first submission (an open INITIATED round, no prior submitted round).
+        // Once a round has been submitted/reviewed, a further submission must go through
+        // resubmitStudyCodeAction (which requires a note) — a positive gate here stops the initial path
+        // from ever overwriting reviewed code or bypassing the note.
+        if (!canSubmitInitialCode(await codeRoundStatusesForStudy(studyId, db))) {
```
nathanstitt, line 555: This `canSubmitInitialCode` read happens before the per-study advisory lock (which only fires later inside `attachCodeToRoundJob → getOrCreateCurrentRoundJob`). Two concurrent initial submits could both read `latestSubmitted == null`, pass the gate, and then serialize on the lock — the second one still opening/attaching to a round the first just submitted. The PR notes claim-inside-lock as deferred, so this is just to flag the exact window: worth a follow-up where the gate is re-checked after the lock is held (or the lock is taken here) so the positive gate is actually atomic with the round decision.

src/server/db/proposal-snapshot.ts
```diff
@@ -0,0 +1,131 @@
...
+        ])
+        .where('id', '=', studyId)
+        .executeTakeFirstOrThrow()
+
+    const maxRow = await db
+        .selectFrom('studyProposalSubmission')
+        .select((eb) => eb.fn.max('version').as('max'))
+        .where('studyId', '=', studyId)
+        .executeTakeFirst()
+    const version = Number(maxRow?.max ?? 0) + 1
```
nathanstitt, line 42: `version = max(existing) + 1` is computed without any lock, and the unique constraint on `(study_id, version)` is the only thing preventing a duplicate. In practice the status guards on finalize/resubmit serialize submissions, so this should never collide — but if it ever did (e.g. a retried mutation racing), the second insert would surface as a raw constraint-violation error rather than a clean `ActionFailure`. Given the advisory lock already exists for code rounds, might be worth taking the same per-study lock here so the version bump is provably serial.

src/server/db/proposal-snapshot.ts
```diff
@@ -0,0 +1,131 @@
...
+    const latest = await latestProposalSnapshotForStudy(studyId, db)
+    if (!latest) return study
+    // OTTER-636: a revision draft renders as its last submitted version, which was change-requested (the
+    // only decision that produces a revision draft). Present that effective status so read-only feedback
+    // / submitted views treat it as a submitted proposal rather than a fresh draft; the base id is left
+    // on the object so those views can still show a "being revised" notice.
+    const isRevisionDraft = study.status === 'DRAFT' && study.proposalRevisionBaseSubmissionId != null
+    return {
+        ...study,
+        status: isRevisionDraft ? 'CHANGE-REQUESTED' : study.status,
```
nathanstitt, line 114: Presenting a revision draft's effective status as `CHANGE-REQUESTED` is clever and the comment justifies it well (change-requested is the only decision that produces a revision draft). One thing worth double-checking: this overlay is what makes `/review/proposal` safe — `proposalReviewDecision` returns `undefined` for the raw DRAFT row so the page falls through to `renderStudyScreen`, which overlays. That's a fairly load-bearing implicit coupling across three files. A one-line comment pointing at the fallthrough (or a direct assertion/test that a revision draft never renders the mutable row on `/review/proposal`) would make the invariant harder to accidentally break later.

src/server/db/mutations.ts
```diff
@@ -71,52 +73,41 @@ export async function getOrCreateCurrentRoundJob(
     studyId: string,
     { backdateMs = 0 }: { backdateMs?: number } = {},
 ): Promise<RoundJob> {
+    // OTTER-636 (Finding 10): serialize round creation per study so concurrent launch/upload/delete/
+    // submit can't each read "no open round" and mint duplicates that split files/notes/scans across
+    // jobs. Transaction-scoped — the mutating actions that call this run inside a transaction.
+    await sql`select pg_advisory_xact_lock(hashtext(${studyId}))`.execute(db)
```
nathanstitt, line 79: `hashtext()` returns a 32-bit int, so distinct study ids can hash-collide and briefly serialize two unrelated studies' round creation. Harmless for correctness (it's a coarse lock, and the work inside is short) — just noting it's a global 32-bit keyspace rather than per-study isolation, in case round-creation ever becomes hot enough for the collisions to matter.

src/lib/proposal-snapshot.ts
```diff
@@ -0,0 +1,73 @@
+import { z } from 'zod'
+
+// OTTER-636: the single validated shape of an immutable submitted-proposal snapshot. Every write to and
```
nathanstitt, line 3: Nice call making this the single validated funnel for every snapshot read/write — the `do NOT hand-build partial snapshot objects` note plus the round-trip test is exactly the guardrail this needs. The Zod schema catching a legacy/malformed row loudly (rather than rendering partial data to a reviewer) is a really solid choice for this immutable-snapshot model.

src/lib/status-labels.ts
```diff
@@ -30,8 +30,17 @@ const COLORS = {
...
 export const REVIEWER_STATUS_LABELS: Partial<Record<AllStatus, StatusLabel>> = {
-    // note: there is no 'DRAFT' label here even though that status exists on studies
-    // BECAUSE a reviewer should never see a DRAFT study
+    // OTTER-636: a reviewer never sees a FRESH draft (a brand-new proposal, filtered out of reviewer
+    // listings), but they DO see a REVISION draft: a change-requested proposal the researcher has begun
+    // editing. It shows here as a non-actionable "Proposal Draft" while they revise; reviewer actions
+    // stay unavailable until they resubmit (→ PENDING-REVIEW).
+    DRAFT: {
+        stage: 'Proposal',
+        label: 'Proposal Draft',
```
nathanstitt, line 39: Conceptual gap vs OTTER-636 (code-draft representation). This adds the reviewer-visible "Proposal Draft" label — the proposal half of the draft concept, and it's done well. But the ticket is symmetric: the code triggers (launch IDE, upload/delete/edit a code file on an approved study) are specified as "study status must change to == Code Draft." There's no Code Draft counterpart label anywhere — neither here nor in RESEARCHER_STATUS_LABELS. An open draft round is INITIATED-only with no researcher label, so `resolvePillStatus` (pill.ts) falls through to the prior study-status label (e.g. Approved). Net effect: a researcher who opens the IDE on an approved study sees no visible state change; the immutable-round model exists underneath, but the user-facing half of the code-draft concept isn't rendered. The PR description lists "dashboard code-draft representation" as deferred — since "the pill flips to Code Draft" is one of the card's most visible acceptance criteria, worth promoting that to an explicit, linked follow-up card rather than a description bullet, so the scope boundary is agreed rather than implicit.

nathanstitt:
## Conceptual review against OTTER-636

Reviewed the overall concept against the ticket (including the architecture comment and the deferred-items list). Summary — the architecture is right; the one real shortfall is completeness on the code-draft UX. Details below (one concrete point is left as an inline comment on `status-labels.ts`).

### Where the concept matches the ticket well
- **Fresh-vs-revision discriminator** — the ticket (§2) explicitly asks for "an explicit discriminator, not a guess based on timestamps." `proposalRevisionBaseSubmissionId` + the "never infer from `submittedAt`" rule in the predicate module is exactly that.
- **Immutable snapshots solve §1** (reviewers must never see in-progress draft content). Repointing every reviewer read through `overlaidWithLatestProposalSnapshot` + the snapshot table + backfill matches the prescribed approach.
- **"Real edit" detection matches the trigger/non-trigger table** — gating on `dirtyElements/dirtyLeaves` while excluding `collaboration`/`history-merge` tags faithfully implements "selection/focus/hydration is *not* a trigger." This is the subtle part of the card and it's handled correctly.
- **Immutable code rounds + per-study advisory lock** address §4 as described.

### The main conceptual gap: code-draft is modeled but not surfaced
The ticket is symmetric — proposal *and* code both must visibly enter "draft" on a real edit. The proposal half lands (visible "Proposal Draft" state). The **code half is plumbing without a face**: there's no "Code Draft" pill/label, so a researcher who opens the IDE / uploads a file on an approved study — the card's headline code trigger — sees no visible state change (falls through to the prior "Approved" label). The PR already lists "dashboard code-draft representation" as deferred. That's fine for a staged PR (the card itself says multi-PR), but since "the pill flips to Code Draft" is arguably the single most visible acceptance criterion, it should be an explicit, linked follow-up card, not a description bullet. (Left as an inline note on `status-labels.ts`.)

### Smaller notes
- **Proposal status change is literal; code status change is a reinterpretation.** Proposals actually flip `study.status → DRAFT`; code reinterprets "status" as a derived display concept (no `CODE-DRAFT` enum). Defensible given the round model is the source of truth, but a deliberate deviation from the card's literal wording — worth stating for sign-off.
- **Server-side rejection of decisions on a revision draft is *mostly* already covered.** `claimInitialProposalReviewStudy` gates on `status = 'PENDING-REVIEW'`, so approve/reject/needs-clarification on a revision DRAFT already throws server-side, not just hidden in the UI. Worth confirming there's no *other* proposal-decision path that keys off something looser before treating the deferred "server-side rejection on revision drafts" bullet as fully open.
- **Verification honesty matches §6** — the collaborative/two-user flows genuinely can't be unit-tested; the PR says so and a manual test plan is attached to the card. Consistent with the ticket's "confidence comes incrementally" expectation.

### Bottom line
Land it as "proposal-draft concept complete + code-round data model," but make the deferred **code-draft representation** (and any remaining server-side revision-draft decision gaps) named, linked follow-up cards agreed as out-of-scope-for-this-PR, rather than leaving them as description bullets.

nathanstitt (commented):
I think this approach is sound but think we should discuss it throughly a bit more, made some comments on the card

therealmarv:
This is something for the future. If we need it again we can reopen this one. Closing it for now.

PR #900, opened 2026-07-20
Enforce Key Generation for Existing Users Without One

chrisbendel:
Functionally implemented in #881, this PR adds tests to lock down the intended behavior

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15,12 +15,16 @@ overrides:
   vite: ^8.0.16
   ws: ^8.21.0
   hono@<4.12.25: '>=4.12.25'
-  js-yaml@<=4.1.1: '>=4.1.2'
+  js-yaml@>=4.0.0 <4.3.0: '>=4.3.0 <5'
   nodemailer@<=8.0.8: '>=8.0.9'
   nodemailer@<9.0.1: '>=9.0.1'
   '@opentelemetry/core@<2.8.0': '>=2.8.0'
   '@babel/core@<=7.29.0': '>=7.29.1'
   shell-quote@<1.8.5: '>=1.8.5'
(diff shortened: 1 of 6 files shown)


therealmarv (approved):
Approving. Good to merge overall.

On the infosec side: I worked through the same Trivy failures independently and landed on the exact same conclusion, the same six transitive bumps via workspace overrides (axios 1.18.1, brace-expansion 5.0.7, js-yaml 4.3.0, shell-quote 1.10.0, fast-uri 3.1.4, sharp 0.35.3), including overriding Next's sharp pin. So that part looks right to me. 👍

A few minor, non-blocking nits on the rest:

1. In the two dual-role sign-in tests, the `deleteFrom('userPublicKey')` doesn't actually remove anything. `mockDualRoleSessionWithTestData` seeds the account through a lab org and the enclave side only adds a membership, so there's never a key at that point. Harmless, but it reads like the helper creates a key when it doesn't. Might be worth dropping it or leaving a quick comment.

2. The two "merges a second email..." tests assert the key-status outcome (`needsUserKey`) but not the merge itself, i.e. that the new org membership row was created or that the Clerk `createEmailAddress`/`updateEmailAddress` calls fired. Since the goal is locking down the behavior, an assertion on the actual merge would make the tests match their names and guard that path too.

3. The copy tweak in `regenerate-key-view.tsx` ("You need this key to access your outputs in every organization you belong to") is nice and accurate, but it's a user-facing change that isn't mentioned in the description (which says the PR just adds tests). Worth a line in the PR body.

4. Tiny process one: the title could use the `OTTER-656:` prefix to match our convention.

None of these are blockers, just polish. Nice work locking this down.

nathanstitt (approved):
looks great, thank Chris, I'm going to merge so we can get into the new release

PR #901, opened 2026-07-20
enforce draft privacy from Data Organizations on the backend

jbwilson8:
[Issue:](https://openstax.atlassian.net/browse/OTTER-596) Unsubmitted drafts must not be readable by Data Organizations

### Summary
Unsubmitted study drafts (proposal content and uploaded code) were hidden from Data Organizations (DOs) only in the UI and list actions — the per-record backend paths were not enforcing this. A DO (enclave) member who knew a study or job ID could read draft proposal text, documents, and code directly via server actions **and** the download routes (`/dl/study-code`, `/dl/study-documents`, `/dl/results`).

This PR closes that gap on the server so a draft stays private to its Research Lab until it is submitted. No frontend changes.

### Approach
- Gated the two enclave (reviewer) `view` ability rules in [`src/lib/permissions.ts`](src/lib/permissions.ts) with a positive `SUBMITTED_STUDY_STATUSES` allowlist (`status: { $in: [...] }`).
  - Chose an allowlist over `$ne: 'DRAFT'` deliberately: the CASL subject is assembled from middleware-returned fields, so `$in` **fails closed** when `status` is absent (denies), whereas `$ne` would **fail open** (grant). A security fix should err toward denial.
  - Lab rules (`submittedByOrgId`) are untouched, so Research Labs keep full access to their own drafts. CASL's additive semantics keep dual-role users working.
- Threaded the study `status` into every enclave-reachable read path so the gate has the data it needs: all `view Study` / `view StudyJob` middlewares, the shared queries (`getStudyJobInfo`, `latestJobForStudyQuery`, `getProposalFeedbackForStudy`), the download-route guards in [`src/server/auth.ts`](src/server/auth.ts), and their feeder queries (`jobInfoForJobId`, `studyInfoForStudyId`, `getInfoForStudyJobId`).

### Acceptance criteria
- [x] **Drafts are private to their Research Lab** — a DO cannot access an unsubmitted draft's proposal, documents, or code by known ID.
- [x] **Submitted studies are unaffected** — DOs keep access across the review lifecycle, including `CHANGE-REQUESTED` resubmissions.
- [x] **Research Lab collaboration is unaffected** — lab members still see/edit/submit their lab's drafts (OTTER-497 / OTTER-550 behavior preserved).
- [x] **Other labs stay locked out** — Lab B still cannot see Lab A's drafts.

### Testing
- New unit tests in `permissions.test.ts`, `study.actions.test.ts`, and `study-job.actions.test.ts` cover all four criteria (DO denied on DRAFT for both proposal and code file; DO allowed once submitted and on `CHANGE-REQUESTED`; lab retains draft access; other lab locked out).

### DIFF
--- a/src/lib/permissions.ts
+++ b/src/lib/permissions.ts
@@ -11,19 +11,27 @@ import {
 
 export { subject, type AppAbility, type PermissionsActionSubjectMap, type PermissionsSubjectToObjectMap, toRecord }
 
-// Every study status except DRAFT, i.e. "has been submitted at least once". Used to gate what a
-// Data Organization (enclave reviewer) may read: unsubmitted drafts belong to the Research Lab only
-// (OTTER-596). This is a positive allowlist on purpose — the ability subject is assembled from
-// middleware-returned fields, and a mongo `$in` fails CLOSED when `status` is absent (denies),
-// whereas `$ne: 'DRAFT'` would fail OPEN (grant). A new StudyStatus must be added here or DOs will
-// not be able to view studies in that status — the safe direction for a fix about hiding content.
-const SUBMITTED_STUDY_STATUSES: StudyStatus[] = [
-    'APPROVED',
-    'ARCHIVED',
-    'CHANGE-REQUESTED',
-    'PENDING-REVIEW',
-    'REJECTED',
-]
+// Which study statuses a Data Organization (enclave reviewer) may read: every status except DRAFT,
+// i.e. "has been submitted at least once". Unsubmitted drafts belong to the Research Lab only
+// (OTTER-596). Declared as a Record<StudyStatus, …> so adding a new StudyStatus is a TypeScript error
+// until someone makes an explicit visible/hidden decision here — the one constant the whole read
+// boundary hinges on.
+const ENCLAVE_VIEWABLE_STUDY_STATUS: Record<StudyStatus, boolean> = {
+    DRAFT: false,
+    'PENDING-REVIEW': true,
+    'CHANGE-REQUESTED': true,
+    APPROVED: true,
+    REJECTED: true,
+    ARCHIVED: true,
+}
+
+// Derived allowlist used as a positive `$in` condition. Positive on purpose — the ability subject is
+// assembled from middleware-returned fields, and a mongo `$in` fails CLOSED when `status` is absent
+// (denies), whereas `$ne: 'DRAFT'` would fail OPEN (grant): the safe direction for a fix about hiding
+// content.
+const SUBMITTED_STUDY_STATUSES = (Object.keys(ENCLAVE_VIEWABLE_STUDY_STATUS) as StudyStatus[]).filter(
+    (status) => ENCLAVE_VIEWABLE_STUDY_STATUS[status],
+)
 
 export function defineAbilityFor(session: UserSession) {
     const { isSiAdmin } = session.user
--- a/src/lib/permissions.ts
+++ b/src/lib/permissions.ts
@@ -1,4 +1,5 @@
 import { type UserSession, isLabOrg, isEnclaveOrg, isOrgAdmin } from './types'
+import type { StudyStatus } from '@/database/types'
 import { AbilityBuilder, createMongoAbility, subject } from '@casl/ability'
 import {
     AppAbility,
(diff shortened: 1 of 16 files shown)


src/lib/permissions.ts
```diff
@@ -43,8 +58,11 @@ export function defineAbilityFor(session: UserSession) {
     // researchers need to be able to view enclave orgs to create studies
     permit('view', 'Org')
 
-    permit('view', 'Study', { orgId: { $in: usersReviewerOrgIds } })
-    permit('view', 'StudyJob', { orgId: { $in: usersReviewerOrgIds } })
+    // Enclave (Data Organization) reviewers may only view SUBMITTED studies/jobs. Unsubmitted drafts
+    // (proposal content + uploaded code) stay private to the submitting Research Lab, which retains
+    // access via the submittedByOrgId rules below (OTTER-596).
+    permit('view', 'Study', { orgId: { $in: usersReviewerOrgIds }, status: { $in: SUBMITTED_STUDY_STATUSES } })
```
therealmarv, line 72: **[High] A Data Organization can still reach a draft by first changing its status.**

This gate is solid, but it only controls reads. The status a study sits in is not itself protected. `approveStudyProposalAction` and `rejectStudyProposalAction` (`src/server/actions/study.actions.ts:433` and `:463`) authorize on the `approve`/`reject Study` rules (lines 92-93 of this file), which check `orgId` only and never look at status. A reviewer of the draft's assigned enclave who knows the draft id can call either action directly:

- `performStudyProposalApproval` only guards `status === 'APPROVED' || approvedAt`, so a DRAFT passes and flips to `APPROVED` (proposal-only drafts return early at `study.actions.ts:378` and approve cleanly).
- `markStudyRejected` has no status guard at all, so a DRAFT flips to `REJECTED`.

Both land the study in `SUBMITTED_STUDY_STATUSES`, this gate then opens, and the DO can read the proposal and code it was never meant to see. That is the exact scenario the card forbids.

The fix is small and self-contained: require the pre-decision status (`PENDING-REVIEW`) server-side in these two actions, the way `claimInitialProposalReviewStudy` already does at `study.actions.ts:511-529`. This is the one item I would treat as blocking for the card.

src/lib/permissions.ts
```diff
@@ -10,6 +11,20 @@ import {
 
 export { subject, type AppAbility, type PermissionsActionSubjectMap, type PermissionsSubjectToObjectMap, toRecord }
 
+// Every study status except DRAFT, i.e. "has been submitted at least once". Used to gate what a
+// Data Organization (enclave reviewer) may read: unsubmitted drafts belong to the Research Lab only
+// (OTTER-596). This is a positive allowlist on purpose — the ability subject is assembled from
+// middleware-returned fields, and a mongo `$in` fails CLOSED when `status` is absent (denies),
+// whereas `$ne: 'DRAFT'` would fail OPEN (grant). A new StudyStatus must be added here or DOs will
+// not be able to view studies in that status — the safe direction for a fix about hiding content.
+const SUBMITTED_STUDY_STATUSES: StudyStatus[] = [
```
therealmarv: **[Low] Consider making this allowlist exhaustiveness-checked.**

This is a plain `StudyStatus[]` literal, so if a new submitted status is ever added to the `StudyStatus` union, TypeScript will not flag its absence here. Because the gate fails closed, the effect would be safe (DOs silently over-denied on the new status) but it would be an easy-to-miss regression in the reviewer workflow. A `Record<StudyStatus, boolean>` map or a `satisfies` construct would force whoever adds a status to make an explicit include/exclude decision. Not blocking, just future-proofing the one constant the whole boundary hinges on.

therealmarv (changes requested):
Thanks for this, the core of the change is genuinely good. Gating the enclave `view` rules with a positive `$in` allowlist so a missing `status` fails closed is exactly the right call for a privacy fix, the comment explaining the `$in`-over-`$ne` choice is excellent, and threading `status` through the read paths and download guards is thorough. Lab access via `submittedByOrgId`, dual-role users, and SI admins all still work.

I'm requesting changes for one thing, and it is a small one.

**[High] The read gate can be bypassed by changing a draft's status.** Details are inline on `permissions.ts` line 64. In short: `approveStudyProposalAction` / `rejectStudyProposalAction` don't check status, so a DO reviewer with a known draft id can flip a DRAFT to APPROVED/REJECTED and then read it. Requiring `PENDING-REVIEW` in those two actions closes it, and that is the only item I'd treat as blocking.

There is also one **[Low]** optional hardening note inline on the `SUBMITTED_STUDY_STATUSES` definition.

### Test coverage gaps (worth adding alongside the fix)

- **[Medium] No negative test for the bypass above.** There is good coverage that a DO is denied `getStudyAction` on a draft and that a lab teammate keeps access, but nothing asserts that a DO is denied `approveStudyProposalAction` / `rejectStudyProposalAction` on a DRAFT, nor that the draft's status is left unchanged when the call is rejected. That test is what would lock in the fix and stop a future regression from silently reopening the hole.

- **[Medium] No route-level test for the download guards this PR modified.** The PR adds status threading to the `/dl/study-code` and `/dl/study-documents` guards in `auth.ts` and their feeder queries, but the closest existing test exercises `fetchStudyJobCodeFileAction` (the server action), not the routes themselves. A test that a DO gets a 401 on `/dl/study-code` and `/dl/study-documents` for a DRAFT job, and a redirect/200 once the study is submitted, would directly cover the paths the PR claims to protect.

- **[Low] Acceptance criterion 3 is only partially covered.** The "Research Lab collaboration is unaffected" tests confirm a lab member can still read a draft, but not that they can still edit or submit/resubmit one after this change. A short assertion that a lab member's update/submit still succeeds would round out that criterion.

Everything else I looked at is either pre-existing or outside this card's scope, so I've deliberately kept it off this review.

therealmarv (approved):
thanks, looks nice now! 🟢

PR #904, opened 2026-07-20
OTTER-527: prevent deleting the last usable code environment 

Rod0352:
Fixes [OTTER-527](https://openstax.atlassian.net/browse/OTTER-527)

### Problem

In DO Admin → Settings, the delete icon on a code environment is clickable even when
it's the only usable (non-testing) base image — clicking shows the confirmation popup,
then the server rejects the delete with an error toast.

Root cause: `362e453d` (sample-data-format refactor) collapsed the per-language
`canDelete` guard from `9c109416` into a blanket `images.length > 1` and dropped the
tests that protected it. The server-side guard in `deleteOrgCodeEnvAction` survived,
which is why the delete failed noisily instead of destroying data.

### Change

The delete icon is now hidden, and the server action refuses, when a **non-testing**
env is either:

1. the last non-testing env for its language (restores the pre-regression rule), or
2. the only non-testing env for its language whose latest scan passed
   (`SCAN-COMPLETE`).

Rule 2 is intentionally **new behavior**, not part of the old guard — it implements the
ticket's "successfully passed scans" wording. Without it, an org with one passed and one
failed R image could delete the passed one, leaving the failed image as the language's
default (the [OTTER-525](https://openstax.atlassian.net/browse/OTTER-525) failure mode,
since the default is the newest non-testing env per language).

Unchanged: testing envs are always deletable; a scan-failed env is deletable whenever
another non-testing env for its language exists.

Also: `SuretyGuard` accepts an optional `label` prop rendered as `aria-label` on the
trash `ActionIcon` (used here as `Delete <env name>`); other call sites are unaffected.

### Tests

- Component: delete hidden for the only env, for the last non-testing env per language,
  and for the only scan-passed env; shown when multiple non-testing envs exist.
- Actions: deleting the only scan-passed env errors; deleting a failed env with a
  passed sibling works; deleting a passed env with a passed sibling works.
- `pnpm run checks` and the unit suite pass locally.

[OTTER-527]: https://openstax.atlassian.net/browse/OTTER-527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OTTER-525]: https://openstax.atlassian.net/browse/OTTER-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/admin/settings/code-envs.actions.ts
+++ b/src/app/[orgSlug]/admin/settings/code-envs.actions.ts
@@ -383,18 +383,47 @@ export const deleteOrgCodeEnvAction = new Action('deleteOrgCodeEnvAction', { per
     .requireAbilityTo('update', 'Org')
     .handler(async ({ params: { orgSlug }, db, ...codeEnv }) => {
         if (!codeEnv.isTesting) {
-            const nonTestingForLanguage = await db
+            const otherNonTesting = await db
                 .selectFrom('orgCodeEnv')
-                .select(({ fn }) => [fn.count<number>('id').as('count')])
-                .where('orgId', '=', codeEnv.orgId)
-                .where('language', '=', codeEnv.language)
-                .where('isTesting', '=', false)
-                .where('id', '!=', codeEnv.id)
-                .executeTakeFirstOrThrow()
+                .select((eb) => [
+                    'orgCodeEnv.id',
+                    eb
+                        .selectFrom('codeScan')
+                        .select('codeScan.status')
+                        .whereRef('codeScan.codeEnvId', '=', 'orgCodeEnv.id')
+                        .orderBy('codeScan.createdAt', 'desc')
+                        .limit(1)
+                        .as('latestScanStatus'),
+                ])
+                .where('orgCodeEnv.orgId', '=', codeEnv.orgId)
+                .where('orgCodeEnv.language', '=', codeEnv.language)
+                .where('orgCodeEnv.isTesting', '=', false)
+                .where('orgCodeEnv.id', '!=', codeEnv.id)
+                .execute()
 
-            if (Number(nonTestingForLanguage.count) === 0) {
+            if (otherNonTesting.length === 0) {
                 throw new Error(
                     `Cannot delete the last non-testing ${codeEnv.language} code environment. At least one non-testing code environment must exist for each language.`,
+                )
+            }
+
+            // OTTER-527: an env whose latest scan passed must not be deleted unless another
+            // non-testing env for the language also passed, otherwise the language's default
+            // image would fall back to one that failed (or never finished) scanning.
+            const latestScan = await db
+                .selectFrom('codeScan')
+                .select('status')
+                .where('codeEnvId', '=', codeEnv.id)
+                .orderBy('createdAt', 'desc')
+                .limit(1)
+                .executeTakeFirst()
+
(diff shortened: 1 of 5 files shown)


jbwilson8 (commented):
Restores the dropped per-language guard and adds the scan-passed rule cleanly. I verified the client button-visibility and the server action agree (both derive the latest scan by `createdAt desc, limit 1`), so they won't diverge. Good test coverage of both layers.

**🔴 Hard blockers:** none.

**🟡 Non-blocking:**

1. **Concurrent-delete race** ([code-envs.actions.ts:424](https://github.com/safeinsights/management-app/blob/a03989d61e5da9ca3c63bdb1e064493822e45313/src/app/%5BorgSlug%5D/admin/settings/code-envs.actions.ts#L424)) — the guard checks "does another passed env exist?" and then deletes, with no lock between the two steps, so a concurrent delete can invalidate the check. Example: an org has exactly two passed R envs, A and B, and two deletes fire at nearly the same time. Request 1 checks A, sees B still there → allowed. Request 2 checks B, sees A still there → allowed. Both delete → zero passed envs remain, the exact state this guard exists to prevent. Low probability (needs two near-simultaneous deletes), but wrapping the check + delete in one transaction (or serializable isolation) would close it.

2. **UX: hidden vs. disabled** ([code-envs.tsx:252](https://github.com/safeinsights/management-app/blob/a03989d61e5da9ca3c63bdb1e064493822e45313/src/app/%5BorgSlug%5D/admin/settings/code-envs.tsx#L252)) — hiding the delete icon gives the admin no reason why; a disabled icon with an explanatory tooltip may be clearer. Subjective.

jbwilson8 (approved):

src/app/[orgSlug]/admin/settings/code-envs.actions.ts
```diff
@@ -383,20 +383,49 @@ export const deleteOrgCodeEnvAction = new Action('deleteOrgCodeEnvAction', { per
...
+            // OTTER-527: an env whose latest scan passed must not be deleted unless another
+            // non-testing env for the language also passed, otherwise the language's default
+            // image would fall back to one that failed (or never finished) scanning.
+            const latestScan = await db
+                .selectFrom('codeScan')
+                .select('status')
+                .where('codeEnvId', '=', codeEnv.id)
+                .orderBy('createdAt', 'desc')
+                .limit(1)
+                .executeTakeFirst()
```
Comment (nathanstitt) [inline src/app/[orgSlug]/admin/settings/code-envs.actions.ts]: The `otherNonTesting` query above already runs this exact correlated subquery (`latestScanStatus` per env) — this second round trip re-derives the same thing for just the target env. We should modify the first query to drop the `id != codeEnv.id` filter and then partition the rows in JS

nathanstitt (approved):
looks good, but consider if we should use a single query

src/app/[orgSlug]/admin/settings/code-envs.actions.ts
Comment (Rod0352) [reply src/app/[orgSlug]/admin/settings/code-envs.actions.ts]: adjusted in [168f546](https://github.com/safeinsights/management-app/pull/904/commits/168f546cdb1b74cc51927e8a47ffd907d54e4167)

PR #899, opened 2026-07-16
 OTTER-639: Streamline invitation acceptance and add join confirmation banner

stetradis:
This pull request introduces a new feature that displays a one-time confirmation banner when a user successfully joins an organization. The banner appears on relevant dashboards after onboarding and is managed via session storage to ensure it is shown only once. The implementation includes new components, supporting utilities, tests, and updates to onboarding flows.

**New "Joined Organization" Banner Feature:**

- Added a `JoinedOrgBanner` component that checks session storage for a flag indicating the user has just joined an organization, displays a success banner with the org name, and clears the flag after display. Includes a reveal delay to avoid transient displays during onboarding redirects.
- Integrated `JoinedOrgBanner` into both the organization dashboard and user dashboard pages so users see the confirmation after joining. 

**Onboarding Flow Updates:**

- Updated the invitation acceptance and sign-up flows to set the "just joined org" flag in session storage by calling `markOrgJoined(orgName)` after a successful join, ensuring the banner displays on the user's first dashboard visit. 
- Improved the invitation acceptance logic to handle signed-in users more robustly, redirecting or showing the sign-out panel as appropriate.

**Supporting Utilities and Testing:**

- Added `markOrgJoined` utility and storage key constant to `joined-org.ts` for managing the session storage flag. 
- Created comprehensive unit tests for the `JoinedOrgBanner` component, covering correct display, dismissal, and persistence across remounts.
- Added an end-to-end test to verify the banner appears once after joining an organization and does not repeat on reload. 

**Sign-out Behavior Improvements:**

- Refactored `useSignOut` to use soft navigation and clear caches after sign-out, improving user experience and preventing stale data issues/intermediate flashes.

PR #906, opened 2026-07-21
OTTER-671: always land users on their dashboard after sign-in

Rod0352:
split by mechanism per card discussion:

- App-initiated logout (Sign Out button, inactivity auto-logout): no
  redirect_url is captured; next sign-in lands on /dashboard
- Direct link while signed out (shared study link, bookmark): unchanged —
  proxy still captures the destination and sign-in returns the user there
- Post-login fallbacks default to /dashboard instead of the root stub page

Tests: new use-sign-out.test.ts, new sign-in-form dashboard-fallback test.
proxy.test.ts "metadata re-sync" fails on pristine main in a clean sandbox
too — pre-existing.


jbwilson8 (commented):
Clean, focused change with good test coverage, and it actually reduces open-redirect surface. Verified all three `useSignOut` call sites (invite panel, inactivity, navbar) behave correctly under the new default.

**🔴 Hard blockers:** none.

**🟡 Non-blocking — confirm before merge:**

1. **Conflicts with PR #899 on `use-sign-out.ts`** ([use-sign-out.ts:16](https://github.com/safeinsights/management-app/blob/bff1ea927e8bf3c1fab3a286dd294fe7232ce078/src/hooks/use-sign-out.ts#L16)) — both open PRs rewrite this file with opposing intent: #899 keeps capturing the current page into `redirect_url`, this PR removes it. Coordinate merge order; whoever's second must keep #899's soft-nav while dropping the capture.

2. **Confirm the mid-task/deep-link trade-off is intended** ([proxy.ts:97](https://github.com/safeinsights/management-app/blob/bff1ea927e8bf3c1fab3a286dd294fe7232ce078/src/proxy.ts#L97)) — a user logged out by inactivity mid-study, or opening a shared deep link while logged out, now lands on `/dashboard` and loses their place. Called out in the ticket as desired — just flagging for explicit product sign-off.

jbwilson8 (approved):

nathanstitt:
@Rod0352 I'm going to move discussion onto the card, but do not merge until it's resolved please.

Rod0352:
> @Rod0352 I'm going to move discussion onto the card, but do not merge until it's resolved please.

Iris signed off on splitting by mechanism, and
543c1c6 implements that. Your example works unchanged. a signed-out
user opening a shared study link still lands on that study after sign-in. Only
app-initiated logout (sign-out button, inactivity timeout) goes to My Dashboard.

PR #907, opened 2026-07-21
Otter 590 update study code section

jbwilson8:
[Issue](https://openstax.atlassian.net/browse/OTTER-590?focusedCommentId=45203): fix study-code link styling

## Summary

Corrects the styling of the "View approved initial request" link in the study-code
section of the researcher view page, and extracts a reusable `LinkWithIcon`
component so the style lives in one place instead of being re-hardcoded per link.

Previously the link rendered at `fw={700}` in the theme's default link color
(purple). Per design it should be **14px, semi-bold (600), `#0058BD`**.

## What changed

- **New `LinkWithIcon` component** in [links.tsx](src/components/links.tsx),
  alongside the existing `Link`/`ButtonLink`. It renders a text link with an icon
  next to it, with:
  - dynamic text (`children`) and a dynamic `icon` prop
  - `iconPosition` (`'leading' | 'trailing'`, default `trailing`)
  - the standard link style baked in as **overridable** defaults —
    `c="blue.7"` (`#0058BD`), `fz="sm"` (14px), `fw={600}` — using theme tokens
    rather than hardcoded values
  - native anchor attributes (`href`, `target`, `rel`, `data-testid`) forwarded
    via Mantine's `ElementProps`
- **Repointed the two researcher view-page links** to `LinkWithIcon`
  ([code-post-decision-view.tsx](src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx)
  and [code-post-submission-view.tsx](src/app/[orgSlug]/study/[studyId]/view/code-post-submission-view.tsx)),
  which also fixes the color/weight in both the post-submission and
  change-requested states.

## Why

`#0058BD` already exists in the theme as `blue[7]` and 14px is the `sm` token, so
the values now reference the theme (leaning into the team's theme
standardization). The one value without a theme primitive — weight `600` — is now
defined once, in `LinkWithIcon`, instead of being hardcoded at each call site.
The same icon+text link shape is repeated across the app, so the component is
available for those other links (reviewer side, proposal form, edit-and-resubmit)
in follow-up work.

## Notes / out of scope

- Only the two researcher view-page links are migrated here; the other
  candidate links are left untouched.
- The `LinkWithIcon` default style is overridable, so callers needing a different
  weight/size/color can still pass props.

## Screenshots

| Before | After |
| --- | --- |
| Screenshot 2026-07-21 at 5 13 46 PM | Screenshot 2026-07-21 at 5 16 16 PM |

### DIFF
--- a/src/components/links.tsx
+++ b/src/components/links.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { Anchor as MantineAnchor, AnchorProps, Button, ButtonProps } from '@mantine/core'
+import { Anchor as MantineAnchor, AnchorProps, ElementProps, Button, ButtonProps } from '@mantine/core'
 import { FC, ReactNode } from 'react'
 import NextLink from 'next/link'
 import type { Route } from 'next'
@@ -14,6 +14,28 @@ export type LinkProps = AnchorProps & {
 export const Link: FC<LinkProps> = ({ href, target, children, ...anchorProps }) => (
     <MantineAnchor component={NextLink} href={href} target={target} {...anchorProps}>
         {children}
+    </MantineAnchor>
+)
+
+export type LinkWithIconProps = AnchorProps &
+    ElementProps<'a', keyof AnchorProps> & {
+        icon: ReactNode
+        iconPosition?: 'leading' | 'trailing'
+        children: ReactNode
+    }
+
+export const LinkWithIcon: FC<LinkWithIconProps> = ({ icon, iconPosition = 'trailing', children, ...anchorProps }) => (
+    <MantineAnchor
+        c="blue.7"
+        fz="sm"
+        fw={600}
+        display="inline-flex"
+        style={{ alignItems: 'center', gap: 4, whiteSpace: 'nowrap', flexShrink: 0 }}
+        {...anchorProps}
+    >
+        {iconPosition === 'leading' && icon}
+        {children}
+        {iconPosition === 'trailing' && icon}
     </MantineAnchor>
 )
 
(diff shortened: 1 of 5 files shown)


src/components/links.tsx
```diff
@@ -17,6 +17,28 @@ export const Link: FC<LinkProps> = ({ href, target, children, ...anchorProps })
...
+        children: ReactNode
+    }
+
+export const LinkWithIcon: FC<LinkWithIconProps> = ({ icon, iconPosition = 'trailing', children, ...anchorProps }) => (
+    <MantineAnchor
+        c="blue.7"
+        fz="sm"
+        fw={600}
+        display="inline-flex"
+        style={{ alignItems: 'center', gap: 4, whiteSpace: 'nowrap', flexShrink: 0 }}
```
nathanstitt: Since `{...anchorProps}` is spread *after* this `style`, a caller passing their own `style` prop replaces this whole object rather than merging into it — so they'd silently lose `alignItems`/`gap`/`whiteSpace`/`flexShrink` and the icon+text layout would break. The PR frames the defaults as "overridable", which is true for `c`/`fz`/`fw` (scalar props), but `style` and `display` are the two that don't compose cleanly. If that matters, we could merge instead: `style={{ ...baseStyle, ...anchorProps.style }}`. Probably fine for the current call sites, just a sharp edge for the reuse this component is meant to enable.

nathanstitt (approved):
looks good, I think it's fine not to do the style pointed out below but left in case you think it's worthwhile

PR #908, opened 2026-07-23
security: resolve pnpm audit vulnerabilities (next, hono, body-parser)

nathanstitt:
Resolves all 14 findings from `pnpm audit` (4 high, 9 moderate, 1 low). `pnpm audit` reports **no known vulnerabilities**, and the app builds and unit tests pass.

## Direct dependency fix

- **next → `^16.2.11`** — patches 4 high + 5 moderate advisories: Turbopack middleware/proxy bypass, Server Actions DoS, SSRF in Server Actions on custom servers, SSRF in rewrites via attacker-controlled destination hostname, plus cache-confusion and image-optimization DoS moderates.

## Removed Panda CSS (eliminates the remaining advisories at the source)

The `hono`, `@hono/node-server`, and `body-parser` advisories all came in **exclusively** through `@pandacss/dev → @pandacss/mcp → @modelcontextprotocol/sdk`. Panda turned out to be used in only two files (the root page and its loading skeleton) for simple layout wrappers, while the app styles with Mantine.

Rather than pin those dev-only transitives, this PR **removes Panda entirely**:
- Rewrote `src/app/(root)/page.tsx` and `loading.tsx` to use Mantine `Flex`/`Stack`.
- Deleted `panda.config.ts`, `postcss.config.cjs`, `src/styles/`, the `@pandacss/dev` devDep, and the `prepare: panda codegen` script.
- Deleted the `hono` / `@hono/node-server` / `body-parser` overrides — the vulnerable packages are no longer in the tree.

## Review feedback (thanks @therealmarv)

- **Babel 8 / Node 22 compatibility** — the `pnpm update` had incidentally pulled `@babel/core@8` (engine `^22.18.0 || >=24.11.0`, narrower than this repo's `node >=22`) into Vitest's Istanbul coverage path. Added a parent-specific override pinning `@vitest/coverage-istanbul>@babel/core` back to `7.29.7`, so Node 22.0–22.17 stay supported.
- **Unbounded hono override major range** — moot; the hono/node-server overrides are gone with Panda.

Verified: `pnpm run checks` (typecheck + lint + action validation), `pnpm run build`, and `pnpm run test:unit` (1978 passing) all pass.


pnpm-lock.yaml
```diff
@@ -9829,67 +10043,65 @@ snapshots:
...
       react-refresh: 0.18.0
-      vite: 8.0.16(@types/node@25.9.1)(esbuild@0.28.1)(jiti@2.7.0)(terser@5.48.0)(tsx@4.21.0)(yaml@2.9.0)
+      vite: 8.0.16(@types/node@25.9.1)(esbuild@0.28.1)(jiti@2.7.0)(terser@5.49.0)(tsx@4.21.0)(yaml@2.9.0)
     transitivePeerDependencies:
       - supports-color
 
   '@vitest/coverage-istanbul@4.1.6(vitest@4.1.6)':
     dependencies:
-      '@babel/core': 7.29.7
+      '@babel/core': 8.0.1
```
therealmarv: This unrelated lockfile re-resolution moves Vitest's Istanbul coverage path from Babel 7 to Babel 8. `@babel/core@8.0.1` declares Node `^22.18.0 || >=24.11.0`, while this repository currently advertises `node >=22`. CI uses the rolling Node 22 release, so it passes without testing Node 22.0 through 22.17.

If support for all Node 22 releases is intentional, this can stay scoped to Vitest with a parent-specific override such as `'@vitest/coverage-istanbul>@babel/core': 7.29.7`; the previous lockfile already resolved that compatible version. pnpm's parent-specific override syntax is documented here: https://pnpm.io/settings#overrides. Alternatively, the repository's declared Node minimum should be raised to `>=22.18.0`.

pnpm-workspace.yaml
```diff
@@ -14,16 +14,20 @@ overrides:
...
     minimatch: ^10.2.1
     picomatch: ^4.0.4
     postcss: ^8.5.10
     vite: ^8.0.16
     ws: ^8.21.0
-    'hono@<4.12.25': '>=4.12.25'
+    'hono@<4.12.27': '>=4.12.27'
+    # @hono/node-server <2.0.5: path traversal in serve-static on Windows via encoded backslash
+    # (%5C). Same dev-only @pandacss/mcp chain; Trivy scans dev deps.
+    '@hono/node-server@<2.0.5': '>=2.0.5'
```
therealmarv: Looking at repository usage, I do not see any `panda mcp`, `panda init-mcp`, MCP client configuration, or direct `@pandacss/mcp` use. The package is installed because the Panda CLI depends on it, but Panda dynamically imports it only for the MCP-specific commands; normal `panda codegen` and application builds use separate Panda packages.

That makes this advisory effectively unreachable here: the unused Panda MCP path reaches `@modelcontextprotocol/sdk`, which imports `getRequestListener`, while the vulnerability is in `serveStatic`. The upstream issue documents that exact reachability and version conflict: https://github.com/honojs/node-server/issues/381

The versioning risk still remains because this override forces the SDK from its declared `^1.19.9` dependency onto the 2.x major, and the replacement range is open to future majors. Since Panda MCP is not used, I think the safer temporary fix is to remove this override and add a documented `auditConfig.ignoreGhsas` entry for `GHSA-frvp-7c67-39w9`, then remove the exception when a patched 1.x release or compatible SDK update is available. pnpm documents GHSA exceptions here: https://pnpm.io/cli/audit#--ignore.

If project policy requires installing a patched version instead of accepting this unreachable advisory, bounding the override to `<3` would at least avoid an unexpected future major, but it would still be an unsupported dependency override for functionality this repository does not use.

therealmarv (commented):
Overall this PR looks okay and the security updates are well scoped, but I found two dependency-resolution risks during code review: the Hono override crosses the MCP SDK's declared major-version boundary, and the lockfile refresh introduces Babel 8 with a narrower Node 22 compatibility range than the repository declares. I left inline comments with potential fixes and upstream references. Neither appears to affect the production application path directly, but both are worth considering before merge.

pnpm-lock.yaml
nathanstitt: Good catch — fixed in 681dd09. Added a parent-specific override `'@vitest/coverage-istanbul>@babel/core': 7.29.7` per your suggestion, so the coverage path stays on Babel 7 and Node 22.0–22.17 remain supported. The lockfile no longer references `@babel/core@8`.

pnpm-workspace.yaml
nathanstitt: You're right that this path is unreachable, and thanks for the detailed writeup. Rather than pin (or ignore) a dev-only advisory in a package we don't actually use, I removed Panda entirely in 681dd09 — it was only used in two files for layout wrappers, and the app styles with Mantine. That drops the whole `@pandacss/mcp → @modelcontextprotocol/sdk` tree, so `hono`, `@hono/node-server`, and `body-parser` are no longer in the dependency graph and their overrides are deleted. No unsupported major-version override remains.

nathanstitt:
Resolves the following SafeInsights InfoSec (SIINFOSEC) Trivy findings, all now marked Done:

**next 16.2.6 → 16.2.11** (commit c3643967)
- SIINFOSEC-1147, SIINFOSEC-1148, SIINFOSEC-1149, SIINFOSEC-1150 (HIGH)
- SIINFOSEC-1151, SIINFOSEC-1152, SIINFOSEC-1153, SIINFOSEC-1154, SIINFOSEC-1155 (MEDIUM)

**Panda removal — drops hono / @hono/node-server / body-parser from the tree** (commit 681dd095)
- SIINFOSEC-1139, SIINFOSEC-1140, SIINFOSEC-1141 (hono, MEDIUM)
- SIINFOSEC-1138 (@hono/node-server, MEDIUM)
- SIINFOSEC-1128 (body-parser, LOW)

Jira board: https://openstax.atlassian.net/jira/software/projects/SIINFOSEC/boards/160

therealmarv (approved):
thanks, approved!

PR #909, opened 2026-07-23
OTTER-666: Reviewer outputs-pending screen + standardized study headers

stetradis:
## Summary

Adds a reviewer-facing **"outputs pending"** screen for the window after code is approved and executing in the secure enclave, but before results are available.

While the job runs, reviewers now see the study's current execution stage (`packaging` / `queued` / `provisioning` / `running`) with contextual guidance.

Also standardizes study screen headers, layout spacing, and date formatting across the researcher/reviewer study pages.

### New `reviewer-outputs-pending` Screen

* Adds a new `ScreenId`: `reviewer-outputs-pending`.
* Registers it in the screen registry and wires it into `REVIEWER_SCREEN_RULES`.
* Sits just below `reviewer-study-results` (results outrank it) and above `reviewer-code-feedback`.
* Gated on `isExecuting`.
* `ReviewerOutputsPendingScreen` resolves the latest submitted job and its current execution stage, then renders `SecondaryAnalysisView`.
* `currentExecutionStage()` in `study-job-status.ts` returns the most recently recorded `JOB-*` running status and when it started, or `null`.

### New Components

* **`SecondaryAnalysisView`**
  "Secondary analysis study / Review outputs" step layout with previous-step and back-to-studies navigation.

* **`OutputsStatusAlert`**
  Stage-specific copy for each job status, including:

  * A live-updating timestamp (`"5 minutes ago"` within 24 hours).
  * An absolute date after 24 hours.
  * A Slack contact link for stalled jobs.

* **`StatusAlert`**
  Shared informative/action alert primitive.

* **`StudyPageHeader`**
  Shared page title using `order={1}` and `fz={40}`.

### Header / Layout / Date Standardization

* Study screens now use `StudyPageHeader` instead of ad hoc `<Title>` markup (e.g. `study-overview`, `code-review`, etc.).
* Standardizes `xxl` stack spacing across study screens.
* Standardizes proposal-step header divider styling.
* Standardizes date formatting across researcher and reviewer study pages.

## Testing

* Unit: new screen rule; execution-stage resolution; StatusAlert; OutputsStatusAlert (elapsed + absolute date formatting); StudyPageHeader.
* Integration (review/page.test.tsx): review page resolves code-approved + executing &rarr; SecondaryAnalysisView; SI-admin no-code proposal seed corrected.
* E2E (study-flow.spec.ts): reviewer approve-code flow now lands on the outputs-pending screen.


src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx
```diff
@@ -0,0 +1,93 @@
...
+type OutputsStatusAlertProps = {
+    stageStatus: StudyJobStatus
+    startedAt: string | Date
+}
+
+function useStartedWhen(startedAt: string | Date): string {
+    const startedAtMs = new Date(startedAt).getTime()
+    const [nowMs, setNowMs] = useState(() => Date.now())
+    useEffect(() => {
+        const id = setInterval(() => setNowMs(Date.now()), 60_000)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx]: we have a useTimer hook that's probably better to use here.  it has a isEnabled flag.

We should disable the timer once we've crossed `DAY_MINUTES`, no point in re-checking after that pint is there?

src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <>
+                Preparing the code to run in the secure enclave. If it stays in this status for over 1 hour, contact
+                SafeInsights via <SlackLink />
+            </>
+        ),
+    },
+    'JOB-READY': {
+        variant: STATUS_ALERT_VARIANT.informative,
+        title: (when) => `Outputs not ready, code queued ${when}`,
+        body: 'The code is packaged and ready to be picked up by the secure enclave. If it stays in this status for over 1 hour, contact your organization admin.',
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx:56]: Small copy inconsistency worth a second look with design: `JOB-PACKAGING` tells the reviewer to contact **SafeInsights via Slack**, while the other three all say **contact your organization admin**. Given the reviewer here *is* the org's data partner, "contact your organization admin" reads a little oddly for them — is the split intentional, or should all four route to the same place?

src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx
```diff
@@ -0,0 +1,93 @@
+'use client'
+
+import { useEffect, useState, type ReactNode } from 'react'
+import dayjs from 'dayjs'
+import { Anchor } from '@mantine/core'
+import { ArrowSquareOutIcon } from '@phosphor-icons/react/dist/ssr'
+import type { StudyJobStatus } from '@/database/types'
+import { StatusAlert, STATUS_ALERT_VARIANT, type StatusAlertVariant } from '@/components/study/status-alert'
+
+const SAFE_INSIGHTS_SLACK_URL = 'https://openstax.slack.com/archives/C081BN1R8CE'
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx]: This should go into lib/config

src/lib/study-job-status.ts
```diff
@@ -24,6 +24,16 @@ export const STUDY_CODE_RUNNING_JOB_STATUSES: readonly StudyJobStatus[] = [
     'JOB-RUNNING',
 ]
 
+// The most recently recorded JOB-* running status, or null if none — drives the outputs view's stage alert.
+export function currentExecutionStage(
+    statusChanges: ReadonlyArray<{ status: StudyJobStatus; createdAt: Date | string }>,
+): { status: StudyJobStatus; startedAt: Date | string } | null {
+    const stages = statusChanges.filter((c) => STUDY_CODE_RUNNING_JOB_STATUSES.includes(c.status))
+    if (stages.length === 0) return null
+    const latest = stages.reduce((a, b) => (new Date(b.createdAt).getTime() > new Date(a.createdAt).getTime() ? b : a))
```
nathanstitt: The reduce picks the max by `createdAt` timestamp, but keeps the *first* element on ties (`>` is strict). Job status rows are written in quick succession by the enclave — `JOB-READY` → `JOB-PROVISIONING` → `JOB-RUNNING` can land inside the same millisecond, and `timestamptz` at ms resolution will then report equal `createdAt`s. In that case we'd show the earlier stage's copy ("code queued") while the job is actually running.

Since the rows carry v7 ids (time-ordered), a tie-break on `id` — or just taking the last matching element in insertion order, which the query already returns — would make this deterministic. Are the `statusChanges` from `latestSubmittedJobForStudy` ordered? If so, a `findLast` over the filtered list is both simpler and tie-safe.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-pending-screen.tsx
```diff
@@ -0,0 +1,32 @@
...
+    orgSlug,
+}: Pick<ScreenComponentProps, 'study' | 'orgSlug'>) {
+    if (!isSubmittedStudy(study)) {
+        return <AlertNotFound title="Study was not found" message="No such study exists" />
+    }
+
+    const job = await latestSubmittedJobForStudy(study.id)
+    const stage = job ? currentExecutionStage(job.statusChanges) : null
+    if (!stage) {
+        return <AlertNotFound title="No submission found" message="This study has no submitted code to review." />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-pending-screen.tsx:19]: This will show "No submission found" message when (a) there's no submitted job at all, and (b) there *is* a job but `currentExecutionStage` returned `null`.

Should we have 2 different messages?

src/app/[orgSlug]/study/[studyId]/review/secondary-analysis-view.tsx
```diff
@@ -0,0 +1,46 @@
...
+    stageStartedAt,
+    previousHref,
+    dashboardHref,
+}: SecondaryAnalysisViewProps) {
+    return (
+        <Box bg="grey.10">
+            <Stack px="xl" gap="xxl" py="xl">
+                <StudyPageHeader>Secondary analysis study</StudyPageHeader>
+                <ProposalStepHeader
+                    stepLabel="STEP 3"
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/secondary-analysis-view.tsx:30]: `stepLabel="STEP 3"` is hardcoded, matching `code-review.tsx:110` which also uses STEP 3 for the code review step. So the reviewer goes STEP 3 (review code) → STEP 3 (review outputs), which is a little confusing as a wayfinding cue. Is outputs meant to be step 4, or is the repetition deliberate because both live under the same "review" phase?

src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx
```diff
@@ -0,0 +1,93 @@
...
+    if (totalMinutes < DAY_MINUTES) return `${formatElapsed(startedAtMs, nowMs)} ago`
+    return `on ${dayjs(startedAtMs).format('MMM DD, YYYY [at] h:mm A')}`
+}
+
+const SlackLink = () => (
+    <Anchor href={SAFE_INSIGHTS_SLACK_URL} target="_blank" rel="noopener noreferrer" c="blue.7" inherit>
+        Slack
+        <ArrowSquareOutIcon size={16} style={{ marginLeft: 4, verticalAlign: 'text-bottom' }} />
+    </Anchor>
+)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx:38]: maybe lets call this SlackDPSupportLink so it's clear that it's for supporting DPs, not general slack

nathanstitt (approved):
Looks great, ty.  Comments are minor just a bit of code reuse suggestions

src/app/[orgSlug]/study/[studyId]/review/secondary-analysis-view.tsx
Comment (stetradis) [reply src/app/[orgSlug]/study/[studyId]/review/secondary-analysis-view.tsx:30]: I believe this is deliberate based on this [comment](https://openstax.atlassian.net/browse/OTTER-666?focusedCommentId=45249) - when the agreements page is removed (2A, 2B, 2C), I imagine that we would renumber code review to Step 2 instead of Step 3

src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx
Comment (stetradis) [reply src/app/[orgSlug]/study/[studyId]/review/outputs-status-alert.tsx:56]: This matches the AC, but I'll confirm with Iris

nathanstitt:
**QA validation note (OTTER-666) — gap between `CODE-APPROVED` and the first `JOB-*` status**

Validating this on the PR preview (`pr909.qa.safeinsights.org`), I walked a fresh study all the way through: proposal approved → agreements → code submitted → **Approve and run code**.

After the code-approval confirmation, the reviewer lands on the **code post-feedback** screen ("This study code has been approved. You will be notified when the study results are available for review."), not the new *Secondary analysis study / Review outputs* screen. It stays there across reloads, and the DP dashboard shows the study as **Code approved**.

That behavior matches the code as written — it isn't a routing bug:

- `REVIEWER_SCREEN_RULES` gates `reviewer-outputs-pending` on `s.isExecuting`.
- `isExecuting` is driven by `STUDY_CODE_RUNNING_JOB_STATUSES` = `JOB-PACKAGING` / `JOB-READY` / `JOB-PROVISIONING` / `JOB-RUNNING`.
- Immediately after approval the job is at `CODE-APPROVED`, which is in none of those sets, so rule #3 (`reviewer-code-feedback`) wins.

So there's a window — from the moment the DP approves the code until the enclave records its first `JOB-*` status — where the outputs-pending screen does not render. On this preview env the enclave never picked the job up at all, so that window was effectively unbounded and the new screen could not be reached through the UI.

**Question:** should `reviewer-outputs-pending` also cover `CODE-APPROVED` (i.e. treat "approved but not yet packaged" as the first stage of the outputs view), or is landing on the code-feedback screen during that gap the intended behavior?

The ticket's alert copy starts at *Job Packaging*, so `CODE-APPROVED` may simply be out of scope by design — but from the DP's perspective it reads as a dead spot right after the action that's supposed to start the run, which seems to cut against the card's stated goal of "surfacing the different stages a code goes through after being approved."

Worth noting the acceptance criteria for the four `JOB-*` alert states could not be exercised end-to-end here, since no job reached an executing status on this environment.

PR #910, opened 2026-07-24
SHRIMP-271: Possibility to refresh environment variables on restart if staled

sparksam:


chrisbendel:
couple things before this goes in:

- the restart in `restartWorkspaceIfParametersStale` can sit waiting on the stop build for up to 5 min (100 polls × 3s), and that runs inside `ensureWorkspaceAction` which the client is awaiting the whole time. kind of defeats the point of the status polling — can we fire the stop and let the poller kick off the start once it's down, or at least drop that ceiling way down?

- `startWorkspace` swallows its errors (warns and returns). that was ok when it was best-effort, but now if the start fails right after we've already stopped the workspace, we've killed it and still return success — the user just polls a stopped workspace forever. the restart path probably needs the start failure to actually throw.

- restarting a `running` workspace blows away whatever the researcher has open in it. is that what we want on every relaunch where the env drifted? there's no guard for one that's actively in use.

also worth confirming the deploy order with iac#190 — once we start sending container_image/environment on every start build, those have to already be mutable in the active template version or coder rejects the build. so that template needs to land first.

smaller stuff, take it or leave it: `mutableRichParameterValues(codeEnv)` gets computed in both branches, could hoist it out. and the environment staleness check is a raw JSON string compare, so it's sensitive to key order — fine today since it's the same serializer on both sides, just means it'll false-trigger a restart if that ordering ever changes.

src/server/coder/workspaces.ts
```diff
@@ -237,60 +242,117 @@ async function buildWorkspaceEnvironment(codeEnv: Awaited<ReturnType<typeof fetc
...
+    ]
+}
+
+// Whether any parameter we manage differs from what the workspace's latest build ran with.
+function parametersStale(current: CoderBuildParameter[], desired: CoderBuildParameter[]): boolean {
+    return desired.some((d) => current.find((c) => c.name === d.name)?.value !== d.value)
+}
+
+const STOP_POLL_INTERVAL_MS = 3_000
+const STOP_POLL_MAX_ATTEMPTS = 100
```
nathanstitt, line 261: 100 attempts × 3s means `waitForBuildCompletion` can block for up to ~5 minutes. The thing that worries me is where this runs: `restartWorkspaceIfParametersStale` is reached from `getOrCreateCoderWorkspace` → `createUserAndWorkspace` → `ensureWorkspaceAction`, which is a server action awaited directly by the launch button in `use-workspace-launcher.tsx`. We don't set `maxDuration` anywhere, so on a slow stop we'd likely hit the platform's request timeout and the user gets an opaque "launch failed" — after the stop build already went through, leaving the workspace stopped.

Worth considering whether the stop-then-start should be fire-and-forget instead: issue the stop, return, and let the existing `getCoderWorkspaceLaunchStatus` polling drive the rest. That's the pattern the rest of this flow already uses for long waits. If we do keep the inline wait, a much smaller attempt cap (and a comment about the request budget it's sized against) would be safer.

src/server/coder/workspaces.ts
```diff
@@ -237,60 +242,117 @@ async function buildWorkspaceEnvironment(codeEnv: Awaited<ReturnType<typeof fetc
...
+    if (!parametersStale(current, desired)) return
+
+    logger.info(`${logCtx} parameters changed since last build, restarting workspace ${workspace.id}`)
+    const stopBuild = await coderFetch<CoderWorkspaceBuild>(coderWorkspaceBuildPath(workspace.id), {
+        method: 'POST',
+        body: { transition: 'stop' },
+        errorMessage: 'Failed to stop workspace',
+    })
+    await waitForBuildCompletion(stopBuild.id, logCtx)
+    await startWorkspace(workspace.id, desired)
```
nathanstitt, line 310: `startWorkspace` swallows its own errors — it catches and only `logger.warn`s. That was fine when it was a best-effort nudge for an already-stopped workspace, but here it's the second half of a stop→start pair we deliberately initiated. If this start POST fails, we've stopped a workspace that the user was actively using and then return normally, reporting success. The user's IDE session is gone and nothing surfaces why.

Could we let the restart path see the failure — either by threading a flag through `startWorkspace` or calling `coderFetch` directly here — so the launch reports an error rather than silently leaving the workspace down?

src/server/coder/workspaces.ts
```diff
@@ -237,60 +242,117 @@ async function buildWorkspaceEnvironment(codeEnv: Awaited<ReturnType<typeof fetc
...
+
+    const buildStatus = workspaceData.latest_build?.status
+    if (buildStatus === 'stopped') {
+        logger.warn(`Workspace was stopped`)
+        await startWorkspace(workspaceData.id, await mutableRichParameterValues(codeEnv))
+    } else if (buildStatus === 'running') {
+        await restartWorkspaceIfParametersStale(workspaceData, await mutableRichParameterValues(codeEnv), logCtx)
+    }
+
+    return workspaceData
```
nathanstitt, line 340: After a restart we return the `workspaceData` we fetched *before* the stop/start, so `latest_build` still points at the old build id and status `running`. `getCoderWorkspaceLaunchStatus` re-fetches the workspace on its own so polling should recover, but the value this action returns to the client is describing a build that no longer exists. If nothing downstream reads it we could note that; otherwise a re-fetch before returning would keep the payload honest.

src/server/coder/workspaces.ts
```diff
@@ -237,60 +242,117 @@ async function buildWorkspaceEnvironment(codeEnv: Awaited<ReturnType<typeof fetc
...
+): Promise<CoderBuildParameter[]> {
+    return [
+        { name: 'container_image', value: codeEnv.url },
+        { name: 'environment', value: JSON.stringify(await buildWorkspaceEnvironment(codeEnv)) },
+    ]
+}
+
+// Whether any parameter we manage differs from what the workspace's latest build ran with.
+function parametersStale(current: CoderBuildParameter[], desired: CoderBuildParameter[]): boolean {
+    return desired.some((d) => current.find((c) => c.name === d.name)?.value !== d.value)
```
nathanstitt, line 257: The environment parameter is compared as an opaque JSON string here, which gives false positives in two cases I checked  against the real deps:

* Order changes. codeEnv.settings.environment is spread in first and comes from the DB, so if that ever comes back in a
  different order we read "stale" and restart.
* Extra keys. Coder's parameters endpoint returns richer objects than {name, value} (there's a workspace_build_id among others), so a round-tripped value won't string-match a freshly serialized one.

Since a restart now tears down a workspace someone may be actively using, a false positive costs a live session rather than just a wasted rebuild. remeda is already a dependency (we use it in src/server/clerk.ts and src/server/actions/action.ts) and has what's needed:



```suggestion
      import { isDeepEqual, sortBy } from 'remeda'

  // Compare the environment parameter structurally: it's a JSON array whose order isn't
  // guaranteed by the DB, and an order-only difference would restart a healthy workspace
  // out from under whoever is using it.
  function parameterValuesMatch(current: string | undefined, desired: string): boolean {
      if (current === desired) return true
      try {
          const norm = (raw: string) =>
              sortBy(
                  (JSON.parse(raw) as CoderBuildParameter[]).map(({ name, value }) => ({ name, value })),
                  (p) => p.name,
              )
          return isDeepEqual(norm(current ?? ''), norm(desired))
      } catch {
          return false // non-JSON (e.g. container_image) already failed the === above
      }
  }

  function parametersStale(current: CoderBuildParameter[], desired: CoderBuildParameter[]): boolean {
      return desired.some((d) => !parameterValuesMatch(current.find((c) => c.name === d.name)?.value, d.value))
  }

```

src/server/coder/workspaces.ts
```diff
@@ -237,60 +242,117 @@ async function buildWorkspaceEnvironment(codeEnv: Awaited<ReturnType<typeof fetc
...
+    const buildId = workspace.latest_build?.id
+    if (!buildId) return
+
+    let current: CoderBuildParameter[]
+    try {
+        current = await coderFetch<CoderBuildParameter[]>(coderWorkspaceBuildParametersPath(buildId), {
+            errorMessage: 'Failed to get build parameters',
+        })
+    } catch (error) {
+        logger.warn(`${logCtx} could not read build parameters, skipping staleness check:`, error)
```
nathanstitt, line 297: Nice call degrading to the old behavior here rather than failing the launch. One thought: this `catch` also swallows a genuinely broken/renamed parameters endpoint, which would silently disable the whole feature this PR adds — every launch would quietly keep running stale env vars and the only signal is a warn line. Since that's the exact bug we're fixing, it might be worth making this log at `error` so it actually gets noticed.

src/server/coder/workspaces.ts
```diff
@@ -237,60 +242,117 @@ async function buildWorkspaceEnvironment(codeEnv: Awaited<ReturnType<typeof fetc
...
+}
+
+const STOP_POLL_INTERVAL_MS = 3_000
+const STOP_POLL_MAX_ATTEMPTS = 100
+
+// Waits for a build's provisioner job to finish; throws if it fails or doesn't finish in time,
+// since proceeding would make the follow-up start build 409 against the in-progress one.
+async function waitForBuildCompletion(buildId: CoderWorkspaceBuild['id'], logCtx: string): Promise<void> {
+    for (let attempt = 0; attempt < STOP_POLL_MAX_ATTEMPTS; attempt++) {
+        if (attempt > 0) await new Promise((resolve) => setTimeout(resolve, STOP_POLL_INTERVAL_MS))
```
nathanstitt, line 267: Small thing: polling only breaks out on `succeeded`/`failed`/`canceled`. If `build.job` is undefined — which the type allows, `job?` is optional — `jobStatus` is `undefined` and we fall through to the full 100-attempt wait before throwing a timeout. A missing job is arguably an immediate error rather than something to wait 5 minutes on.

src/server/coder.test.ts
```diff
@@ -514,6 +514,126 @@ describe('createUserAndWorkspace', () => {
...
+            const ok = (json: unknown) => Promise.resolve({ ok: true, json: vi.fn().mockResolvedValue(json) })
+            if (url.includes('/users?')) return ok(mockUsersEmailQueryResponse)
+            if (url.includes('/workspace/')) return ok({ id: 'ws-1', latest_build: opts.latestBuild })
+            if (url.includes('/builds') && options?.method === 'POST') {
+                const body = JSON.parse(options.body!)
+                buildPosts.push(body)
+                return ok({ id: `${body.transition}-build-1`, status: 'pending' })
+            }
+            if (url.includes('/parameters')) return ok(opts.currentParams ?? [])
+            if (url.includes('/workspacebuilds/')) return ok({ status: 'stopped', job: { status: 'succeeded' } })
```
nathanstitt, line 566: This catch-all returns `job.status: 'succeeded'` on the very first poll, so `waitForBuildCompletion`'s loop always exits on attempt 0. That means the retry/backoff path, the failed/canceled throw, and the timeout throw are all untested — and it's the part of this PR with the most operational risk (see the note on the attempt cap). A test that returns `running` a couple of times before `succeeded`, plus one that returns `failed` and asserts we throw and never issue the start, would cover the interesting behavior. Worth using fake timers so the 3s sleeps don't make the suite slow.

src/server/coder.test.ts
```diff
@@ -514,6 +514,126 @@ describe('createUserAndWorkspace', () => {
...
+    it('leaves a running workspace alone when its build parameters are current', async () => {
+        const { mock, buildPosts } = routeExistingWorkspace({
+            latestBuild: { id: 'build-0', status: 'running' },
+            currentParams: [{ name: 'study_id', value: 'study123' }, ...refreshedParams()],
+        })
+        ;(global.fetch as unknown as Mock).mockImplementation(mock)
+
+        await createUserAndWorkspace('study123')
+
+        expect(buildPosts).toEqual([])
```
nathanstitt, line 633: Good set of cases. One gap that'd be cheap to add: a running workspace where the parameters fetch *fails*, asserting we leave it alone rather than restarting. That's the documented fallback in `restartWorkspaceIfParametersStale` and it's the behavior most likely to regress if someone later decides the catch should rethrow.

src/server/coder/types.ts
```diff
@@ -76,16 +76,23 @@ export interface CoderAgent {
 }
 
 export interface CoderWorkspaceBuild {
+    id: BuildId
```
nathanstitt, line 79: `CoderWorkspaceBuild.id` is required here while `CoderWorkspace.latest_build.id` below is optional — these describe the same underlying object from two endpoints. The optional one is handled correctly (`if (!buildId) return`), so this is fine, just noting the asymmetry in case it's easier to keep them consistent.

nathanstitt (changes requested):
I'm not seeing a `SHRIMP-271` ticket, is it correct?   I was trying to find the desired functionality but wasn't able to.

I have concerns about this, because it doesn't check if the user is using the workspace.  Currently if they are we'll just shut it down mid-keystroke and restart.  What happens if that occurs?   I'm guessing the coder front-end will throw some kind of error, and will likely not re-attach to the new workspace, leaving them stranded.

Before merging this lets run a through test of the above and document it so PMs can be informed on the side-effects.

chrisbendel:
@nathanstitt  for context the ticket is here: https://openstax.atlassian.net/browse/SHRMP-271

nathanstitt:
> @nathanstitt for context the ticket is here: https://openstax.atlassian.net/browse/SHRMP-271

Thanks, I don't know how I'd missed that!  I'll move discussion to the ticket so PMs can follow it

PR #912, opened 2026-07-26
Add QA endpoints to provision users and create invites

nathanstitt:
Adds two QA endpoints and opens the QA routes up to production.

### `PATCH /api/qa/users/{idOrEmail}`

```json
{
  "orgs": [{ "slug": "openstax", "isAdmin": true }],
  "publicKey": "-----BEGIN PUBLIC KEY-----\nMIIC...",
  "password": "qa-test-1234"
}
```

Fields are optional and omitted ones are untouched. `orgs` is declarative — orgs not listed are removed, `[]` clears all. DB writes run in one transaction; Clerk calls follow the commit. Org changes resync Clerk metadata, since authorization reads memberships from the JWT, not the database. If that resync fails the membership change is rolled back, so a failed request cannot leave the DB and the JWT granting different access.

### `POST /api/qa/invites`

Takes `{ email, orgSlug, isAdmin }` and returns the invite URL, so QA doesn't wait on an inbox. Mirrors `orgAdminInviteUserAction` (lowercasing, 409 when already a member, reuse of an outstanding invite) but skips `onUserInvited`, so no email is sent. A test pins that.

### Production access

The `PROD_ENV` gate is removed; these routes now run everywhere. What limits them instead is `assertQaEmail` — every account they touch must have an email whose local part starts with `qa-`. Anything else gets a 403.

This matters because the operations are destructive: `DELETE` permanently removes DB rows, S3 objects, and the Clerk account; `PATCH` overwrites a real password and public key, and key rotation orphans results wrapped to the old key. So:

- The check reads the **stored** address, not caller input, so passing a real user's id can't bypass it.
- A study has no email, so its **researcher** is checked. The user-facing study delete goes through `deleteStudyCompletely` and is unaffected.
- The dash is required, so `qa@`, `qasim@`, and `qadir@` are all rejected along with `bob-qa@`, `quality@`, and `bob@qa.example.com`. A bare `qa` prefix would classify real given names as QA accounts.

### Blast radius of a user delete

Deleting a QA account removes only the studies it **owns** (`researcher_id`). Studies where it is merely PI or reviewer are **detached** — those columns are nulled — because such a study can belong to a real researcher, and deleting it would destroy their work.

### Audit trail

Every invocation writes to the existing `audit` table, attributed to the acting SI admin and tagged `via: 'qa-api'`. Awaited, not deferred. The password is never recorded, only that one was set.

Destructive routes cannot be transactional across Postgres, S3, and Clerk, so they write an `attempted` row **before** any destructive work and a `succeeded`/`failed` row after. A run that commits the DB delete and then fails against S3 or Clerk therefore still leaves a record that production was touched, along with the error. The target is resolved before the attempt row is written, so a 404 or non-QA target is rejected without polluting the trail.

### Also

- `DELETE /api/qa/users/{id}` accepts an email as well as an id. A segment that is neither a UUID nor an email 404s instead of making Postgres raise on the uuid comparison; the same guard now covers the study route.
- A decodable PEM carrying an unsupported key type (EC, Ed25519) is caller error, so it maps to a 400 rather than surfacing as a 500.
- Extracted `pathForInvitation`, the SPKI key validation, and the QA error-to-status mapping into shared helpers.
- Bumps `brace-expansion` (CVE-2026-14257) and `postcss` (GHSA-r28c-9q8g-f849) past the HIGH advisories Trivy flagged, and restores `@types/node`, without which the crypto call sites fail to typecheck.

### Not included

`org_user` has no unique constraint on `(org_id, user_id)` — the `createIndex` calls in migration `1727379622503` were never awaited. The upsert selects-then-inserts like the rest of the app rather than using `onConflict`. Adding the constraint needs its own migration.

### Testing

`pnpm run test:unit` (2049 passed) and `pnpm run checks` pass. Coverage includes the guard's accept/reject table (with `qasim@`/`qadir@` pinned as rejected), that an internal id can't bypass it, that a non-qa researcher blocks study deletion, that a reviewed-but-not-owned study survives its reviewer's deletion, that the attempt/outcome audit rows are written in order around a failing S3 cleanup, that a failed Clerk sync restores the prior memberships, and that an unsupported key type is a 400.


src/server/qa-cleanup.ts
```diff
@@ -59,14 +76,41 @@ export async function requireQaAdmin(): Promise<QaAuthResult> {
...
+    } else if (UUID_RE.test(identifier)) {
+        user = await query.where('id', '=', identifier).executeTakeFirst()
+    }
+    // A non-uuid, non-email segment never matches: comparing it to the uuid column would
+    // make Postgres raise instead of returning the 404 the caller expects.
+
+    if (!user) throw new QaCleanupNotFoundError(`user ${identifier} not found`)
+    // Checked against the stored address rather than the caller's input, so passing a
+    // real user's id cannot bypass it.
+    assertQaEmail(user.email, `user ${identifier}`)
```
therealmarv, line 110: **Critical:** QA, staging, and production use separate databases, so a call in one environment cannot affect another. Within the environment whose endpoint is called, though, this guard validates only the user being deleted while `deleteUserById` also deletes every study where that user is the PI or reviewer. If the production endpoint deletes a QA account that is assigned to a real study in the production database, that study, its jobs, comments, and files are permanently removed. Please ensure every affected study in the target environment is safe to delete, or avoid cascading deletion for PI/reviewer references.

src/server/qa-cleanup.ts
```diff
@@ -10,21 +17,35 @@
...
+export class QaForbiddenError extends Error {}
+
+/**
+ * The QA routes operate on production, where a stray id would otherwise reach a real
+ * account. Restrict them to addresses whose local part starts with "qa" (qa@, qa-bob@,
+ * QATest@) — the convention QA fixtures follow — so a real user can never be the target.
+ */
+export function assertQaEmail(email: string | null, context: string) {
+    const localPart = (email ?? '').split('@')[0]
+    if (!/^qa/i.test(localPart)) {
```
therealmarv: **High:** `^qa` also matches legitimate names such as `qasim@example.org` and `qadir@example.org`. The databases are isolated by environment, so this cannot cross from QA or staging into production. It can still classify an ordinary account as QA within whichever environment receives the request, including production when the production endpoint is called. Please use an authoritative QA marker or a pattern/domain that cannot overlap with real users in the target environment.

src/app/api/qa/users/[userId]/route.ts
```diff
@@ -9,14 +21,53 @@ export const DELETE = async (_req: Request, { params }: { params: Promise<{ user
...
-        if (error instanceof QaCleanupNotFoundError) {
-            return NextResponse.json({ error: error.message }, { status: 404 })
-        }
-        throw error
+        return qaErrorResponse(error)
     }
 
+    // Audited after the fact: the user row is gone, so this records what was removed
+    // against the admin who removed it.
+    await auditQaInvocation({
```
Comment (therealmarv) [inline src/app/api/qa/users/[userId]/route.ts]: **High:** The audit is written only after DB deletion, S3 cleanup, and Clerk deletion all succeed. These resources are environment-specific, so a QA or staging request cannot alter production. Within the target environment, however, an S3 or Clerk failure after the DB transaction commits leaves changed data without an audit record, and a retry then finds no DB user. This applies to production only when the production endpoint itself is invoked. Please make the invocation durable before destructive work, or otherwise record partial/failing outcomes reliably.

src/server/qa-provision.ts
```diff
@@ -0,0 +1,242 @@
...
+ * sender wrap to a key the owner cannot unwrap.
+ */
+async function applyPublicKey(db: Kysely<DB>, userId: string, pem: string) {
+    let keyData: ArrayBuffer
+    try {
+        keyData = pemToArrayBuffer(pem)
+    } catch {
+        throw new QaInvalidRequestError('publicKey is not a valid PEM encoded key')
+    }
+    await assertValidPublicKey(keyData)
```
therealmarv: **Medium:** A decodable PEM containing a non-RSA or otherwise unsupported SPKI key throws `InvalidPublicKeyError` here. `qaErrorResponse` does not map that error, so invalid caller input produces a 500 rather than a 400. Please translate this validation failure to `QaInvalidRequestError` or handle `InvalidPublicKeyError` in the response mapper, and add coverage for a valid PEM with an unsupported key type.

src/server/qa-provision.ts
```diff
@@ -0,0 +1,242 @@
...
+        }
+        if (update.publicKey) {
+            fingerprint = await applyPublicKey(trx, user.id, update.publicKey)
+        }
+    })
+
+    // Authorization reads org membership and isAdmin from the Clerk JWT's publicMetadata,
+    // not the DB, so a membership change is invisible until the metadata is rewritten.
+    if (resolvedOrgs) {
+        await updateClerkUserMetadata(user.id)
```
therealmarv: **High:** Membership changes commit before this Clerk update. If Clerk is unavailable, the endpoint returns 500 while the DB contains the new memberships and Clerk metadata can continue granting the old permissions; the route also skips its audit entry. Please make this operation recoverable or compensating so a failed request cannot leave authorization state split between the DB and Clerk.

therealmarv (changes requested):
Most of this PR looks solid. The endpoint structure, validation, tests, and shared helper extraction are generally clear and well covered.

QA, staging, and production use separate databases and services, so requests in one environment cannot affect another. The findings below apply only within the environment whose endpoint is called. Because this branch enables these routes in production, the production-specific scenarios require calling the production endpoint against production data.

I found a critical same-environment deletion case and several high-severity failure paths that need to be addressed before merge. In particular, deleting a QA account can remove a non-QA study in the same database when that account is only its PI or reviewer, and the email-prefix guard can classify legitimate users as QA. There are also partial-failure paths where data or authorization state changes in the target environment without the promised audit record.

I left severity-marked comments on the relevant diff lines. The public-key error mapping is less severe, but should also be corrected so invalid client input does not surface as a 500.

src/server/qa-cleanup.ts
nathanstitt: Fixed. `deleteUserById` now deletes only the studies the QA account actually **owns** (`researcher_id`, the non-nullable column). Studies where it is merely PI or reviewer are **detached** instead — `pi_user_id` and `reviewer_id` are nullable, so those references are set to null and the study survives.

That closes the case you described: a QA reviewer assigned to a real researcher’s study can be deleted without taking the study, its jobs, comments, or files with it.

Covered by a new test that assigns a QA account as both reviewer and PI on a study owned by a real researcher, deletes the QA account, and asserts the study still exists with `researcherId` intact and the other two nulled.

One residual worth flagging: `study_review_comment.author_id` / `study_proposal_comment.author_id` are `ON DELETE RESTRICT`, so the QA account’s own comments still have to be removed even when they sit on a study it does not own. There is no way to drop the user row otherwise without a nullable-author migration, which felt out of scope for this PR.

src/server/qa-cleanup.ts
nathanstitt: Fixed — the guard now requires a `qa-` prefix (`/^qa-/i`), so the dash is part of the marker rather than just the letters.

That rejects exactly the collisions you named: `qasim@` and `qadir@` no longer match, and neither does a bare `qa@`. The QA fixture helper (`qaEmail()`) already generated `qa-<random>@test.com`, so the convention was in place — the pattern was simply looser than the convention.

The test table is now explicit about both sides:

- accepts: `qa-reviewer@`, `QA-Test@`, `qa-@`
- rejects: `qa@`, `qa.bob@`, `QATest@`, `qasim@`, `qadir@`, `bob-qa@`, `quality@`, `q@`, `""`, and `bob@qa.example.com`

src/app/api/qa/users/[userId]/route.ts
Comment (nathanstitt) [reply src/app/api/qa/users/[userId]/route.ts]: Fixed. Since Postgres, S3, and Clerk cannot share a transaction, the audit is now two-phase rather than success-only.

Added `auditQaOperation`, which writes an `attempted` row **before** any destructive work and a `succeeded`/`failed` row after, both under the same `recordId`. The failure row carries the error message. So the case you described — DB transaction commits, S3 or Clerk then fails, request 500s — now leaves both an `attempted` and a `failed` row, and a retry that finds no DB user still has the trail explaining what happened.

The attempt row is deliberately not rolled back on failure: it is the evidence that production was touched.

One ordering detail this forced: the target is resolved **before** the attempt row is written, so a 404 or a non-QA target is rejected without writing anything. That meant splitting `findQaStudy` out of `deleteStudyById` for the study route. While there I gave it the same UUID guard `findQaUser` has — a non-uuid study segment would otherwise have made Postgres raise instead of returning a 404.

Two tests cover it: one asserts the rows are `[attempted, succeeded]` in order on the happy path, the other forces `deleteFolderContents` to reject and asserts `[attempted, failed]` with the S3 error captured in the metadata.

The invites route still writes a single row — it creates a pending record rather than destroying anything, and the invite id only exists once the call returns.

src/server/qa-provision.ts
nathanstitt: Fixed — this now maps to a 400.

I translated it at the source rather than in the response mapper: `applyPublicKey` catches `InvalidPublicKeyError` and rethrows it as `QaInvalidRequestError`, right next to the sibling PEM-decode failure that was already handled that way. That keeps `qaErrorResponse` from needing to know about a crypto-layer error type, and keeps the two "bad key input" cases adjacent.

Added the coverage you asked for: a test generates a real ECDSA P-256 key via WebCrypto, exports it as SPKI, and wraps it in a valid PEM envelope. It decodes cleanly and then fails the RSA-OAEP import, which is precisely the path that used to 500.

src/server/qa-provision.ts
nathanstitt: Fixed by compensation, so a failed request is a no-op rather than one that splits authorization state.

Moving the sync inside the transaction does not work: `updateClerkUserMetadata` recomputes the metadata by reading memberships back through the module-level `db`, so inside the transaction it would not see the uncommitted rows and would write stale metadata. Instead the prior memberships are captured before the write, and if the Clerk sync fails they are restored and the error is rethrown. The DB and the JWT therefore agree on the old state, which is recoverable — the caller can simply retry.

The rollback is best-effort and wrapped in its own try/catch: nothing it does may replace the original error, since that error is what tells the caller the request did not take effect. If the rollback itself fails it is logged, and the `failed` audit row from the change above records that production was left mid-change.

Covered by a test that makes `updateClerkUserMetadata` reject, then asserts the memberships match their pre-call state exactly.

Worth noting that writing that test caught a real bug in my first attempt — I had used `.catch()` on the best-effort resync, and the mock returns `undefined`, so that path would have thrown a `TypeError` and masked the original Clerk error in production. It is a proper try/catch now.

therealmarv (approved):
thanks for the fixes. Approved now!

PR #913, opened 2026-07-27
OTTER-647: highlight required fields when left incomplete on blur

therealmarv:
see [OTTER-647](https://openstax.atlassian.net/browse/OTTER-647)

Leaving a required field incomplete currently produces no error state. The user is left with a disabled submit button and no indication of what is missing. This branch makes an incomplete required field raise its error when the user moves on, per the card's acceptance criteria.

**Unit suite, typecheck, eslint, prettier and `validate-actions` all pass**, on CI as well as locally.

## Why the fix is not just a config flag

Three root causes, not one:

1. **`validateInputOnChange` cannot flag a never-touched field.** Mantine triggers change validation from `setFieldValue`. It can raise a required error once a user enters and clears a value, but never for a field focused and blurred while still empty, which is exactly the AC's scenario. Five forms relied on change-only validation.
2. **Blur validation only reaches fields that spread `getInputProps`.** Mantine builds the validating `onBlur` inside `getInputProps` and nowhere else. Seven controls set values by hand via `setFieldValue`, so enabling `validateInputOnBlur` alone does nothing for them.
3. **Disabled submit buttons hide validation.** Twelve call sites gate on `form.isValid()`, which by design does not set `form.errors`, and a disabled button cannot reach submit-time validation either.

## Approach

Shared abstractions carry the fix so the defect classes stop being expressible:

- **A `useForm` wrapper in `@/common`** applying `validateInputOnBlur: true`. Caller options still win, so a form needing narrower coverage can pass a path array. Both Mantine generics are preserved.
- **A `FormField` component built on `Input.Wrapper`.** `Input.Wrapper` was used nowhere in the codebase; the `FormFieldLabel` plus `InputError` pair hand-rolled a primitive that already provides label, description, error, required asterisk and ARIA wiring.
- **`nativeFieldProps`** for standard inputs, **`fieldDescribedBy`** for controls taking a describedby prop under another name, **`widgetBlurHandler`** for composite widgets, and **`revalidateOnBlur`** so blur cannot erase a server-set error.

`FormField` distinguishes three cases, because `Input.Wrapper` does not inject error or ARIA into arbitrary children:

| Case | Applies to | Approach |
| --- | --- | --- |
| Standard Mantine input | `TextInput`, `Select`, `MultiSelect` | Wrapped for label and description; spread `nativeFieldProps` on the input, because its own `Input.Wrapper` shadows the outer context. |
| Composite widget | The Lexical `Editor` | Wrapped, with `fieldDescribedBy` passed to `ariaDescribedBy` (it takes Lexical-style props, not `aria-*`). |
| Already owns its chrome | `Radio.Group`, `PillsInput`, `PinInput` | Not wrapped. `label` / `withAsterisk` / `error` passed straight through so Mantine wires the association. |

### Mantine behaviors that shaped the design

**A null `relatedTarget` is ambiguous, and getting it wrong silently disables the feature.** Composite widgets emit blur as focus moves between their internal parts, so they need a containment guard. But `relatedTarget` is also null when the user clicks a non-focusable part of the page, which IS them moving on. Treating that as "still inside" meant clicking whitespace produced no error at all. `document.hasFocus()` separates an in-page click (validate) from a tab or window switch (skip).

**Mantine wins the prop race on ARIA.** `Input` applies `aria-describedby` from its wrapper's context, spread *after* the caller's props, so a hand-passed value is silently discarded. `nativeFieldProps` passes the error node and description to the input so its own wrapper computes the association, then uses `inputWrapperOrder` to suppress duplicate rendering. `PinInput` and `PasswordInput` needed their own routes: the former exposes `getInputProps(index)` which reaches each digit input, the latter derives `describedBy` from its own `description`.

**Blur revalidation clears errors the client cannot re-derive.** `validateField` clears the error when the client rule passes, which erased Clerk's "invalid code" and "invalid phone number" messages. `revalidateOnBlur` skips when an error is already showing; `clearInputErrorOnChange` drops it as soon as the user edits.

**Blur validation is per-field**, so a cross-field rule surfaces on the field the issue is attached to, not on its partner.

**`Radio.Group` names itself from its rendered label, not from `aria-label`.** It spreads unknown props onto an outer wrapper that carries no role, while the inner element holding `role="radiogroup"` takes `aria-labelledby` from `labelProps.id` and `aria-describedby` from Mantine's own `description` / `error`. A hand-passed `aria-label`, `aria-labelledby`, `aria-describedby` or `aria-invalid` therefore lands on a roleless div and reaches nothing, and the radiogroup is left pointing at a `${id}-label` element that was never rendered. Groups with a visible caption alongside them point `labelProps.id` at it; the two with no caption use the `label` prop. This is the same prop-race hazard as `Input`, reached by a different route.

### Schema and dead code

The Step 1 resolver required four fields Step 1 does not render: three document uploads and `title`. The uploads were orphaned by PR #555, which removed the OpenStax feature flag and deleted the component holding them, cleaning up the readiness schema but not the resolver schema. Enabling blur validation without this would surface four permanently failing, undisplayable errors.

Extracted `step1FieldsSchema` (Data Partner and language only) and pointed the Step 1 resolver at it; `isStep1Valid` is now `form.isValid()`. Pruned the orphan fields plus the client code consuming the inferred type. Title now trims before `min(1)`: a whitespace-only title used to pass schema validation while a separate trimmed check disabled submit. Deleted the dead `use-step1-form.ts`. Normalized the 13 direct `@mantine/form` `useForm` imports to `@/common`.

One regression caught here: removing `validateInputOnChange` broke the Proceed button, because that form is uncontrolled and the provider does not re-render on value change. Change validation is retained explicitly on the two Step 1 fields, with a comment.

### Researcher Lab

- `Editor` gained `error`, `onBlur` and `inputId`. Its existing `id` is the Yjs document name and the `yjs_document` primary key, so it could not double as a DOM id. The error border is ported from the pre-Yjs `editable-text.tsx`. The blur guard lives inside the editor so call sites cannot forget it.
- The proposal form, edit-and-resubmit section and collaborative text field moved onto `FormField`. The resubmission-note sections use `nativeFieldProps` / `fieldDescribedBy`; the collaborative one's blur validation and error border, lost in the Yjs migration, are restored.
- The PI `Select` keeps its composite change handler: its value is `piUserId` while `piName` holds the display label, so spreading `getInputProps('piName')` would inject a label into a Select keyed by user IDs. Blur validation is wired explicitly and the missing error message now renders.
- The PI-must-be-linked rule is attached to the `piName` path rather than the invisible `piUserId`, so the gate is enforceable without being unclearable.
- Step 1's Data Partner is marked required and `aria-required`; the language radio group gained a required indicator, the containment guard, and an error id it actually references.
- Research interests now validates on blur even when the draft is empty, and uses the shared guard rather than its own.

### Data Partner

- `useReviewFeedback` and `useReviewDecision` moved from plain `useState` to `useField({ validateOnBlur: true })`, preserving their surfaces since the former feeds a Yjs-backed editor and two mutation hooks.
- Code review criteria gained a per-criterion validator plus nested-path blur and error wiring. The form previously had no `validate` at all.
- Both decision radio groups gained real labels, error state, and the containment guard. They use Mantine's `label` prop rather than `aria-label`, which also makes `withAsterisk` render their only visible required marker.
- Team invitation permission: `getInputProps('permission', { type: 'checkbox' })` returns `checked` instead of `value`, which `Radio.Group` ignores. The group was therefore uncontrolled and did not follow form state, so a reset after a successful invite left the previous choice selected. Dropping the `type` option fixes it.
- Env-var and command row errors are derived from the value, not per-row `touched` state: these lists are keyed by index, so row state transferred to whichever row slid into that index on delete.
- `addEnvVar` inserts the trimmed pair it validated and applies the schema's own name rule. `addUrl` validates the draft pair, names both halves when empty rather than no-opping, and trims on both paths.
- The starter-code Dropzone gained a touched model.

### Shared account surface

- The SMS phone field gained a validator. It uses `isPossiblePhoneNumber` rather than `isValidPhoneNumber`: the latter rejects Clerk's reserved test range (+1 555 555 0100-0199) that the MFA e2e depends on. "Possible" still catches too-short, too-long and wrong-shape input, accepts extensions, and is per-country.
- Password errors un-suppressed. Both signup and reset pass `error={undefined}` to defer to a requirements list that was hidden when the password was empty, so blurring an empty password showed nothing at all. The list is now the input's `description`, so Mantine wires the ARIA, rendered below the field as before.
- The Terms and Privacy checkbox moved into the form with a `literal(true)` rule and forwards `onBlur`.
- OTP error containers carry per-screen ids.

### Layout

The four submit footers gained `align="flex-start"`. They now carry an outstanding-fields hint under the submit button, and the footer row aligns the other button against that stack. None of the four set an alignment before, so all inherited Mantine's `Group` default of `center`; measured in a browser, that left the left-hand button 12px below the submit button whenever a hint was present. An earlier revision of this description said they switched from `flex-end`, which was wrong: they had no `align` prop at all.

## Deviations from the plan, flagged for review

**Submit gating was not reversed.** The plan called for validate-on-click. That conflicts with OTTER-557, which deliberately gates these buttons and has tests asserting it. Reversing a shipped decision is not this card's call, so an `IncompleteFieldsHint` names the outstanding fields beside the disabled button instead. Blur validation covers fields the user visited; the hint covers fields never visited. If the team wants the button clickable, that is a follow-up card.

**Main-file selection** derives from async workspace state, so `useField` would go stale on every refetch. Derivation stays in `use-ide-files` and `submitDisabledReason` names the missing selection.

## Verification

Blur validation, error association and the composite blur guard are covered by unit tests: the
`useForm` wrapper's default and its per-path override, `nativeFieldProps` across `TextInput`,
`Select` and `MultiSelect`, the four `widgetBlurHandler` cases, per-digit OTP wiring, radio-group
accessible names, the password requirements association, the unlinked-PI gate and the
whitespace-only title. Tests assert computed accessible names rather than raw attributes, because an
attribute can be present and still resolve to nothing.

## Notes for reviewers

- Mantine option names are easy to confuse: `useForm` takes `clearInputErrorOnChange`, while `clearErrorOnChange` is the `useField` option.
- A boolean `error` gives visual styling only. Mantine renders the message, and adds it to `aria-describedby`, only for non-boolean values.
- `Tab` exits a radio group; arrow keys move within it. Worth knowing when testing the guarded groups by hand.
- **The Playwright e2e spec has not been run.** CI is the check to watch, particularly the MFA spec given the phone-validator change.
- Two pre-existing eslint errors in the untracked `src/styles/generated/` output, left over from the Panda CSS removal, are unrelated to this branch but make bare `pnpm run lint` fail.
- The server-side document upload actions in `study-request.ts` are confirmed dead but left for a separate cleanup PR.
- **Known follow-up, deliberately not fixed here.** In the code environment form, uploading starter code whose extension has no command line while the draft command extension box is empty surfaces `Missing command line for extension(s): .py` under Starter Code, rather than an error on the empty box. The message is accurate and submit is correctly blocked; only its placement is wrong. Fixing it means reordering chained `superRefine`s, which changes which issue surfaces first for other input combinations that have no test coverage, so it is left for a follow-up with a proper test matrix.
- Library behavior was checked against current Mantine documentation via context7 and against the installed source.

[OTTER-647]: https://openstax.atlassian.net/browse/OTTER-647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/common.ts
```diff
@@ -1,21 +1,71 @@
...
+ * validating `onBlur` there. Controls wired by hand need their own blur handler.
+ */
+export function useForm<
+    Values extends Record<string, unknown> = Record<string, unknown>,
+    TransformValues extends _TransformValues<Values> = (values: Values) => Values,
+>(input: UseFormInput<Values, TransformValues> = {}): UseFormReturnType<Values, TransformValues> {
+    return mantineUseForm<Values, TransformValues>({
+        validateInputOnBlur: true,
+        ...input,
+        enhanceGetInputProps: (payload) => {
```
nathanstitt, line 41: Small contract wrinkle: `enhanceGetInputProps` sits *after* the `...input` spread, so unlike every other option it's the one thing a caller can't override. The docblock above says "Caller options win", which is true for `validateInputOnBlur` but not for this key.

It's handled correctly today — line 37 calls through to `input.enhanceGetInputProps` — and nothing in the codebase passes one yet, so this is latent rather than broken. Might be worth a one-line note here saying the composition is deliberate, just so a future caller passing their own enhancer doesn't have to read the body to work out that theirs still runs.

src/app/[orgSlug]/study/[studyId]/proposal/missing-fields.ts
```diff
@@ -0,0 +1,34 @@
...
+    const missing: string[] = []
+
+    if (!values.title?.trim()) missing.push(FIELD_LABELS.title)
+    if (values.datasets.length === 0) missing.push(FIELD_LABELS.datasets)
+    if (!hasLexicalContent(values.researchQuestions)) missing.push(FIELD_LABELS.researchQuestions)
+    if (!hasLexicalContent(values.projectSummary)) missing.push(FIELD_LABELS.projectSummary)
+    if (!hasLexicalContent(values.impact)) missing.push(FIELD_LABELS.impact)
+    // Also missing when the name has no linked user: submission requires the id, but no field
+    // displays it, so without this a legacy draft disabled submit while naming nothing (OTTER-647).
+    if (!values.piName?.trim() || !values.piUserId?.trim()) missing.push(FIELD_LABELS.piName)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/missing-fields.ts]: The hint checks `piUserId` for non-emptiness, but `schema.ts` requires it to parse as a UUID (the `superRefine` on `piName`). A `piUserId` that's non-empty but not a valid UUID would drop out of the hint while still failing the schema — so submit stays disabled with nothing named beside it, which is the exact failure mode this hint exists to prevent.

Probably not reachable through the UI today since the Select only ever writes real ids, but a legacy or hand-edited draft could get there. Could we reuse the same UUID check the schema uses, so the two can't drift?

src/app/[orgSlug]/study/[studyId]/proposal/schema.ts
```diff
@@ -30,52 +26,74 @@ function maxWordsLexicalRefine(maxWords: number) {
...
+            .refine((val) => !val || countWordsFromLexical(val) <= WORD_LIMITS.additionalNotes, {
+                message: WORD_LIMIT_ERROR,
+            })
+            .optional()
+            .default(''),
+        piName: z.string().min(1, { message: REQUIRED_FIELD_ERROR }),
+        // No rule of its own: no field displays piUserId, so an error on this path is one the user
+        // cannot see or clear while it still blocks submit (OTTER-647). `default` also absorbs the
+        // `undefined` that hydrating a draft with no PI yields, which a bare `z.string()` rejects.
+        piUserId: z.string().default(''),
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/schema.ts:83]: Worth calling out for reviewers that this is a real loosening: `piUserId` went from `z.string().uuid()` to `z.string().default('')`. The `superRefine` below restores the guarantee (empty `piName` fails `min(1)`, non-empty `piName` demands a valid UUID), so I don't think there's an actual gap — I traced both branches and they're covered.

The reasoning in the comment is sound and the server-side `piUserId: z.string().uuid()` in `study-request.ts` is a good backstop. Just flagging it because "required UUID becomes optional string" is the kind of diff that deserves a second pair of eyes on the cross-field rule.

nathanstitt (commented):
looks good but specs are failing

src/common.ts
therealmarv: Good catch, the docblock was overclaiming. Documented in `76bd430e`:

The placement after the spread is deliberate, so a caller passing its own enhancer can't accidentally drop the server-error guard. A caller's enhancer still runs and still wins on every key except the guarded `onBlur`, which wraps whatever onBlur the caller leaves in place. That's now stated in the docblock rather than only being discoverable from the body.

src/app/[orgSlug]/study/[studyId]/proposal/missing-fields.ts
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/proposal/missing-fields.ts]: You're right, and this was a real gap rather than a theoretical one: `!piUserId?.trim()` passes for any non-empty string, while the schema demands a UUID. So `piUserId: 'not-a-uuid'` dropped the field out of the hint while still failing submit, which is exactly the failure mode the hint exists to prevent.

Fixed in `76bd430e` by doing what you suggested, extracting the check so the two can't drift:

```ts
// schema.ts
export const isLinkedPiUserId = (piUserId: string | undefined) => z.uuid().safeParse(piUserId).success
```

Both the `superRefine` and `missingProposalFields` now call it. Added a test pinning the malformed-id case alongside the empty-id one.

Two side notes while in there: the old inline check used `z.string().uuid()`, the deprecated Zod 4 form, so the shared helper uses `z.uuid()` to match the rest of the codebase. And you're right that the Select only ever writes real ids, so this needs a legacy or hand-edited draft to reach.

src/app/[orgSlug]/study/[studyId]/proposal/schema.ts
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/proposal/schema.ts:83]: Thanks for tracing both branches, your reading matches mine and no code change here.

The loosening was forced: hydrating a draft with no PI yields `undefined` for `piUserId`, which `z.string().uuid()` rejects, and the resulting error sat on a path no field renders. So it blocked submit with nothing the user could see or clear, which is the defect this card is about. `.default('')` absorbs the `undefined` and the `superRefine` moves the gate onto `piName`, the path the Select actually displays.

Worth noting your sibling comment on `missing-fields.ts` found the one place where that guarantee had genuinely leaked: the hint was checking non-emptiness rather than UUID validity. Both now call a shared `isLinkedPiUserId` exported from this file, so the cross-field rule and the hint can't drift apart. That also makes the invariant you traced explicit in one place rather than duplicated in two.

nathanstitt (approved):
I see multiple places where the tests were updated to match the label case, but the `/i` regex flag was also left in place so they are case insensitive. it'd cut down on diff size a lot if they weren't updated

PR #914, opened 2026-07-27
SHRMP-273 Legal info and acknowledgment - feature branch

kswanson33:
Contains work for the overarching epic (SHRMP-273)
- SHRMP-274
- SHRMP-275
- SHRMP-276
- SHRMP-277


src/app/admin/safeinsights/legal/draft-form.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Paper, Title, Button, Flex, Group, Text, Stack } from '@mantine/core'
+import { Dropzone } from '@mantine/dropzone'
+import { UploadIcon, FileArrowUpIcon, ArrowCircleRightIcon } from '@phosphor-icons/react/dist/ssr'
+
+// Access modal from "Upload New {label}" button
+// Page 1: upload draft. Displays if there isn't a current draft
+// Page 2: Shows after finishing Page 1 ("Save Draft" button), or if there is a current draft
+// and contains review + publish logic
+
+export default function DraftForm({ doctype }: { doctype: LegalDocumentType }) {
```
nathanstitt: Nothing imports this file — `tos-pn-upload.tsx` pulls `DraftForm` from `./document-modal`, which has a near-identical copy of this component (plus the `onDraftSaved` callback). If this one is superseded, worth deleting before merge so the two copies can't drift; if it's the one being actively worked on, maybe the `document-modal` copy is the one to drop.

src/app/admin/safeinsights/legal/draft-form.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useState } from '@/common'
+import { LegalDocumentType } from '@/database/types'
+import { uploadFiles } from '@/hooks/upload'
+import { isActionError } from '@/lib/errors'
+import { createLegalDocumentDraftAction } from '@/server/actions/legal-document.actions'
+import { Paper, Title, Button, Flex, Group, Text, Stack } from '@mantine/core'
+import { Dropzone } from '@mantine/dropzone'
+import { UploadIcon, FileArrowUpIcon, ArrowCircleRightIcon } from '@phosphor-icons/react/dist/ssr'
+
+export default function DraftForm({ doctype }: { doctype: LegalDocumentType }) {
```
nathanstitt: This file doesn't appear to be imported anywhere, is it old?

src/server/actions/legal-document.actions.ts
```diff
@@ -0,0 +1,519 @@
...
+    return await Promise.all(
+        latest.map(async (version) => ({
+            type: version.type,
+            versionId: version.versionId,
+            content: await contentOf(version.filePath),
+        })),
+    )
+})
+
+export const fetchLegalDocumentAcknowledgementsAction = new Action('fetchLegalDocumentAcknowledgementsAction')
```
nathanstitt, line 335: This action has no UI consumer yet, assuming you're adding it later?

src/app/admin/safeinsights/legal/tos-pn/upload-modal-pages.tsx
```diff
@@ -0,0 +1,162 @@
...
+        setFile(null)
+    }
+
+    return (
+        <Stack>
+            <Title order={4} pb="sm">
+                Upload your draft document here:
+            </Title>
+            <Paper shadow="xs" p="md">
+                <Dropzone onDrop={handleDrop} p="md">
```
nathanstitt: This dropzone accepts any file type, but the viewer only renders markdown — a PDF dropped here uploads fine and then displays as garbage. `PdfDropzone` next door shows the `accept` + `maxFiles={1}` pattern; an `accept` for markdown (or reusing a generalized version of that component) would close the gap. (I know the plan doc flags this one already.)

src/app/admin/safeinsights/legal/participation/participation-agreements.tsx
```diff
@@ -0,0 +1,134 @@
...
+            </AppModal>
+            <VersionHistoryModal
+                isOpen={Boolean(historyFor)}
+                onClose={closeHistory}
+                title={`${historyFor?.orgName ?? ''} — version history`}
+                scope={{ type, orgId: historyFor?.orgId }}
+            />
+            <LoadingState isVisible={isLoading} />
+            <EmptyState isVisible={!isLoading && agreements.length === 0} />
+            <Table withTableBorder withRowBorders horizontalSpacing="md" verticalSpacing="sm">
```
nathanstitt: We already use `mantine-datatable` in this exact admin section (`src/app/admin/safeinsights/table.tsx`, and `admin/team/users-table-view.tsx`),  `DataTable` gives `fetching` and `noRecordsText` for free

src/app/admin/safeinsights/legal/sla/upload-sla-form.tsx
```diff
@@ -0,0 +1,295 @@
...
+        <Text c="red.8">
+            Acknowledgements of the current version do not carry over. Everyone on this study will be asked to
+            acknowledge again, and researchers cannot submit code until they do.
+        </Text>
+    )
+}
+
+// Publishing cannot be undone and it obligates people, so the confirmation repeats everything about
+// to be written and names who will be asked to acknowledge it.
+const ConfirmPublishModal: FC<{
```
nathanstitt: This `ConfirmPublishModal`, plus `VersionNote` and the publish mutation (create draft → `uploadFiles` → publish → invalidate → close), are nearly line-for-line the same as in `upload-participation-agreement-form.tsx` — the two files only genuinely differ in how the subject is chosen (org picker vs. the study cascade) and the read-back fields. A shared confirm modal and a shared `usePublishAgreement` hook would remove ~150 lines and keep the two confirmation flows from drifting apart.

src/app/admin/safeinsights/legal/tos-pn/tos-pn.tsx
```diff
@@ -0,0 +1,126 @@
...
+            </Anchor>
+            <AppModal title="Review version" isOpen={viewModalOpened} onClose={closeViewModal}>
+                <PreviewDocument url={version.downloadUrl} label={label} />
+            </AppModal>
+            <Text>Published on {publishedOn}</Text>
+        </Group>
+    )
+}
+
+function VersionHistory({ history, doctype }: { history: PublishedVersion[]; doctype: LegalDocumentType }) {
```
nathanstitt: This is a second version-history UI — `VersionHistoryModal` is deliberately scope-agnostic (its `scope` prop and the action already accept a bare `{ type: 'tos' }`), so we could reuse it here and drop `VersionHistory` + `ShowVersion`. That would also take out the `padStart(6, '0')` label ("TOS000001" reads oddly) and one of the two new date formats this PR introduces.

src/app/admin/safeinsights/legal/tos-pn/tos-pn.tsx
```diff
@@ -0,0 +1,126 @@
...
+            </Collapse>
+        </>
+    )
+}
+
+export function TosPnUpload({ doctype }: { doctype: 'tos' | 'pn' }) {
+    const [createModalOpened, { open: openCreateModal, close: closeCreateModal }] = useDisclosure(false)
+
+    const { data, isLoading, isError, error } = useQuery({
+        queryKey: ['legalVersions', doctype],
```
nathanstitt: This caches `fetchLegalDocumentVersionsAction` under `['legalVersions', doctype]` while `version-history-modal.tsx` caches the same action's data under `['legalDocumentVersions', ...]` — the upload mutations invalidate the latter prefix, so they'd miss this cache (and vice versa). Worth converging on one key for the action so future invalidations can't silently skip half the consumers.

src/app/admin/safeinsights/legal/version-history-modal.tsx
```diff
@@ -0,0 +1,92 @@
...
+import type { ActionSuccessType } from '@/lib/types'
+import type { LegalDocumentTypeValue } from '@/schema/legal-document'
+import { fetchLegalDocumentVersionsAction } from '@/server/actions/legal-document.actions'
+import { Anchor, Stack, Table, Text } from '@mantine/core'
+
+type Scope = { type: LegalDocumentTypeValue; orgId?: string; studyId?: string }
+
+type Version = NonNullable<ActionSuccessType<typeof fetchLegalDocumentVersionsAction>['current']>
+
+const formatPublishedOn = (publishedAt: Version['publishedAt']) =>
```
nathanstitt: while I personally also prefer `toLocaleDateString(dateStyle: 'medium')`, the app convention for dates is `dayjs(...).format('MMM DD, YYYY')` (studies table, job status, proposal headers)

src/app/admin/safeinsights/legal/tos-pn/upload-modal-pages.tsx
```diff
@@ -0,0 +1,162 @@
...
+        }
+        await uploadFiles([[file, result.upload]])
+
+        onDraftSaved()
+    }
+
+    const onRemove = () => {
+        setFile(null)
+    }
+
```
nathanstitt, line 80: we should use a useMutation for the above actions, it has onError handle that should be wired up to either form or toast variations.

src/app/admin/safeinsights/legal/participation/upload-participation-agreement-form.tsx
```diff
@@ -0,0 +1,227 @@
...
+            // Publish last, so a failed upload leaves a replaceable draft rather than a live
+            // agreement with no file behind it.
+            const { version, upload } = actionResult(
+                await createLegalDocumentDraftAction({ type, orgId, fileName: file.name }),
+            )
+            await uploadFiles([[file, upload]])
+            return actionResult(await publishLegalDocumentVersionAction({ versionId: version.id, signedAt }))
+        },
+        // Wrapped because react-query's second arg is the variables, which reportError reads as a title.
+        onError: (error: unknown) => reportError(error),
```
nathanstitt: we have a `reportMutationError` that wraps for you in componetns/errors

src/database/migrations/1785860103751_legal_documents.ts
```diff
@@ -0,0 +1,93 @@
+import { type Kysely, sql } from 'kysely'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function up(db: Kysely<any>): Promise<void> {
+    await db.schema.createType('legal_document_type').asEnum(['tos', 'pn', 'ropa', 'dopa', 'sla']).execute()
```
nathanstitt: all other enums we've crated are upper case:

```suggestion
    await db.schema.createType('legal_document_type').asEnum(['TOS', 'PN', 'ROPA', 'DOPA', 'SLA']).execute()
```

nathanstitt (commented):
looking good KC, just a bunch of notes to think over. ty!

src/app/admin/safeinsights/legal/participation/participation-agreements.tsx
chrisbendel: Switched to DataTable, with fetching and noRecordsText. Same in the SLA table.

src/app/admin/safeinsights/legal/sla/upload-sla-form.tsx
chrisbendel: Pulled out into publish-agreement.tsx: usePublishAgreement plus a shared ConfirmPublishModal that takes subject and consequence. Both forms use it now.

src/app/admin/safeinsights/legal/tos-pn/tos-pn.tsx
chrisbendel: Both consumers go through legalDocumentQueryKeys in schema/legal-document.ts now. versionsForType is a real prefix of versions, so a publish invalidation reaches every mounted scope.

src/app/admin/safeinsights/legal/version-history-modal.tsx
chrisbendel: Using dayjs MMM DD, YYYY. signedAt stays a bare YYYY-MM-DD string since it never goes through a timezone.

src/app/admin/safeinsights/legal/participation/upload-participation-agreement-form.tsx
chrisbendel: Done, the shared publish mutation uses reportMutationError. Swapped the tos/pn one over too.

src/database/migrations/1785860103751_legal_documents.ts
chrisbendel: Enum is uppercase now: TOS, PN, ROPA, DOPA, SLA.

src/app/admin/safeinsights/legal/tos-pn/tos-pn.tsx
chrisbendel: Reusing VersionHistoryModal here now. VersionHistory and ShowVersion are gone, the panel keeps a single current-version line (bare version number, MMM DD, YYYY) plus a Version History link, so the padStart label and the second date format went with them.

One wrinkle: tos/pn are markdown, so a signed URL in a new tab gives raw source or a download. The Document column derives from legalDocumentFormats now, markdown renders through PreviewDocument in place and pdf keeps the anchor. Signed on only shows for the scoped types since tos/pn never carry a signature date. PreviewDocument moved up to legal/preview-document.tsx so the shared modal is not importing out of the tos-pn folder.

Also took the two adjacent ones in that file: reportMutationError, and isSettled on ConfirmPublishForm so a settled publish cannot fire twice.

src/components/legal/require-acknowledgement.tsx
```diff
@@ -0,0 +1,86 @@
...
+        isPending,
+        error,
+    } = useMutation({
+        // One row per document, so the evidence stays per-document however many checkboxes were
+        // shown. actionResult is what makes a failure visible at all: actions resolve with
+        // { error } rather than rejecting, so without it a refused acknowledgement would run
+        // onSuccess and silently reopen the modal saying nothing. A partial failure is safe to
+        // retry — the acks are idempotent and the modal re-renders with whatever is outstanding.
+        mutationFn: async (pending: PendingLegalDocument[]) => {
+            await Promise.all(
```
nathanstitt: `Promise.all` rejects on the first failure, but the other `acknowledgeLegalDocumentAction` calls have already been dispatched and will keep running. So when a user owes both a ToS and a Privacy Notice and one call fails, we can land in a state where one ack row was written, the mutation reports an error, and `onSuccess` never fires — so the query isn't invalidated and the modal re-renders from stale data still listing *both* documents.

The comment above says a partial failure is safe to retry because the acks are idempotent, which is true of the write itself. The gap is the display: the user is told to re-acknowledge something the server already recorded, and the error text is whichever one lost the race.

`Promise.allSettled` would let us invalidate regardless and surface only the genuinely-failed ones — the refetch then narrows the modal to what's actually outstanding, which is the behavior the comment is describing.

src/components/legal/require-acknowledgement.tsx
```diff
@@ -0,0 +1,86 @@
...
+            setIsChecked(false)
+            await queryClient.invalidateQueries({ queryKey: legalDocumentQueryKeys.pendingAcknowledgements() })
+        },
+    })
+
+    return {
+        documents,
+        isChecked,
+        setIsChecked,
+        onContinue: () => acknowledge(documents),
```
nathanstitt: `onContinue` sends `documents` — the full pending list — rather than what the user actually consented to. There's a single checkbox covering every document, so today those are the same set.

The thing that makes me raise it: `documents` is query data that can refetch while the modal is open. If a new document is published between render and click, React Query can swap the array underneath, and we'd record consent to a document whose text was never displayed. Snapshotting the list that was rendered (or keying the checkbox to the version ids it covered) would close that, and these rows are the compliance evidence, so it seems worth being strict about.

src/components/legal/acknowledgement-modal.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    label={legalAcknowledgementCheckboxLabel(documents)}
+                />
+
+                <AcknowledgementError error={error} />
+
+                <Group justify="flex-end">
+                    <Button variant="subtle" onClick={onSignOut} disabled={isSubmitting}>
+                        Sign out
+                    </Button>
+                    <Button onClick={onContinue} disabled={!isChecked} loading={isSubmitting}>
```
nathanstitt: Continue is gated on `!isChecked` but not on `isSubmitting`. `loading` renders the spinner and Mantine does block pointer clicks on a loading button, but it doesn't set `disabled`, so the button stays keyboard-focusable and Enter/Space still fires `onClick`.

Sign out right next door correctly uses `disabled={isSubmitting}`. Worth matching it (`disabled={!isChecked || isSubmitting}`) so a double-tap can't fire a second round of writes mid-flight.

src/database/migrations/1785860103751_legal_documents.ts
```diff
@@ -0,0 +1,97 @@
...
+             OR (type = 'SLA' AND study_id IS NOT NULL AND org_id IS NULL)`,
+        )
+        .execute()
+
+    // NULLS NOT DISTINCT (PG 15+) is required: by default Postgres treats NULLs as distinct, so a
+    // plain UNIQUE would allow duplicate ('TOS', NULL, NULL) rows.
+    await sql`
+        ALTER TABLE legal_document
+        ADD CONSTRAINT legal_document_scope_unique
+        UNIQUE NULLS NOT DISTINCT (type, org_id, study_id)
```
nathanstitt, line 29: nice! I had no idea what this was `UNIQUE NULLS NOT DISTINCT` but comment cleared it up.

nathanstitt (commented):
looking really nice, only thing I'd call out as critical is the snapshotting agreed on docs so they don't silently change while signing

src/components/legal/require-acknowledgement.tsx
chrisbendel: went the other way on this - theres only one document per modal now, so theres no Promise.all left to settle. ack one, invalidate, refetch hands you the next one if theres a next one.

allSettled would have worked but it keeps the partial-success state around and i couldnt come up with a decent error string for "one of these failed". this way its one call and one error.

src/components/legal/require-acknowledgement.tsx
chrisbendel: yeah this was the real one. modal renders `documents[0]` and the checkbox stores the versionId it was ticked against instead of a boolean, so `isChecked` is `consentedVersionId === document?.versionId`. if a refetch swaps the doc underneath, the tick stops matching and Continue re-disables. Continue sends `consentedVersionId`, not whatevers in the list at click time.

nice side effect - the `setIsChecked(false)` in onSuccess went away, the derivation handles it.

src/components/legal/acknowledgement-modal.tsx
chrisbendel: done, `disabled={!isChecked || isSubmitting}`. added a test since `loading` on its own doesnt set disabled.

chrisbendel:
heads up, the ToS/PN modal is now one document per modal instead of both stacked under one checkbox. flagged on SHRMP-275 for UX.

src/components/legal/require-acknowledgement.tsx
```diff
@@ -0,0 +1,87 @@
...
+            await queryClient.invalidateQueries({ queryKey: legalDocumentQueryKeys.nextPendingAcknowledgement() })
+        },
+    })
+
+    return {
+        document,
+        // Consent is keyed to the version that was on screen rather than kept as a bare boolean. The
+        // query can refetch while the modal is open, so a document published mid-read would otherwise
+        // inherit a tick given to the one before it — and these rows are the compliance evidence.
+        isChecked: consentedVersionId === document?.versionId,
```
nathanstitt, line 52: This is a much better answer than the `Promise.allSettled` I suggested,  good call ty!

src/components/legal/require-acknowledgement.tsx
```diff
@@ -0,0 +1,87 @@
...
+        queryKey: legalDocumentQueryKeys.nextPendingAcknowledgement(),
+        queryFn: () => fetchNextPendingLegalAcknowledgementAction(),
+        enabled: Boolean(session),
+    })
+
+    // An unreadable document leaves the gate open, which is the right call — the user can do nothing
+    // about it and the compliance artifact is the ack row, not the blocking. But a gate that has
+    // quietly stopped asking must not also be invisible to us.
+    useEffect(() => {
+        if (readError) captureException(readError)
```
nathanstitt: Nice detail that wasn't in my original feedback: a `readError` leaves the gate open (correct — the user can't act on it, and the ack row is the artifact, not the blocking) but reports to Sentry so a gate that has quietly stopped asking isn't also invisible to us.

That's the failure mode I'd actually worry about in production, since it's silent and fail-open by design. Good that it's instrumented.

src/components/legal/acknowledgement-modal.tsx
```diff
@@ -0,0 +1,89 @@
...
+                />
+
+                <AcknowledgementError error={error} />
+
+                <Group justify="flex-end">
+                    <Button variant="subtle" onClick={onSignOut} disabled={isSubmitting}>
+                        Sign out
+                    </Button>
+                    {/* Mantine's `loading` blocks pointer clicks but leaves the button keyboard-focusable. */}
+                    <Button onClick={onContinue} disabled={!isChecked || isSubmitting} loading={isSubmitting}>
```
nathanstitt, line 82: `disabled={!isChecked || isSubmitting}` — matches Sign out next door, and the comment above it records *why* (`loading` blocks pointer clicks but leaves the button keyboard-focusable, so Enter/Space would still fire). Exactly the kind of non-obvious reason that earns a comment under the repo's "why, not what" rule.

src/server/actions/legal-document.actions.ts
```diff
@@ -0,0 +1,556 @@
...
+        const next = latest.find((version) => !acknowledgedVersionIds.has(version.versionId))
+        if (!next) return null
+
+        // S3 is read only once something is actually outstanding. Every page load runs this action and
+        // the overwhelmingly common answer is "nothing pending".
+        return {
+            type: next.type,
+            versionId: next.versionId,
+            isUpdate: acknowledgedDocumentIds.has(next.legalDocumentId),
+            content: await contentOf(next.filePath),
```
nathanstitt: Deferring the S3 read until something is genuinely outstanding is the right shape — every page load runs this action and the overwhelmingly common answer is "nothing pending", so fetching content eagerly would have put an S3 round-trip on the hot path for all users.

The `isUpdate` derivation via `acknowledgedDocumentIds` is a nice touch too: distinguishing "has been updated" from "is now available" costs one extra Set from data already in hand, and it's the difference between a modal that reads naturally and one that tells a returning user a document they've seen for months is new.

src/app/admin/safeinsights/legal/publish-agreement.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * Upload a signed agreement and publish it as the document's next version.
+ *
+ * Shared by the participation and study-level forms, which differ only in how the subject is chosen
+ * and which caches go stale — everything from "make a draft" to "it is live" is the same. Publish
+ * runs last so a failed upload leaves a replaceable draft rather than a live agreement with no file
+ * behind it; a document that already has versions gets a new one rather than a second document.
+ */
+const usePublishAgreement = ({
```
nathanstitt: This is the duplication comment I felt most strongly about last round, and the resolution goes further than asked. `usePublishAgreement` owns the create-draft → upload → publish → invalidate sequence, `useAgreementUpload` parameterises only the genuinely-varying part (`scopeFor`, `invalidateKeys`), and `ConfirmPublishModal` is shared with the caller supplying just `subject` and `consequence`. The two forms now differ in what they actually differ in.

Two things I'd have missed and am glad are written down: publish running last so a failed upload leaves a replaceable draft rather than a live agreement with no file behind it, and `isSettled` being success-based rather than pending-based because the form stays mounted until the parent closes it. That second one is exactly the kind of thing that comes back as a double-publish bug six months later.

All fifteen of my previous comments are addressed — including the ones I'd have accepted a "won't fix" on. Thanks for working through them.

bin/seed-dev-legal.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+    console.log('\nEveryone but the `legal` role is acknowledged up to date.')
+    await db.destroy()
+}
+
+main().catch(async (error) => {
+    console.error(error)
+    await db.destroy()
+    process.exit(1)
+})
```
nathanstitt, line 41: should this be in a proper seed file in `src/database/seeds` ? those get ran on every deployment so they can populate everywhere, but need to be written to skip if they're already been applied

bin/seed-dev-studies.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    console.log(`\nSeeded ${created} studies across ${pairs.length} org pairs.`)
+    await db.destroy()
+}
+
+main().catch(async (error) => {
+    console.error(error)
+    await db.destroy()
+    process.exit(1)
+})
```
nathanstitt, line 85: same comment, seems like this might be good to be a db seed

nathanstitt (approved):
looks good, not finding anything else of note other than it seems like maybe we should move the manual things into seeds, but I'm not sure if that's valid or not.  thanks so much, this is looking really nice

bin/seed-dev-legal.ts
chrisbendel: kept it out of src/database/seeds on purpose - those run in deployed envs via the migrator lambda, and tos/pn are global. publishing one there means everyone on staging is blocked until they tick a document whose body says "Seeded for end-to-end tests".

the worse half is what it does to get around that. seedLegalDocuments acks every other user to the current versions so they arent blocked, and those ack rows are the compliance evidence for the whole epic. dont want to be writing consent records for real people on staging.

thats what assertDisposableDatabase is doing, it refuses any host that isnt localhost/docker/CI. staging getting a real ToS should be an admin publishing one through the Legal tab i think

bin/seed-dev-studies.ts
chrisbendel: this ones closer since its just demo data. two things stopped me. it calls seedStudyFor out of tests/e2e.seed which pulls in @/server/aws, and that bundle is what broke the migrator lambda last time with the CJS require. and its additive, so every deploy adds another N studies per org pair unless i give the seeded ones an identity to skip on.

did add the testingDataAllowed gate to it though, the script had no guard at all and would run against whatever DATABASE_URL was exported. if we do want studies on staging i think a one off run beats every deploy

PR #917, opened 2026-07-28
OTTER-685: Log code environment changes and show them in a history modal

nathanstitt:
Closes [OTTER-685](https://openstax.atlassian.net/browse/OTTER-685).

Edits to code environments can break studies later, but nothing recorded who changed what — `orgCodeEnv` has no `updated_at` and the actions wrote no audit events. This records a per-field before/after diff on create, update and delete, and surfaces it behind a history icon beside the pencil in org settings.

## Notable decisions

**Audit writes are inline, not `deferred()`.** Every other handler in `events.ts` defers via `after()`, which runs post-response — after the transaction has committed *or rolled back* — and `after()` does not unschedule on error. A mutation that failed partway would still emit a row claiming it succeeded. These write on the caller's transactional executor instead, so the audit row commits and rolls back with the change. A test forces that rollback and asserts no row is written.

**The update middleware only loaded 6 of the 10 editable columns**, so there was no "before" value for `name`, `language`, `commandLines`, `isTesting` or `settings`. It now takes a namespaced snapshot of the full prior row rather than widening the shared loader, which `fetchStarterCodeAction` also uses and which feeds every key into the CASL subject.

**Field comparison is structural.** Kysely parses a fresh object per jsonb read and Postgres reorders jsonb keys on storage, so both reference equality and `JSON.stringify` comparison report false positives on every save.

**The history query filters on the middleware-resolved code env id, never the client-supplied param** — `audit` has no org column to scope by, so trusting the param would expose other orgs' history. Covered by a cross-org test.

## Scope notes

- The starter-code flag records that the server wiped and rewrote the S3 prefix, not that the new files landed — the browser uploads them after the action returns. If that upload fails the row claims files that aren't in S3, which is a plausible cause of the failures this is meant to diagnose; fixing it needs a separate post-upload confirmation step.
- Deletes are logged with full final state but have no UI, since the row is gone from the listing.
- Two migrations: `CODE_ENV` added to the `audit_record_type` enum (alone in its file — Postgres forbids using a new enum value in the transaction that adds it) and an index on `audit (record_type, record_id, created_at DESC)`.

## Testing

`pnpm run checks` and the full unit suite pass — 2008 tests, 0 failures. New coverage includes the no-op save (logs nothing), starter-code replacement with unchanged filenames, rollback safety and cross-org denial.

Not yet exercised end-to-end in a running app.

[OTTER-685]: https://openstax.atlassian.net/browse/OTTER-685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/[orgSlug]/admin/settings/code-envs.actions.ts
+++ b/src/app/[orgSlug]/admin/settings/code-envs.actions.ts
@@ -36,6 +36,15 @@ import logger from '@/lib/logger'
 import type { DB } from '@/database/types'
 import type { Kysely } from 'kysely'
 import { Routes } from '@/lib/routes'
+import { AUDITED_CODE_ENV_FIELDS, diffFields } from '@/lib/audit-diff'
+import { onCodeEnvCreated, onCodeEnvDeleted, onCodeEnvUpdated } from '@/server/events'
+
+const priorCodeEnvFor = (db: Kysely<DB>, codeEnvId: string) =>
+    db
+        .selectFrom('orgCodeEnv')
+        .select([...AUDITED_CODE_ENV_FIELDS])
+        .where('id', '=', codeEnvId)
+        .executeTakeFirstOrThrow()
 
 const codeEnvFromOrgAndId = async ({
     params: { orgSlug, codeEnvId },
@@ -96,7 +105,7 @@ export const createOrgCodeEnvAction = new Action('createOrgCodeEnvAction', { per
     .params(createOrgCodeEnvSchema)
     .middleware(orgIdFromSlug)
     .requireAbilityTo('update', 'Org')
-    .handler(async ({ params, orgId, db }) => {
+    .handler(async ({ params, orgId, db, session }) => {
         const { orgSlug, starterCodeFileNames, sampleDataPath, sampleDataUploaded, dataSourceIds, ...fieldValues } =
             params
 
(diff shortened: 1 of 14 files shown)


src/app/[orgSlug]/admin/settings/code-envs.actions.ts
```diff
@@ -181,7 +199,10 @@ export const updateOrgCodeEnvAction = new Action('updateOrgCodeEnvAction', { per
     .params(updateOrgCodeEnvSchema)
     .middleware(async (args) => {
         const { codeEnv, orgId } = await codeEnvFromOrgAndId(args)
-        return { ...codeEnv, orgId }
+        // Namespaced rather than flattened: the handler already destructures several of
+        // these columns as prev*, and every middleware key is fed into the CASL subject.
+        const priorCodeEnv = await priorCodeEnvFor(args.db, codeEnv.id)
```
Comment (Rod0352) [inline src/app/[orgSlug]/admin/settings/code-envs.actions.ts]: This runs before the ability check, and when that check fails the error message
includes `JSON.stringify(abilityArgs)` — which now contains `priorCodeEnv.settings` —
and gets returned to the caller. So a non-admin org member can call update, get
denied, and read the org's env vars out of the error. Loading `priorCodeEnv` in
the handler instead (after the permission check) fixes it with no other changes.

Rod0352 (approved):
Lgtm :) one problem worth looking into possibly

PR #919, opened 2026-07-28
OTTER-642: enforce one study_job_file row per artifact slot

therealmarv:

## Problem

When a study errors out, each log file is listed twice for both the reviewer and the researcher, and after release the two copies can disagree (one approved, one still locked).

Each duplicate is a real `study_job_file` row. The storage path is derived from the job and the artifact type, so a re-delivered ingest webhook overwrote the same S3 object while inserting a second row describing it. An errored run ends at `JOB-ERRORED` and never reaches the `RUN-COMPLETE` guard in `results/route.ts`, so the retry was never rejected and it also appended a second `JOB-ERRORED` and re-sent the reviewer email.

## Fix

An artifact slot is one `(study_job_id, path, file_type)`. A partial unique index makes a second row for a slot impossible, so there is one rule rather than a read-side collapse to remember at every call site.

1. **Migration `1780600000000_dedupe_study_job_file_artifacts.ts`** collapses each slot to its newest row and adds the index. The key-preservation sequence comes from the closed PR #876: pick the survivor, copy every recipient key onto it (`ON CONFLICT (study_job_file_id, file_path, fingerprint) DO NOTHING`), repoint `source_id`, then delete the rest.
2. **`storeJobFile` keys its lookup on the same slot**, so the row it updates, the row the index protects and the row the migration keeps are one rule instead of three that can drift. Its insert uses `ON CONFLICT DO NOTHING` and recovers the loser of a concurrent first delivery as the repeat it effectively is.
3. **`results/route.ts` finalizes the outcome once**, in a single transaction behind a `FOR UPDATE` on the job row: check whether this callback's outcome was already recorded, check the round is still open, insert. The lock is taken after the uploads and released before the email, so neither S3 nor SMTP runs inside it. That is the distinction from #876, which held a transaction open across the upload itself.

Also moves the `MAIN-CODE`/`SUPPLEMENTAL-CODE` list into `file-type-helpers.ts` as `isCodeFileType`, since the migration, the index and `study-code-files.ts` all need the same notion of a code file. No behavior change.

## Not replacing content someone can already open

A re-delivery is dropped when the artifact's keys have already been wrapped for recipients, falling back to whether the round is decided. Keys are the thing being protected: they were wrapped from the AES keys inside the ciphertext stored at that path right now, so replacing the object leaves them unwrappable and the bucket is unversioned, so the old bytes are gone.

Round status alone was the wrong predicate, and this was found in review. `approveJobCode` persists recipient keys alongside `CODE-APPROVED` and that round stays open afterwards, so a reviewer approving code can share scan logs while `roundIsClosed` is still false. Under the old guard a delayed scanner or containerizer delivery would then overwrite an artifact a researcher could already open, permanently. Keying on the keys also closes the narrower window in `approveStudyJobFilesAction`, where keys are written just before `FILES-APPROVED`.

A scan or packaging callback stores its log twice, encrypted for the researcher and plaintext for the reviewer's parsed statuses and download. Protecting only the encrypted half would let a re-delivery after sharing refuse the ciphertext and replace the readable copy, leaving the reviewer's PASSED/FAILED describing findings the researcher cannot open. `encryptAndStoreLog` now reports whether it stored (or null when it never tried, where there is no counterpart to disagree with), and both routes skip the plaintext write only when the encrypted half was actually refused.

## Why the migration is safe where #876 was not

#876 was approved and then closed over the risk of its migration. Two changes remove that risk rather than accept it:

- **The key includes `file_type`.** Runs before `e10c2cd7` (2025-06-26) wrote logs and results both to `results/encrypted-results.zip`, and approved logs and results both to `results/approved/{name}`. Those rows differ only by type, so a `(study_job_id, path)` index could not be created without deleting semantically distinct records. Including the type leaves them in the table untouched, and the migration only removes rows that are redundant descriptions of one object.
- **The index is partial on non-code types.** `study_job_file` also holds `MAIN-CODE` and `SUPPLEMENTAL-CODE`, whose paths run through `sanitizeFileName`, so two uploads can resolve to one path. Under a total index that would 500 the submit, which is why #876 needed path collapsing and a `SELECT ... FOR UPDATE` on the study row. Excluding code types drops both, and matches Iris's note that submitted code stays out of scope.

Two further hazards in the backfill, both found in review and fixed:

- **`source_id` is a self-reference with no `ON DELETE` clause**, so `NO ACTION` applies. Approval records the encrypted row an `APPROVED-*` row came from, and on a job that already held duplicates that reference can point at one the backfill drops, so the delete raised a foreign key violation and aborted the migration. That is the likeliest shape for this card's data, since the reported bug had a reviewer release one of two duplicate log rows. `source_id` now moves to the survivor first, verified by removing the repoint and watching the test fail on `study_job_file_source_id_fkey`.
- **The copy and the delete were separate snapshots** under `READ COMMITTED`, so an approval committing a recipient key against a loser in between was visible to the delete but not the copy, and the cascade took a key that was never carried forward. Both tables are now locked in `SHARE ROW EXCLUSIVE` for the transaction, which also covers the index creation.

The migration logs what it collapsed plus how many same-path slots with differing types it left behind. That count cannot be recovered afterwards and is the only input a duplicate-display follow-up would need.

## Behavior

Confirmed with @nathanstitt and Iris before this rework:

- First error: `JOB-ERRORED` recorded, timestamp reflects that delivery, logs saved.
- Second error while the round is open and nothing has been shared: log content is overwritten in place, the timestamp and `JOB-ERRORED` stay from the first, no new email. Content is never compared, so the stored log is always the latest sent and no trace of the earlier one is kept.
- Once the artifact's keys are shared, or the round is decided: a late re-delivery is dropped and logged as a warning. The response says the artifacts were dropped rather than "already received", which claimed more than actually happened.
- A late delivery carrying a slot the job never had keeps that artifact, since nothing was shared for it to protect, but records no outcome: appending `RUN-COMPLETE` after a decision would break the round-closing invariant and email the reviewer about a decided study.
- **Exception, unchanged from `main`:** a repeat delivery on a job that already has `RUN-COMPLETE` is rejected by the guard at the top of the route with `422`, before artifacts are parsed. So the success-with-detail responses apply to errored rounds, not completed runs. Left as is deliberately: that guard is what stops a completed run's blob being overwritten under already-shared keys, and its test pins the 422.

## Known limits

- **Approval-side serialization is not closed.** Recipient keys are invisible until the approval transaction commits, and `approveJobCode` holds an external image-build trigger between `CODE-APPROVED` and its key inserts, so a callback landing inside that window can still replace an artifact whose keys are about to be committed. Separately, `approveStudyJobFilesAction` does not take the job lock the results route takes, so a reviewer's all-or-nothing decision can be recorded against an artifact set a late callback has since added to (leaving a stored artifact with no researcher keys). Both predate this branch, which narrows rather than widens them, and closing either means reaching into the approval path. Own card.

- **Status ordering is a heuristic, not proof of ownership.** The outcome lookup is scoped to statuses recorded after the job already held these artifacts, which removes the common masking case. It cannot be conclusive: `JOB-ERRORED` is written by the scanner, the containerizer and the generic `PUT /api/job/[jobId]` route, and `jobStatusChange` records no producer, so a callback that stored its log and then died can still be masked by another producer's status written afterwards. The real fix is to make the artifact write and the outcome write one transaction so presence implies announcement, which means splitting `storeJobFile` into upload and record. Own card, not a fourth guess here.
- **An errored delivery whose email fails is treated as announced.** The status commits before the send, so a retry finds it and does not try again. This is **not** the same as `main`, which retries the email for log-only errors because it never suppresses those repeats, though only as a side effect of the duplicate-email bug this PR fixes. A durable fix is an outbox.
- **The scanner and containerizer routes still append their status after a decision.** They ignore `storeJobFile`'s result, so a delayed `JOB-ERRORED` correctly leaves the shared artifact alone but is still recorded, and `JOB-ERRORED` outranks `FILES-APPROVED` on the dashboard. Predates the branch. Own card.
- **The same logical run log can still occupy two rows across historical path changes.** Encrypted run logs have lived at `results/encrypted-results.zip`, then `results/encrypted-logs.zip`, now `results/encrypted-code-run-log.zip`. Same job and type at different paths is legal under the index, and the read-side collapse this replaced could not merge them either, since it keyed on path. Not fixable in the migration: those `path` values address real objects at the old keys, so rewriting them would break fetching. Requires a legacy job to receive a fresh delivery.
- **Two windows remain around the S3 upload.** Keys can be wrapped, or the round closed, between the guard and the upload; and two concurrent uploads to one key can leave the row naming one payload while holding the other's bytes. Closing either needs a lock held across the upload.
- **The contention that `FOR UPDATE` protects is not unit-tested.** `pg-transactional-tests` runs each test on one connection inside a single transaction, so two requests cannot contend and any assertion would come down to query interleaving. Sequential repeats are covered; the concurrent path is enforced by the database.
- **A legacy job holding two types on one path still displays that object twice**, and **Iris's "a set of errors per run" is not implementable at the database layer**: paths are job-scoped, and a `CODE-CHANGES-REQUESTED` resubmit reuses the job, so run two overwrites run one in S3. Run-scoped paths fix both and would let this index become plain and unqualified.

## Tests

- `study_job_file_artifact_slot.test.ts`: the index rejects a second row for a slot, permits differing types on one path, and leaves code files unconstrained. The backfill is exercised on temporary tables of the same names inside one transaction (`pg_temp` precedes the search path, `ON COMMIT DROP` keeps the pooled connection clean), asserting the newest row survives with every recipient key carried onto it including one the survivor already held, and that an approved row referencing a dropped duplicate is repointed rather than blocking the delete.
- `storage.test.ts`: a repeat delivery updates rather than duplicates; an artifact whose keys are already shared is not replaced even on an open round; a late delivery on a decided round is ignored and reports `stored: false`; a first-time artifact after a decision is still stored; a legacy row of another type on the results path is left alone; two first deliveries collapse to one row with exactly one reported as new.
- `job-scan-results/route.test.ts`: a repeat `CODE-SCANNED` does not duplicate log rows, and once the encrypted half is shared a re-delivery leaves the plaintext half's stored bytes untouched (asserted on content, since both halves update in place).
- `results/route.test.ts`: a re-delivered errored run leaves one log row and one `JOB-ERRORED` and sends no second email; a retry carrying nothing new still records an outcome that was never written; an earlier stage's `JOB-ERRORED` does not mask a run failure that was never announced; a delivery on a decided round stores its unseen artifact but records no status and sends no email; a delivery carrying anything new still completes normally.

Full unit suite green: 241 files, 2268 tests, against `main` merged in.


src/hooks/use-encrypted-files-panel.ts
```diff
@@ -75,13 +75,16 @@ export function useEncryptedFilesPanel({ job, onFilesApproved, isReviewer }: Opt
...
+    // Matched on storage path rather than row id: the action resolves a researcher's keys by path, so
+    // it can answer through a different duplicate row than the one dedupe kept, and an id comparison
+    // would then hide an artifact they can actually open. Matching the fileType instead would be too
+    // wide, since a job can hold rows of one type at more than one path (the run-log path changed
+    // twice historically) and the ones with no keys would render as rows that never unlock.
+    const accessiblePathSet = useMemo(() => new Set((encryptedFiles ?? []).map((f) => f.path)), [encryptedFiles])
     const visibleRows = useMemo(
-        () => (isReviewer ? encryptedRows : encryptedRows.filter((f) => accessibleIdSet.has(f.id))),
-        [isReviewer, encryptedRows, accessibleIdSet],
+        () => (isReviewer ? encryptedRows : encryptedRows.filter((f) => accessiblePathSet.has(f.path))),
```
nathanstitt: Visibility moves to path-matching here, but the "shared with lab" indicator ~30 lines below doesn't — `sharedIdSet.has(f.id)` on the locked branch and `sharedIdSet.has(f.sourceId)` on the decrypted one, both fed by `getSharedFileIdsForJob`, which returns `studyJobFileRecipientKey.studyJobFileId`, i.e. row ids.

So the two halves of this hook now key on different things. On a job holding duplicates where the keys hang off the row dedupe *didn't* keep, the artifact renders `locked` for a reviewer even though it was shared. `dedupeJobArtifactFiles` preferring the keyed row makes that mostly unreachable today — but "mostly unreachable" is the reasoning the description explicitly declines to lean on elsewhere, and the id-vs-path split is exactly what this change is fixing on the other side.

Could we resolve `sharedFileIds` to paths (or have `getSharedFileIdsForJob` return paths) so both sides agree?

src/server/storage.ts
```diff
@@ -28,15 +30,90 @@ export async function urlForStudyDocumentFile(info: MinimalStudyInfo, fileType:
...
+
+    // If the write below fails (or the process dies between these two writes), the S3 object is
     // orphaned with no row pointing at it. Left to an S3 lifecycle/sweeper rather than a 2-phase commit.
     await storeS3File(info, file.stream(), path)
 
-    return await db
+    if (existing) {
+        const updated = await db
+            .updateTable('studyJobFile')
+            .set({ name: file.name, fileType })
```
nathanstitt: The update sets `name` and `fileType` but not `sourceId`, while the insert below does. For the current callers that's fine (none of them pass one), but it makes the two branches quietly disagree: a future caller that passes `sourceId` gets it honored on first delivery and silently dropped on the re-delivery path. Worth either setting it in the `.set()` too, or dropping the parameter since nothing uses it today.

src/server/storage.ts
```diff
@@ -28,15 +30,90 @@ export async function urlForStudyDocumentFile(info: MinimalStudyInfo, fileType:
...
+
+    // Once the round is decided the artifact has been released, and a late re-delivery is stale by
+    // definition. Overwriting the object would strand the researcher: their per-file keys were wrapped
+    // from the AES keys of the ciphertext being replaced, so an already-released file would stop
+    // decrypting. Ignore the delivery entirely. A first-ever arrival for a path is still stored -
+    // nothing was released for it, so there is nothing to protect and dropping it would lose data.
+    //
+    // The sender gets a success response either way (it has nothing to do differently), so log the
+    // drop: without it, content that was accepted but never stored leaves no trace to diagnose from.
+    if (existing && (await roundIsClosed(info.studyJobId))) {
```
nathanstitt: Worth thinking through what this does to the `CODE-CHANGES-REQUESTED` rescan case the PR body calls out as the thing #876 got wrong.

`ROUND_CLOSING_JOB_STATUSES` is `FILES-APPROVED`/`FILES-REJECTED`, and per `getOrCreateCurrentRoundJob` those open a *new* job — so a job that has one shouldn't be receiving fresh scan/run deliveries at all, and this guard should only ever fire on genuinely late re-deliveries. That reads correct to me. But it does mean the guard is effectively dead for the scanner and containerizer paths (which write on an open round), so the only protection for those is the in-place update. That seems intended given the trade explained in the description — flagging it mostly so the asymmetry is a deliberate, recorded choice rather than something we rediscover later.

src/app/api/job/[jobId]/results/route.ts
```diff
@@ -44,12 +44,25 @@ export const POST = wrapApiOrgAction(async (req: Request, { params }: { params:
...
+        stored.push(await storeStudyEncryptedResultsFile(info, results))
+    }
+
+    // OTTER-642: an errored run ends at JOB-ERRORED and never reaches the RUN-COMPLETE guard above, so
+    // a re-delivered error used to append a second JOB-ERRORED and re-send the reviewer email. When
+    // every artifact this delivery carried was already on file it is a repeat: the artifacts were
+    // refreshed in place above, and the run's outcome is announced only once. A delivery carrying
+    // anything the job did not have (including the retry that completes a half-stored one) is not a
+    // repeat and still records its status and email, so no delivery is ever permanently rejected.
+    const isRepeatDelivery = stored.length > 0 && stored.every((artifact) => !artifact.isNew)
```
Comment (nathanstitt) [inline src/app/api/job/[jobId]/results/route.ts]: One edge worth naming: when the round is already closed, `storeJobFile` returns `isNew: false` *without storing anything* (the `roundIsClosed` early return). So a late delivery on a decided round lands here as a repeat and returns `200 { detail: 'artifacts already received' }` — which is accurate for the row, but the artifacts it carried were dropped, not received. The `logger.warn` in storage is the only trace.

Probably the right behavior (the sender has nothing to do differently either way), but the response text reads as a stronger promise than what happened. Maybe worth distinguishing the two so an on-call reading the TOA's logs isn't misled.

src/server/storage.ts
```diff
@@ -28,15 +30,90 @@ export async function urlForStudyDocumentFile(info: MinimalStudyInfo, fileType:
...
+ *
+ * Two races are accepted rather than closed. The check and the write here are separate statements,
+ * not one atomic upsert, so two concurrent deliveries can both find no row and both insert; the read
+ * side collapses that in `dedupeJobArtifactFiles`. And the round can close between the check below
+ * and the upload, so a re-delivery landing in that window still replaces a just-released ciphertext.
+ * Both windows are narrow, both are strictly better than the unconditional overwrite this replaced,
+ * and closing them properly means a unique index plus row locks held across an S3 upload, which is
+ * the destructive migration this approach exists to avoid.
+ */
 async function storeJobFile(info: MinimalJobInfo, path: string, file: File, fileType: FileType, sourceId?: string) {
```
nathanstitt: `storeJobFile` now issues two extra round trips before the upload on every single delivery — the `existing` lookup, plus `roundIsClosed` when a row is found. Both are cheap and indexed-ish, and this is a webhook path rather than a hot loop, so I don't think it matters; just noting that the check is unconditional rather than only on the retry path, in case ingest volume ever changes that calculus.

src/lib/file-type-helpers.ts
```diff
@@ -77,3 +83,61 @@ export function jobHasEncryptedArtifacts(files: { fileType: FileType }[]): boole
...
+ */
+export function dedupeJobArtifactFiles<T extends DedupableJobFile>(files: T[]): T[] {
+    const winnerByPath = new Map<string, T>()
+    for (const file of files) {
+        if (isCodeFileType(file.fileType)) continue
+        const current = winnerByPath.get(file.path)
+        winnerByPath.set(file.path, current ? preferredArtifact(current, file) : file)
+    }
+
+    return files.filter((file) => isCodeFileType(file.fileType) || winnerByPath.get(file.path) === file)
```
nathanstitt: The filter uses reference identity (`winnerByPath.get(file.path) === file`) against the same array it built the map from, which is correct here. Just flagging the sharp edge for whoever touches this next: if a caller ever passes an array where the same object appears twice (e.g. a `concat` of overlapping lists), both occurrences pass the identity test and the row comes back doubled — the exact failure this function exists to prevent. A `Set` of winners, or deduping by id on the way out, would be immune to that. Non-blocking, since no current caller does it.

src/server/db/queries.ts
```diff
@@ -31,8 +32,27 @@ export async function siUser(throwIfNotFound = true): Promise<SiUser | null> {
...
 }
 
+// Whether this artifact has been released to researchers, which is what `dedupeJobArtifactFiles`
+// needs to avoid collapsing away the row their wrapped keys hang off. Raw SQL so one definition can
+// be reused by both file-list subqueries below: a typed `eb.exists(...)` callback would have to be
+// written against each outer query's own table union. Column names are snake_case here because the
+// CamelCasePlugin does not rewrite raw fragments; the alias is transformed, matching the other columns.
+// The correlated `study_job_file.id` is the unaliased table name, so a subquery that ever selects
+// `studyJobFile as f` has to correlate this fragment itself rather than reuse it as-is.
+const hasRecipientKeys = sql<boolean>`exists (
```
nathanstitt: Nice touch documenting the snake_case-in-raw-SQL trap and the unaliased-`study_job_file` correlation constraint. Since that constraint is silent (a subquery aliasing the table would compile and correlate to the wrong scope), it might be worth a `getStudyJobFileOfType`-style comment at the *use* sites too — but this is already far more than most raw fragments get.

src/server/actions/study-job.actions.ts
```diff
@@ -263,34 +264,42 @@ export const fetchEncryptedJobFilesAction = new Action('fetchEncryptedJobFilesAc
...
-            )
-            .where('fingerprint', '=', userKey.fingerprint)
+            .innerJoin('studyJobFile', 'studyJobFile.id', 'studyJobFileRecipientKey.studyJobFileId')
+            .select(['studyJobFile.path', 'studyJobFileRecipientKey.filePath', 'studyJobFileRecipientKey.crypt'])
+            .where('studyJobFile.studyJobId', '=', studyJob.studyJobId)
+            .where('studyJobFileRecipientKey.fingerprint', '=', userKey.fingerprint)
+            // Oldest first, so when duplicate rows carry a key for the same inner file the newest one
+            // wins the merge below. Without an ORDER BY the winner is whatever the planner returned
+            // last, which can leave a researcher decrypting on one page load and failing on the next.
+            .orderBy('studyJobFile.createdAt', 'asc')
```
nathanstitt: Good catch adding the explicit ordering here — an unordered merge that silently picks whatever the planner returned last is precisely the kind of thing that manifests as "works on my page load, fails on theirs" and takes a week to pin down. The comment explaining *why* the ORDER BY exists (rather than just that it does) is what makes this survive the next refactor.

nathanstitt (changes requested):
Claude left a lot of other comments but I think we can ignore them, and focus on a simpler question: 

*Do we need this complexity, or can we simplify it down to a unique index + error checking (or upsert)*

It seems like we can avoid all this by:
 1.  writing a migration that deletes existing dupes and adds a uniq index
 2. update the insert to be an upsert (or try/catch?)

what do you think?

therealmarv:
Yes, it does that now. The read-side dedupe is gone, and what's left is a migration, a unique index, and an idempotent write.

Two deviations from the literal version. The index is keyed on `(study_job_id, path, file_type)` rather than `(study_job_id, path)`, because jobs from before June 2025 wrote the run log and the results to the same path, so a path-only index can't be built without deleting rows that mean different things. And it's partial, excluding `MAIN-CODE`/`SUPPLEMENTAL-CODE`, since `sanitizeFileName` lets two uploads resolve to one path and a total index would 500 the submit instead of writing a harmless duplicate.

For the write I used `ON CONFLICT DO NOTHING` rather than try/catch, because a raised violation aborts the surrounding transaction and the recovery lookup after the catch wouldn't run.

What I deliberately left alone is listed at the bottom of the description.

src/server/storage.ts
```diff
@@ -28,14 +30,157 @@ export async function urlForStudyDocumentFile(info: MinimalStudyInfo, fileType:
...
+    await storeS3File(info, file.stream(), path)
+
+    if (existing) return { ...(await renameArtifactRow(existing.id, file.name)), isNew: false, stored: true }
+
+    // DO NOTHING rather than letting the unique index raise: a raised violation aborts the surrounding
+    // transaction, so the recovery below could not run if a caller ever wrapped this in one. Left
+    // untargeted because inferring a partial index as the arbiter means restating its predicate here,
+    // and a second place to keep in sync is exactly what this design is removing. The only other
+    // unique index on the table is the primary key, which a fresh v7uuid cannot collide with.
+    const inserted = await db
```
nathanstitt, line 153: `isNew` looks dead now. It's computed on all four return paths and documented in the docblock above, but the only consumer — the results route — declares it in the `deliveries` type and never reads it; repeat detection moved to the `artifactsHeldSince` comparison. Grepping the non-test source at this SHA, `isNew` appears only in this file and in that unused type annotation.

Worth dropping it and the paragraph describing it, or wiring it back if it's meant to be the repeat signal. Right now it reads as load-bearing to anyone reasoning about the flag, and a future caller could reasonably trust it to mean something the route has stopped relying on.

src/server/storage.ts
```diff
@@ -28,14 +30,157 @@ export async function urlForStudyDocumentFile(info: MinimalStudyInfo, fileType:
...
+}
 
+// Why a re-delivery must not replace this artifact, or null when replacing it is safe.
+//
+// Sharing is checked first and separately from the round, because the two do not coincide: a reviewer
+// approving CODE can already re-wrap keys for researchers (approveJobCode persists them alongside
+// CODE-APPROVED), and that round stays open afterwards. A round-status-only guard therefore left every
+// artifact shared at code approval, scan logs included, replaceable by a delayed scanner or
+// containerizer delivery. Keys are the thing being protected, so key existence is the question to ask.
+async function unreplaceableReason(studyJobFileId: string, studyJobId: string): Promise<string | null> {
```
nathanstitt, line 71: Good change — checking "are keys already wrapped" before "is the round decided" closes the hole the old description had to list as a residual limit, since `approveJobCode` persists keys while the round is still open.

One cost worth naming: this is now two sequential round trips on the *hot* path, before every single delivery that finds an existing row, where the old code did one. They're independent, so `Promise.all` would halve the latency — though it'd mean always paying for both rather than short-circuiting on the first. Probably not worth it at webhook volume; flagging in case ingest ever gets chattier.

src/database/migrations/1780600000000_dedupe_study_job_file_artifacts.ts
```diff
@@ -0,0 +1,165 @@
...
+    // COMMITTED: an approval committing a recipient key against a loser in between would be visible
+    // to the delete but not to the copy, so the cascade would take a key that was never carried
+    // forward and a researcher would lose a released file. SHARE ROW EXCLUSIVE blocks writers while
+    // still allowing plain reads, and both tables are named because the key insert is the write that
+    // has to be shut out. The table is small and this runs once, so the wait is momentary.
+    await sql`LOCK TABLE study_job_file, study_job_file_recipient_key IN SHARE ROW EXCLUSIVE MODE`.execute(db)
+
+    // Keys move before the delete, not after: study_job_file_recipient_key cascades on
+    // study_job_file_id, so deleting a row takes its per-recipient wrapped AES keys with it. Those
+    // keys are a researcher's only way into an already-released file and exist nowhere else, so a
```
nathanstitt, line 99: Copying keys onto the survivor before the delete — rather than choosing a survivor that already holds keys — is the right call, and stronger than what I'd suggested. It handles keys spread across *multiple* duplicates, where any single-survivor tiebreak would drop some.

`ON CONFLICT DO NOTHING` against the `(study_job_file_id, file_path, fingerprint)` unique constraint is the correct dedupe key here. One thing worth stating in the comment: if two losers hold *different* `crypt` values for the same `(file_path, fingerprint)`, this keeps whichever the planner inserts first and silently discards the other. Both wrap the same AES key from the same ciphertext so either decrypts — but that's a real assumption, and it's currently implicit.

src/database/migrations/1780600000000_dedupe_study_job_file_artifacts.ts
```diff
@@ -0,0 +1,165 @@
...
+        DELETE FROM study_job_file
+         WHERE id IN (SELECT id FROM ranked WHERE rn > 1)
+    `.execute(db)
+}
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    const { redundantRows, slotsAffected, crossTypeSlots } = await inventory(db)
+
+    // console.warn because it is the channel eslint allows and because an irreversible delete is
+    // worth a warning rather than a debug line.
```
nathanstitt, line 150: The `inventory()` pre-flight logging what's about to be destroyed — and specifically calling out `crossTypeSlots` as the number that can't be recovered afterwards — is exactly the right instinct for an irreversible migration.

One suggestion: `console.warn` fires *after* the counts are read but the collapse runs unconditionally right after. For a migration that deletes rows and cascades keys, would it be worth failing loudly instead if `crossTypeSlots` comes back non-zero in production, rather than logging and proceeding? The comment says that case "needs its own card" — a hard stop would guarantee someone actually opens it, where a warning in a deploy log is easy to scroll past.

nathanstitt (approved):
looks great @therealmarv but probably we should remove the `isNew` as outlined below, then fine to merge

src/server/storage.ts
therealmarv: You're right, it was dead. Nothing read it: the route declared it in the `deliveries` type and only ever used `stored` and `createdAt`, so the flag and the paragraph describing it were both lying about being load-bearing.

Dropped in 592a8b0d, along with the type annotation and the three test assertions that were the only remaining readers. The docblock now covers just `stored` and `createdAt`: https://github.com/safeinsights/management-app/blob/592a8b0d/src/server/storage.ts#L98

src/server/storage.ts
therealmarv: Leaving it sequential, but your latency point is stronger than the short-circuit makes it look. The short-circuit only pays off when keys *are* wrapped, which is the rare case. The common delivery finds an existing row with no keys on an open round, so it runs both checks anyway, exactly as you describe.

Still not worth `Promise.all` here: two indexed existence checks on a webhook that fires a handful of times per job, against the cost of always paying for both and of having to hold both results to pick the reason for the log line. If ingest ever gets chattier this is the right first change though.

src/database/migrations/1780600000000_dedupe_study_job_file_artifacts.ts
therealmarv: Agreed that assumption shouldn't stay implicit, since it's the whole reason a plain `DO NOTHING` is enough. Written into the comment in 592a8b0d: every duplicate describes the same S3 object, so the AES key being wrapped is the same and either wrapping opens it, which is why keeping whichever lands first is safe rather than merely convenient.

https://github.com/safeinsights/management-app/blob/592a8b0d/src/database/migrations/1780600000000_dedupe_study_job_file_artifacts.ts#L102

src/database/migrations/1780600000000_dedupe_study_job_file_artifacts.ts
therealmarv: Not adding the hard stop, and chasing your question turned up a wrong claim in my own comment, which is what made a hard stop look reasonable.

The comment said `crossTypeSlots` "cannot be recovered afterwards". That's false. Those rows are left in place, so the same `GROUP BY study_job_id, path HAVING count(DISTINCT file_type) > 1` still answers it at any time after the migration. The count that genuinely can't be recomputed is the redundant-row one, because those rows are gone. Corrected in 592a8b0d: https://github.com/safeinsights/management-app/blob/592a8b0d/src/database/migrations/1780600000000_dedupe_study_job_file_artifacts.ts#L35

Which also answers the suggestion: nothing is destroyed on the cross-type count's account, so failing the migration over it would block a deploy on a display nicety, and the workaround under pressure would be to comment the check out. The number staying queryable is what makes logging it sufficient.

PR #920, opened 2026-07-29
OTTER-674: hide the autosave indicator while a field error is showing

Rod0352:
## Why

An empty-field error did not dismiss "All changes saved", so both showed at once. The label
also pushed the resubmission note's error message down, away from the input field.

## What changed

`SaveStatusIndicator` gains an `isVisible` prop; every autosave surface passes `!error`, so
the error always wins the slot: the collaborative editor footer (all proposal text fields +
resubmission note), the proposal form's title/datasets/PI indicators, and the single-user
note section. With the label gone the error sits directly under the field — no layout changes
needed. Unit tests cover the ticket's repro (type → clear) and the hidden-indicator states.

### DIFF
--- a/src/components/editable-text/collaborative-editor.tsx
+++ b/src/components/editable-text/collaborative-editor.tsx
@@ -27,9 +27,9 @@ import { Toolbar } from './toolbar'
 import { EscapeFocusPlugin } from './escape-focus-plugin'
 import { widgetBlurHandler } from '@/components/form-field'
 
-function SaveStatus({ provider }: { provider: HocuspocusProvider | null }) {
+function SaveStatus({ provider, isVisible }: { provider: HocuspocusProvider | null; isVisible: boolean }) {
     const status = useProviderSaveStatus(provider)
-    return <SaveStatusIndicator status={status} />
+    return <SaveStatusIndicator status={status} isVisible={isVisible} />
 }
 
 type ActiveEditor = { userId: string; name: string; color: string; focusing: boolean }
@@ -416,7 +416,9 @@ export function CollaborativeEditor({
                 </Paper>
                 <Stack gap={4} mt={4}>
                     <Group align="center" wrap="nowrap">
-                        <SaveStatus provider={activeProvider} />
+                        {/* The caller renders the error message right below this row; hiding the
+                            indicator keeps the two from co-existing (OTTER-674). */}
+                        <SaveStatus provider={activeProvider} isVisible={!error} />
                         {footerRight && <Box ml="auto">{footerRight}</Box>}
                     </Group>
                     <ActiveEditorsList providerRef={providerRef} currentUserId={userId} />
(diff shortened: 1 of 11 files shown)


src/components/study/collaborative-resubmission-note-section.tsx
```diff
@@ -108,7 +108,7 @@ export const CollaborativeResubmissionNoteSection: FC<CollaborativeResubmissionN
                         <Box id={fieldErrorId('resubmissionNote')}>
                             <InputError error={error} />
                         </Box>
-                        <SingleUserSaveStatus isVisible={singleUserEditing} autosaveStatus={autosaveStatus} />
+                        <SingleUserSaveStatus isVisible={singleUserEditing && !error} autosaveStatus={autosaveStatus} />
```
nathanstitt, line 111: Worth noting a small inconsistency this leaves behind: the single-user sibling, `resubmission-note-section.tsx`, already solves OTTER-674 a different way — `const footerStatus = error ? <InputError .../> : <SaveStatusIndicator .../>` — so after this PR the codebase has two idioms for one rule (`isVisible={!error}` in three places, a ternary swap in the fourth).

Since `isVisible` is what CLAUDE.md actually prescribes ("have the component accept an `isVisible` prop and return null when it shouldn't render"), converting that one too would make all four surfaces grep-able as a single behavior. It's outside this diff's stated scope, so equally fine to leave — just flagging it so the next person touching an autosave surface doesn't have to guess which pattern applies.

src/components/editable-text/collaborative-editor.tsx
```diff
@@ -416,7 +416,7 @@ export function CollaborativeEditor({
                 </Paper>
                 <Stack gap={4} mt={4}>
                     <Group align="center" wrap="nowrap">
-                        <SaveStatus provider={activeProvider} />
+                        <SaveStatus provider={activeProvider} isVisible={!error} />
```
nathanstitt, line 423: `error` is typed `React.ReactNode` (line 188), so `!error` is a truthiness check on a node rather than on a message string. Correct for every current call site, but `ReactNode` admits a couple of falsy-but-present values — notably `0` and `''` — where an "error" would read as absent and the saved label would stay up.

Nobody realistically passes `0` as an error, so this isn't a live bug. Just noting the invariant ("callers pass undefined/null when there's no error") is currently implicit. `Boolean(error)` wouldn't change anything; making it airtight would mean narrowing the prop to `string | undefined`.

src/components/save-status.tsx
```diff
@@ -27,12 +27,15 @@ const SavedLabel: FC = () => {
...
+    /** False while the field's validation error is showing — the error takes the slot (OTTER-674). */
+    isVisible?: boolean
 }
 
 // Single autosave indicator shared across every surface (collaborative editor,
 // proposal fields, resubmission note). Renders nothing until there is something
 // to report so it can be dropped under any field unconditionally.
-export const SaveStatusIndicator: FC<SaveStatusIndicatorProps> = ({ status }) => {
+export const SaveStatusIndicator: FC<SaveStatusIndicatorProps> = ({ status, isVisible = true }) => {
+    if (!isVisible) return null
```
nathanstitt, line 38: Good call defaulting `isVisible` to `true` — the call sites that don't pass it keep working, so this stays purely additive.

The guard's placement is also the right one: a hidden indicator short-circuits before `status` is read at all. Worth preserving that ordering if this function ever grows more states.

nathanstitt (approved):
looks good just a few comments around isVisible logic

PR #921, opened 2026-07-30
OTTER-667: Add job-errored screen with security key decryption

stetradis:
## Summary

- Adds the "Job Errored - Before Decryption" screen for Data Partners, redirecting reviewers here when `resultsErrored` is true and no files decision has been made
- Implements the full security key entry flow: section header, input with error states, View/Decrypting button, "Lost your key?" popover, and decryption via the existing `useDecryptFiles` hook
- All security key sub-components (`FormSectionHeader`, `SecurityKeyInput`, `SecurityKeyViewButton`, `LostKeyPopover`) are standalone reusable components with their own test files
- Adds an optional `onError` callback to `useDecryptFiles` so callers can suppress the default toast when handling errors inline

## Changes

**Routing & screen registration**
- `screens.ts` — new `'reviewer-outputs-errored'` screen ID
- `reviewer-screen-rules.ts` — new rule: `resultsErrored && !resultsApproved && !resultsRejected` takes precedence over `hasResults`
- `registry.ts` — registers `ReviewerOutputsErroredScreen`

**Screen & form**
- `reviewer-outputs-errored-screen.tsx` — server component assembling the page header, step header, action alert ("Code errored - {Date}"), security key form, and "Previous step" navigation
- `security-key-form.tsx` — client component wiring the hook to the reusable sub-components
- `use-security-key-form.ts` — form state machine: empty validation, delegates decryption to `useDecryptFiles`, maps errors to Figma copy, refocuses input on failure

**Reusable components**
- `FormSectionHeader` — title with optional required indicator, divider, description
- `SecurityKeyInput` — `Textarea` with `forwardRef`, `autoComplete="off"`, `aria-required`, error with `role="alert"` and `aria-describedby`
- `SecurityKeyViewButton` — View/Decrypting states with `Loader`, `disabled`, `aria-live="polite"` announcement
- `LostKeyPopover` — keyboard-operable popover (Enter/Space/Escape, focus return), new-tab link with `aria-label`

**Shared changes**
- `status-alert.tsx` — updated action variant icon (`WarningCircleIcon`) and colors to match Figma
- `use-decrypt-files.ts` — optional `onError` callback; existing callers are unaffected
- `theme.ts` — `lg` spacing overridden to `1.5rem` (24px), added `red.10` and `yellow.10` color stops
- `generate-keys.tsx` — updated `gap={24}` to `gap="lg"` for consistency

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx
@@ -1,12 +1,12 @@
 import dayjs from 'dayjs'
 import { Box, Group, Stack } from '@mantine/core'
 import { CaretLeftIcon } from '@phosphor-icons/react/dist/ssr'
-import { isSubmittedStudy } from '@/schema/study'
 import { AlertNotFound } from '@/components/errors'
 import { ButtonLink } from '@/components/links'
 import { StudyPageHeader } from '@/components/study/study-page-header'
 import { ProposalStepHeader } from '@/components/study/proposal-step-header'
 import { StatusAlert, STATUS_ALERT_VARIANT } from '@/components/study/status-alert'
+import { SecurityKeyForm } from '@/components/study/security-key-form'
 import { Routes } from '@/lib/routes'
 import { latestSubmittedJobForStudy } from '@/server/db/queries'
 import type { ScreenComponentProps } from './types'
@@ -30,10 +30,6 @@ export async function ReviewerOutputsErroredScreen({
     study,
     orgSlug,
 }: Pick<ScreenComponentProps, 'study' | 'orgSlug'>) {
-    if (!isSubmittedStudy(study)) {
-        return <AlertNotFound title="Study was not found" message="No such study exists" />
-    }
-
     const job = await latestSubmittedJobForStudy(study.id)
     if (!job) {
         return <AlertNotFound title="No submission found" message="This study has no submitted code to review." />
@@ -51,9 +47,12 @@ export async function ReviewerOutputsErroredScreen({
                 <ProposalStepHeader
                     stepLabel="STEP 3"
                     heading="Review outputs"
-                    studyTitle={study.title}
+                    studyTitle={study.title ?? ''}
                     banner={<ErroredBanner erroredAt={erroredAt} />}
                 />
+
+                <SecurityKeyForm />
+
                 <Group>
                     <ButtonLink
                         href={Routes.studyReviewCode({ orgSlug, studyId: study.id })}
--- a/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx
@@ -11,12 +11,6 @@ import { Routes } from '@/lib/routes'
 import { latestSubmittedJobForStudy } from '@/server/db/queries'
 import type { ScreenComponentProps } from './types'
 
-function erroredTimestamp(
-    statusChanges: ReadonlyArray<{ status: string; createdAt: Date | string }>,
-): Date | string | null {
-    return statusChanges.find((c) => c.status === 'JOB-ERRORED')?.createdAt ?? null
-}
-
 const ErroredBanner = ({ erroredAt }: { erroredAt: Date | string }) => (
     <StatusAlert
         variant={STATUS_ALERT_VARIANT.action}
@@ -30,12 +24,16 @@ export async function ReviewerOutputsErroredScreen({
     study,
     orgSlug,
 }: Pick<ScreenComponentProps, 'study' | 'orgSlug'>) {
+    // Uses the same "latest submitted job" anchor as the state machine's latestJob()
+    // so the job here always matches the one that set state.resultsErrored.
+    // The not-found guards below are unreachable via normal routing but protect against
+    // direct URL navigation that bypasses the state machine.
     const job = await latestSubmittedJobForStudy(study.id)
     if (!job) {
         return <AlertNotFound title="No submission found" message="This study has no submitted code to review." />
     }
 
-    const erroredAt = erroredTimestamp(job.statusChanges)
+    const erroredAt = job.statusChanges.find((c) => c.status === 'JOB-ERRORED')?.createdAt ?? null
     if (!erroredAt) {
         return <AlertNotFound title="No error found" message="This study has not encountered an error." />
     }
--- a/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx
@@ -1,0 +1,69 @@
+import dayjs from 'dayjs'
+import { Box, Group, Stack } from '@mantine/core'
+import { CaretLeftIcon } from '@phosphor-icons/react/dist/ssr'
+import { isSubmittedStudy } from '@/schema/study'
+import { AlertNotFound } from '@/components/errors'
+import { ButtonLink } from '@/components/links'
+import { StudyPageHeader } from '@/components/study/study-page-header'
+import { ProposalStepHeader } from '@/components/study/proposal-step-header'
+import { StatusAlert, STATUS_ALERT_VARIANT } from '@/components/study/status-alert'
+import { Routes } from '@/lib/routes'
+import { latestSubmittedJobForStudy } from '@/server/db/queries'
+import type { ScreenComponentProps } from './types'
+
+function erroredTimestamp(
+    statusChanges: ReadonlyArray<{ status: string; createdAt: Date | string }>,
+): Date | string | null {
+    return statusChanges.find((c) => c.status === 'JOB-ERRORED')?.createdAt ?? null
+}
+
+const ErroredBanner = ({ erroredAt }: { erroredAt: Date | string }) => (
+    <StatusAlert
+        variant={STATUS_ALERT_VARIANT.action}
+        title={`Code errored \u2022 ${dayjs(erroredAt).format('MMM DD, YYYY')}`}
+    >
+        Enter your security key below to access the outputs and see what went wrong.
+    </StatusAlert>
+)
+
+export async function ReviewerOutputsErroredScreen({
+    study,
+    orgSlug,
+}: Pick<ScreenComponentProps, 'study' | 'orgSlug'>) {
+    if (!isSubmittedStudy(study)) {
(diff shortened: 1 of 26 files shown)


src/hooks/use-decrypt-files.ts
```diff
@@ -81,7 +82,11 @@ export function useDecryptFiles(options: {
         if (err instanceof KeyParseError || err instanceof DecryptionError) {
             form.setFieldError('privateKey', err.message)
         }
-        reportMutationError('decryption failed')(err)
+        if (onError) {
+            onError(err)
```
nathanstitt, line 86: (Anchoring here, but this is about `if (!encryptedFiles) return []` a few lines down in `mutationFn` — that line is outside the diff.)

That early-return has a different meaning now that a caller treats `onSuccess` as "the key was accepted". `mutationFn` resolves to `[]` — and therefore fires `onSuccess` — in two cases that aren't successes:

- `encryptedFiles` is still `undefined` (query in flight, or the query errored)
- the action legitimately returned `[]` — `fetchEncryptedJobFilesAction` does that when the user has no public key on file, or when the job has no encrypted artifacts

In `useSecurityKeyForm` that means the reviewer types literal garbage, gets "Security key accepted.", and no key was ever parsed. The `isLoadingFiles` guard in the hook narrows the first case but doesn't close it (`isLoading` is false for an errored query), and it does nothing for the second.

Could we make the no-files case explicit rather than a silent empty success? Something like throwing when `!encryptedFiles?.length` so it lands in `onError`, or having `decryptFiles` be the only path that can resolve. Worth a look — the existing `useEncryptedFilesPanel` caller only sets `decryptedFiles`, so it never noticed, but this new caller renders an affirmative claim off it.

src/components/study/use-security-key-form.ts
```diff
@@ -0,0 +1,65 @@
...
+
+        const trimmed = value.trim()
+        if (!trimmed) {
+            setError(ERRORS.empty)
+            setSuccessMessage(undefined)
+            inputRef.current?.focus()
+            return
+        }
+
+        if (isLoadingFiles) return
```
nathanstitt, line 58: This bails out silently — if the encrypted-files query is still loading, the click does nothing at all: no error, no spinner, no state change. From the reviewer's side that reads as a dead button.

Since it sits *below* the empty-value check, we've already cleared the "blank" case, so the user has typed a real key and gets zero feedback. Could we either disable the View button while `isLoadingFiles` (the button already has a `isDecrypting` disabled state we could widen), or set a transient "Loading outputs…" state so the click is visibly acknowledged?

src/components/study/use-security-key-form.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const SUCCESS_MESSAGE = 'Security key accepted.'
+
+export function useSecurityKeyForm({ job }: { job: LatestJobForStudy }) {
+    const [value, setValue] = useState('')
+    const [error, setError] = useState<string>()
+    const [successMessage, setSuccessMessage] = useState<string>()
+    const inputRef = useRef<HTMLTextAreaElement>(null)
+
+    const { data: encryptedFiles, isLoading: isLoadingFiles } = useQuery({
```
nathanstitt, line 22: The `useQuery` here has no error handling — if `fetchEncryptedJobFilesAction` rejects, `encryptedFiles` stays `undefined`, `isLoadingFiles` flips to `false`, and `handleSubmit` sails past the guard on line 49 straight into `decrypt(trimmed)`. That hits the `if (!encryptedFiles) return []` branch and reports "Security key accepted." on a failed fetch.

The sibling `useEncryptedFilesPanel` handles this by catching in the `queryFn`, reporting to Sentry, and surfacing a field error. Worth mirroring that here so a fetch failure is distinguishable from a good key.

src/lib/study-screen/reviewer-screen-rules.ts
```diff
@@ -5,9 +5,13 @@ import type { ScreenRuleEntry } from './screen-rules'
 // Transcribes the legacy review/page.tsx cascade with the ?from= cases removed (those are routing,
 // not screen-selection). See docs/plans/2026-06-23-reviewer-screen-state-machine-design.md §4.
 export const REVIEWER_SCREEN_RULES = [
-    // 1. Results exist → results-only Study Details (OTTER-538). Out-ranks the code decision
-    //    (CODE-APPROVED is always present once results land), mirroring legacy
-    //    `decisionMade = hasLiveCodeDecision && !hasResultsStatus`.
+    // 1a. Job errored, no files decision yet → errored outputs view (OTTER-667). The reviewer
+    //     enters their security key to decrypt error logs before making a files decision.
+    ['reviewer-outputs-errored', { when: (s) => s.resultsErrored && !s.resultsApproved && !s.resultsRejected }],
```
nathanstitt: This predicate is character-for-character the one already exported as `isErroredResultHiddenFromResearcher` in `state.ts` (`s.resultsErrored && !s.resultsApproved && !s.resultsRejected`), which the researcher table imports and uses on its line 11.

Since the two tables now have to agree on "has a files decision been made" for routing to stay coherent, could we import the shared helper here too rather than restating it? The name is researcher-flavoured, so maybe worth renaming to something role-neutral like `awaitingFilesDecisionOnError` and having both tables call it — that way a future change to the definition can't drift the two roles apart.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { StudyPageHeader } from '@/components/study/study-page-header'
+import { ProposalStepHeader } from '@/components/study/proposal-step-header'
+import { StatusAlert, STATUS_ALERT_VARIANT } from '@/components/study/status-alert'
+import { SecurityKeyForm } from '@/components/study/security-key-form'
+import { Routes } from '@/lib/routes'
+import { latestSubmittedJobForStudy } from '@/server/db/queries'
+import type { ScreenComponentProps } from './types'
+
+function erroredTimestamp(
+    statusChanges: ReadonlyArray<{ status: string; createdAt: Date | string }>,
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx]: The `status: string` here widens away the `StudyJobStatus` union that `job.statusChanges` actually carries, so a typo in the `'JOB-ERRORED'` literal on line 17 would type-check fine and just silently never match (rendering the "No error found" not-found page).

Could we type it as `ReadonlyArray<{ status: StudyJobStatus; createdAt: Date | string }>` — or just drop the standalone helper and inline `job.statusChanges.find(...)` so it infers from the query type directly?

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx
```diff
@@ -0,0 +1,68 @@
...
+    orgSlug,
+}: Pick<ScreenComponentProps, 'study' | 'orgSlug'>) {
+    const job = await latestSubmittedJobForStudy(study.id)
+    if (!job) {
+        return <AlertNotFound title="No submission found" message="This study has no submitted code to review." />
+    }
+
+    const erroredAt = erroredTimestamp(job.statusChanges)
+    if (!erroredAt) {
+        return <AlertNotFound title="No error found" message="This study has not encountered an error." />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-errored-screen.tsx:38]: Worth thinking about whether this branch is reachable in a way that hurts. The screen rule selects this screen off `state.resultsErrored`, which is derived from the *latest* job's status set — but the screen independently re-fetches via `latestSubmittedJobForStudy`. Those are two different queries (`latestJobForStudyQuery` vs. the submitted-filtered variant), so a study whose newest job is `INITIATED`-only could route here off the older errored job and then land on the not-found page.

If that's genuinely unreachable it'd be good to say so in a comment; otherwise this renders a dead-end for a state the router thinks is valid.

src/components/study/lost-key-popover.tsx
```diff
@@ -0,0 +1,79 @@
...
+                            if (e.key === 'Enter' || e.key === ' ') {
+                                e.preventDefault()
+                                toggle()
+                            }
+                            if (e.key === 'Escape' && opened) {
+                                e.stopPropagation()
+                                closeAndRestoreFocus()
+                            }
+                        }}
+                        role="button"
```
nathanstitt: Putting `role="button"` + `tabIndex={0}` + key handling on an SVG works, but we're hand-rolling what `ActionIcon`/`UnstyledButton` gives for free — and we lose the native bits (Space-on-keyup semantics, `:focus-visible` ring, disabled handling, form-agnostic activation). Mantine's `Popover.Target` is happy to wrap an `ActionIcon` with the icon inside.

The tests here are good and cover the behaviour, so this is a maintainability nudge rather than a defect — the custom version has to keep re-earning that coverage.

src/components/study/security-key-input.tsx
```diff
@@ -0,0 +1,25 @@
+import { Textarea, type TextareaProps } from '@mantine/core'
+import { forwardRef } from 'react'
+
+type SecurityKeyInputProps = Omit<TextareaProps, 'autoComplete' | 'aria-required'>
+
+export const SecurityKeyInput = forwardRef<HTMLTextAreaElement, SecurityKeyInputProps>(
+    ({ error, disabled, ...props }, ref) => {
+        const borderColor = disabled ? undefined : error ? 'var(--mantine-color-red-7)' : 'var(--mantine-color-blue-7)'
```
nathanstitt, line 9: Hardcoding a blue border on the idle state means this input renders as though it's focused even when it isn't, which is a bit of a hijack of the focus affordance. Is that deliberate from the Figma, or should the blue be the `:focus` state only? If it's intentional it'd be worth a `// why` comment, since it reads like a bug at a glance.

src/theme.ts
```diff
@@ -140,6 +142,7 @@ export const theme = createTheme({
     primaryColor: 'purple',
     spacing: {
         xs: '0.5rem',
+        lg: '1.5rem',
```
nathanstitt: Redefining `lg` shifts spacing for every existing `gap="lg"` / `p="lg"` / `m="lg"` in the app, not just the new screens — Mantine's default `lg` is `1.25rem`, so everything currently on `lg` grows by 4px.

The PR body frames this as a design-token alignment, which is plausibly correct, but it's a global visual change riding along in a feature PR. Did we sweep the existing `lg` usages to confirm none regress? If this came out of the OTTER-661 token work it might belong on that branch instead, where it'd get looked at as a theme change rather than as a footnote.

nathanstitt (approved):
Looks good ty, comments are all minor

src/theme.ts
stetradis: This was implemented to align theme spacing with the figma dimensions, which should be applied globally - but agree that this doesn't need to ship as part of this feature, it looks like the same fix is landing in this [PR](https://github.com/safeinsights/management-app/pull/923/changes#diff-0e21fb1c9519a397050defc2148a2d1b966a0982cc4cf196e1e6b007f8d095e5R252) - will revert here 😃

PR #922, opened 2026-07-30
Otter 668 outputs available

Rod0352:
Summary

* Adds the "Outputs Available - Before Decryption" screen for Data Partners, shown when the run has completed (`RUN-COMPLETE`) and no files decision exists
* Action alert "Outputs are available for review • {Date}" with body copy naming the submitting lab; reuses the OTTER-667 `SecurityKeyForm` unmodified
* Post-decryption review/approval is OTTER-675 — this screen stops at a successful decrypt, like 667's errored flow

Changes
Routing & screen registration

* `screens.ts` — new `'reviewer-outputs-available'` screen ID
* `reviewer-screen-rules.ts` — new rule between errored and study-results: `hasResults` with no decision and no error
* `registry.ts` — registers `ReviewerOutputsAvailableScreen`

Screen

* `reviewer-outputs-available-screen.tsx` — page header, step header, action alert, `SecurityKeyForm`, "Previous step" → code page; not-found guards for missing job or missing `RUN-COMPLETE`

Tests

* New screen tests (banner, lab name, form, nav, guards) and rule-table cases; stale expectation in `resolve.test.ts` updated
* `study-flow.spec.ts` — results review e2e now drives the decrypt-only screen; approve step deferred to OTTER-675

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.tsx
@@ -1,0 +1,75 @@
+import dayjs from 'dayjs'
+import { Box, Group, Stack } from '@mantine/core'
+import { CaretLeftIcon } from '@phosphor-icons/react/dist/ssr'
+import { AlertNotFound } from '@/components/errors'
+import { ButtonLink } from '@/components/links'
+import { StudyPageHeader } from '@/components/study/study-page-header'
+import { ProposalStepHeader } from '@/components/study/proposal-step-header'
+import { StatusAlert, STATUS_ALERT_VARIANT } from '@/components/study/status-alert'
+import { SecurityKeyForm } from '@/components/study/security-key-form'
+import { Routes } from '@/lib/routes'
+import { latestSubmittedJobForStudy } from '@/server/db/queries'
+import type { ScreenComponentProps } from './types'
+
+// statusChanges arrive newest-first (see latestJobForStudyQuery ordering), so `find` picks the
+// most recent RUN-COMPLETE — the moment the outputs became available for review.
+function availableTimestamp(
+    statusChanges: ReadonlyArray<{ status: string; createdAt: Date | string }>,
+): Date | string | null {
+    return statusChanges.find((c) => c.status === 'RUN-COMPLETE')?.createdAt ?? null
+}
+
+const AvailableBanner = ({ availableAt, labName }: { availableAt: Date | string; labName: string }) => (
+    <StatusAlert
+        variant={STATUS_ALERT_VARIANT.action}
+        title={`Outputs are available for review \u2022 ${dayjs(availableAt).format('MMM DD, YYYY')}`}
+    >
+        Enter your security key to decrypt the outputs, review them, and then share with {labName}.
+    </StatusAlert>
+)
+
+export async function ReviewerOutputsAvailableScreen({
+    study,
+    orgSlug,
+}: Pick<ScreenComponentProps, 'study' | 'orgSlug'>) {
+    const job = await latestSubmittedJobForStudy(study.id)
+    if (!job) {
+        return <AlertNotFound title="No submission found" message="This study has no submitted code to review." />
+    }
+
+    const availableAt = availableTimestamp(job.statusChanges)
+    if (!availableAt) {
+        return (
+            <AlertNotFound
+                title="Outputs not found"
+                message="This study does not have outputs available for review yet."
+            />
+        )
+    }
+
+    return (
+        <Box bg="grey.10">
+            <Stack px="xl" gap="xxl" py="xl">
+                <StudyPageHeader>Secondary analysis study</StudyPageHeader>
+                <ProposalStepHeader
+                    stepLabel="STEP 3"
+                    heading="Review outputs"
+                    studyTitle={study.title ?? ''}
+                    banner={<AvailableBanner availableAt={availableAt} labName={study.submittingLabName} />}
+                />
+
+                <SecurityKeyForm job={job} />
+
+                <Group>
+                    <ButtonLink
+                        href={Routes.studyReviewCode({ orgSlug, studyId: study.id })}
+                        variant="subtle"
+                        leftSection={<CaretLeftIcon />}
+                    >
+                        Previous step
+                    </ButtonLink>
+                </Group>
+            </Stack>
+        </Box>
+    )
+}
--- a/src/components/study/security-key-form.tsx
+++ b/src/components/study/security-key-form.tsx
@@ -1,8 +1,9 @@
 'use client'
 
-import { Button, Paper, Stack, Textarea } from '@mantine/core'
+import { Button, Paper, Stack } from '@mantine/core'
 import { FormSectionHeader } from '@/components/study/form-section-header'
 import { LostKeyPopover } from '@/components/study/lost-key-popover'
+import { SecurityKeyInput } from '@/components/study/security-key-input'
 
 export function SecurityKeyForm() {
     return (
@@ -13,12 +14,7 @@ export function SecurityKeyForm() {
                     description="This key is required to access the outputs. It was issued to you during sign-up."
                     required
                 />
-                <Textarea
-                    autoComplete="off"
-                    aria-required
-                    styles={{ input: { minHeight: 72, borderColor: 'var(--mantine-color-blue-7)' } }}
-                    maw={800}
-                />
+                <SecurityKeyInput />
                 <div>
                     <Button size="sm">View</Button>
                 </div>
--- a/src/components/study/security-key-form.tsx
+++ b/src/components/study/security-key-form.tsx
@@ -1,10 +1,8 @@
 'use client'
 
-import { Anchor, Button, Group, Paper, Popover, Stack, Text, Textarea } from '@mantine/core'
-import { useDisclosure } from '@mantine/hooks'
-import { InfoIcon } from '@phosphor-icons/react/dist/ssr'
-import { Routes } from '@/lib/routes'
+import { Button, Paper, Stack, Textarea } from '@mantine/core'
 import { FormSectionHeader } from '@/components/study/form-section-header'
+import { LostKeyPopover } from '@/components/study/lost-key-popover'
 
 export function SecurityKeyForm() {
     return (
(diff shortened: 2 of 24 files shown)


src/lib/study-screen/reviewer-screen-rules.ts
```diff
@@ -5,9 +5,22 @@ import type { ScreenRuleEntry } from './screen-rules'
 // Transcribes the legacy review/page.tsx cascade with the ?from= cases removed (those are routing,
 // not screen-selection). See docs/plans/2026-06-23-reviewer-screen-state-machine-design.md §4.
 export const REVIEWER_SCREEN_RULES = [
-    // 1. Results exist → results-only Study Details (OTTER-538). Out-ranks the code decision
-    //    (CODE-APPROVED is always present once results land), mirroring legacy
-    //    `decisionMade = hasLiveCodeDecision && !hasResultsStatus`.
+    // 1a. Job errored, no files decision yet → errored outputs view (OTTER-667). The reviewer
+    //     enters their security key to decrypt error logs before making a files decision.
+    ['reviewer-outputs-errored', { when: (s) => s.resultsErrored && !s.resultsApproved && !s.resultsRejected }],
```
nathanstitt, line 10: This branch is behind `main` and this line silently reverts #921. `main` currently reads:

```ts
import { awaitingFilesDecisionOnError } from './state'
// ...
['reviewer-outputs-errored', { when: (s) => awaitingFilesDecisionOnError(s) }],
```

The inlined predicate here is *behaviourally* identical today, but `awaitingFilesDecisionOnError` exists specifically so this rule and `isExecuting` in `state.ts:102` stay in lockstep — `isExecuting` calls it too, and the comment above it explains why a bare `JOB-ERRORED` must not close the execution window. Inlining re-opens the drift the helper was extracted to prevent.

Worth merging `main` in and re-resolving so we keep the helper (and so the diff shrinks to just the outputs-available work — right now `reviewer-outputs-errored-screen.tsx`, `security-key-*`, `lost-key-popover`, `form-section-header`, `use-decrypt-files` and the theme/status-alert edits all show up as new here even though they're already on `main`).

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { Routes } from '@/lib/routes'
+import { latestSubmittedJobForStudy } from '@/server/db/queries'
+import type { ScreenComponentProps } from './types'
+
+// statusChanges arrive newest-first (see latestJobForStudyQuery ordering), so `find` picks the
+// most recent RUN-COMPLETE — the moment the outputs became available for review.
+function availableTimestamp(
+    statusChanges: ReadonlyArray<{ status: string; createdAt: Date | string }>,
+): Date | string | null {
+    return statusChanges.find((c) => c.status === 'RUN-COMPLETE')?.createdAt ?? null
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.tsx:19]: There's a mismatch between the rule that routes here and the guard that renders here, and it can strand a study on an error screen.

The 1b rule (`reviewer-screen-rules.ts:17`) fires on `s.hasResults && !approved && !rejected && !errored`. But `hasResults` is `STUDY_RESULTS_JOB_STATUSES` = `['RUN-COMPLETE', 'FILES-APPROVED', 'FILES-REJECTED', 'JOB-ERRORED']`, so those three negations exist only to subtract the other three back out. `projectStudyState` already computes exactly that residue: `resultsDisplayStatus` (`state.ts:90`), priority-ordered `FILES-APPROVED > FILES-REJECTED > JOB-ERRORED > RUN-COMPLETE`.

Could we key the rule on that instead?

```ts
['reviewer-outputs-available', { when: (s) => s.resultsDisplayStatus === 'RUN-COMPLETE' }],
```

Same set today, but it stops the rule re-deriving a fact the state machine already owns — and more importantly it makes routing key on **the same fact this screen's guard checks**. Right now `resolveScreen` derives state from the state-machine's `latestJob()`, while this screen independently re-queries `latestSubmittedJobForStudy(study.id)`. Two anchors, one decision: if they ever diverge we route here on one job and then render `"Outputs not found"` from another.

Worth considering the same for the guard while we're in here. `code-decision-screen.tsx:15` and `reviewer-code-feedback-screen.tsx:51` both take the `raw` prop the screen already receives and call `projectStudyState(raw)` rather than re-querying. If this screen did the same for the *decision* (is there a `RUN-COMPLETE`?) and used `latestSubmittedJobForStudy` only for the job payload `SecurityKeyForm` needs, the divergence can't happen at all.

(The errored screen carries a comment about exactly this anchoring — "Uses the same 'latest submitted job' anchor as the state machine's `latestJob()` so the job here always matches". Rather than copying that comment over, I'd rather the code not need it.)

src/lib/study-screen/reviewer-screen-rules.ts
```diff
@@ -5,9 +5,22 @@ import type { ScreenRuleEntry } from './screen-rules'
...
+    //     RUN-COMPLETE also landed.
+    [
+        'reviewer-outputs-available',
+        { when: (s) => s.hasResults && !s.resultsApproved && !s.resultsRejected && !s.resultsErrored },
+    ],
+
+    // 1c. Results exist → results-only Study Details (OTTER-538). Out-ranks the code decision
+    //     (CODE-APPROVED is always present once results land), mirroring legacy
+    //     `decisionMade = hasLiveCodeDecision && !hasResultsStatus`.
     ['reviewer-study-results', { when: (s) => s.hasResults }],
```
nathanstitt, line 24: With 1a and 1b both landing above it, is `reviewer-study-results` still reachable in any state the 1a/1b predicates don't already claim? Walking it: rule 3 requires `hasResults`; 1a takes errored-undecided; 1b takes `hasResults && !approved && !rejected && !errored`. So rule 3 only survives when a decision exists (approved/rejected) *or* when errored+decided. That's the intent per the new tests, but the comment above still reads "Results exist → results-only Study Details" which now overstates what it catches. Might be worth rewording to "results exist **and a files decision was recorded**" so the next reader doesn't have to re-derive the interaction.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { ProposalStepHeader } from '@/components/study/proposal-step-header'
+import { StatusAlert, STATUS_ALERT_VARIANT } from '@/components/study/status-alert'
+import { SecurityKeyForm } from '@/components/study/security-key-form'
+import { Routes } from '@/lib/routes'
+import { latestSubmittedJobForStudy } from '@/server/db/queries'
+import type { ScreenComponentProps } from './types'
+
+// statusChanges arrive newest-first (see latestJobForStudyQuery ordering), so `find` picks the
+// most recent RUN-COMPLETE — the moment the outputs became available for review.
+function availableTimestamp(
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.tsx:16]: `availableTimestamp` here and `erroredTimestamp` in `reviewer-outputs-errored-screen.tsx:14` are the same function with a different status literal, and the two screens are otherwise near-identical (same `Box`/`Stack`/`StudyPageHeader`/`ProposalStepHeader`/`SecurityKeyForm`/`Previous step` scaffold, same two not-found guards). Now that there are two, could we pull a shared `firstStatusAt(statusChanges, status)` helper — or even a single `ReviewerOutputsDecryptScreen` taking `{ status, banner }` — so a third variant doesn't mean a third copy?

Also a small type nit: this signature widens `status` to `string` where the errored screen's does too, but `job.statusChanges` is typed with `StudyJobStatus`. Narrowing the param to `StudyJobStatus` would let a typo like `'RUN_COMPLETE'` fail at compile time instead of silently returning `null` and rendering "Outputs not found".

tests/study-flow.spec.ts
```diff
@@ -281,78 +281,67 @@ function uploadResults(jobId: string): void {
...
-    await expect(page.getByText(/The code errored/i)).toBeVisible()
-    await expect(page.getByText(/Job ID/i)).toBeVisible()
-
-    // Verify logs row exists in the results table (JobResults now renders EncryptedFilesPanel,
-    // which lists the log artifact by its file-type label before decryption).
-    await expect(page.getByText('Code Run Log')).toBeVisible()
-}
+// OTTER-675: restore once the reviewer approve flow is built — the researcher's
+// errored view is gated on file approval (isErroredResultHiddenFromResearcher).
+// async function verifyFailedStatusDisplay(page: Page, studyTitle: string): Promise<void> {
```
nathanstitt, line 332: Commenting out `verifyFailedStatusDisplay` (and its call site at line 465) drops the only e2e coverage of the researcher's errored-study view — the dashboard `Errored` pill, the "The code errored" copy, the Job ID, and the `Code Run Log` row. The OTTER-675 note explains *why*, which is good, but commented-out code tends to rot: by the time 675 lands, `getByPlaceholder`-era selectors in here may no longer match anything.

Could we use `test.fixme()` (or `test.skip()` with the same OTTER-675 annotation) on a real test instead? That keeps it type-checked and lint-visible, and Playwright will report it as skipped rather than it vanishing silently from the suite.

tests/study-flow.spec.ts
```diff
@@ -281,78 +281,67 @@ function uploadResults(jobId: string): void {
...
     await expect(page.getByText('Review Studies')).toBeVisible()
     await viewStudyDetails(page, studyTitle)
     await page.waitForURL(/\/review$/)
 
+    await expect(page.getByText(/Outputs are available for review/)).toBeVisible()
+    await expect(page.getByRole('heading', { name: /security key/i })).toBeVisible()
+
     const privateKey = await readTestSupportFile('private_key.pem')
-    const privateKeyTextarea = page.getByPlaceholder('Enter your Results Key to access encrypted content.')
+    const privateKeyTextarea = page.getByRole('textbox')
```
nathanstitt, line 297: `page.getByRole('textbox')` is unscoped — it'll throw strict-mode if any other textbox ever renders on this screen (a search field in the header, a future notes box). The old selector was specific for a reason. Since `SecurityKeyInput` renders inside a labelled section, could we anchor it, e.g. `page.getByRole('textbox', { name: /security key/i })` — or scope to the form `Paper` — so this doesn't become a flake source later? Same at line 319.

src/components/study/security-key-form.tsx
```diff
@@ -0,0 +1,50 @@
...
+                />
+                <SecurityKeyViewButton isDecrypting={isDecrypting} onClick={handleSubmit} />
+                <SuccessMessage message={successMessage} />
+                <LostKeyPopover />
+            </Stack>
+        </Paper>
+    )
+}
+
+// TODO: remove this after 675 is implemented. this is temporary for testing.
```
nathanstitt, line 42: Small thing since this is already on `main`, but flagging while we're here: `SUCCESS_MESSAGE` is the assertion both e2e tests now hang their success check on (`await expect(page.getByText('Security key accepted.')).toBeVisible()`). When 675 removes this TODO block, both e2e helpers break. Might be worth a note here that the e2e suite depends on it, so whoever deletes it knows to update `study-flow.spec.ts` in the same change.

nathanstitt (changes requested):
Looks good overall but check the comment about stranding a a study on an error screen, that does seem like something that could occur.  The rest are pretty minor though

PR #923, opened 2026-07-30
OTTER-661: bring the Mantine theme in line with the design tokens

nathanstitt:
Brings `src/theme.ts` in line with the design tokens and adds the semantic layer the card asks for. Supersedes #896 (see below).

**Source:** the *Interim design tokens - handoff - Jul 2026* DTCG file, which supersedes the Figma export linked on the card. Its handoff note is explicit that it is hand-corrected and that a fresh Figma export would reintroduce two issues it fixes (line-height units, prototype styles).

## What's in it

- **Palette** — all nine ramps updated; **Navy** and **Turquoise** added. Grey and Purple keep their 950 step in slot 10; the malformed 11th entries on green/yellow are gone.
- **Semantic layer** (`src/theme/tokens.ts`) — each role bound to a ramp position exactly as the handoff states it, emitted as `--si-color-*` via `cssVariablesResolver`. Components reference roles, never shades.
- **Scales the theme never defined** — spacing, radius, shadows, and the composed type styles. These silently used Mantine's defaults before.
- **Components** (`src/components/ui/`) — `Heading`/`Text` plus Button/Alert/TextInput overrides, carried over from #896.
- **Ladle catalog** — Foundations and Components stories that read every value back out of the live theme, so the catalog *is* the verification surface.

## Bug this fixes along the way

Two call sites read `theme.colors.grey[10]` — the old out-of-contract 11th entry — as a **light** surface (`#F1F3F5`). Under the token file slot 10 is Grey 950 `#212529`, so a naive port would have turned modal headers, table headers, and the Ladle canvas near-black. Both are repointed at the semantic surface they actually meant.

## Relationship to #896

#896's *structure* is carried over intact — it referenced semantic tokens rather than raw shades, so it ported by rebinding. Its navy-primary palette is not: the handoff keeps purple as the brand color. One difference worth noting — #896 flagged its bindings as *inferred, needing UX sign-off*; these are stated outright in the handoff file, so that caveat is gone.

## Open questions

1. **Font family.** The card flags Inter @ 500 and says *confirm with design first*. **Not done here** — the July handoff specifies Open Sans, so this keeps Open Sans. Needs a decision on which source wins.
2. **Yellow badge contrast.** `<Badge color="yellow">` resolves to Yellow 500 `#ffdd75` with white text via `primaryShade: 5` — roughly **1.4:1**, illegible. The new `warning.*` tokens are the fix (yellow.9 on yellow.0, AA-compliant), but migrating call sites is out of scope here. Same class of issue for `color="green"`.
3. **Shadow color.** The handoff decomposes shadows into blur/spread/offset/opacity with **no color**; these compose over black. Worth confirming with design. Its `xs` blur (3px) is also larger than `sm` (2px), which looks like an authoring slip on their end.
4. **Catalog sidebar title** is `AA Design system` purely to sort first. Functional but ugly; Ladle supports explicit ordering in `config.mjs` if preferred.
5. **`spacing.lg` shifts 20px → 24px** (Mantine default → token), affecting existing `p="lg"` at ~44 call sites. `radius.md` shifts 10px → 8px at 6 sites. Every other scale key already matched.
6. **Not a migration.** Existing call sites still pass old color names as string props; those degrade quietly. Only the two `theme.colors.X[n]` reads were remapped, because that form throws. The follow-up card covers the rest.

## Verification

`pnpm run checks` and `pnpm run test:unit` (2160 passed) green. Catalog built and loaded in a browser — ramps, semantic tokens, type scale, radius/shadow/spacing all render from the live theme.

Visual spot-check of key app pages is still outstanding; unit tests don't assert on theme values, so green there confirms nothing broke, not that it looks right.

Companion infra PR (publishes this catalog on every PR preview): safeinsights/iac#191

### DIFF
--- a/.ladle/backgrounds.ts
+++ b/.ladle/backgrounds.ts
@@ -8,7 +8,7 @@
 // the app-canvas grey via `pageBackgroundArgTypes` on their meta, matching the AppShellMain area.
 
 const WHITE = '#ffffff'
-const APP_CANVAS = 'var(--mantine-color-grey-10)'
+const APP_CANVAS = 'var(--si-color-surface-canvas)'
 const NAVY = 'var(--mantine-color-purple-8)'
 
 const control = {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,14 +65,27 @@ pipeline {
                         echo "Not a merge commit"
                     }
                 }
-                sh """
-                    printenv
-                    [ -d ./cicd ] && find ./cicd -maxdepth 1 -name '*.zip' -delete
-                    aws s3 sync s3://si-mgmt-app-build/scripts ./cicd
-                    cd cicd
-                    unzip -o *.zip
-                    ./deploy
-                """
+                // The deploy reports the PR preview and component catalog URLs back to GitHub, which
+                // needs a token in the environment. 'github-app-safeinsights' is the GitHub App the
+                // Branch Source plugin already uses to post build statuses; the plugin exchanges the
+                // app key for a short-lived installation token, exposed here as the password.
+                // Posting needs pull_requests:write and statuses:write on the installation.
+                //
+                // `printenv` was dropped from this step: it predates the credential and would now
+                // echo the token into the build log.
+                withCredentials([usernamePassword(
+                    credentialsId: 'github-app-safeinsights',
+                    usernameVariable: 'GITHUB_APP_USER',
+                    passwordVariable: 'GITHUB_TOKEN'
+                )]) {
+                    sh """
+                        [ -d ./cicd ] && find ./cicd -maxdepth 1 -name '*.zip' -delete
+                        aws s3 sync s3://si-mgmt-app-build/scripts ./cicd
+                        cd cicd
+                        unzip -o *.zip
+                        ./deploy
+                    """
+                }
             }
         }
     }
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -25,6 +25,7 @@ const eslintConfig = [
             '!.ladle',
             '!.ladle/**',
             '.ladle/dist/**',
+            '.ladle/dist-standalone/**',
             'CHANGELOG.md',
             'test-results/**',
             'tests/coverage/**',
--- a/src/components/modals/app-modal.tsx
+++ b/src/components/modals/app-modal.tsx
@@ -37,7 +37,7 @@ export function AppModal({
             styles={{
                 header: {
                     padding: '0px 40px',
-                    backgroundColor: theme.colors.grey[10],
+                    backgroundColor: theme.colors.charcoal[0],
                     ...styles?.header,
                 },
                 body: {
--- a/src/components/ui/components.stories.tsx
+++ b/src/components/ui/components.stories.tsx
@@ -1,0 +1,86 @@
+import type { Story } from '@ladle/react'
+import { Alert, Button, Group, Stack } from '@mantine/core'
+import { Heading, Text } from '@/components/ui'
+
+// Variant matrices mirroring the Figma component pages so they can be compared side by side.
+//   Button 4216:2 · Alert 61:5826
+// Geometry is transcribed from Figma; colors are mapped by role through the semantic tokens,
+// because the Figma component pages are still drawn on a stale palette.
+
+const meta = { title: 'AA Design system / Components' }
+export default meta
+
+const SIZES = ['xs', 'sm', 'md', 'lg', 'xl'] as const
+
+// Figma's variant axis -> the Mantine equivalent. `error` is a color, not a variant, in Mantine.
+const BUTTON_VARIANTS = [
+    { label: 'filled', props: { variant: 'filled' } },
+    { label: 'error (Figma) → color="red"', props: { variant: 'filled', color: 'red' } },
+    { label: 'outline', props: { variant: 'outline' } },
+    { label: 'subtle', props: { variant: 'subtle' } },
+    { label: 'Disable (Figma) → disabled', props: { disabled: true } },
+] as const
+
+const ALERT_COLORS = [
+    { label: 'Red', color: 'red' },
+    { label: 'Green', color: 'green' },
+    { label: 'Purple', color: 'purple' },
+    { label: 'Blue', color: 'blue' },
+    { label: 'Yellow', color: 'yellow' },
+] as const
+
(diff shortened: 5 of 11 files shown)


Jenkinsfile
```diff
@@ -65,14 +65,27 @@ pipeline {
...
-                """
+                // The deploy reports the PR preview and component catalog URLs back to GitHub, which
+                // needs a token in the environment. 'github-app-safeinsights' is the GitHub App the
+                // Branch Source plugin already uses to post build statuses; the plugin exchanges the
+                // app key for a short-lived installation token, exposed here as the password.
+                // Posting needs pull_requests:write and statuses:write on the installation.
+                //
+                // `printenv` was dropped from this step: it predates the credential and would now
+                // echo the token into the build log.
+                withCredentials([usernamePassword(
```
nathanstitt: Please move the Jenkinsfile credential change and the `dist-standalone` ignore into their own PR.

This is CI infrastructure for the preview-catalog deploy. It is not part of the theme. Two problems come from carrying it here:

- #1045 has a byte-identical `Jenkinsfile` and `.ladle/backgrounds.ts`. One of the two PRs will conflict on a file neither card owns.
- A theme revert now also reverts the deploy credential.

The token change needs its own review. It adds `GITHUB_TOKEN` to a shell step that runs `./deploy`, a script pulled from S3 at build time.

src/lib/constants.ts
```diff
@@ -8,10 +8,10 @@ export const PROPOSAL_GRID_SPAN = {
 }
 
 export const ENCLAVE_BG = 'purple.6'
```
nathanstitt: These four constants are raw shade strings. Three of them have a semantic token added by this same PR.

```suggestion
export const ENCLAVE_BG = semanticColor('surface.sidenav.dp')
export const LAB_BG = semanticColor('surface.sidenav.rl')

// Exported so Ladle's canvas shares the app's source of truth.
export const APP_MAIN_BG = semanticColor('surface.page')
```

`tokens.ts` states that components reference roles, never raw shades. These are the app's four largest surfaces. They are the call sites that most need the indirection.

`headerBg: 'purple.8'` maps to no token. The theme is navy-primary now. Please confirm with design whether the top nav stays purple.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx
```diff
@@ -520,7 +520,7 @@ describe('SubmittedCodeSection — Security scan log', () => {
         expect(row).toHaveTextContent('Vulnerabilities found')
         const icon = row.querySelector('[data-icon="warning"]')
         expect(icon).not.toBeNull()
-        expect(icon?.outerHTML).toContain('red')
+        expect(icon?.outerHTML).toContain('--si-color-error-text')
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.test.tsx]: `renderWithProviders` mounts `MantineProvider` without `cssVariablesResolver`, so no `--si-color-*` variable is defined in any unit test.

Please add it in `tests/unit.helpers.tsx`:

```tsx
<MantineProvider theme={theme} cssVariablesResolver={cssVariablesResolver}>
```

The test at line 523 asserts the literal string `--si-color-error-text` appears in the markup. That passes even when the token resolves to nothing. With the resolver mounted, a token deleted from `semanticShades` fails a test instead of shipping an uncolored icon.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx
```diff
@@ -715,8 +715,8 @@ const SONARQUBE_LABELS: ScanStatusLabels = {
 // pairing the design system already applies to WarningCircle (see StatusAlert's action variant)
 // rather than introducing a new treatment. Provisional along with the labels above.
 const SCAN_ICON_COLORS: Partial<Record<ScanToolStatus, string>> = {
-    FAILED: 'var(--mantine-color-red-9)',
-    INDETERMINATE: 'var(--mantine-color-yellow-10)',
+    FAILED: 'var(--si-color-error-text)',
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-interactive.tsx]: Use the typed `semanticColor` helper here, not a hand-written `var()` string.

```suggestion
    FAILED: semanticColor('error.text'),
    INDETERMINATE: semanticColor('warning.text'),
```

`semanticColor` takes a `SemanticToken`, so a renamed or deleted token becomes a type error. The raw string form fails silently. Five call sites in this PR write the string by hand and three use the helper, which sets two conventions for one job. The same change applies to `code-evaluation-section.tsx:111-112`.

src/lib/status-labels.ts
```diff
@@ -13,11 +13,11 @@ export type StatusLabel = {
 }
 
 const COLORS = {
-    draft: { bg: 'grey.10', c: 'gray.9' },
+    draft: { bg: 'grey.0', c: 'gray.9' },
```
nathanstitt: This PR repointed three rows here and left the other four on Mantine's built-in ramps. `gray` and `dark` are not in `theme.colors`, so they resolve against Mantine's defaults, not the library.

```suggestion
    draft: { bg: 'grey.0', c: 'charcoal.9' },
    needsReview: { bg: 'purple.1', c: 'purple.7' },
    underReview: { bg: 'yellow.0', c: 'charcoal.9' },
    rejected: { bg: 'red.0', c: 'red.7' },
    approved: { bg: 'green.0', c: 'green.7' },
    clarification: { bg: 'blue.1', c: 'blue.8' },
    default: { bg: 'charcoal.0', c: 'charcoal.7' },
```

Pill colors are one map. Half on the library and half off it is the state a later reader will copy from.

.ladle/backgrounds.ts
```diff
@@ -8,7 +8,7 @@
 // the app-canvas grey via `pageBackgroundArgTypes` on their meta, matching the AppShellMain area.
 
 const WHITE = '#ffffff'
-const APP_CANVAS = 'var(--mantine-color-grey-10)'
+const APP_CANVAS = 'var(--si-color-surface-page)'
 const NAVY = 'var(--mantine-color-purple-8)'
```
nathanstitt: `NAVY` points at purple. Under the new palette `purple[8]` is `#100a4c`, and `navy` is now its own ramp.

```suggestion
const NAVY = 'var(--si-color-surface-sidenav)'
```

That resolves to `navy.7`, the sidenav surface. If the focused shell must stay purple, please rename the constant to `PURPLE`.

sparksam (approved):

PR #924, opened 2026-08-03
OTTER-675: review decrypted outputs and submit a sharing decision

therealmarv:
see [OTTER-675](https://openstax.atlassian.net/browse/OTTER-675)

**Now targets `main` and includes it.** `origin/main` (with #921) is merged in, so the diff is this card alone.

## What this adds

The post-decryption half of the Data Partner "job errored" outputs screen: once a valid security key decrypts the artifacts, the key form gives way to an outputs table, a Decision section, and a confirmation modal that records the reviewer's files decision.

## Key design point: this is a client phase, not a new screen

Decryption happens in the browser and changes no server state, so "advancing" to the review view is a local phase flip inside the screen #921 registered (`reviewer-outputs-errored`). `reviewer-screen-rules.ts` and `screens.ts` are untouched.

The card's AC reads "the user must be redirected to this page when status == job-errored AND they have successfully validated their security key", which sounds like routing; it is really this two-part client gate. Holding both phases in one component is also what keeps the decrypted plaintext in memory and off the server.

```
OutputsReviewPanel
  ├─ locked   : errored banner + SecurityKeyForm          (OTTER-667)
  └─ unlocked : review banner + files table + Decision    (this card)
```

Page and section headers are identical either side of the flip; only the banner and the body below it change.

## Changes

**Migrations (2)**
- `study_job_file_activity`, the per-file view/download log behind the "Last activity" column. Addressed by `(study_job_file_id, file_path)` following `study_job_file_recipient_key`: a `study_job_file` row is the encrypted *archive*, and what the reviewer opens is one inner file, so keying on the archive alone would smear one file's activity across every sibling extracted from the same zip.
- `'RESULTS'` added to `study_review_comment_kind`, plus a `study_job_id`-required CHECK mirroring the existing CODE one. The outputs decision is stored as a `studyReviewComment` exactly like a code decision.

**Server**
- `submitOutputsDecisionAction`, one action so the feedback and the files status land together and a rationale can never be orphaned from the status that acted on it. The study and org are derived from the job id, never taken alongside it. Gated on the job having reached a terminal result and having no decision yet, with the word cap derived from the job's own status rather than the request.
- `assertSharesEveryArtifact` refuses a "share outputs" that carries no usable keys or omits an artifact, so an approval the lab cannot open is never recorded.
- `approveStudyJobFilesAction` / `rejectStudyJobFilesAction` were carrying the same derive-from-job flaw and are fixed alongside.
- `recordJobFileActivityAction` / `fetchJobFileActivityAction`, scoped so a forged id cannot attach activity to another study's file. `latestActivityPerJobFile` collapses each file to its newest action via `DISTINCT ON`.
- `normalizeFeedbackToLexical` moved from `study.actions.ts` into `@/lib/lexical` and shared rather than duplicated.

**Yjs**
- New `outputs-review-feedback-{jobId}` document kind, mirrored into `services/editor/auth.ts`. Job-keyed like `code-review-feedback`, but persistence is refused once the job has a files decision, so a stale tab cannot recreate the draft the submit deletes.

**Client**
- `OutputsFilesSection` / `OutputsFileRow`: semantic table in a scroll container, 50-char truncation, tooltip on hover *and* focus, per-row `aria-label="Download {file}"`, `Download all` (≥2 files) zipping in the browser via `@zip.js/zip.js`.
- `OutputsDecisionSection`: the code-review Lexical + Yjs editor, word counter, native radio group.
- `SubmitOutputsDecisionModal`, `useOutputsDecision`, `useOutputsFiles`, `focusFirstInvalid`.
- `SecurityKeyForm` now reports decrypted files through `onDecrypted`; a key that decrypts nothing is no longer treated as success.

## Notable findings

**Focus was not returning to "Submit decision" when the modal closed.** Caught in review. The modal returned null while closed, which unmounted `ModalBase` and took the `useFocusReturn` effect inside it along with it, leaving focus on `<body>` after Escape, Cancel and X alike. The comment justifying that early return was itself wrong: `keepMounted` defaults to false and `ModalBaseContent` renders children inside the exit transition, so every open already mounts fresh copy and the body text could never have been a stale cached announcement. The modal now stays mounted and `opened={decision !== null}` lets Mantine's own `returnFocus` do the work. Asserted in the e2e run rather than jsdom, where the Lexical editor this flow needs is not typeable.

**"Modal open" was not a state of its own.** Also from review: an `isOpen` flag beside the selected decision made "open with nothing chosen" representable, and two guards were paying for the gap. Collapsed into a single `confirming: OutputsDecision | null`, which made both dead: the mutation's null check and the modal's own re-check. `focusFirstInvalid` now returns the field it moved to, so its scan doubles as the "is anything invalid" test instead of a second identical pass.

**The unique constraint alone would not have blocked a double decision.** `FILES-APPROVED`/`FILES-REJECTED` are themselves round-opening events for `codeSubmissionVersion`, so after the first decision the round advances and a second attempt would slot in beside it rather than collide, writing a phantom decision for a round whose code was never resubmitted. There is now an explicit guard; the constraint still covers the genuine same-round race.

**Mantine's `Radio.Group` swallows an `id` prop without rendering it**, so `document.getElementById` found nothing and the submit-time focus jump was a silent no-op. The id now lives on a wrapper we own. Found by rendering the real DOM, not by unit tests, which had built the wrapper by hand.

**Mantine's `Radio` renders `description` visually but never wires `aria-describedby`.** The AC requires the association, so the description is wrapped in an element we own and referenced explicitly.

## Testing

- 247 files / 2332 unit tests pass; `pnpm run checks` clean (typecheck, eslint, prettier, action validation).
- The screen test drives a genuine decrypt: it seeds a real encrypted artifact and enters the real test private key, so the phase flip is exercised end to end rather than through a stubbed callback.
- E2E: `Error log review` runs decrypt, blank-submit validation, the modal's three dismissal paths with focus assertions, then the real submit, and re-enables `verifyFailedStatusDisplay`, which #922 had commented out pending this card.
- Every state was also inspected in a browser against the Figma frames, which is how the two Mantine issues above and a missing counter unit were found.

## Where the submitted feedback is read

This card writes the decision feedback as a `studyReviewComment` with `reviewKind: 'RESULTS'`. Nothing reads that kind yet, by design; the read paths are separate cards in this epic:

- [OTTER-677](https://openstax.atlassian.net/browse/OTTER-677) DP post-review page, which declares this card a dependency and reuses this card's outputs table. It owns the "Feedback & Notes" panel and the "View outputs again" re-decrypt section (Figma `12287:25855` / `25966`).
- [OTTER-696](https://openstax.atlassian.net/browse/OTTER-696) / [OTTER-697](https://openstax.atlassian.net/browse/OTTER-697) the Research Lab views for the errored flow, outputs-and-feedback and feedback-only respectively.

Until 677 lands, submitting redirects to `/review` and the reviewer screen rules resolve to the existing post-decision screen. The redirect target is expected to change when 677 adds its rule; no change is needed here for that.

Two things worth flagging to PM before this epic closes: OTTER-697 currently has an empty description, so the lab-side read of feedback-only is unspecified; and OTTER-677 will need text-override props on `SecurityKeyForm` for its "View outputs again" heading and body.

## For Iris (non-blocking copy flags)

- AC says "character limit" / `maxCharacter`; every errored-flow frame says `words`, including the unit next to the count. Shipping words, which review agreed reads as the AC being wrong rather than the build. Still wants an explicit sign-off before the card is validated.
- AC says "Outputs files"; frame `25892` says "Output files". Shipping "Output files".
- Frames `25572`/`25724` label the table card "Review outputs", which collides with the STEP 3 page heading. Not shipped.
- The confirmation-modal frames still read "Modal header"; shipping the AC's "Submit your decision?".
- No 675 frame renders the `Title: {study title}` line the reused section header includes; kept, since the AC says reuse it unmodified.
- The feedback field is the shared Lexical editor, so it carries a formatting toolbar the Figma textarea does not show.

## One thing for QA

The "All changes saved" autosave indicator cannot be verified on a PR preview: no editor websocket connects there, so Yjs autosave is not running. The doc kind is registered and parsed correctly and the decision submits without it, so this reads as environmental, but it is the one AC row worth re-checking on QA before the card closes.

[OTTER-677]: https://openstax.atlassian.net/browse/OTTER-677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/study/submit-outputs-decision-modal.test.tsx
+++ b/src/components/study/submit-outputs-decision-modal.test.tsx
@@ -14,7 +14,6 @@ const renderModal = (decision: OutputsDecision | null, overrides: Record<string,
     const props = {
         decision,
         labName: LAB,
-        isOpen: true,
         isSubmitting: false,
         onClose: vi.fn(),
         onConfirm: vi.fn(),
@@ -49,13 +48,9 @@ describe('SubmitOutputsDecisionModal', () => {
         expect(confirmationBody('share-feedback-only', 'Acme Lab')).toContain('Acme Lab')
     })
 
+    // A null decision IS the closed state: there is no separate open flag to disagree with it.
     it('stays closed until a decision has been picked', () => {
         renderModal(null)
-        expect(screen.queryByText('Submit your decision?')).toBeNull()
-    })
-
-    it('stays closed when isOpen is false', () => {
-        renderModal('share-outputs', { isOpen: false })
         expect(screen.queryByText('Submit your decision?')).toBeNull()
     })
 
@@ -65,25 +60,16 @@ describe('SubmitOutputsDecisionModal', () => {
     // survive precisely that sequence.
     it('shows the updated body after the decision changes between opens', async () => {
         const Harness = () => {
-            const [decision, setDecision] = useState<OutputsDecision>('share-outputs')
-            const [isOpen, setIsOpen] = useState(true)
+            const [confirming, setConfirming] = useState<OutputsDecision | null>('share-outputs')
             return (
                 <>
-                    <button onClick={() => setIsOpen(false)}>close it</button>
-                    <button
-                        onClick={() => {
-                            setDecision('share-feedback-only')
-                            setIsOpen(true)
-                        }}
-                    >
-                        reopen with feedback only
-                    </button>
+                    <button onClick={() => setConfirming(null)}>close it</button>
+                    <button onClick={() => setConfirming('share-feedback-only')}>reopen with feedback only</button>
                     <SubmitOutputsDecisionModal
-                        decision={decision}
+                        decision={confirming}
                         labName={LAB}
-                        isOpen={isOpen}
                         isSubmitting={false}
-                        onClose={() => setIsOpen(false)}
+                        onClose={() => setConfirming(null)}
                         onConfirm={vi.fn()}
                     />
                 </>
@@ -100,6 +86,9 @@ describe('SubmitOutputsDecisionModal', () => {
         expect(screen.queryByText(SHARE_OUTPUTS_BODY)).toBeNull()
     })
 
+    // Focus restoration on close is the trigger owner's job (the modal unmounts itself, which
+    // pre-empts Mantine's focus-return effect), so it is asserted in
+    // reviewer-outputs-errored-screen.test.tsx where the real Submit decision button exists.
     it('closes without submitting from Cancel', async () => {
         const { onClose, onConfirm } = renderModal('share-outputs')
 
--- a/src/components/study/submit-outputs-decision-modal.test.tsx
+++ b/src/components/study/submit-outputs-decision-modal.test.tsx
@@ -1,0 +1,150 @@
+import { useState } from 'react'
+import { describe, expect, it, renderWithProviders, screen, userEvent, vi } from '@/tests/unit.helpers'
+import { SubmitOutputsDecisionModal, confirmationBody } from './submit-outputs-decision-modal'
+import type { OutputsDecision } from '@/lib/outputs-review'
+
+const LAB = 'Rice Lab'
+
+const SHARE_OUTPUTS_BODY =
+    'You are sharing the output files and your feedback with Rice Lab. You will not be able to make changes after submitting.'
+const FEEDBACK_ONLY_BODY =
+    'You are sharing your feedback only. The output files will not be shared with Rice Lab. You will not be able to make changes after submitting.'
+
+const renderModal = (decision: OutputsDecision | null, overrides: Record<string, unknown> = {}) => {
+    const props = {
+        decision,
+        labName: LAB,
+        isOpen: true,
+        isSubmitting: false,
+        onClose: vi.fn(),
+        onConfirm: vi.fn(),
+        ...overrides,
+    }
+    renderWithProviders(<SubmitOutputsDecisionModal {...props} />)
+    return props
+}
+
+describe('SubmitOutputsDecisionModal', () => {
+    it('renders the title and all three controls', () => {
+        renderModal('share-outputs')
+
+        expect(screen.getByText('Submit your decision?')).toBeInTheDocument()
+        expect(screen.getByRole('button', { name: 'Cancel' })).toBeInTheDocument()
+        expect(screen.getByRole('button', { name: 'Submit decision' })).toBeInTheDocument()
+        expect(screen.getByRole('button', { name: 'Close' })).toBeInTheDocument()
+    })
+
+    it('states that the files and feedback are being shared for the share-outputs path', () => {
+        renderModal('share-outputs')
+        expect(screen.getByText(SHARE_OUTPUTS_BODY)).toBeInTheDocument()
+    })
+
+    it('states that the files are withheld for the feedback-only path', () => {
+        renderModal('share-feedback-only')
+        expect(screen.getByText(FEEDBACK_ONLY_BODY)).toBeInTheDocument()
+    })
+
+    it('interpolates the lab name into both bodies', () => {
+        expect(confirmationBody('share-outputs', 'Acme Lab')).toContain('Acme Lab')
+        expect(confirmationBody('share-feedback-only', 'Acme Lab')).toContain('Acme Lab')
+    })
+
+    it('stays closed until a decision has been picked', () => {
+        renderModal(null)
+        expect(screen.queryByText('Submit your decision?')).toBeNull()
+    })
+
+    it('stays closed when isOpen is false', () => {
+        renderModal('share-outputs', { isOpen: false })
+        expect(screen.queryByText('Submit your decision?')).toBeNull()
+    })
+
+    // Mantine keeps modal children mounted, so a cached body from the previously chosen option is
+    // exactly what a screen reader would re-announce on the next open. Driven through a real
+    // close → change → reopen cycle rather than a prop rerender, because a stale body would
+    // survive precisely that sequence.
+    it('shows the updated body after the decision changes between opens', async () => {
+        const Harness = () => {
+            const [decision, setDecision] = useState<OutputsDecision>('share-outputs')
+            const [isOpen, setIsOpen] = useState(true)
+            return (
+                <>
+                    <button onClick={() => setIsOpen(false)}>close it</button>
+                    <button
+                        onClick={() => {
+                            setDecision('share-feedback-only')
+                            setIsOpen(true)
+                        }}
+                    >
+                        reopen with feedback only
+                    </button>
+                    <SubmitOutputsDecisionModal
+                        decision={decision}
+                        labName={LAB}
+                        isOpen={isOpen}
+                        isSubmitting={false}
+                        onClose={() => setIsOpen(false)}
+                        onConfirm={vi.fn()}
+                    />
+                </>
+            )
+        }
+
+        renderWithProviders(<Harness />)
+        expect(screen.getByText(SHARE_OUTPUTS_BODY)).toBeInTheDocument()
+
+        await userEvent.click(screen.getByRole('button', { name: 'close it' }))
+        await userEvent.click(screen.getByRole('button', { name: 'reopen with feedback only' }))
+
+        expect(screen.getByText(FEEDBACK_ONLY_BODY)).toBeInTheDocument()
+        expect(screen.queryByText(SHARE_OUTPUTS_BODY)).toBeNull()
+    })
+
+    it('closes without submitting from Cancel', async () => {
+        const { onClose, onConfirm } = renderModal('share-outputs')
+
+        await userEvent.click(screen.getByRole('button', { name: 'Cancel' }))
+
+        expect(onClose).toHaveBeenCalledTimes(1)
+        expect(onConfirm).not.toHaveBeenCalled()
+    })
+
+    it('closes without submitting from the X', async () => {
+        const { onClose, onConfirm } = renderModal('share-outputs')
+
+        await userEvent.click(screen.getByRole('button', { name: 'Close' }))
+
+        expect(onClose).toHaveBeenCalledTimes(1)
+        expect(onConfirm).not.toHaveBeenCalled()
+    })
+
+    it('closes on Escape', async () => {
+        const { onClose } = renderModal('share-outputs')
+
+        await userEvent.keyboard('{Escape}')
+
+        expect(onClose).toHaveBeenCalled()
+    })
+
+    it('submits from the confirm button', async () => {
+        const { onConfirm } = renderModal('share-outputs')
+
+        await userEvent.click(screen.getByRole('button', { name: 'Submit decision' }))
+
+        expect(onConfirm).toHaveBeenCalledTimes(1)
+    })
+
+    it('exposes itself as a modal dialog labelled by its title', () => {
+        renderModal('share-outputs')
+
+        const dialog = screen.getByRole('dialog')
+        expect(dialog).toHaveAttribute('aria-modal', 'true')
+        expect(dialog).toHaveAccessibleName('Submit your decision?')
+    })
+
+    it('disables Cancel while the submission is in flight', () => {
+        renderModal('share-outputs', { isSubmitting: true })
+
+        expect(screen.getByRole('button', { name: 'Cancel' })).toBeDisabled()
+    })
+})
--- a/src/components/study/submit-outputs-decision-modal.test.tsx
+++ b/src/components/study/submit-outputs-decision-modal.test.tsx
@@ -49,8 +49,11 @@ describe('SubmitOutputsDecisionModal', () => {
     })
 
     // A null decision IS the closed state: there is no separate open flag to disagree with it.
-    it('stays closed until a decision has been picked', () => {
+    // The component stays mounted while closed (that is what keeps Mantine's focus return alive),
+    // so "closed" means no dialog is rendered, not that nothing is.
+    it('shows no dialog until a decision has been picked', () => {
         renderModal(null)
+        expect(screen.queryByRole('dialog')).toBeNull()
         expect(screen.queryByText('Submit your decision?')).toBeNull()
     })
 
@@ -86,9 +89,8 @@ describe('SubmitOutputsDecisionModal', () => {
         expect(screen.queryByText(SHARE_OUTPUTS_BODY)).toBeNull()
     })
 
-    // Focus restoration on close is the trigger owner's job (the modal unmounts itself, which
-    // pre-empts Mantine's focus-return effect), so it is asserted in
-    // reviewer-outputs-errored-screen.test.tsx where the real Submit decision button exists.
+    // Focus restoration is Mantine's own returnFocus, which works because this component stays
+    // mounted across close; the e2e covers it end to end against a real trigger.
     it('closes without submitting from Cancel', async () => {
         const { onClose, onConfirm } = renderModal('share-outputs')
 
--- a/src/components/study/submit-outputs-decision-modal.tsx
+++ b/src/components/study/submit-outputs-decision-modal.tsx
@@ -10,10 +10,13 @@ export const confirmationBody = (decision: OutputsDecision, labName: string): st
         : `You are sharing your feedback only. The output files will not be shared with ${labName}. You will not be able to make changes after submitting.`
 
 type SubmitOutputsDecisionModalProps = {
-    /** null keeps the modal closed; submit only opens it once an option is picked. */
+    /**
+     * The decision being confirmed, or null when the modal is closed. A single nullable value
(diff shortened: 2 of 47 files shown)


src/components/study/submit-outputs-decision-modal.tsx
```diff
@@ -0,0 +1,46 @@
...
+    labName,
+    isOpen,
+    isSubmitting,
+    onClose,
+    onConfirm,
+}) => {
+    // Unmounting while closed is what keeps the body text honest: Mantine's Modal keeps its
+    // children mounted, so a cached body from the previously chosen option would be what a
+    // screen reader re-announces on the next open.
+    if (!isOpen || !decision) return null
```
nathanstitt: Verified against the PR preview (pr924) as a Data Partner: this early return means **focus is never restored to "Submit decision" when the modal closes** — via Escape, Cancel, *or* X. In all three cases `document.activeElement` ends up on `<body>`.

The card asks for this explicitly, twice — in the Confirmation Modal AC ("Focus must return to the 'Submit decision' button on close, whether closed via X, Cancel, or Submit") and again in the unit-test requirements.

The cause looks like the interaction between this line and Mantine: `AppModal` would normally restore focus to the trigger on close, but returning `null` unmounts the whole subtree first, so that effect never gets to run. The comment above is right that keeping it mounted caches the body text — I confirmed the stale-text behaviour is genuinely fixed (RB1 → close → RB2 → reopen shows the updated copy). So this isn't a case of "just drop the early return"; we'd want to keep the remount *and* restore focus.

One option is for the panel that owns the trigger to hold a ref and refocus it in the `onClose`/`closeModal` path, which keeps the fresh-body-text property intact. Happy to be wrong about the mechanism — but the symptom reproduces every time.

src/components/study/submit-outputs-decision-modal.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+    it('closes without submitting from the X', async () => {
+        const { onClose, onConfirm } = renderModal('share-outputs')
+
+        await userEvent.click(screen.getByRole('button', { name: 'Close' }))
+
+        expect(onClose).toHaveBeenCalledTimes(1)
+        expect(onConfirm).not.toHaveBeenCalled()
+    })
+
+    it('closes on Escape', async () => {
```
nathanstitt, line 112: Related to the focus finding on the component: these tests assert the modal *closes* on Escape/Cancel, but nothing asserts where focus lands afterwards — which is why the missing focus restoration got through despite otherwise thorough coverage.

Worth an assertion along the lines of `expect(trigger).toHaveFocus()` after each of the three close paths (X, Cancel, Escape). The card lists that as its own unit-test row.

src/lib/outputs-review.ts
```diff
@@ -0,0 +1,65 @@
...
+export const toOutputsReviewDecision = (decision: OutputsDecision): ReviewDecision =>
+    OUTPUTS_DECISION_TO_REVIEW[decision]
+
+export const OUTPUTS_FEEDBACK_MIN_WORDS = 1
+
+/**
+ * Feedback length caps, keyed by the run outcome being reviewed. An errored run gets the shorter
+ * cap because the reviewer is explaining a failure, not summarizing results.
+ */
+export const ERRORED_OUTPUTS_FEEDBACK_MAX_WORDS = 300
```
nathanstitt, line 26: Flagging for the record rather than as a defect, since the PR description already calls it out: the card's AC says a **300-character** limit and names `{maxCharacter}`, while this ships **words**. Confirmed on the preview — "alpha beta gamma" (16 characters) displays as `3/300`.

Agreed that the Figma frames all render `0/300 words` and that `WordCounter`/`countWords` is what the repo has, so words is the defensible read. Just noting that this is the one place the build knowingly departs from the written AC, so it needs Iris's sign-off before the card can be validated rather than being settled in review.

src/lib/outputs-review.ts
```diff
@@ -0,0 +1,65 @@
...
+        ? ERRORED_OUTPUTS_FEEDBACK_MAX_WORDS
+        : COMPLETED_OUTPUTS_FEEDBACK_MAX_WORDS
+}
+
+// File names are truncated in the table so a long name cannot push the other columns off screen.
+export const OUTPUTS_FILE_NAME_MAX_LENGTH = 50
+
+export const OUTPUTS_DECISION_ERRORS = {
+    feedbackEmpty: (labName: string) => `Enter your feedback for ${labName} before submitting.`,
+    feedbackTooLong: (maxWords: number) => `Feedback exceeds the ${maxWords} word limit. Shorten it to continue.`,
```
nathanstitt, line 70: Same words-vs-characters point, surfacing in user-visible copy: this renders as "Feedback exceeds the 300 word limit." whereas the AC's error text is "Feedback exceeds the {maxCharacter} character limit. Shorten it to continue."

Consistent with the decision above, so no change requested — it just means the copy check against the card will show a diff, and whoever signs off should know it's intentional.

src/hooks/use-outputs-decision.ts
```diff
@@ -0,0 +1,138 @@
...
+        },
+        onError: reportMutationError('Failed to submit your decision'),
+        onSuccess: () => {
+            queryClient.invalidateQueries({ queryKey: ['org-studies', orgSlug] })
+            setIsOpen(false)
+            // push() alone would be a no-op here: /review is already the current URL, so the
+            // decrypted review form would stay mounted with plaintext on screen even though the
+            // decision is final. refresh() is what re-runs the server components and lets the
+            // reviewer screen rules resolve to the post-decision screen.
+            router.push(Routes.studyReview({ orgSlug, studyId }))
```
nathanstitt, line 91: Nice detail, and it verified end-to-end on the preview: submitting advanced the study to "The results and logs have been approved and shared with the researcher" in one go. The `push()`-alone no-op reasoning in the comment is a real trap and worth having written down.

One thing I could **not** verify on a PR preview, flagging so it gets a second look on QA rather than being assumed: the "All changes saved" autosave indicator never appeared. No editor websocket connects in the preview at all (zero websocket requests), so Yjs autosave simply isn't running there — the doc kind is registered and parsed correctly in `collaboration-documents.ts`, and the decision still submitted and persisted without it, so this reads as environmental rather than a code issue. It's the one AC row I'd want re-checked against QA before the card closes.

src/hooks/use-outputs-decision.ts
```diff
@@ -0,0 +1,138 @@
...
+    const router = useRouter()
+    const queryClient = useQueryClient()
+
+    const [feedback, setFeedback] = useState('')
+    const [selected, setSelected] = useState<OutputsDecision | null>(null)
+    // Errors surface only after the user has left a field or pressed Submit; a pristine form
+    // must not open with everything flagged red.
+    const [showFeedbackError, setShowFeedbackError] = useState(false)
+    const [showDecisionError, setShowDecisionError] = useState(false)
+    const [isOpen, setIsOpen] = useState(false)
```
nathanstitt: This hook hand-rolls `touched` tracking, validation and error state that this repo already has a convention for — and there is a near-identical precedent in-tree.

`src/app/[orgSlug]/study/[studyId]/edit-and-resubmit/schema.ts` validates a **Lexical feedback field with a 300-word cap and min-1-word required rule** — the same shape as this card — through a zod schema, consumed by `useForm` in `src/contexts/edit-code-resubmit/context.tsx`. It even shares `countWordsFromLexical` with this hook. 49 files in the repo call `useForm`.

Worth noting `@/common`'s `useForm` wrapper exists *specifically* to default `validateInputOnBlur: true`, and its doc comment says: "Import `useForm` from here rather than from `@mantine/form`, otherwise the default is silently bypassed." Surfacing a required-field error once the user leaves the field is exactly what `showFeedbackError` / `showDecisionError` reimplement here.

Sketch, following the resubmit-note precedent:

```ts
// outputs-review schema, co-located like edit-and-resubmit/schema.ts
export const outputsDecisionSchema = z.object({
    feedback: z
        .string()
        .refine((v) => countWordsFromLexical(v) >= OUTPUTS_FEEDBACK_MIN_WORDS, {
            message: OUTPUTS_DECISION_ERRORS.feedbackEmpty(labName),
        })
        .refine((v) => countWordsFromLexical(v) <= maxWords, {
            message: OUTPUTS_DECISION_ERRORS.feedbackTooLong(maxWords),
        }),
    decision: z.enum(["share-outputs", "share-feedback-only"], {
        message: OUTPUTS_DECISION_ERRORS.decisionMissing,
    }),
})

const form = useForm({
    validate: zodResolver(outputsDecisionSchema),
    initialValues: { feedback: "", decision: null },
})
```

That removes `showFeedbackError`, `showDecisionError`, `resolveFeedbackError`, the `invalid` record in `attemptSubmit`, and the separate `decisionError` derivation — `form.errors` and `form.isTouched()` cover all of it, and `attemptSubmit` collapses to roughly:

```ts
const attemptSubmit = form.onSubmit(
    (values) => setConfirming(values.decision),
    (errors) => focusFirstInvalid(FIELD_ORDER, (id) => id in errors),
)
```

`focusFirstInvalid` still earns its place — Mantine has no equivalent, and the onValidationFailure callback hands you exactly the error map it needs.

**One caveat, so this is not oversold as a drop-in:** the same `@/common` doc comment notes the blur-validation default "only reaches inputs that spread `form.getInputProps(path)`", and a Lexical contenteditable and a `Radio.Group` are hand-wired. So the touched-on-blur behaviour needs explicit `onBlur` wiring either way — the win is centralised validation rules and one error source, not zero glue.

Separately, and independent of whether the form library is adopted: `isOpen` and `selected` are independent state, but `isOpen: true` is only legal when `selected !== null`. Nothing enforces it — it holds because `attemptSubmit` returns early — and two guards pay for the gap: `confirmSubmit` bails on `selected === null` (line 118) and `SubmitOutputsDecisionModal` re-checks `!isOpen || !decision`. Collapsing the pair makes the illegal state unrepresentable and both guards dead code:

```ts
const [confirming, setConfirming] = useState<OutputsDecision | null>(null)
// null = closed; non-null = open, with the decision it is confirming
```

To be explicit about what is *not* a problem, since I first wrote this up as a state-count complaint and that was wrong: separate `useState` calls for independent values is correct, and computing `wordCount` / `isEmpty` / `isOverLimit` during render rather than storing and syncing them in an effect is the right call. Gating errors on touched-or-submitted is deliberate UX, and two fields legitimately need two flags.

No behavioural bug: blank-submit, both radio variants, the over-limit state and a real submit all verified against the pr924 preview.

src/lib/outputs-review.ts
nathanstitt: I think AC is incorrect, we count words not chars

nathanstitt (changes requested):
looks good overall but think the useOutputsDecision hook could be greatly simplified by leveraging existing logic as pointed out, lets do that and then take another look

src/components/study/submit-outputs-decision-modal.tsx
therealmarv: Fixed in 1c1eaf97, and you were right about the mechanism. Checking it against the Mantine source turned up something better than the ref approach, because my comment above was simply wrong.

`keepMounted` defaults to false and `ModalBaseContent` renders children inside the exit `Transition`, so Mantine already unmounts the content once the modal closes. The body text could never have been a stale cached announcement, which means the early return was solving a problem that did not exist while breaking focus return: `useFocusReturn` lives in `useModal` inside `ModalBase`, so unmounting the wrapper took the restore effect with it.

So instead of restoring focus by hand, the modal now stays mounted and `opened={decision !== null}` lets Mantine's own `returnFocus` do it: [submit-outputs-decision-modal.tsx](https://github.com/safeinsights/management-app/blob/1c1eaf973d6e511ed3d028ca5f32112485d3eb79/src/components/study/submit-outputs-decision-modal.tsx). The one thing retained is the last non-null decision, so the body does not blank out mid-animation while the modal fades.

Verified in a browser on all three paths: Escape, Cancel and X each leave `document.activeElement` on `outputs-submit-decision`, and reopening after switching the radio shows the updated copy with no stale text and a single dialog in the DOM.

src/components/study/submit-outputs-decision-modal.test.tsx
therealmarv: Right that the gap in coverage is what let it through. Covered, though not in this file, and it is worth saying why.

Reaching the modal requires real feedback text, and the Lexical editor is not typeable in jsdom: the repo's own `single-user-editor.test.tsx` drives it through Lexical's `editor.update()` rather than typing. I tried it here and at screen level (including forcing `singleUserEditing` so the editable surface renders instead of the collaborative skeleton) and the modal never opened, because the word count stayed at 0. Asserting focus in this file would have meant building a harness with its own trigger, which tests the harness rather than the code that restores focus.

So the three paths are asserted in the e2e, where the editor does accept typing: [study-flow.spec.ts](https://github.com/safeinsights/management-app/blob/1c1eaf973d6e511ed3d028ca5f32112485d3eb79/tests/study-flow.spec.ts) now does Escape, then the X, then reopens for the real submit, checking `toBeFocused()` on the trigger after each. Green in CI on 1c1eaf97.

I left a comment in this file pointing at where it is covered, so the next person does not read the omission as an oversight.

src/lib/outputs-review.ts
therealmarv: Agreed, and thanks for saying so plainly. Worth adding one piece of evidence I found while inspecting the frames: the errored-flow ones render the unit next to the count, literally `0/300 words`. We were dropping it, so I added an optional `unit` to `WordCounter` (default off, so the other call sites are untouched) and the counter now reads `9/300 words`, matching the design.

Still leaving it as needing Iris's explicit sign-off rather than settled here, and it is called out in the PR description. No code change beyond the unit.

src/lib/outputs-review.ts
therealmarv: Noted, and left as is for consistency with the counter. The copy diff against the card is intentional and listed in the PR description so whoever signs off sees it rather than discovering it.

src/hooks/use-outputs-decision.ts
therealmarv: Thanks, and the autosave observation is the useful half. Your reading matches mine: no editor websocket reaches a PR preview, so Yjs autosave cannot run there, and the decision persisting without it is consistent with that being environmental.

I have added it to the PR description as an explicit QA item rather than leaving it implied, since it is the one AC row this PR genuinely cannot demonstrate.

src/hooks/use-outputs-decision.ts
therealmarv: Took the second half, declined the first, and I want to be concrete about why rather than just asserting scope.

**Adopted: the illegal state.** You are right that `isOpen: true` with `selected === null` was unrepresentable-in-practice but not unrepresentable-in-types, and that two guards were paying for it. Collapsed to `confirming: OutputsDecision | null` in 5acda0f0, which killed both: `confirmSubmit`'s null check and the modal's `!isOpen || !decision`. The modal now takes one nullable prop instead of two, so the illegal combination cannot be constructed at the boundary either. I also used `focusFirstInvalid`'s return value as the "is anything invalid" test, dropping the duplicate `FIELD_ORDER.some(...)` scan.

**Declined: the zod plus `useForm` swap.** I read `edit-and-resubmit/schema.ts` and `ResubmissionNoteSection` before deciding, and the precedent surfaces its over-limit *message* only on blur: `onBlur={() => noteForm.validateField('resubmissionNote')}`, with the counter turning red live. Our AC wants the error state itself live while typing: *"If more than {maxCharacters} are inserted, allow the user to continue typing but return an error state while highlighting the field"*, plus a unit-test row for *"Error state clears if user edits back under"*.

Mantine validates per field, not per rule, so neither default gets both behaviours: `validateInputOnBlur` loses the live over-limit, and `validateInputOnChange` would also flag the pristine empty field, which the same AC forbids. Splitting them (schema for empty and decision, derived value for over-limit) means one field's error arriving from two sources, which reads worse than the single `resolveFeedbackError` with its two ordered branches.

Combined with the caveat you already flagged, that the blur glue stays manual for a Lexical contenteditable and a `Radio.Group` either way, the win shrinks to centralising two rules that are currently four lines. Happy to revisit if the live-over-limit requirement gets relaxed, or if you think Mantine can express per-rule validation timing and I have missed it.

**One correction to my own code that your review led to.** Chasing the focus finding, I checked the Mantine source and found the comment on this hook's sibling was wrong about `keepMounted`. Details in the modal thread; the upshot is the manual focus restore is gone in favour of Mantine's `returnFocus`, verified in a browser on all three dismissal paths.

nathanstitt (approved):
lgtm, thanks!

PR #926, opened 2026-08-03
Initialize posthog integration

rnathuji:
This PR is intended to get a basic integration with PostHog deployed so we can evaluate and iterate as desired. Specifically it:

* Adds dependencies for server and client side events
* Instruments code for an example server-side event: `study_proposal_submitted`
* Instruments code for an example client-side event: `team_member_invited`
    * For now client-side events are sent directly to posthog. They do recommend a reverse proxy per [here](https://posthog.com/docs/libraries/next-js#configuring-a-reverse-proxy-to-posthog), and we may want to consider their [managed reverse proxy](https://posthog.com/docs/advanced/proxy/managed-reverse-proxy) to avoid load on our system, but that likely requires some conversation so deferring for this initial POC rollout.
    * **NOTE:**  I did try the nextjs rewrite approach per [this documentation](https://posthog.com/docs/advanced/proxy/nextjs#add-rewrites-to-nextconfigjs) but reverted for this PR since:
        1. It would impose load on our system and likely isn't a good final solution for us
        2.  I ran into issues with the AWS deployment mangling the compressed data and (figuring if we employ a reverse proxy in the flow we'd likely eventually adopt the managed reverse proxy) decided it wasn't worth the headache of dealing with that for this issue / PR :man_shrugging: 
* Integrates use of client-side [identify()](https://posthog.com/docs/references/posthog-js?filter=identification#PostHog-identify) interface so events for authenticated users can be stitched together


src/server/actions/study-request.ts
```diff
@@ -439,6 +440,13 @@ export const finalizeStudySubmissionAction = new Action('finalizeStudySubmission
 
         onStudyCreated({ userId, studyId })
 
+        const posthog = await getPostHogClient()
+        await posthog.captureImmediate({
```
nathanstitt: This `await` puts a PostHog round-trip in the critical path of study submission,  the default request timeout is 10s, so a slow/degraded PostHog delays the user's response by that much.  Can we move this to `src/server/events.ts` in a `after()` block like the other events from there?

src/server/posthog.ts
```diff
@@ -0,0 +1,20 @@
...
+let client: PostHog | null = null
+
+export async function getPostHogClient(): Promise<PostHog> {
+    if (client) return client
+
+    const postHogProjectToken = (await getConfigValue('POSTHOG_PROJECT_TOKEN', false)) ?? ''
+
+    // flushAt and flushInterval are set per https://posthog.com/docs/libraries/node#short-lived-processes-like-serverless-environments ,
+    // but they shouldn't matter if we consistently use captureImmediate
+    client = new PostHog(postHogProjectToken, {
```
nathanstitt, line 15: When `POSTHOG_PROJECT_TOKEN` isn't configured (local dev, CI), `new PostHog('')`  logs `"You must pass your PostHog project's api key. The client will be disabled."` at error level and then warns on every `captureImmediate`. So unconfigured environments like dev get noisy.

Maybe we can return `null` when there's no token and guard at the call site, like how `mailgun.ts` does it?

src/app/layout.tsx
```diff
@@ -34,10 +35,13 @@ export default async function RootLayout({
 }: Readonly<{
     children: ReactNode
 }>) {
+    await connection() // force server rendering so we can access config vars
```
nathanstitt: Using `connection()` forces NextJS into using dynamic rendering vs SPA for every route and will be much slower.  

 We solved the same problem for  how to pass env vars to the the client with Clerk, for example `dashboard/layout.tsx`, has a ClerkProvider. That works with a shared build because those layouts are already dynamic, so the read happens in the container, not at build time.

Perhaps we can rename ClerkProvider to be `EnvironmentProvider` and pass both clerk and Posthog envs using it?

src/components/posthog-user-provider.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { useSession } from '@/hooks/session'
+
+export default function PostHogUserProvider() {
+    const { isLoaded, session } = useSession()
+
+    useEffect(() => {
+        if (isLoaded && session) {
+            posthog.identify(session.user.id)
+        }
+    }, [isLoaded, session])
```
nathanstitt: Depending on the `session` object means the effect re-runs (and re-identifies) whenever the session object identity changes — `useSession` builds a new object when Clerk metadata syncs. The actual input is the id, so `session?.user.id` as the dep would make identify fire exactly once per user.

src/hooks/use-posthog-init.ts
```diff
@@ -0,0 +1,16 @@
...
+import posthog from 'posthog-js'
+
+let initialized = false
+
+export function usePostHogInit(projectToken: string) {
+    useEffect(() => {
+        if (!projectToken || initialized) return
+        initialized = true
+        posthog.init(projectToken, {
+            api_host: 'https://us.i.posthog.com',
```
nathanstitt: This host is duplicated in `src/server/posthog.ts` — worth a shared constant so client and server can't drift, especially if we adopt the reverse proxy later and need to change it in one place.

nathanstitt (commented):
Looking good @rnathuji - had a suggestion on how to possibly send the ENV vars up the stack a bit nicer, see what you think

src/app/layout.tsx
rnathuji: While looking into changes based on this comment, I noticed that the `await connection()` has already landed on `main` [here](https://github.com/safeinsights/management-app/blob/95fb40e4a3a5462cd17f10411fd584cbcf2bda08/src/app/layout.tsx#L43) (there's currently a merge conflict on that file with this PR). I can still look into your approach, but I thought I'd point that out 🤷‍♂️ .

src/server/actions/study-request.ts
rnathuji: Addressed in [this commit](https://github.com/safeinsights/management-app/pull/926/commits/4f857e70d4236557ba2512616098ced3759c43c5).

src/server/posthog.ts
rnathuji: Addressed in [this commit](https://github.com/safeinsights/management-app/pull/926/commits/fbaf763193f5235f1799fcb41bce6167df74d069).

src/components/posthog-user-provider.tsx
rnathuji: Addressed in [this commit](https://github.com/safeinsights/management-app/pull/926/commits/7029136e59c99a8d09eb6969e44e707966cb92e5).

src/hooks/use-posthog-init.ts
rnathuji: Addressed in [this commit](https://github.com/safeinsights/management-app/pull/926/commits/a3aaf6e7b2b30ee3f426bd238598a825175fca0f).

src/app/layout.tsx
rnathuji: See latest diff against `main` after [this merge commit](https://github.com/safeinsights/management-app/pull/926/commits/1c9da8adaf873250f4b4078c6be4f327bc109541). Let me know if you think it's better to stick with what's there now given `await connection()` is already there on `main`, or if you'd like me to make a change to create an `EnvironmentProvider` given that approach is now orthogonal to your original concern.

src/server/events.ts
```diff
@@ -101,6 +102,12 @@ type StudyEvent = { studyId: string; userId: string }
 export const onStudyCreated = deferred(async ({ studyId, userId }: StudyEvent) => {
     await audit({ userId, eventType: 'CREATED', recordType: 'STUDY', recordId: studyId })
     await email.sendStudyProposalEmails(studyId)
+
+    await capturePostHogEvent({
```
nathanstitt, line 106: exactly what I was thinking, ty!

src/hooks/use-sign-out.ts
```diff
@@ -3,13 +3,15 @@ import { useRouter } from 'next/navigation'
...
 import { Routes } from '@/lib/routes'
+import posthog from 'posthog-js'
 
 export function useSignOut(options?: { redirectAfterSignOut: string }) {
     const { signOut } = useClerk()
     const router = useRouter()
     const queryClient = useQueryClient()
 
     return async () => {
+        posthog.reset()
```
nathanstitt, line 14: Good catch pairing `reset()` with sign-out.  I think we might need to do something here for a few others as well, I'll have to investigate that

src/app/layout.tsx
nathanstitt: Oh yes, CSP did require that.  Given that we're using that lets go ahead and leave this in then, it's the simpler change, and removing it will require a larger refactor

nathanstitt (approved):
Looks good, thanks Ripial.  We just merged a fix for the security checks in `main` so if you update to that tests should pass.

PR #928, opened 2026-08-04
Otter 676 outputs available after decryption

Rod0352:
OTTER-676: DP Outputs Refactoring — Outputs Available, After Decryption

After a completed run, the reviewer's outputs-available screen now uses the shared OutputsReviewPanel (OTTER-675): security key gate → decrypted files table, feedback, and the share-outputs / feedback-only decision with confirmation modals.

Stacked PR: base is OTTER-675-outputs-post-decryption-review, and the diff carries OTTER-668's commit. Merge #922 and #924 first, then retarget this to main — only the OTTER-676 commits are new.

Changes
reviewer-outputs-available-screen.tsx: swapped the static key-form layout for OutputsReviewPanel, with the completed-run feedback cap (1500).
review-before-sharing-banner.tsx (new): post-decryption warning banner extracted from the errored screen so both flows share it.
reviewer-outputs-errored-screen.tsx: imports the shared banner; no behavior change.

Notes: the ticket says 1500 characters; the existing panel and server validation count words (cap already defined as 1500 words in #924). Kept words for consistency — flag if characters are actually required.

Tests

15 new tests for the screen (gating on key validation, wrong key/no artifacts, exact banner copy + footnote a11y, blank-submit validation, 1500 cap; decrypts real ENCRYPTED-RESULT artifacts). Regression run over touched areas: 328 passed. typecheck/lint/validate-actions clean.

Depends on: #922, #924

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
@@ -36,11 +36,6 @@ vi.mock('@/server/actions/study-job.actions', async () => {
     return { ...actual, fetchEncryptedJobFilesAction: vi.fn(async () => []) }
 })
 
-// Real in-browser RSA/AES decryption drives the phase flip in these tests, and on a loaded CI
-// runner (sharded workers + coverage instrumentation) that chain can outlast the default 5s
-// per-test budget. The waitFor timeouts below stay inside this ceiling.
-vi.setConfig({ testTimeout: 30_000 })
-
 const toArrayBuffer = (str: string): ArrayBuffer => {
     const buf = Buffer.from(str, 'utf-8')
     return buf.buffer.slice(buf.byteOffset, buf.byteOffset + buf.byteLength)
@@ -108,10 +103,6 @@ const unlock = async () => {
     fireEvent.click(screen.getByRole('button', { name: 'View' }))
 }
 
-// Generous asynchronous wait: CI shards run 4 workers with coverage instrumentation, and the
-// decrypt flow under test is CPU-bound, so the RTL default of 1s is far too tight there.
-const WAIT = { timeout: 15_000 }
-
 // Matches on an element's full textContent, so text split across child nodes (or differing
 // whitespace) can never produce a false negative the way an exact-string matcher can.
 const textIncludes = (needle: string) => (_: string, element: Element | null) =>
@@ -169,9 +160,7 @@ describe('ReviewerOutputsAvailableScreen before decryption', () => {
         fireEvent.change(screen.getByRole('textbox'), { target: { value: 'not-a-real-key' } })
         fireEvent.click(screen.getByRole('button', { name: 'View' }))
 
-        expect(
-            await screen.findByText(textIncludes('Invalid key. Check that you copied'), undefined, WAIT),
-        ).toBeInTheDocument()
+        expect(await screen.findByText(textIncludes('Invalid key. Check that you copied'))).toBeInTheDocument()
         expect(screen.queryByTestId('outputs-files-section')).toBeNull()
     })
 
@@ -185,6 +174,10 @@ describe('ReviewerOutputsAvailableScreen before decryption', () => {
         )
     })
 
+    // This test and the next are defensive-only: reviewer-screen-rules (rule 1b) routes this
+    // screen solely for a RUN-COMPLETE job with no files decision, so neither state can reach it
+    // through the resolved flow. Rendering the component directly is what makes them reachable
+    // here — they pin the guards, not a reviewer-visible state.
     it('shows a not-found alert when the study has no submitted job', async () => {
         const { org, study } = await setupStudyAction({ orgSlug: 'openstax', orgType: 'enclave', createJob: false })
         ;(useParams as Mock).mockReturnValue({ orgSlug: org.slug, studyId: study.id })
@@ -213,12 +206,12 @@ describe('ReviewerOutputsAvailableScreen after decryption', () => {
         await doRender(study, org.slug)
         await waitFor(() => expect(vi.mocked(fetchEncryptedJobFilesAction)).toHaveBeenCalled())
         await unlock()
-        await waitFor(() => expect(screen.getByTestId('outputs-files-section')).toBeInTheDocument(), WAIT)
+        await waitFor(() => expect(screen.getByTestId('outputs-files-section')).toBeInTheDocument())
         // Quiesce: the files table fires the last-activity query on mount. Waiting for its answer
         // here means no test ends with that DB query in flight — an in-flight query racing the
         // per-test transaction rollback closes the shared client and poisons every later test
         // (the deferred-callback race documented in tests/vitest.setup.ts, in query form).
-        await waitFor(() => expect(screen.getAllByText('No activity yet').length).toBeGreaterThan(0), WAIT)
+        await waitFor(() => expect(screen.getAllByText('No activity yet').length).toBeGreaterThan(0))
         return { org, study, job }
     }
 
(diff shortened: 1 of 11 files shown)


tests/study-flow.spec.ts
```diff
@@ -476,21 +475,19 @@ test('Reviewer approves submitted code', async ({ browser, studyFeatures }) => {
     })
 })
 
-// Owns the reviewer results decrypt+approve surface. Seeds a JOB-READY job, uploads
-// an encrypted result via the debug script (no runner), then drives the UI.
+// Owns the outputs-available surface end to end (OTTER-668 + OTTER-676): decrypt, the
+// validation gate, and sharing the outputs through the confirmation modal. Seeds a
+// JOB-READY job, uploads an encrypted result via the debug script (no runner), then
+// drives the UI. Researcher's approved-results view is owned by its own card.
```
nathanstitt: This comment says the researcher's approved-results view is "owned by its own card", but I can't find the card that owns it. The old body of this test had the only assertion on that view:

```ts
await expect(page.getByText(/results of your study have been approved/i)).toBeVisible()
```

and grepping the spec at this commit for `results of your study have been approved` returns nothing — the string is gone from the e2e suite entirely. The nearest candidate (`Researcher resubmits code`, ~line 734) *seeds* FILES-APPROVED directly and drives the resubmit form; it never renders or asserts the approved-results message.

So the net effect is that the researcher-facing half of the happy path lost its only e2e coverage here. Note the errored test right below deliberately keeps its researcher leg (`verifyFailedStatusDisplay`), which is the asymmetry that makes this stand out.

Could we either add the researcher leg back after `reviewerSharesOutputs` (the decision has already been submitted at that point, so the state is there for free), or point the comment at whichever card actually owns it?

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.tsx
```diff
@@ -0,0 +1,66 @@
...
+export async function ReviewerOutputsAvailableScreen({
+    study,
+    orgSlug,
+}: Pick<ScreenComponentProps, 'study' | 'orgSlug'>) {
+    const job = await latestSubmittedJobForStudy(study.id)
+    if (!job) {
+        return <AlertNotFound title="No submission found" message="This study has no submitted code to review." />
+    }
+
+    const availableAt = availableTimestamp(job.statusChanges)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.tsx:44]: Worth a thought on the two not-found branches: this screen only renders because `reviewer-screen-rules` matched `hasResults && !approved && !rejected && !errored`, which — given `hasResults` is RUN-COMPLETE/FILES-APPROVED/FILES-REJECTED/JOB-ERRORED — narrows to exactly RUN-COMPLETE. So by construction there *is* a submitted job with a RUN-COMPLETE row, and both `AlertNotFound`s should be unreachable in the resolved flow.

That's fine as defensive coding (the errored screen does the same), but the two tests that cover them (`no submitted job`, `job has not completed a run`) call the component directly, bypassing the rule table. They're really asserting the guard, not a state a reviewer can reach. Might be worth a line noting these are defensive-only, so a future reader doesn't infer the rule table can actually route here.

tests/study-flow.spec.ts
```diff
@@ -281,30 +281,29 @@ function uploadResults(jobId: string): void {
...
     await expect(page.getByText('Review Studies')).toBeVisible()
     await viewStudyDetails(page, studyTitle)
     await page.waitForURL(/\/review$/)
 
+    await expect(page.getByText(/Outputs are available for review/)).toBeVisible()
+    await expect(page.getByRole('heading', { name: /security key/i })).toBeVisible()
+
     const privateKey = await readTestSupportFile('private_key.pem')
-    const privateKeyTextarea = page.getByPlaceholder('Enter your Results Key to access encrypted content.')
+    const privateKeyTextarea = page.getByRole('textbox')
```
nathanstitt: Swapping the specific placeholder locator for a bare `page.getByRole('textbox')` makes this strict-mode-fragile — it resolves the whole page, so the moment the locked phase gains any second textbox this throws "strict mode violation: resolved to N elements" rather than failing on a meaningful assertion.

The old `getByPlaceholder('Enter your Results Key…')` was doing real work. The form still has `placeholder="Enter your security key"` (`security-key-form.tsx:38`), so we could keep the intent and stay specific:

```ts
const privateKeyTextarea = page.getByPlaceholder('Enter your security key')
```

Same applies to the identical line in `reviewerDecryptsErrorLogs` (~321). Given the zero-tolerance flakiness rule in CLAUDE.md, pinning these seems worth it.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
```diff
@@ -0,0 +1,303 @@
...
+        await setupDecrypted([
+            { name: 'results.csv', content: 'a,b\n1,2' },
+            { name: 'summary.txt', content: 'all good' },
+        ])
+
+        expect(screen.getByRole('button', { name: 'results.csv' })).toBeInTheDocument()
+        expect(screen.getByRole('button', { name: 'summary.txt' })).toBeInTheDocument()
+        // Waits for the activity query: the cell stays blank until the answer is in, so it never
+        // claims "No activity yet" on the strength of an unresolved request.
+        await waitFor(() => expect(screen.getAllByText('No activity yet')).toHaveLength(2), { timeout: 15_000 })
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx]: Small consistency nit — this is the one place that inlines `{ timeout: 15_000 }` instead of using the `WAIT` constant defined at line 113 for exactly this purpose. Reusing `WAIT` here keeps the "CI is slow, one knob" rationale in a single spot:

```ts
await waitFor(() => expect(screen.getAllByText('No activity yet')).toHaveLength(2), WAIT)
```

src/lib/study-screen/reviewer-screen-rules.ts
```diff
@@ -9,7 +9,16 @@ export const REVIEWER_SCREEN_RULES = [
     //     enters their security key to decrypt error logs before making a files decision.
     ['reviewer-outputs-errored', { when: (s) => s.resultsErrored && !s.resultsApproved && !s.resultsRejected }],
 
-    // 1b. Results exist → results-only Study Details (OTTER-538). Out-ranks the code decision
+    // 1b. Run completed, no files decision yet ("results pending review") → outputs-available view
+    //     (OTTER-668). The reviewer enters their security key to decrypt the outputs before making
+    //     a files decision. Sits below 1a so an errored run keeps the errored view even when a
+    //     RUN-COMPLETE also landed.
+    [
+        'reviewer-outputs-available',
```
nathanstitt, line 18: Nice — the `!s.resultsErrored` clause here is what makes the table genuinely order-independent, matching the file header's "every `when` reads only StudyState" claim. Rule 1a would have won anyway by position, but spelling out the exclusion means the two rules stay disjoint if anyone reorders them, and it lines up with `outputsFeedbackMaxWords` deriving the same JOB-ERRORED-wins precedence server-side. Good that the `run complete AND errored` test pins the behaviour rather than trusting ordering.

tests/study-flow.spec.ts
nathanstitt: if it wasn't perfectly matching the placeholder you can use regex:  ` const privateKeyTextarea = page.getByPlaceholder(/enter your results key/i)`

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx]: using a huge timeout like this is a big red flag  that something is wrong.  If we legit are not rendering in 15s, then we have a much larger issue than this test and we need to deeply investigate why

nathanstitt (approved):
looks good but found a few places where tests could be tightened up.  fine to merge after that

PR #929, opened 2026-08-04
OTTER-687: carry the code-approved page forward instead of ending at the dashboard

therealmarv:
see [OTTER-687](https://openstax.atlassian.net/browse/OTTER-687)

[#950](https://github.com/safeinsights/management-app/pull/950) ([OTTER-686](https://openstax.atlassian.net/browse/OTTER-686)) has landed on main and is merged into this branch, so the blocker that kept this in draft is gone. Both halves are live: the researcher's forward link now has `outputs-pending` to point at.

## What this changes

Both roles' code-approved pages ended the flow with "Go to dashboard". They now offer **"Next step"** into the outputs screen this epic added, per the Figma frames ([DP](https://www.figma.com/design/BuIOVeT1vEj2aKLuOHAwvl/66-Study-code-feedback--CRATE-DO-?node-id=5217-8205), [RL](https://www.figma.com/design/BuIOVeT1vEj2aKLuOHAwvl/66-Study-code-feedback--CRATE-DO-?node-id=5217-8728)), which both show Previous step / Next step and no dashboard CTA.

The forward button **replaces** the dashboard one rather than sitting beside it. A change request still outranks both: "Edit and resubmit" is the next step in that case, not a step to skip.

## Key design point: ask the table, don't restate it

The outputs screens are not routes of their own. Each role reaches them at its canonical study URL (`/view` for the researcher, `/review` for the reviewer) and the screen-rule tables decide what renders there. So "is there a next step" is exactly **"does that URL resolve to a screen other than the one we are on"**, which is what the new `hasNextStepFromCode` asks:

```
resolveScreen(role, state).screen !== currentScreen
```

Two things fall out of that:

- **It cannot self-link.** A button pointing at the page it sits on is the only failure mode worth guarding here, and it is precisely the case where the predicate says no. Concretely, a DP whose code is approved but whose job has not reached `JOB-PACKAGING` yet still resolves `/review` to the code screen, so they keep "Go to dashboard" until the enclave picks the job up.
- **It stayed current as the epic landed screens.** #950 registered the researcher's `outputs-pending` screen, and `/view` started resolving past the code screen for an executing study with no change to the predicate. No new screen id, no new rule, no route definition touched.

It also **subsumes the OTTER-614 guard it replaces**, which is why `awaitingFilesDecisionOnError` is gone from `code-decision-screen`, and it does so in a way that survived #950. The invariant is that while an error is still hidden from the researcher, the table holds them on a screen that does not disclose it, so whichever screen the forward link points at is one they were already entitled to land on. An errored-while-executing study resolves to `outputs-pending`, which shows "Outputs not ready, code processing started" and discloses nothing, so the link is correct there. Results present resolves forward; a decided-against study stays put.

## Changes

- `hasNextStepFromCode` in `src/lib/study-screen/next-step.ts`, exported from the barrel alongside the other resolvers. It takes `(role, state, currentScreen)`; screen selection is pure state, so there is no routing context to thread through.
- **Researcher** (`code-decision-screen` → `code-post-decision-view`): `resultsHref` becomes `nextStepHref`, driven by the predicate. `ProceedToResultsAction` → `NextStepAction`, copy "Proceed to step 5" → "Next step", testid `cta-proceed-to-results` → `cta-next-step`.
- **Reviewer** (`reviewer-code-feedback-screen` → `post-feedback-view`): new optional `nextStepHref`. Only the CODE screen passes it, so every proposal usage keeps "Go to dashboard" untouched. The existing `{previousHref && ...}` inline conditional moved into the components' own visibility guards, per the JSX rule in CLAUDE.md, and the button row's `justify` is now a named value above the JSX.
- Reused `projectStudyState(raw)` in the reviewer screen instead of projecting twice.

No route definitions, permissions, or migrations touched.

## Review follow-ups

- **Dropped the discarded `ctx` argument, then the parameter itself.** `resolveScreen` ignores routing context by design (screen choice is pure state), so requiring one meant call sites built an object only to see it thrown away, and it implied routing context feeds screen selection. The first pass made it optional and took it off `hasNextStepFromCode`. The follow-up review pointed out that the only thing then keeping `ScreenRuleCtx` alive was an argument nothing read, so the parameter is gone entirely, `render-screen.tsx` no longer builds an object for it (which made its `studyId` arg dead too), and the type is deleted. `screen-rules.ts` now says a rule sees only `StudyState`, never routing context, in contrast to `resolveDashboardAction`, whose rules genuinely do build hrefs.
- **Named the button row's layout intent.** `justify` keyed on `previousHref` while the row now has three possible occupants, so the ternary no longer described what it was laying out. It is a `buttonRowJustify` value above the JSX with a comment: the row splits only when there is a left button, and the forward link and dashboard button are mutually exclusive occupants of the right slot.
- **Documented the walk-back behavior.** On `/review/code` the two resolvers deliberately disagree: `resolveReviewerCodeScreen` restricts candidates to the code screens so a results study lands on the code page, while `hasNextStepFromCode` asks the full table and answers "results". That is what gives a reviewer who walked back both Previous and Next step. The comment now covers it, not just the live-screen suppression case.
- **One shared `StudyState` test fixture.** `src/lib/study-screen/state.fixture.ts` now holds the single blank-slate literal; the six resolver test files each keep their own named helper as a thin baseline wrapper over it. Adding a field to `StudyState` updates one object instead of breaking every test file that enumerated the type.

## The #950 merge

Everything on this branch that would have flipped meaning under #950 was dealt with before the merge, so landing it was mechanical rather than a rewrite:

- **The one conflict was two lines, and resolved as predicted.** #950 rewrote the OTTER-598 errored test in `view/page.test.tsx` to assert the `outputs-pending` screen instead of `CodePostDecisionView`. This branch had already stopped asserting `nextStepHref` inside that block, so the conflict came down to assertions #950 deletes outright and the resolution was "take theirs". The equivalent forward-link assertion survives in the packaging-error test in the same file, which #950 left alone and which stays correct: that shape errors with no execution substatus, so `isExecuting` is false, `/view` still resolves to `code-approved`, and there is still no step forward to offer.
- **The `hasNextStepFromCode` doc comment states the invariant, not one release's outcome.** It previously said a bare `JOB-ERRORED` resolves back to the code screen for the researcher, which held only until #950, because `isExecuting` stays true precisely while a files decision is outstanding (`state.ts`). The comment now says what is true on both sides of the merge.
- **The errored test is pinned to the shape that does not move.** `next-step.test.ts` names its errored case "errored before execution started" and sets `isExecuting: false` explicitly, which is the packaging-failure shape `projectStudyState` produces and which resolves to `code-approved` either side of #950.
- **The deferred researcher-executing assertions are in.** They flip from false to true once #950 registers the screen, so they could not be written green beforehand. Both are now present and passing: `is true while the enclave runs the job and /view resolves to outputs-pending (OTTER-686)` in `next-step.test.ts`, and `forwards to /view once the enclave is running the job` in `view/code/page.test.tsx`. That is the coverage for the card's RL acceptance row.

## Testing

- Full unit suite green: 265 files / 2585 tests. `pnpm run checks` clean (typecheck, eslint, prettier, action validation).
- `next-step.test.ts` covers both roles across approved / executing / results / errored-before-execution / decided-against.
- Screen-level coverage in `review/code/page.test.tsx` and `view/code/page.test.tsx` asserting the forward href is present for a results study and for a running job, and absent while the canonical URL still resolves to the code step.
- View-level cases for the CTA swap on both roles, plus resubmit still outranking the forward link.
- Verified all six button-row combinations in a browser against local preview renders: researcher with the forward link, researcher without it, researcher change-requested with a forward href passed in anyway (resubmit wins, no dashboard), and the reviewer's three rows. The lone "Next step" in the reviewer's no-Previous case lands on the same right edge as the "Go to dashboard" button it replaces, so the right-aligned single-button row matches what was there before.
- E2E: no changes needed. Every `go-to-dashboard` assertion in `study-flow.spec.ts` sits on a proposal decision, a code change request, or a code rejection, none of which resolve past their own screen.

## For Iris (non-blocking)

- The DP row of the AC reads "building the RL outputs screen is handled by this card" while linking OTTER-666. Looks like a copy-paste slip; the RL row links 686 correctly. Implemented as DP → 666's screen, RL → 686's screen.
- The AC says the current copy is "Back to dashboard"; it is actually "Go to dashboard" on both pages. No behavior question, just flagging that the strings differ.
- Both Figma frames label the left link "Previous step", while the DP page currently renders "Previous". Left as-is since the AC covers only the forward button. Happy to align it here if you want it in scope.
- After #950 the researcher walks from a screen labeled "STEP 4 Study code" to one labeled "STEP 4 Verify outputs", and this PR's button is what puts them one click apart. #950 defers the renumbering to the agreements-page removal card, so this is only more visible with this PR, not caused by it.

[OTTER-687]: https://openstax.atlassian.net/browse/OTTER-687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OTTER-686]: https://openstax.atlassian.net/browse/OTTER-686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/study-screen/next-step.ts
```diff
@@ -0,0 +1,28 @@
...
+ * It also subsumes the guards this replaces — a bare JOB-ERRORED still resolves back to the code
+ * screen for the researcher, so the button stays hidden while the reviewer triages the error.
+ */
+export function hasNextStepFromCode(
+    role: StudyRole,
+    state: StudyState,
+    currentScreen: ScreenId,
+    ctx: ScreenRuleCtx,
+): boolean {
+    return resolveScreen(role, state, ctx).screen !== currentScreen
```
nathanstitt: The `ctx` parameter is threaded all the way through here but never actually influences the answer — `resolveScreen` takes it as `_ctx` and ignores it outright:

```ts
export function resolveScreen(role: StudyRole, state: StudyState, _ctx: ScreenRuleCtx): ScreenDescriptor {
    const rules = role === 'reviewer' ? REVIEWER_SCREEN_RULES : RESEARCHER_SCREEN_RULES
    const [screen] = rules.find(([, rule]) => rule.when(state))!
    return { screen }
}
```

and the file header above it is explicit that this is deliberate ("Pure state → screen: the URL never influences which screen renders"). Every `when` in both tables reads only `StudyState`.

So the required `ctx` argument is dead weight that both call sites have to construct — including the researcher one, which builds `{ orgSlug, studyId, returnTo }` purely to satisfy the signature. It also mildly implies to a reader that routing context feeds screen selection, which is the one thing the design is careful *not* to do.

We could drop it and let the two call sites get shorter:

```ts
export function hasNextStepFromCode(role: StudyRole, state: StudyState, currentScreen: ScreenId): boolean {
    return resolveScreen(role, state, { orgSlug: '', studyId: '' }).screen !== currentScreen
}
```

…though the nicer version is probably to give `resolveScreen` an optional ctx, since nothing reads it. Either way, worth not requiring callers to build an object that's discarded. Non-blocking if you'd rather keep the signature aligned with `resolveScreen` for future-proofing — but then a one-line comment saying so would help.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -270,8 +289,9 @@ export function PostFeedbackView({
                 />
                 <FeedbackAndNotesSection entries={entries} alwaysExpandLatest={isCode} />
                 <Group justify={previousHref ? 'space-between' : 'flex-end'}>
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: The `justify` still keys only on `previousHref`, but the button row now has three possible occupants instead of two. Tracing the combinations:

- `previousHref` set + a right button → `space-between`, correct.
- neither `previousHref` nor `nextStepHref` → `flex-end`, dashboard on the right, correct.
- **no `previousHref` but `nextStepHref` set** → `flex-end`, "Next step" right-aligned.

That third row is the live `/review` code screen (`readOnlyCodeStep` unset ⇒ no Previous), and it's exactly the case this PR newly creates. Right-aligned is arguably fine, but it's now decided by a condition that no longer names the thing it's laying out — the ternary reads as "is there a left button" when what it means is "is the row split".

Since `PreviousButton`/`NextStepButton`/`GoToDashboardButton` all self-hide now, the `justify` is the one spot still doing conditional logic from the outside. Might read better as an explicit derived value above the JSX, per the "move logic out" rule:

```ts
const justify = previousHref ? 'space-between' : 'flex-end'
```

with a comment on why a lone forward button right-aligns. Purely cosmetic — flagging because the layout intent is now less obvious than it was with two buttons.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx
```diff
@@ -21,6 +21,14 @@ export async function ReviewerCodeFeedbackScreen({ study, raw, orgSlug, descript
         ? Routes.studyReviewerAgreements({ orgSlug, studyId: study.id })
         : undefined
 
+    // OTTER-687: forward to the DP outputs screen, which lives at bare /review. Suppressed while
+    // /review still resolves to this screen (code approved, enclave not started yet), where the
+    // button would only point back at the page it sits on.
+    const state = projectStudyState(raw)
+    const nextStepHref = hasNextStepFromCode('reviewer', state, descriptor.screen, { orgSlug, studyId: study.id })
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx]: Worth confirming the walk-back case is intended, because the predicate and the `previousHref` above it disagree about which screen we're on.

On `/review/code`, the descriptor comes from `resolveReviewerCodeStep` → `resolveReviewerCodeScreen`, which deliberately restricts candidates to the code screens so a results study lands on `reviewer-code-feedback` rather than looping to results. Meanwhile `hasNextStepFromCode` calls the *full* `resolveScreen`, which for that same study returns `reviewer-study-results`. Different ⇒ `nextStepHref` is set.

So a reviewer who walked *back* from results to `/review/code` now sees both "Previous" and "Next step", where Next returns them to the results screen they came from. That's coherent as breadcrumb navigation, and the `forwards to /review for a results study` test locks it in — I'm reasonably confident it's deliberate. But the comment directly above only describes the live-screen suppression case ("code approved, enclave not started yet") and doesn't mention the walk-back at all, which is the more interesting of the two behaviors this enables.

A sentence noting that the walk-back intentionally offers a forward step back to where they came from would save the next reader the same trace through two different resolvers.

src/lib/study-screen/next-step.test.ts
```diff
@@ -0,0 +1,75 @@
+import { describe, expect, it } from 'vitest'
+import { hasNextStepFromCode } from './next-step'
+import type { StudyState } from './state.types'
+
+const CTX = { orgSlug: 'org', studyId: 'study-id' }
+
+// An approved-code study that has not started running: the state both roles' code screens show.
+const state = (overrides: Partial<StudyState> = {}): StudyState => ({
```
nathanstitt: The `state()` factory spells out all 20 `StudyState` fields as a literal. That makes each case readable, but it means this file has to be touched every time a field is added to `StudyState` — and because the object is typed as `StudyState` (not `Partial`), a new required field breaks this test file rather than the code under test.

Since `projectStudyState` is the real producer of these objects, would it be worth a shared fixture — or reusing whatever the existing screen-rule tests build? `reviewer-screen-rules.test.ts` already has an `st()` helper doing something similar, so there may be an opportunity to share one rather than maintain a third copy.

Not blocking, and the explicitness genuinely helps here — mostly flagging the maintenance coupling.

src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx
```diff
@@ -140,15 +141,16 @@ const EditAndResubmitAction: FC<{ isVisible: boolean; href: Route }> = ({ isVisi
...
 }
 
-function DecisionActions({ decision, previousHref, dashboardHref, resubmitHref, resultsHref }: DecisionActionsProps) {
+function DecisionActions({ decision, previousHref, dashboardHref, resubmitHref, nextStepHref }: DecisionActionsProps) {
     const showResubmit = decision === 'CODE-CHANGES-REQUESTED'
-    // Once results exist, continue forward to Step 5 instead of ending at the dashboard.
-    const showProceedToResults = !showResubmit && !!resultsHref
+    // When the flow continues past this screen, carry the user forward instead of ending at the
+    // dashboard. Resubmit outranks it: a change request is the flow, not a step to skip.
+    const showNextStep = !showResubmit && !!nextStepHref
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/code-post-decision-view.tsx:148]: Nice call making resubmit outrank the forward link rather than letting both render. `showNextStep = !showResubmit && !!nextStepHref` keeps the three actions provably mutually exclusive (dashboard already carries `!showResubmit && !showNextStep`), so there's no state where two primary CTAs compete — and `keeps "Edit and resubmit" over "Next step" for CODE-CHANGES-REQUESTED` pins it with a `nextStepHref` deliberately passed in, which is the case that would actually regress. The reasoning in the comment ("a change request is the flow, not a step to skip") is the right framing too.

nathanstitt (commented):
sorry, I missed that this was draft and reviewed, leaving as unedited comments in case they're useful

src/lib/study-screen/next-step.ts
therealmarv: Agreed, and I took the version you preferred: `ctx` is gone from `hasNextStepFromCode` entirely, and `resolveScreen` accepts it optionally.

[`next-step.ts#L20`](https://github.com/safeinsights/management-app/blob/74f2092a38ce5dc0cc445a5ce588c8faca026df6/src/lib/study-screen/next-step.ts#L20):

```ts
export function hasNextStepFromCode(role: StudyRole, state: StudyState, currentScreen: ScreenId): boolean {
    return resolveScreen(role, state).screen !== currentScreen
}
```

[`resolve.ts#L12`](https://github.com/safeinsights/management-app/blob/74f2092a38ce5dc0cc445a5ce588c8faca026df6/src/lib/study-screen/resolve.ts#L12) is `_ctx?: ScreenRuleCtx`, with the reason stated where a reader will hit it rather than left implicit:

> `_ctx` is optional because of that purity: no rule reads it, so callers that don't already have one (hasNextStepFromCode) shouldn't have to build an object only to see it discarded.

Optional rather than removed so the existing caller in `render-screen.tsx` and the resolver tests keep passing theirs unchanged. Both of my call sites lost their throwaway object: the researcher one no longer constructs `{ orgSlug, studyId, returnTo }` at all.

You were right about the second-order cost too. Requiring the argument was the part that suggested routing context feeds screen selection, which is the one thing the table is careful not to do.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx]: Fixed, and your reading of the three rows is exactly right. It is now a named value above the JSX ([`post-feedback-view.tsx#L262-L264`](https://github.com/safeinsights/management-app/blob/74f2092a38ce5dc0cc445a5ce588c8faca026df6/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/review/post-feedback-view.tsx#L262-L264), used at [L294](https://github.com/safeinsights/management-app/blob/74f2092a38ce5dc0cc445a5ce588c8faca026df6/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/review/post-feedback-view.tsx#L294)):

```ts
// The row splits only when there is a left button: the forward link and the dashboard button are
// mutually exclusive and both sit on the right, so a lone one right-aligns either way.
const buttonRowJustify = previousHref ? 'space-between' : 'flex-end'
```

The comment names the reason the condition stays keyed on `previousHref`: the two right-hand buttons can never co-occur, so the row is never genuinely three-occupant at render time. Two possible occupants, one per side, which is what `previousHref` is deciding.

I also checked the third row visually rather than trusting the reasoning, since right-aligned was only "arguably fine". Rendering that exact combination (`nextStepHref` set, no `previousHref`) at 1440px, the lone "Next step" occupies x=1305–1408; the "Go to dashboard" button it replaces in the neither-link row sits at x=1257–1408. Same right edge, so nothing shifts position from what was there before, it just gets narrower.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx]: Confirmed intended, and you traced it correctly through both resolvers. The comment now says so ([`reviewer-code-feedback-screen.tsx#L24-L34`](https://github.com/safeinsights/management-app/blob/74f2092a38ce5dc0cc445a5ce588c8faca026df6/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/_screens/reviewer-code-feedback-screen.tsx#L24-L34)):

> On the walk-back route the two resolvers deliberately disagree, and the forward link is the point of that: `resolveReviewerCodeScreen` restricts its candidates to the code screens so a results study lands here instead of looping to results, while `hasNextStepFromCode` asks the full table and still answers "results". So a reviewer who walked back from results gets Previous and Next step, with Next returning them to the screen they came from.

Your framing of it as breadcrumb navigation is what I'd been treating as obvious and shouldn't have: with both links present the code step reads as a waypoint in the flow rather than a dead end, which is the behavior the walk-back route exists for. The disagreement between the two resolvers is load-bearing, so it deserved to be written down next to the call rather than inferred from two files.

I also rendered that row to be sure it looks like navigation and not a duplicate CTA: Previous at x=32–144, Next step at x=1305–1408, `space-between`.

src/lib/study-screen/next-step.test.ts
therealmarv: Worth it, yes, and it turned out to be a fifth copy rather than a third. `resolve.test.ts`, `consistency.test.ts`, `reviewer-screen-rules.test.ts` and `pill.test.ts` each already spelled `StudyState` out, plus `dashboard.test.ts` for the `DashboardState` subset.

[`state.fixture.ts`](https://github.com/safeinsights/management-app/blob/74f2092a38ce5dc0cc445a5ce588c8faca026df6/src/lib/study-screen/state.fixture.ts) now holds the one blank-slate literal, and all six files layer their own baseline on top, keeping their existing helper names so no call site changed:

- `resolve.test.ts` / `consistency.test.ts`: baseline was already the blank slate, so `state`/`full` just delegate.
- `reviewer-screen-rules.test.ts`: `st` overrides `status` / `isDraft` / `displayStatus` for PENDING-REVIEW.
- `pill.test.ts`: six overrides for a submitted-code study.
- `next-step.test.ts`: eleven for an approved-code study.
- `dashboard.test.ts`: `DashboardState` is `Omit<StudyState, 'submissionRound' | 'hasSavedEdits' | 'hasSavedCodeEdits'>`, so the fixture satisfies it structurally and the extra fields go unread.

Net -97 lines across the six, and a new required `StudyState` field now breaks one object instead of six test files. Test counts unchanged at 251 files / 2402 tests, which is the check that the baselines were reproduced faithfully.

I deliberately did not build it on `projectStudyState` even though it is the real producer: that would mean authoring `RawStudyState` rows with jobs and status changes for cases that only care about one boolean, and the resolver tests are unit tests of the tables rather than of the projection (`state.test.ts` covers that). Happy to split the five pre-existing files into their own commit if you'd rather keep this PR's diff to the navigation change.

therealmarv:
No apology needed, the review on a draft was useful and all four points are addressed in 74f2092a. Replies are in the threads; short version:

- `ctx` dropped from `hasNextStepFromCode`, optional on `resolveScreen`, with the reason written down there.
- The button row's `justify` is a named value above the JSX explaining what the condition actually decides.
- The walk-back behavior is documented next to the call instead of left to a two-resolver trace.
- One shared `StudyState` fixture replacing what turned out to be five duplicated literals across the study-screen tests.

Still draft for the same reason as before: the RL's forward link needs [OTTER-686](https://openstax.atlassian.net/browse/OTTER-686)'s screen to exist before `/view` resolves past the code page. The DP half works against main today.

Verification since the review: `pnpm run checks` clean, full unit suite green at 251 files / 2402 tests, and I rendered all six button-row combinations locally to confirm the layout, including the no-Previous row you flagged (the lone "Next step" keeps the same right edge as the dashboard button it replaces).

[OTTER-686]: https://openstax.atlassian.net/browse/OTTER-686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/lib/study-screen/resolve.ts
```diff
@@ -7,7 +7,9 @@ import { DASHBOARD_RULES, type DashboardRuleCtx } from './dashboard-rules'
 
 // Pure state → screen: the URL never influences which screen renders. Walking back to an earlier
 // screen of an advanced study is addressed by its own route instead (see resolveResearcherCodeScreen).
-export function resolveScreen(role: StudyRole, state: StudyState, _ctx: ScreenRuleCtx): ScreenDescriptor {
+// `_ctx` is optional because of that purity: no rule reads it, so callers that don't already have one
+// (hasNextStepFromCode) shouldn't have to build an object only to see it discarded.
+export function resolveScreen(role: StudyRole, state: StudyState, _ctx?: ScreenRuleCtx): ScreenDescriptor {
```
nathanstitt: This landed the way I hoped, and putting the reason in the comment where a reader hits it is better than what I suggested.

One leftover the change surfaces: `ScreenRuleCtx` is now referenced in exactly two places in non-test code — its own definition in `screen-rules.ts:7`, and this parameter that nothing reads. Neither rule table takes it, and `hasNextStepFromCode` no longer passes one. So the only thing keeping the type alive is the optional argument `render-screen.tsx` still constructs.

Not asking you to rip it out in this PR — the header comment in `screen-rules.ts` still frames it as shared rule-table context, which is now slightly stale. Either dropping the param entirely (making `render-screen.tsx`'s object unnecessary too) or a note that it's retained as a forward-compat hook would stop the next reader wondering which rules consume it. Your call which.

src/lib/study-screen/state.fixture.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Blank-slate StudyState for the resolver tests: a fresh draft with every downstream fact off.
+ *
+ * The point is that this is the only place in the test tree that spells `StudyState` out as a
+ * literal, so adding a field to the type means updating one object here rather than breaking every
+ * test file that happened to enumerate it. Tests whose cases share a different starting point layer
+ * their own baseline on top (see each test file's local helper) instead of restating the whole shape.
+ */
+export const studyState = (overrides: Partial<StudyState> = {}): StudyState => ({
```
nathanstitt, line 11: This is the change I'm happiest about in the PR. Five copies collapsing to one is more than I expected when I flagged it, and the way each file keeps its own helper name means the diff stays reviewable — no call site churn hiding a semantic change.

I checked the part that actually mattered: every consumer places `...overrides` last, so per-file baselines still win where they should, and the `st`/`pill`/`next-step` baselines reproduce their previous literals. Full `src/lib/study-screen/` suite is 121 passing, and the review/view component suites are 334 passing, so the consolidation is behavior-neutral.

On splitting the five pre-existing files into their own commit — I'd leave it. The consolidation is what makes the sixth file cheap to add, so it belongs with the change that motivated it, and separating them would make the net -97 harder to read, not easier.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx
```diff
@@ -21,6 +21,20 @@ export async function ReviewerCodeFeedbackScreen({ study, raw, orgSlug, descript
         ? Routes.studyReviewerAgreements({ orgSlug, studyId: study.id })
         : undefined
 
+    // OTTER-687: forward to the DP outputs screen, which lives at bare /review. Suppressed while
+    // /review still resolves to this screen (code approved, enclave not started yet), where the
+    // button would only point back at the page it sits on.
+    //
+    // On the walk-back route the two resolvers deliberately disagree, and the forward link is the
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-code-feedback-screen.tsx:28]: The added paragraph is exactly what was missing — it names the resolver disagreement as load-bearing rather than leaving it to be inferred across two files. Good.

Thanks for rendering it rather than just reasoning about it; "looks like navigation, not a duplicate CTA" was the actual question and coordinates settle it.

src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx
```diff
@@ -240,6 +259,9 @@ export function PostFeedbackView({
     const crumbs = buildCrumbs({ orgSlug, studyId: study.id, kind, crumbLast: kindCopy.crumbLast })
     const banner = <DecisionBanner decision={decision} kind={kind} />
     const isCode = kind === 'CODE'
+    // The row splits only when there is a left button: the forward link and the dashboard button are
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/post-feedback-view.tsx:262]: Good — and your comment is more accurate than my original reading was. I described the row as having three possible occupants; the point that the forward link and the dashboard button are mutually exclusive means it's never actually three at render time, which is what justifies keeping the condition on `previousHref`. Naming that in the comment is the right resolution.

The x-coordinate check confirming the lone "Next step" keeps the same right edge as the button it replaces is a nice touch — that was the part I couldn't be sure of from reading.

src/lib/study-screen/next-step.ts
```diff
@@ -0,0 +1,26 @@
...
+ * outputs epic registers screens, the states that gain a forward step gain it here without this file
+ * changing. Both of the epic's outputs screens landed that way, with no edit here: a study whose
+ * results are in forwards on the strength of the results rule, and one still running in the enclave
+ * began forwarding for the researcher when OTTER-686 registered `outputs-pending`.
+ *
+ * The same delegation subsumes the OTTER-614 guard this replaces. While an error is still hidden
+ * from the researcher the table holds them on a screen that does not disclose it, so whichever
+ * screen the forward link points at is one they were already entitled to land on.
+ */
+export function hasNextStepFromCode(role: StudyRole, state: StudyState, currentScreen: ScreenId): boolean {
```
nathanstitt, line 24: Worth calling out that delegating to the table rather than restating its predicates is what makes this hold up: the docblock's claim that OTTER-686's `outputs-pending` started forwarding with no edit here is the kind of thing that's usually aspirational, and here it's demonstrably true because the predicate is one comparison against `resolveScreen`.

Unrelated to the code: `infosec` is currently red on this branch, but it's the `nanoid` CVE-2026-67213 finding, not anything you introduced — the override pin landed on `main` in #955 earlier today. Merging `main` should clear it.

nathanstitt (approved):
Looks good, thanks Marvin!

src/lib/study-screen/resolve.ts
therealmarv: Dropped it entirely rather than documenting it. A forward-compat hook nothing consumes is the thing that generates this question again in six months, and you had to trace it twice already.

[`resolve.ts#L7-L11`](https://github.com/safeinsights/management-app/blob/62b8b904e029cd7b309a8a3b75527f44d1cf83b8/src/lib/study-screen/resolve.ts#L7-L11):

```ts
// Routing context is therefore not a parameter here: no rule reads it, and taking one would imply
// the URL feeds screen selection. Contrast resolveDashboardAction, whose rules do build hrefs.
export function resolveScreen(role: StudyRole, state: StudyState): ScreenDescriptor {
```

`ScreenRuleCtx` is gone from `screen-rules.ts`, and the header there no longer frames the tables as taking context: it now says a rule sees only `StudyState`, with the `resolveDashboardAction` contrast named in both places so the next reader sees why one resolver takes a ctx and the other cannot ([`screen-rules.ts#L4-L8`](https://github.com/safeinsights/management-app/blob/62b8b904e029cd7b309a8a3b75527f44d1cf83b8/src/lib/study-screen/screen-rules.ts#L4-L8)).

One thing your note did not mention that fell out of it: `render-screen.tsx` was the only remaining passer, and `studyId` in `renderStudyScreen`'s signature existed solely to populate that object. Both are gone, so the three page callers stopped passing `studyId` too ([`render-screen.tsx#L39-L41`](https://github.com/safeinsights/management-app/blob/62b8b904e029cd7b309a8a3b75527f44d1cf83b8/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/_screens/render-screen.tsx#L39-L41)). Net -29 lines across 9 files, no behavior change, full suite green at 265 files / 2585 tests.

`resolveDashboardAction` keeps its ctx untouched, since its rules actually call `action(ctx)`.

src/lib/study-screen/next-step.ts
therealmarv: Thanks for chasing down the `infosec` red and pinning it on `nanoid` rather than leaving me to. main is merged in now, so #955's override pin is on the branch and that check should clear on this push.

Worth recording that the merge exercised the delegation claim rather than just asserting it: #950 is in, and the two assertions I had to defer (`is true while the enclave runs the job and /view resolves to outputs-pending` in `next-step.test.ts`, `forwards to /view once the enclave is running the job` in `view/code/page.test.tsx`) went green with no change to `next-step.ts`. The predicate started answering "yes" for executing researchers purely because the table gained a screen. The only conflict in the whole merge was the two-line one I flagged in the description, resolved take-theirs as predicted.

src/lib/study-screen/state.fixture.ts
therealmarv: Keeping them together then, thanks. Agreed that the net -97 only reads as a win next to the change that motivated it.

PR #931, opened 2026-08-04
Split local e2e by infrastructure

therealmarv:
Running e2e locally was broken for developers whose PostgreSQL runs only inside Docker. The previous launcher required host PostgreSQL tools and a reachable `localhost:5432`, while the main development Compose stack does not publish PostgreSQL to the host.

This PR splits local e2e by infrastructure and makes the Docker path deterministic across its containers and host Playwright process.

## Entrypoints

| Command | Infrastructure | App |
| --- | --- | --- |
| `./bin/docker-e2e` | Dedicated PostgreSQL and SeaweedFS from `docker-compose.e2e.yml` on ports 5433 and 8334 | Dedicated Docker app on port 4101 |
| `./bin/local-e2e` | Developer-managed PostgreSQL and SeaweedFS on ports 5432 and 8333 | Main-compatible host app on port 4100 |
| `pnpm run test:e2e` | Already-running infrastructure, used by CI | CI-managed server |

`pnpm run test:e2e:up` remains an alias for `./bin/local-e2e`.

## Docker e2e

`./bin/docker-e2e` starts dedicated PostgreSQL and SeaweedFS containers, creates the local bucket, migrates and seeds inside Docker, builds and starts the e2e app in Compose, and then runs Playwright on the host. Host PostgreSQL tools are not required, and Playwright UI and headed modes remain interactive.

The effective environment is built once with this precedence:

```text
Docker-owned values > .env.test > .env > ambient values
```

`.env` remains the base and `.env.test` overrides ordinary configuration when present. Docker then forces its database URL, S3 endpoints, local AWS credentials, bucket, fake Clerk mode, and app endpoint so environment files cannot redirect one participant to dev infrastructure.

Supported topology customization is limited to `E2E_PG_PORT`, `E2E_S3_PORT`, `E2E_APP_PORT`, and `COMPOSE_PROJECT_NAME`. Port values are validated before startup.

The seed and app containers load optional `.env` and `.env.test` files in that order. Compose keeps explicit overrides for container-side database and S3 endpoints, fake Clerk, package-manager settings, blank `AWS_PROFILE`, and other Docker-owned values. It clears `NO_TESTING_DATA` for seeding so local configuration cannot suppress the fixed users required by the suite.

Remote database-secret, application-secret, AWS profile, and Claude API selectors are removed from Docker e2e host processes. The regular environment started by `docker compose up` is unchanged.

The app uses a disposable anonymous Docker volume at `/app/.next`, masking the bind-mounted checkout. Docker e2e therefore neither reads nor writes the host checkout's `.next` directory. The launcher removes the app container and anonymous build volume after every run, including failures and interrupts, while preserving the named dependency, database, and object-storage volumes.

The disposable app is killed before removal rather than waiting through Docker's graceful stop timeout. Measured cleanup fell from 10 seconds to 0-1 second per run.

Playwright starts only after the app healthcheck succeeds and does not manage a second host app in Docker mode.

The implementation keeps four scripts with distinct responsibilities:

| File | Responsibility |
| --- | --- |
| `bin/local-e2e` | Developer-managed infrastructure workflow |
| `bin/docker-e2e` | Docker Compose lifecycle, seeding, app startup, and Playwright orchestration |
| `bin/docker-e2e-environment.ts` | Testable dotenv merge and authoritative Docker values |
| `bin/app-test` | Next.js build/server command used on the local host or inside the Docker app container |

Available stack commands:

- `./bin/docker-e2e --clean`
- `./bin/docker-e2e --down`
- `./bin/docker-e2e --prune`
- `./bin/docker-e2e --print-project-name`
- `./bin/docker-e2e --print-environment`

## Local e2e

`./bin/local-e2e` retains the environment behavior from `main`:

- It sources `.env.test` when present.
- Migration and seeding receive a generated `si_mgmnt_test` URL scoped only to those commands.
- Playwright and `app-test` inherit `.env.test` or the normal environment-loading behavior.
- It does not receive Docker-specific database or S3 defaults.

The local launcher still avoids `pnpm run db:migrate` because that command regenerates tracked database types. It runs Kysely migration, seeds, and `bin/seed-environment.ts` directly instead.

Playwright arguments are accepted with or without a leading `--` on the local launcher. Docker keeps strict parsing for its own management options.

## Performance evidence

Two baseline warm runs averaged 74.39 seconds. The stable phase measurements were:

| Phase | Average |
| --- | ---: |
| Database preflight | 1.0s |
| Infrastructure and bucket | 1.5s |
| Migration and seeding | 14.5s |
| App build and startup | 26.5s |
| Playwright and reporting | 19.5s |
| App cleanup | 10.0s |

Detailed measurements attributed 11 seconds of seeding to `seed-environment.ts` and 24 seconds of app startup to `next build`. Immediate disposable cleanup removes 9-10 seconds from the baseline.

Disabling V8 coverage did not improve Playwright performance: two successful coverage-off samples measured 20 and 21 seconds versus 19 and 20 seconds in the baseline. Coverage therefore remains enabled by default.

## CI

CI remains unchanged. `.github/workflows/checks.yml` is unchanged from `main`; CI continues to run `pnpm run test:e2e` against workflow-managed services and `bin/ci-server`.

## Verification

- `pnpm run checks`
- Docker environment unit tests: 3/3 passed
- Shell syntax and Compose configuration validation passed
- Conflicting ambient CI, AWS, database, and port values produced the authoritative Docker environment
- Corrected app healthcheck reached healthy state before Playwright
- Targeted error-log review passed at `retries: 0`
- Full Docker e2e passed 28/28 three consecutive times at `retries: 0` across two coverage-off measurements and one default run
- Host `.next` content hash remained unchanged across successful Docker e2e runs
- App container and anonymous `.next` volume were absent after each run

The local e2e path has not been executed on this machine because it does not have the required host PostgreSQL setup. The Monocart report still records zero local standalone coverage bytes because its existing URL filter does not match standalone chunk URLs; test execution is unaffected.


bin/docker-e2e
```diff
@@ -0,0 +1,156 @@
...
+if [ "$clean" = true ]; then
+    echo "Cleaning test DB + object storage (dependencies preserved)..."
+    cleanup_app
+    dc down
+    docker volume rm "${COMPOSE_PROJECT_NAME}_pgdata-e2e" "${COMPOSE_PROJECT_NAME}_seaweedfs-data-e2e" 2>/dev/null || true
+else
+    # Rebuild the DB volume when an applied migration no longer exists after a branch switch.
+    dc up -d db-e2e --wait 2>/dev/null || true
+    applied=$(dc exec -T db-e2e psql -U si -d si_test -t -c "SELECT name FROM kysely_migration" 2>/dev/null | tr -d ' ' | grep -v '^$' || echo "")
+    filesystem=$(ls src/database/migrations/*.ts 2>/dev/null | xargs -n1 basename | sed 's/\.ts$//')
```
nathanstitt, line 127: This block can silently destroy the test DB volume. If `ls src/database/migrations/*.ts` returns nothing, `filesystem` is empty, so the very first applied migration fails the `grep -qx` and we drop into the rebuild branch — `docker volume rm ..._pgdata-e2e`.

The `ls` is relative, but by this point we've re-exec'd through `pnpm exec tsx`, so it depends on cwd surviving that hop. Everything else in the script uses the absolute `$repo_root`; this one line doesn't. Could we use `"$repo_root/src/database/migrations"` and, more importantly, bail out rather than delete when the glob comes back empty? Something like:

```bash
filesystem=$(ls "$repo_root"/src/database/migrations/*.ts 2>/dev/null | xargs -n1 basename | sed 's/\.ts$//')
[ -z "$filesystem" ] && { echo "No migrations found; skipping mismatch check." >&2; applied=""; }
```

An empty list means "we couldn't read the migrations," which is not the same as "every applied migration is stale."

bin/docker-e2e-environment.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+    return selected
+}
+
+export function dockerE2EEnvironment({
+    ambient = process.env,
+    envFile = readEnvironmentFile('.env'),
+    envTestFile = readEnvironmentFile('.env.test'),
+}: DockerE2EEnvironmentOptions = {}): Record<string, string | undefined> {
+    const environment = { ...ambient, ...envFile, ...envTestFile }
```
nathanstitt, line 32: The precedence here is inverted relative to what the PR description and README promise. The description says:

> Docker-owned values > .env.test > .env > ambient values

and the README tells devs to "Customize the published ports with `E2E_PG_PORT`, `E2E_S3_PORT`, or `E2E_APP_PORT`" — which reads as a shell-export workflow. But `{ ...ambient, ...envFile, ...envTestFile }` puts the *files* last, so they win over the shell.

So `E2E_PG_PORT=5599 ./bin/docker-e2e` is silently ignored by anyone who has that key in their `.env.test`, which is exactly the dev this PR is aimed at. Usually ambient wins over dotenv files (that's what `dotenv` itself does by default — it won't clobber an existing `process.env` key). Worth flipping to `{ ...envFile, ...envTestFile, ...ambient }`, or documenting the reverse ordering deliberately if it's intentional.

bin/docker-e2e
```diff
@@ -0,0 +1,156 @@
...
+
+repo_root=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)
+cd "$repo_root"
+
+# Re-enter this script after the TypeScript loader has merged .env + .env.test and applied
+# authoritative Docker database, S3, auth, and port values. Keeping the merge in TypeScript
+# avoids sourcing dotenv syntax in Bash and keeps its precedence rules unit-testable.
+if [[ "${1:-}" != "--environment-ready" ]]; then
+    unset CI AWS_PROFILE
+    exec pnpm exec tsx bin/docker-e2e-environment.ts ./bin/docker-e2e --environment-ready "$@"
```
nathanstitt, line 24: The re-exec happens before any argument parsing, so purely informational flags pay the full startup cost: `./bin/docker-e2e --help` spawns `pnpm exec tsx`, reads and merges both dotenv files, and validates ports — all to print a usage block. Same for `--print-project-name`.

It also means `--help` can *fail* instead of helping: an out-of-range `E2E_PG_PORT` in `.env.test` throws from `port()` before usage is ever reached. Could we handle `--help`/`-h`/`--print-project-name` above this line, before the trampoline?

bin/app-test
```diff
@@ -1,58 +1,50 @@
...
-# ClerkProvider guards on a publishable key being present; the fake ignores its value, so
-# a dummy pk_test key satisfies the guard if .env.test doesn't provide one.
 export NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY="${NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY:-pk_test_ZHVtbXktZm9yLWUyZS5jbGVyay5hY2NvdW50cy5kZXYk}"
 
 if [ "${1:-}" != "--no-build" ]; then
-    echo "app-test: building (E2E_FAKE_CLERK=1)…"
-    env -u AWS_PROFILE pnpm exec next build
+    echo "app-test: building (E2E_FAKE_CLERK=1)..."
+    if [ "${E2E_MODE:-}" = "docker" ]; then
+        ./node_modules/.bin/next build
```
nathanstitt, line 39: Any reason the docker branch calls `./node_modules/.bin/next` directly instead of `pnpm exec next`? Two paths to the same binary is the sort of thing that drifts.

If it's to avoid corepack's network fetch inside the container, a comment saying so would help — otherwise the next person reading this will "tidy" it back to `pnpm exec` and get an intermittent container failure. Note the local branch also wraps it in `env -u AWS_PROFILE` while the docker branch doesn't; the `unset AWS_PROFILE` on line 33 covers both, so the `env -u` may now be redundant on the local side too.

docker-compose.e2e.yml
```diff
@@ -0,0 +1,183 @@
...
+            # Override .env's dev DATABASE_URL to target the isolated test DB by service DNS.
+            - DATABASE_URL=postgres://si:si@db-e2e:5432/si_test
+            - E2E_FAKE_CLERK=1
+            - NODE_ENV=test
+            # A local override must never disable the fixed users required by the e2e suite.
+            - NO_TESTING_DATA=
+            # Drop any AWS_PROFILE from .env so the SDK never blocks on expired SSO.
+            - AWS_PROFILE=
+            # Corepack provisions pnpm at runtime; cache it inside the node_modules volume
+            # (mirrors docker-compose.testunit.yml) and never prompt.
```
nathanstitt, line 158: `docker-compose.testunit.yml` doesn't exist in the repo — this is the only reference to it anywhere. Either it was renamed/removed at some point, or the file was planned and never landed.

Small thing, but a comment pointing at a nonexistent file as precedent is worse than no comment, since it sends the reader looking for context that isn't there. Worth dropping the parenthetical or pointing it at whatever actually establishes the pattern.

src/docker-e2e-environment.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, expect, it } from 'vitest'
+import { dockerE2EEnvironment } from '../bin/docker-e2e-environment'
```
nathanstitt, line 2: This test lives in `src/` but reaches up into `bin/` — presumably because `vitest.config.ts` only includes `src/**` and `services/**`, so it wouldn't be collected under `bin/`.

CONVENTIONS.md says to place tests next to their source, and everything else in `src/` is application code that ships. Rather than parking a build-tooling test at the app root, could we add `bin/**/*.test.ts` to the vitest `include` and move it to `bin/docker-e2e-environment.test.ts`? Keeps the convention intact and stops this file from showing up in app coverage.

bin/local-e2e
```diff
@@ -0,0 +1,122 @@
...
+note "Postgres reachable on ${PGHOST}:${PGPORT}"
+
+if ! nc -z -w 1 "$PGHOST" "$S3_PORT" >/dev/null 2>&1; then
+    die_local_infra "SeaweedFS not reachable on ${PGHOST}:${S3_PORT} (try 'pnpm run s3:local')."
+fi
+note "SeaweedFS reachable on ${PGHOST}:${S3_PORT}"
+
+if ! psql "postgres://${PGHOST}:${PGPORT}/postgres" -tAc "SELECT 1 FROM pg_database WHERE datname='${TEST_DB}'" 2>/dev/null | grep -q 1; then
+    note "Creating database ${TEST_DB}…"
+    createdb -h "$PGHOST" -p "$PGPORT" ${PGUSER:+-U "$PGUSER"} "$TEST_DB"
```
nathanstitt, line 103: Unquoted `${PGUSER:+-U "$PGUSER"}` — the inner quotes don't survive word-splitting here, so a `PGUSER` containing a space splits into extra arguments. The script it replaces had the same shape, so this isn't a regression, but since we're rewriting the file it'd be a good moment to fix it:

```bash
createdb_args=(-h "$PGHOST" -p "$PGPORT")
[ -n "${PGUSER:-}" ] && createdb_args+=(-U "$PGUSER")
createdb "${createdb_args[@]}" "$TEST_DB"
```

src/hooks/use-outputs-decision.ts
```diff
@@ -84,11 +83,8 @@ export function useOutputsDecision({
...
             queryClient.invalidateQueries({ queryKey: ['org-studies', orgSlug] })
             setConfirming(null)
-            // push() alone would be a no-op here: /review is already the current URL, so the
-            // decrypted review form would stay mounted with plaintext on screen even though the
-            // decision is final. refresh() is what re-runs the server components and lets the
-            // reviewer screen rules resolve to the post-decision screen.
-            router.push(Routes.studyReview({ orgSlug, studyId }))
+            // The current URL is already /review. A same-URL push can race this refresh and leave
+            // the decrypted form mounted after the decision commits.
             router.refresh()
```
nathanstitt, line 92: This change (dropping the `router.push` and the now-unused `Routes` import) looks unrelated to splitting the e2e entrypoints — it's a behavioral change to the outputs-review flow riding along in an infrastructure PR.

The new comment also reverses the reasoning of the one it replaces: the old comment said `push()` alone was a no-op and `refresh()` was the fix, and now we're saying the `push()` actively races the `refresh()`. Both can't be the whole story, and this is security-adjacent code (it governs whether decrypted plaintext stays on screen after a final decision). Worth splitting into its own PR where the race can be reviewed and tested on its merits.

nathanstitt (commented):
This seems like a lot of churn for something that could be fixed by publishing a port on the dev Postgres `ports: ['5433:5432']` as noted in the review.  Do you feel the changes are worthwhile outside that?

PR #932, opened 2026-08-04
OTTER-642 follow-up: keep re-deliveries from replacing artifacts under review

therealmarv:

## Why

QA found that on an errored study the reviewer can read the logs, each listed exactly once, but the researcher cannot decrypt them: `Private key is not valid for these results, check with your administrator`. The researcher's key is fine, which was confirmed by the same account and the same key opening other studies. So the mismatch is per study, not per account.

Wrapped recipient keys are written **last**. The reviewer's browser fetches a ciphertext once at page mount, keeps each inner file's AES key in memory through the mandatory feedback, and persists the wrap only when the decision is submitted. Until that moment the artifact was still replaceable, because the only guards were "keys already wrapped" (`artifactWasShared`) and "round already decided" (`roundIsClosed`). A re-delivery landing inside that window replaced the object, and the wrap that followed described bytes that no longer existed. The researcher then holds a key that unwraps correctly and decrypts to nothing, which surfaces as the message above.

An errored run is the exposed case. `RUN-COMPLETE` re-deliveries are turned away by the results route's repeat guard, whereas `JOB-ERRORED` neither blocks that route nor closes a round, so the artifact stayed replaceable for the whole review. Repeated delivery on the errored path is not exotic either: it is what OTTER-642 was originally filed about.

## What changed

`unreplaceableReason` in `src/server/storage.ts` now asks two further questions before letting a re-delivery replace an artifact:

- **has it been opened**, via the `study_job_file_activity` rows OTTER-675 already writes. An opened artifact is one whose AES keys may already sit in someone's browser.
- **was the run outcome announced after the job already held it**, using `OUTPUTS_REVIEWABLE_JOB_STATUSES`. That is the moment reviewers were told there is something to review, and it lands before anyone can have opened anything, so it is the check that actually closes the window.

The announcement check is scoped to statuses recorded at or after the artifact's `createdAt`, mirroring the scoping the results route applies to its own announcement lookup. `JOB-ERRORED` is shared with the scanner, the containerizer and the generic status route, and one of those can land *before* the run's own log arrives, so a bare status lookup would read someone else's earlier failure as proof this artifact is under review and refuse its first genuine refresh. A test pins that.

Deliberately **not** done: extending the results route's early `alreadyReceived` 422 from `RUN-COMPLETE` to `JOB-ERRORED`. Same producer ambiguity, worse consequence: a pre-existing `JOB-ERRORED` from the containerizer would make the route reject the run's real error log before it is ever stored, losing it. The artifact-level guard reaches the same outcome without that risk.

## Researcher decryption test gap

This path had no real coverage. `study-job.actions.test.ts` asserts on placeholder crypt strings (`'wrapped-for-researcher'`), and the errored e2e stops at asserting `Code Run Log` is visible without ever entering the researcher's key, so the splice of recipient keys into a manifest the researcher is not named in only ever ran on a deployed environment.

`src/hooks/use-decrypt-files.test.tsx` now runs a genuine round trip: an artifact encrypted for the enclave's keypair, read the way the reviewer's browser reads it, re-wrapped for a **separately generated** researcher keypair, then decrypted through the real `fetchEncryptedJobFilesAction`, `ResultsReader` and `EncryptedFilesPanel`. The researcher's pair is deliberately not the shared fixture key every seeded user holds, since a researcher who shares the enclave's key is a manifest recipient by accident and would pass without the recipient-key path running at all. Only the object-store read is mocked.

Three cases: the happy path, a typed key that is not the enrolled one (the message QA saw), and keys wrapped against a superseded copy of the ciphertext (this bug, from the researcher's side). The superseded case renames the inner file so the manifest lookup misses deterministically; a same-name replacement fails on AES padding instead, which CBC does not guarantee to detect, and a probabilistic assertion has no place in the suite.

## Note for QA tooling

`/api/qa` study-state driving reuses these storage helpers, so re-attaching files to a QA study whose outcome has already been announced is now refused and logged rather than silently overwriting. Reaching a fresh artifact state means a fresh study.

## Tests

`./bin/docker-test`: 2393 passed, 251 files. `typecheck`, `validate-actions`, and eslint/prettier over the touched files are clean. No e2e change: adding a researcher decrypt step there is worth doing, but not in a PR whose full suite I have not run against a deployed stack.

## Not covered here

The message itself. A key that does not match the enrolled one still reports "not valid for these results" rather than "this key does not match the security key registered to your account", and no screen shows the enrolled fingerprint. That is a separate card, along with the finding that `bin/seed-environment.ts` and `bin/find-or-update-clerk-test-users.ts` enroll the shared test key for `admin` and `reviewer` only, never for `researcher`, and never reconcile a key row that has drifted.

### DIFF
--- a/src/hooks/use-decrypt-files.test.tsx
+++ b/src/hooks/use-decrypt-files.test.tsx
@@ -1,0 +1,186 @@
+import { describe, expect, it, vi } from 'vitest'
+import { fireEvent, screen, waitFor } from '@testing-library/react'
+import {
+    db,
+    insertTestStudyJobData,
+    mockSessionWithTestData,
+    readTestSupportFile,
+    renderWithProviders,
+} from '@/tests/unit.helpers'
+import { EncryptedFilesPanel } from '@/components/encrypted-files-panel'
+import { latestJobForStudy } from '@/server/db/queries'
+import { insertSharedFileKeys } from '@/server/results-sharing'
+import { fetchFileContents } from '@/server/storage'
+import { ResultsWriter } from 'si-encryption/job-results/writer'
+import { ResultsReader } from 'si-encryption/job-results/reader'
+import { wrapAesKey } from 'si-encryption/job-results/crypto'
+import { fingerprintKeyData, pemToArrayBuffer } from 'si-encryption/util'
+import { generateKeyPair } from 'si-encryption/util/keypair'
+
+// Only the object-store read is mocked: the ciphertext is real, and so is every step between it and
+// the screen (fetchEncryptedJobFilesAction, the recipient-key lookup, ResultsReader, the panel).
+vi.mock('@/server/storage', async (importOriginal) => ({
+    ...(await importOriginal<typeof import('@/server/storage')>()),
+    fetchFileContents: vi.fn(),
+}))
+
+const KEY_PLACEHOLDER = 'Enter your Results Key to access encrypted content.'
+const MISMATCH_MESSAGE = 'Private key is not valid for these results, check with your administrator'
+
+// The researcher's own pair, generated once because RSA-4096 keygen is the slowest thing here. It is
+// deliberately NOT the shared fixture pair every seeded user holds: a researcher who shares the
+// enclave's key is a manifest recipient by accident, and would pass this suite without the
+// recipient-key path ever running.
+const researcher = await generateKeyPair()
+const researcherPrivateKeyPem = `-----BEGIN PRIVATE KEY-----\n${researcher.privateKeyString}\n-----END PRIVATE KEY-----`
+
+// The enclave writes the manifest against the reviewing org's keys, which is the fixture pair here.
+const enclaveRecipient = async () => {
+    const publicKey = pemToArrayBuffer(await readTestSupportFile('public_key.pem'))
+    return { publicKey, fingerprint: await fingerprintKeyData(publicKey) }
+}
+
+const toArrayBuffer = (str: string): ArrayBuffer => {
+    const buf = Buffer.from(str, 'utf-8')
+    return buf.buffer.slice(buf.byteOffset, buf.byteOffset + buf.byteLength)
+}
+
+// One encrypted artifact as TOA produces it: whole zip, embedded manifest, keys for the enclave only.
+async function encryptForEnclave(files: { name: string; content: string }[]) {
+    const recipient = await enclaveRecipient()
+    const writer = new ResultsWriter([recipient])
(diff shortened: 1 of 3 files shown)


therealmarv:
## Scope note: what this does and does not fix for the QA report

Referring to [OTTER-675 comment 45880](https://openstax.atlassian.net/browse/OTTER-675?focusedCommentId=45880), where the researcher sees both logs listed once each and then fails to decrypt with `Private key is not valid for these results, check with your administrator`.

**Forward-only, and conditional on the cause.** This PR stops a re-delivery from replacing an artifact once the run outcome is announced or once anyone has opened it, so on a study created after the deploy the keys the reviewer wraps always describe the bytes the researcher will read. It cannot repair the two repro studies (`019fcd9f-f197-7573-b2a6-2664d3eaf6f7`, `019fcd48-047d-7061-88ed-fdd12be37c1a`) or any study already shared: those recipient-key rows were wrapped from a copy that no longer exists, the bucket is unversioned, and the AES key for the replaced bytes went with them. Those two studies stay unreadable for the researcher, so the criterion has to be retested on a fresh study.

**If the cause is key drift on the shared QA researcher account instead, this PR changes nothing for that symptom.** Drift was the reading in comment 45880 and in the OTTER-642 thread. The observation that the same account and key open other studies argues against it, but that is only conclusive if those studies were opened with that same key while that same fingerprint was enrolled.

### Retest order that distinguishes the two

Deploy this, then retest one fresh errored study **without touching the account**:

- researcher decrypts, so the stale wrap was the cause and the guard holds;
- researcher still fails, so the cause is the account key, and the `qar fix-account` proposed in 45880 is the fix.

Reprovisioning and deploying at the same time destroys the signal, which is worth stating because 45880's own remedy is the reprovision.

### One diagnostic this adds regardless

A refused re-delivery now logs `ignoring re-delivered ENCRYPTED-CODE-RUN-LOG for job <id>: its outputs are already under review`. An accepted overwrite previously left no trace anywhere, which is why this went unexplained across two QA cycles. From the deploy onward, forcing a single errored study and grepping the app log answers whether re-deliveries happen on the errored path at all, with no database access needed.

### Why this was invisible until now

PR #919 removed the visible symptom of a re-delivery, the duplicate row, while leaving the destructive overwrite in place. A second delivery on those studies would therefore not show up anywhere in the UI, which is part of why the failure looked like it had to be a key problem.

The third test added here reproduces the reported experience exactly (files listed, key valid for the account, decryption fails) when the wrapped keys describe a superseded copy, so the failure mode is pinned in the suite even though the already-deployed studies stay broken.

therealmarv:
I will close this for now, if this is still needed we can reopen it.

PR #933, opened 2026-08-04
OTTER-675 follow-up: announce the autosave status in a live region

therealmarv:
see [OTTER-675](https://openstax.atlassian.net/browse/OTTER-675?focusedCommentId=45895)

Follow-up to #924, which QA rejected on a single acceptance-criteria row.

## The failure

The Decision input AC requires: *&#34;&#39;All changes saved&#39; must render in an `aria-live=&#34;polite&#34;` region so it&#39;s announced without stealing focus; confirm debounce/throttle so the announcement doesn&#39;t fire so frequently it becomes noisy.&#34;*

Autosave itself worked, and the indicator appeared, but QA found it had no `aria-live` and no `role=&#34;status&#34;` ancestor, so a screen reader user got no confirmation that their work was saved.

The root cause is not a missing attribute. `SaveStatusIndicator` returned `null` until it had something to report, so the wrapper and its text entered the DOM in the same step. A live region is only announced when content it **already owns** changes; a region mounted together with its content is silent in most AT/browser pairs. Adding `aria-live` to the label as it was would not have fixed it.

## The fix

`src/components/save-status.tsx`, the shared indicator, so every surface that uses it (collaborative editor, proposal fields, resubmission note) gets the fix:

* The region is now mounted unconditionally and starts empty; only the label inside it comes and goes. `role=&#34;status&#34;` and `aria-live=&#34;polite&#34;` are both spelled out, plus `aria-atomic=&#34;true&#34;`, because AT/browser pairs differ on which one they act on.
* `&#34;Saving…&#34;` is rendered **outside** the region on purpose. It flips back to `&#34;saved&#34;` on every pause in typing, so announcing it as well would produce two interruptions per save cycle, which is exactly the noise the AC asks us to hold down. It stays plain visible text that a screen reader still reads when the user navigates to it. This gives one announcement per save cycle with no timer.
* The `isVisible={false}` behavior from OTTER-674 is unchanged: while a field&#39;s validation error owns the slot, no label is drawn, and the region simply sits empty. The announcement is deferred to when the error clears, not dropped.

No visual change on any surface. The region is a zero-size empty element when it has nothing to say, and `data-testid=&#34;autosave-status&#34;` stays on the label, so existing tests and the e2e `getByText(/All changes saved/i)` are unaffected.

## One announcement per save source

Raised in review, and the more substantive half of this PR. Mounting the region unconditionally means every surface that drops a `SaveStatusIndicator` gains a permanent region, and the proposal form drops three of them (title, datasets, PI). All three mirror **one** Yjs provider (`useProviderSaveStatus(yjsForm.provider)`, gated per field on `editedKeys`), so once a researcher has touched all three, every save cycle filled three live regions in the same tick.

Live regions are announced independently; there is no deduplication across regions in NVDA, JAWS or VoiceOver. That is three announcements per save, which is precisely the noise the AC asks us to hold down.

* `SaveStatusIndicator` takes an `announce` prop (default `true`). With `announce={false}` the visible label is unchanged and only that indicator&#39;s own live region is dropped.
* `SaveStatusAnnouncer`, a visually hidden `role=&#34;status&#34;` region, announces once for a surface whose indicators share a save source. `announcedSaveStatus()` collapses the per-field statuses into the one status it reports, so it never claims a save the user cannot see confirmed on screen.
* The proposal form&#39;s three fields now pass `announce={false}` and announce through a single announcer. The four collaborative text editors on that page own separate providers, so they keep their own regions and are unaffected.

The regions carry no accessible name. Several screen readers speak a live region&#39;s name ahead of its content ([MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/status_role), [ARIA22](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22)), which would turn every save into &#34;&lt;name&gt;, All changes saved&#34;. Disambiguation for tests is done with `data-testid` (`autosave-live-region` / `autosave-announcer`), which costs no speech.

## Two further live-region fixes

* `collaborative-resubmission-note-section.tsx` gated the whole indicator behind `singleUserEditing && !error`, so a validation error unmounted the region and clearing it handed the region back with &#34;All changes saved&#34; already inside, which is silent. The error gate now goes through the indicator&#39;s `isVisible` instead; only the collaborative-mode gate still unmounts, which is correct because the editor draws its own indicator there.
* `resubmission-note-section.tsx` rendered the indicator **inside** `&lt;Box id={fieldErrorId(&#39;resubmissionNote&#39;)}&gt;`, the node the textarea&#39;s `aria-describedby` points at. It is not reachable today, because Mantine only folds the error id into `describedBy` when the error is truthy and the indicator is hidden in exactly that state, but a live region inside a field&#39;s description is a trap for the next edit. The indicator is now a sibling of the error node, with the row layout unchanged.

## Also addressed

QA logged this under *Also observed* rather than as a failure, noting it was &#34;worth a look if you touch this area&#34;:

&gt; The radio group is flagged without `aria-invalid`. On a failed submit, the group gets `aria-describedby` pointing at &#34;Select an option before submitting&#34;, but neither the group nor the inputs get `aria-invalid` [...] the feedback field does set `aria-invalid` and the two are now inconsistent.

The decision radios now carry `aria-invalid` while the unselected error is showing, which satisfies the `aria-invalid` + `aria-describedby` branch the Navigation AC allows. Note on placement: per ARIA, the invalid state belongs on the `role=&#34;radiogroup&#34;` element, but Mantine renders that element itself inside `Radio.Group` (`InputsGroupFieldset`) and passes nothing through to it, so the inputs are the only reachable target. Left as a comment in the code.

## Deliberately not changed

* **The two copy deviations** (&#34;Output files&#34; vs &#34;Outputs files&#34;, and words vs characters for the 300 limit). Both follow Figma, both were flagged for Iris&#39;s sign-off on #924, and that sign-off is still outstanding. Unchanged here so the decision stays with PM.
* **`cursor: pointer` on the file name in its default state.** It follows from the file name being a real ``, which the AC separately requires. QA recorded no user impact.

## Testing

* Nine cases in `save-status.test.tsx`: the saved label announces from a polite, atomic live region; the region stays mounted and empty while idle; the saving label is kept out of it; the region empties when a validation error takes the slot; the region carries no accessible name; `announce={false}` drops the region and keeps the label; plus the announcer and `announcedSaveStatus`.
* `proposal/form.test.tsx` (new) asserts at page level that the three fields announce through exactly one region. The isolated component tests cannot catch cross-field duplication.
* `collaborative-resubmission-note-section.test.tsx` pins that the region survives a validation error and that collaborative mode mounts none.
* `resubmission-note-section.test.tsx` pins that no live region sits inside the textarea&#39;s description.
* Two cases in `outputs-decision-section.test.tsx` for the radio `aria-invalid` state, present with an error and absent without.
* Full unit suite green twice (251 files, 2404 tests), plus `pnpm run checks`.

### Browser verification

Both components were driven in a real Chromium session against a locally served build, re-running the exact DOM queries from the QA report. 19/19 checks passed:

| check | before (QA) | after |
| --- | --- | --- |
| `el.closest(&#39;[aria-live]&#39;)` on &#34;All changes saved&#34; | `null` | `polite` |
| `el.closest(&#39;[role=&#34;status&#34;],[role=&#34;alert&#34;]&#39;)` | `null` | `status` |
| region `aria-atomic` | n/a | `true` |
| region accessible name (must stay absent) | n/a | `null` |
| region content across a real save cycle | n/a | `&#34;&#34;` → `&#34;All changes saved&#34;` |
| &#34;Saving…&#34; ancestors (must stay outside the region) | n/a | `null` / `null` |
| `announce={false}`: label rendered, no region | n/a | holds, pixel-identical to `announce={true}` |
| announcer size and positioning | n/a | 1x1, `position: absolute`, no layout impact |
| `isVisible=false`: label absent, region mounted and empty | n/a | holds |
| live region inside the field&#39;s `aria-describedby` target | present | absent |
| resubmission note footer row, saved / saving / error | n/a | unchanged, no layout shift |
| decision radios `aria-invalid` on failed submit | `[null, null]` | `[&#34;true&#34;, &#34;true&#34;]` |
| decision radios `aria-invalid` with no error | n/a | `[null, null]`, not `&#34;false&#34;` |

Re-checked alongside these, all still holding: two native radios sharing one `name`, the radiogroup&#39;s accessible name &#34;Sharing decision&#34;, per-option `aria-describedby`, the group error reachable through `aria-describedby`, the feedback error&#39;s own polite live region, `aria-required` on the editor, and both options unselected on load.

### DIFF
--- a/src/components/save-status.test.tsx
+++ b/src/components/save-status.test.tsx
@@ -1,7 +1,7 @@
 import { describe, expect, it } from 'vitest'
 import { renderWithProviders, screen } from '@/tests/unit.helpers'
 import { theme } from '@/theme'
-import { SaveStatusIndicator } from './save-status'
+import { SaveStatusAnnouncer, SaveStatusIndicator, announcedSaveStatus } from './save-status'
 
 describe('SaveStatusIndicator', () => {
     it('renders nothing while idle', () => {
@@ -59,5 +59,48 @@ describe('SaveStatusIndicator', () => {
     it('empties the live region when hidden by a validation error (OTTER-674)', () => {
         renderWithProviders(<SaveStatusIndicator status="saved" isVisible={false} />)
         expect(screen.getByRole('status')).toBeEmptyDOMElement()
+    })
+
+    it('leaves the live region unnamed, so the name is not spoken ahead of the save (OTTER-675)', () => {
+        // Several screen readers read a live region's accessible name before its content, which
+        // would turn each save into "<name>, All changes saved". Tests select the region by
+        // data-testid instead.
+        renderWithProviders(<SaveStatusIndicator status="saved" />)
+        const region = screen.getByTestId('autosave-live-region')
+        expect(region).not.toHaveAttribute('aria-label')
+        expect(region).not.toHaveAttribute('aria-labelledby')
+    })
+
+    it('drops its own live region when announce is false, keeping the visible label (OTTER-675)', () => {
+        renderWithProviders(<SaveStatusIndicator status="saved" announce={false} />)
+        expect(screen.getByTestId('autosave-status')).toHaveTextContent('All changes saved')
+        expect(screen.queryByRole('status')).not.toBeInTheDocument()
+        expect(screen.getByTestId('autosave-status').closest('[aria-live]')).toBeNull()
+    })
+})
+
+describe('SaveStatusAnnouncer', () => {
+    it('announces the save from a polite, atomic region', () => {
+        renderWithProviders(<SaveStatusAnnouncer status="saved" />)
+        const region = screen.getByRole('status')
+        expect(region).toHaveAttribute('aria-live', 'polite')
+        expect(region).toHaveAttribute('aria-atomic', 'true')
+        expect(region).toHaveTextContent('All changes saved')
+    })
+
+    it('stays mounted and empty until a save lands', () => {
+        renderWithProviders(<SaveStatusAnnouncer status="saving" />)
+        expect(screen.getByRole('status')).toBeEmptyDOMElement()
+    })
+})
+
+describe('announcedSaveStatus', () => {
+    it('announces once when any field of a shared save source is saved', () => {
+        expect(announcedSaveStatus(['idle', 'saved', 'idle'])).toBe('saved')
+    })
+
+    it('stays idle while no field is showing a save', () => {
+        expect(announcedSaveStatus(['idle', 'saving', 'idle'])).toBe('idle')
+        expect(announcedSaveStatus([])).toBe('idle')
     })
 })
--- a/src/components/save-status.test.tsx
+++ b/src/components/save-status.test.tsx
@@ -34,4 +34,30 @@ describe('SaveStatusIndicator', () => {
         renderWithProviders(<SaveStatusIndicator status="saved" isVisible={false} />)
         expect(screen.queryByTestId('autosave-status')).not.toBeInTheDocument()
     })
+
+    it('announces "All changes saved" from a polite live region (OTTER-675)', () => {
+        renderWithProviders(<SaveStatusIndicator status="saved" />)
+        const region = screen.getByRole('status')
+        expect(region).toHaveAttribute('aria-live', 'polite')
+        expect(region).toHaveAttribute('aria-atomic', 'true')
+        expect(region).toHaveTextContent('All changes saved')
+    })
+
+    it('keeps the live region mounted and empty while idle, so the save is announced (OTTER-675)', () => {
+        // A live region is only announced when content it already owns changes, so the region
+        // has to be in the DOM before the label lands in it.
+        renderWithProviders(<SaveStatusIndicator status="idle" />)
+        expect(screen.getByRole('status')).toBeEmptyDOMElement()
+    })
+
+    it('leaves the transient saving label out of the live region (OTTER-675)', () => {
+        renderWithProviders(<SaveStatusIndicator status="saving" />)
+        expect(screen.getByTestId('autosave-status').closest('[aria-live]')).toBeNull()
+        expect(screen.getByRole('status')).toBeEmptyDOMElement()
+    })
+
+    it('empties the live region when hidden by a validation error (OTTER-674)', () => {
+        renderWithProviders(<SaveStatusIndicator status="saved" isVisible={false} />)
+        expect(screen.getByRole('status')).toBeEmptyDOMElement()
+    })
 })
--- a/src/components/save-status.tsx
+++ b/src/components/save-status.tsx
@@ -1,10 +1,12 @@
 'use client'
 
-import { FC } from 'react'
-import { Box, Group, Text, useMantineTheme } from '@mantine/core'
+import { FC, ReactNode } from 'react'
+import { Box, Group, Text, VisuallyHidden, useMantineTheme } from '@mantine/core'
 import { CheckCircleIcon } from '@phosphor-icons/react/dist/ssr'
 
 export type SaveStatusValue = 'idle' | 'saving' | 'saved'
+
+export const SAVED_LABEL = 'All changes saved'
 
 const SavingLabel: FC<{ isVisible: boolean }> = ({ isVisible }) => {
     if (!isVisible) return null
@@ -25,22 +27,71 @@ const SavedLabel: FC<{ isVisible: boolean }> = ({ isVisible }) => {
         <Group gap={8} wrap="nowrap" data-testid="autosave-status">
             <CheckCircleIcon size={16} color={theme.colors.green[9]} weight="fill" />
             <Text size="xs" c="green.9" fw={600}>
-                All changes saved
+                {SAVED_LABEL}
             </Text>
         </Group>
     )
 }
 
+// Holds the saved label, or passes it straight through on a surface that announces from one
+// shared region instead (see SaveStatusAnnouncer). Carries no accessible name on purpose:
+// several screen readers speak a live region's name ahead of its content, which would turn one
+// announcement into "Autosave status, All changes saved" on every save. `data-testid` gives tests
+// something unambiguous to select, since a page legitimately holds one region per save source.
+const SavedLiveRegion: FC<{ isVisible: boolean; children: ReactNode }> = ({ isVisible, children }) => {
+    if (!isVisible) return <>{children}</>
+
+    return (
+        <Box role="status" aria-live="polite" aria-atomic="true" data-testid="autosave-live-region">
+            {children}
+        </Box>
+    )
+}
+
+// Empty until a save lands, so the region owns its content before the text arrives.
+const announcedText = (status: SaveStatusValue) => (status === 'saved' ? SAVED_LABEL : '')
+
+/**
+ * Announces a save once for a surface that draws several indicators from a single save source.
+ *
+ * Live regions are announced independently, so three field indicators mirroring one Yjs provider
+ * mean three announcements per save cycle. Pair this with `announce={false}` on those indicators:
+ * they keep their visible labels and this one region does the talking.
+ */
+export const SaveStatusAnnouncer: FC<{ status: SaveStatusValue }> = ({ status }) => (
+    <VisuallyHidden role="status" aria-live="polite" aria-atomic="true" data-testid="autosave-announcer">
+        {announcedText(status)}
+    </VisuallyHidden>
+)
+
+/**
+ * Collapses the per-field statuses of one save source into the single status to announce.
+ *
+ * Each field is already gated on whether the user edited it and whether a validation error owns
+ * its row, so "any field is showing saved" is what the announcement should follow: it never
+ * claims a save the user cannot see confirmed somewhere on screen.
+ */
+export const announcedSaveStatus = (statuses: SaveStatusValue[]): SaveStatusValue =>
+    statuses.includes('saved') ? 'saved' : 'idle'
+
 interface SaveStatusIndicatorProps {
     status: SaveStatusValue
-    /** False while the field's validation error is showing — the error takes the slot (OTTER-674). */
+    /** False while the field's validation error is showing, since the error takes the slot (OTTER-674). */
     isVisible?: boolean
+    /**
+     * False on a surface where several indicators share one save source and a single
+     * {@link SaveStatusAnnouncer} announces for all of them. The visible label is unchanged;
+     * only this indicator's own live region is dropped.
+     */
+    announce?: boolean
 }
 
 // Single autosave indicator shared across every surface (collaborative editor,
 // proposal fields, resubmission note). Draws no visible label until there is something to
 // report, so it can be dropped under any field unconditionally.
-export const SaveStatusIndicator: FC<SaveStatusIndicatorProps> = ({ status, isVisible = true }) => {
+export const SaveStatusIndicator: FC<SaveStatusIndicatorProps> = ({ status, isVisible = true, announce = true }) => {
+    // Collapsing a hidden indicator to 'idle' empties the region rather than unmounting it, so a
+    // save that lands while an error owns the slot is announced when the error clears, not lost.
     const shown = isVisible ? status : 'idle'
 
     return (
@@ -55,9 +106,9 @@ export const SaveStatusIndicator: FC<SaveStatusIndicatorProps> = ({ status, isVi
                 with its text (what returning null while idle did) left screen reader users with
                 no confirmation that their work was saved. `role="status"` and `aria-live` are
                 both spelled out because AT/browser pairs differ on which one they act on. */}
-            <Box role="status" aria-live="polite" aria-atomic="true">
+            <SavedLiveRegion isVisible={announce}>
                 <SavedLabel isVisible={shown === 'saved'} />
-            </Box>
+            </SavedLiveRegion>
         </>
     )
 }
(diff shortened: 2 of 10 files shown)


src/components/save-status.tsx
```diff
@@ -32,11 +38,26 @@ interface SaveStatusIndicatorProps {
...
+                "saved" on every pause in typing, so announcing it too would double the
+                interruptions the AC asks us to hold down. It stays plain visible text, which a
+                screen reader still reads when the user navigates to it (OTTER-675). */}
+            <SavingLabel isVisible={shown === 'saving'} />
+            {/* The region is mounted unconditionally and starts empty. A live region is only
+                announced when content it already owns changes, so mounting the wrapper together
+                with its text (what returning null while idle did) left screen reader users with
+                no confirmation that their work was saved. `role="status"` and `aria-live` are
+                both spelled out because AT/browser pairs differ on which one they act on. */}
+            <Box role="status" aria-live="polite" aria-atomic="true">
```
nathanstitt: Now that the region is mounted unconditionally, every surface that drops a `SaveStatusIndicator` gains a permanent `role="status"` element — and `proposal/form.tsx` drops three of them, then renders `ProposalFooter` → `IncompleteFieldsHint`, which is its own `role="status"` (`src/components/incomplete-fields-hint.tsx:30`). That page goes from one status region to four.

No test breaks today (the only `getByRole('status')` callers are `security-key-view-button.test.tsx` and `research-details-section.test.tsx`, neither on this page), so this isn't a bug right now. But it does mean a future `getByRole('status')` on any proposal-page test will hit an ambiguous-match error, and AT users get several silent regions where there used to be one. Might be worth an `aria-label` on the region (e.g. `aria-label="Autosave status"`) so the multiple regions are at least distinguishable when someone does go looking.

src/components/save-status.tsx
```diff
@@ -32,11 +38,26 @@ interface SaveStatusIndicatorProps {
...
-// proposal fields, resubmission note). Renders nothing until there is something
-// to report so it can be dropped under any field unconditionally.
+// proposal fields, resubmission note). Draws no visible label until there is something to
+// report, so it can be dropped under any field unconditionally.
 export const SaveStatusIndicator: FC<SaveStatusIndicatorProps> = ({ status, isVisible = true }) => {
-    if (!isVisible) return null
-    if (status === 'saving') return <SavingLabel />
-    if (status === 'saved') return <SavedLabel />
-    return null
+    const shown = isVisible ? status : 'idle'
```
nathanstitt, line 95: Small thing on the `shown` collapse: mapping `isVisible=false` onto `'idle'` means the region empties, which is what we want — but it also silently drops the distinction between "idle" and "hidden because a validation error owns the slot". That's fine for rendering, though it does mean if `status` flips `saved → saving → saved` entirely while an error is showing, the region stays empty the whole time and then announces on the first render where the error clears. Probably the desired behavior, just worth being explicit that the announcement is deferred rather than lost.

src/components/save-status.tsx
```diff
@@ -32,11 +38,26 @@ interface SaveStatusIndicatorProps {
...
+                interruptions the AC asks us to hold down. It stays plain visible text, which a
+                screen reader still reads when the user navigates to it (OTTER-675). */}
+            <SavingLabel isVisible={shown === 'saving'} />
+            {/* The region is mounted unconditionally and starts empty. A live region is only
+                announced when content it already owns changes, so mounting the wrapper together
+                with its text (what returning null while idle did) left screen reader users with
+                no confirmation that their work was saved. `role="status"` and `aria-live` are
+                both spelled out because AT/browser pairs differ on which one they act on. */}
+            <Box role="status" aria-live="polite" aria-atomic="true">
+                <SavedLabel isVisible={shown === 'saved'} />
```
nathanstitt, line 110: One call site is worth re-checking under this change, since it's where the always-mounted region lands somewhere structurally load-bearing: `resubmission-note-section.tsx:71` renders `<SaveStatusIndicator>` **inside** `<Box id={fieldErrorId('resubmissionNote')}>`, and that id is exactly what the textarea's `aria-describedby` points at (wired through `nativeFieldProps`). So the live region is now a permanent child of the element the input is described by.

Practical effect: focusing the textarea computes its description from that box's whole subtree, which now includes "All changes saved" once a save lands — so the description can read as "<error message> All changes saved", and the saved text is re-read on every refocus rather than only when it changes. The browser verification covered the collaborative editor and the decision radios; this third surface has a different DOM shape and may not behave the same. Might want the region as a sibling of that box rather than a child.

src/components/study/outputs-decision-section.tsx
```diff
@@ -75,13 +75,20 @@ const DecisionRadioGroup: FC<DecisionRadioGroupProps> = ({ value, onChange, onBl
...
+    // Without it the group was flagged visually and via `aria-describedby` but never announced as
+    // invalid, unlike the feedback editor right above it (OTTER-675).
     const options = buildDecisionOptions(labName).map((option) => (
         <Radio
             key={option.value}
             value={option.value}
             label={option.title}
             description={<span id={descriptionId(option.value)}>{option.description}</span>}
             aria-describedby={descriptionId(option.value)}
+            aria-invalid={error ? true : undefined}
```
nathanstitt, line 91: Nice that this matches the existing house pattern exactly — `terms-checkbox.tsx:60` uses the identical `error ? true : undefined` shape, and the `|| undefined` variants elsewhere (`otp-input.tsx:44`, `code-env-form.tsx:117`) all deliberately avoid emitting `aria-invalid="false"`. Consistent with what's already here, and the test on the negative case pins that it stays absent rather than `"false"`, which is the part people usually get wrong.

src/components/save-status.test.tsx
```diff
@@ -34,4 +34,30 @@ describe('SaveStatusIndicator', () => {
...
+        expect(region).toHaveAttribute('aria-live', 'polite')
+        expect(region).toHaveAttribute('aria-atomic', 'true')
+        expect(region).toHaveTextContent('All changes saved')
+    })
+
+    it('keeps the live region mounted and empty while idle, so the save is announced (OTTER-675)', () => {
+        // A live region is only announced when content it already owns changes, so the region
+        // has to be in the DOM before the label lands in it.
+        renderWithProviders(<SaveStatusIndicator status="idle" />)
+        expect(screen.getByRole('status')).toBeEmptyDOMElement()
```
nathanstitt, line 50: These three `getByRole('status')` assertions are exact-match today because the component is rendered in isolation. Given the point about multiple status regions on the real proposal page, it might be worth one integration-level test at the `proposal/form.tsx` (or `resubmission-note-section`) level asserting the autosave region is findable among its neighbors — otherwise the isolation-only coverage won't catch the ambiguity if someone later queries by role on a full page.

src/components/save-status.test.tsx
```diff
@@ -34,4 +34,30 @@ describe('SaveStatusIndicator', () => {
...
+    })
+
+    it('leaves the transient saving label out of the live region (OTTER-675)', () => {
+        renderWithProviders(<SaveStatusIndicator status="saving" />)
+        expect(screen.getByTestId('autosave-status').closest('[aria-live]')).toBeNull()
+        expect(screen.getByRole('status')).toBeEmptyDOMElement()
```
nathanstitt, line 56: This is the assertion that actually pins the design decision from the PR description ("Saving…" stays out of the region to avoid two interruptions per cycle) — good that it's checked from both directions, `closest('[aria-live]')` being null *and* the region being empty. That's the pair that would catch a future refactor accidentally nesting the saving label back inside.

nathanstitt (approved):
looks good, thanks @therealmarv !

src/components/save-status.tsx
therealmarv: Fixed, though not the way you suggested, and the underlying problem was worse than "several silent regions".

Those three indicators on `proposal/form.tsx` are not independent. All three read the same `fieldsSaveStatus` from one Yjs provider, gated per field only on `editedKeys`. So once a researcher has touched title, datasets and PI, every save cycle filled all three regions in the same tick. Screen readers do not deduplicate across live regions, so that was **three** "All changes saved" announcements per pause in typing, which is exactly the noise the AC asks us to hold down. Before this PR nothing announced, so I introduced it.

`SaveStatusIndicator` now takes `announce` (default `true`). The three fields pass `announce={false}`, keeping their visible labels but dropping their own regions, and a single visually hidden `SaveStatusAnnouncer` announces once for the form. The four collaborative editors on that page own separate providers, so they keep their own regions and are correct as they are. See [save-status.tsx](https://github.com/safeinsights/management-app/blob/f6f39ffbf58b84b9d9b1b08570c6e788d89d4430/src/components/save-status.tsx#L36-L83) and [form.tsx](https://github.com/safeinsights/management-app/blob/f6f39ffbf58b84b9d9b1b08570c6e788d89d4430/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/proposal/form.tsx#L81-L96).

On `aria-label` specifically: I checked before adding it and it would have made things worse. Several screen readers speak a live region's accessible name **ahead of its content** ([MDN status_role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/status_role), [ARIA22](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22)), so `aria-label="Autosave status"` would have turned each save into "Autosave status, All changes saved". MDN's guidance is to omit the name unless it adds real value. The regions stay unnamed, and there is a test pinning that.

Your test-ambiguity point still stands, so the disambiguation is `data-testid` instead, which costs no speech: `autosave-live-region` on the indicator's region, `autosave-announcer` on the shared one.

Verified in Chromium: region attributes are `status` / `polite` / `true` with a null accessible name on every instance.

src/components/save-status.tsx
therealmarv: Right on both counts, and deferred rather than lost is the intended behavior. Made it explicit in two places rather than leaving it implicit in the collapse: a [comment on the line itself](https://github.com/safeinsights/management-app/blob/f6f39ffbf58b84b9d9b1b08570c6e788d89d4430/src/components/save-status.tsx#L90-L92) and a sentence in the PR description.

Dropping the idle/hidden distinction is deliberate. The only consumer is "does the region hold text", and collapsing to `'idle'` is what keeps the region mounted rather than unmounted, which is the whole fix. Worth noting the inverse of your scenario is the bug that motivated it: unmounting on error and remounting on clear hands the region back with its text already inside, and that is silent. Which turned out to be live in `collaborative-resubmission-note-section.tsx`, now fixed in this push.

src/components/save-status.tsx
therealmarv: Good catch on the call site, and I moved the region out. The described effect does not actually occur today, though, so I want to be precise about what was and was not wrong.

Mantine only folds the error id into `describedBy` when the error is truthy. In `@mantine/core`'s `Input/InputWrapper/InputWrapper.cjs`:

```js
const hasError = !!error && typeof error !== "boolean";
const hasDescription = !!description;
const _describedBy = `${hasError ? errorId : ""} ${hasDescription ? descriptionId : ""}`;
```

And in exactly that state the indicator is passed `isVisible={!error}`, which collapses it to `'idle'` and empties the region. So when the box is referenced the region holds nothing, and when the region holds "All changes saved" the box is not referenced. The description never reads as "&lt;error&gt; All changes saved", and nothing is re-read on refocus. I verified both halves in Chromium: `aria-describedby` is `null` in the saved and saving states and `resubmissionNote-error` only in the error state.

That said, a live region sitting in a field's description subtree is a trap for whoever next edits that `isVisible` gate, and it costs nothing to remove. The indicator is now a sibling of the error node inside a plain wrapper, so the `Group` still has two children and the row layout is unchanged: [resubmission-note-section.tsx](https://github.com/safeinsights/management-app/blob/f6f39ffbf58b84b9d9b1b08570c6e788d89d4430/src/components/study/resubmission-note-section.tsx#L68-L78), with a test pinning that no `[aria-live]` lives under `#resubmissionNote-error`.

Your instinct that this third surface behaves differently was right, just not for that reason. Its sibling, `collaborative-resubmission-note-section.tsx`, gated the whole indicator behind `singleUserEditing && !error` and so unmounted the region on a validation error. Clearing the error remounted it with "All changes saved" already inside, which is silent. That is the same defect this PR exists to fix, still live on that surface. Fixed in this push, and screenshot-verified across saved / saving / error.

src/components/save-status.test.tsx
therealmarv: Added, and it earned its keep immediately.

[`proposal/form.test.tsx`](https://github.com/safeinsights/management-app/blob/f6f39ffbf58b84b9d9b1b08570c6e788d89d4430/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/proposal/form.test.tsx) renders the real form inside `ProposalProvider` and asserts the three fields announce through exactly one region. Writing it is what surfaced that all three mirror one Yjs provider and were announcing three times per save, which the isolated tests could not have caught. See the thread above.

One wrinkle worth recording, since it bit me. My first version counted `autosave-live-region` page-wide and passed in isolation but failed in the full suite. The four collaborative text editors below the fields mount their own regions behind an `ssr: false` dynamic import, so under load they had appeared by the time the assertion ran and in a fast isolated run they had not. Their regions are correct (separate providers, separate save cycles), so the fix was to scope the count to the shared announcer's own testid rather than to add a wait. Full suite green twice at that assertion.

PR #934, opened 2026-08-04
Fix stale org claims blocking researcher decryption

nathanstitt:
Related: [OTTER-675](https://openstax.atlassian.net/browse/OTTER-675)

Researchers got "Private key is not valid for these results" on shared outputs despite having valid `study_job_file_recipient_key` rows. The key was fine — the server withheld the wrapped keys. Two defects:

**1. Stale org claims could never be cleared.** `updateClerkUserMetadata` wrote via `users.updateUserMetadata`, which PATCHes `/users/:id/metadata` and deep-merges. `orgs` is keyed by slug, so a revoked membership's key survived every rewrite. Now uses `users.updateUser`, which replaces `publicMetadata` outright. `session.ts` had a second inline copy of the same write on the sign-in path — it now calls the shared helper.

This is a general authorization fix: removing a user from an org previously did not revoke the JWT claim on any environment.

**2. Dual-membership users were misrouted.** `fetchEncryptedJobFilesAction` inferred researcher vs reviewer from `session.orgs`. That claim goes stale after an org removal, and a dual-role lab+enclave user is legitimately both — either way they took the reviewer path and got `recipientKeys: {}`, so decrypt failed. The caller now passes an explicit `type: 'researcher' | 'reviewer'` and the server serves only that role's keys. No org inference remains in the action.

Note: declaring `'reviewer'` is not separately entitlement-checked. The manifest is encrypted to the enclave's public keys, so a researcher asking for that path receives ciphertext they cannot decrypt.

**Also:** JSON results rendered as one line overflowing the modal. `logViewer` now precedes `codeViewer` so structured run logs (written as `.json`) reach the table renderer instead of being claimed by extension, and other JSON is pretty-printed and soft-wrapped.

## Testing

Full suite passes (2277 tests). New coverage:
- Dual-role user asking as researcher receives their wrapped keys — verified this fails on the previous logic with exactly the reported symptom.
- Same user asking as reviewer still gets the manifest path.
- A stale enclave claim no longer affects the researcher path (the QA scenario, without the manual metadata cleanup).
- `updateClerkUserMetadata` calls `updateUser`, not `updateUserMetadata`, and drops orgs the user has left.
- JSON viewer: log table, pretty-printing, malformed-JSON fallback, and that source code still scrolls rather than wraps.

## Follow-ups (not in this PR)

- `marshalSession` trusts a `format: v3` claim without re-reading memberships, so claims issued before this fix keep their stale orgs until something forces a refresh. This stops new ones being created; it does not expire existing ones. A backfill rewriting `publicMetadata` for all users would clear them.
- `qa-cleanup.ts` deletes `orgUser` rows with no Clerk metadata sync at all.

### DIFF
--- a/src/components/file-viewers/index.tsx
+++ b/src/components/file-viewers/index.tsx
@@ -7,7 +7,10 @@ import { textViewer } from './text-viewer'
 export { CodeViewer } from './code-viewer'
 export { ImageViewer } from './image-viewer'
 
-const viewers = [codeViewer, csvViewer, logViewer, textViewer]
+// Most specific first. logViewer inspects content rather than extension, so it must precede
+// codeViewer: run logs are written as .json, and matching on the extension alone would render a
+// structured log as a single minified line instead of the timestamp/message table.
+const viewers = [logViewer, codeViewer, csvViewer, textViewer]
 
 export const FileViewer: FC<{ path: string; text: string }> = ({ path, text }) => {
     for (const viewer of viewers) {
--- a/src/components/study/use-security-key-form.ts
+++ b/src/components/study/use-security-key-form.ts
@@ -20,10 +20,12 @@ export function useSecurityKeyForm({ job }: { job: LatestJobForStudy }) {
     const inputRef = useRef<HTMLTextAreaElement>(null)
 
     const { data: encryptedFiles, isLoading: isLoadingFiles } = useQuery({
-        queryKey: ['encrypted-files', job.id],
+        // Researcher-only flow: the study author entering their key to read shared outputs. The role
+        // is part of the key so this never collides with the reviewer view's cache.
+        queryKey: ['encrypted-files', job.id, 'researcher'],
         queryFn: async () => {
             try {
-                return await fetchEncryptedJobFilesAction({ jobId: job.id })
+                return await fetchEncryptedJobFilesAction({ jobId: job.id, type: 'researcher' })
             } catch (error) {
                 Sentry.captureException(error)
                 throw error
--- a/src/components/study/use-security-key-form.ts
+++ b/src/components/study/use-security-key-form.ts
@@ -15,24 +15,30 @@ const ERRORS = {
 type UseSecurityKeyFormOptions = {
     job: LatestJobForStudy
     /**
+     * Which key set to ask the server for. It cannot be inferred here: this form serves both the
+     * reviewer's outputs step (manifest recipient, no wrapped keys) and the researcher reading
+     * shared outputs (wrapped keys only), and a dual-role user is legitimately both.
+     */
+    type: 'researcher' | 'reviewer'
+    /**
      * Handed the decrypted plaintext on a successful key. The caller owns it from here: the
      * files carry raw AES keys (see JobFileInfo) and must stay in memory, never persisted.
      */
     onDecrypted: (files: JobFileInfo[]) => void
 }
 
-export function useSecurityKeyForm({ job, onDecrypted }: UseSecurityKeyFormOptions) {
+export function useSecurityKeyForm({ job, type, onDecrypted }: UseSecurityKeyFormOptions) {
     const [value, setValue] = useState('')
     const [error, setError] = useState<string>()
     const inputRef = useRef<HTMLTextAreaElement>(null)
 
     const { data: encryptedFiles, isLoading: isLoadingFiles } = useQuery({
-        // Researcher-only flow: the study author entering their key to read shared outputs. The role
-        // is part of the key so this never collides with the reviewer view's cache.
-        queryKey: ['encrypted-files', job.id, 'researcher'],
+        // The role is part of the key: the action returns a different key set per role, so a
+        // dual-role user switching views must not be served the other's cache.
+        queryKey: ['encrypted-files', job.id, type],
         queryFn: async () => {
             try {
-                return await fetchEncryptedJobFilesAction({ jobId: job.id, type: 'researcher' })
+                return await fetchEncryptedJobFilesAction({ jobId: job.id, type })
             } catch (error) {
                 Sentry.captureException(error)
                 throw error
--- a/src/hooks/use-encrypted-files-panel.ts
+++ b/src/hooks/use-encrypted-files-panel.ts
@@ -43,11 +43,15 @@ export function useEncryptedFilesPanel({ job, onFilesApproved, isReviewer }: Opt
         enabled: isJobApproved,
     })
 
+    // The role is part of the key: the action returns a different key set per role, so a dual-role
+    // user switching between the review and researcher views must not be served the other's cache.
+    const fetchAs = isReviewer ? 'reviewer' : 'researcher'
+
     const { isLoading: isLoadingBlob, data: encryptedFiles } = useQuery({
-        queryKey: ['encrypted-files', job.id],
+        queryKey: ['encrypted-files', job.id, fetchAs],
         queryFn: async () => {
             try {
-                return await fetchEncryptedJobFilesAction({ jobId: job.id })
+                return await fetchEncryptedJobFilesAction({ jobId: job.id, type: fetchAs })
             } catch (error) {
                 Sentry.captureException(error)
                 form.setFieldError('privateKey', 'Failed to fetch encrypted files, please try again later.')
--- a/src/lib/file-content-helpers.ts
+++ b/src/lib/file-content-helpers.ts
@@ -51,50 +51,15 @@ export function parseLogMessages(text: string): LogEntry[] | null {
 // Results are frequently written as minified JSON, which renders as one enormous line. Re-indent it
 // for display. Returns null when the text isn't valid JSON, so callers fall back to the raw text
 // rather than showing nothing.
-//
-// The round trip through JSON.parse is lossy for numbers: integers past Number.MAX_SAFE_INTEGER and
-// high-precision floats get rounded, and 1.50/1e5 re-serialize as 1.5/100000. This is the surface a
-// reviewer uses to judge whether output is safe to disclose, so a number that disagrees with the
-// file is worse than an ugly one. Re-serialize and compare against the minified input; on any
-// disagreement return null and let the caller show the bytes as they are.
 export function formatJson(text: string): string | null {
     const trimmed = text.trim()
     if (!trimmed.startsWith('{') && !trimmed.startsWith('[')) return null
 
     try {
-        const parsed: unknown = JSON.parse(trimmed)
-        if (JSON.stringify(parsed) !== minifyJson(trimmed)) return null
-        return JSON.stringify(parsed, null, 2)
+        return JSON.stringify(JSON.parse(trimmed), null, 2)
     } catch {
         return null
     }
-}
-
-// Strip insignificant whitespace so the fidelity check compares values rather than formatting.
-// String contents are preserved verbatim, since whitespace inside them is significant.
-function minifyJson(json: string): string {
-    let out = ''
-    let inString = false
-    let escaped = false
-
-    for (const char of json) {
-        if (inString) {
-            out += char
-            if (escaped) escaped = false
-            else if (char === '\\') escaped = true
-            else if (char === '"') inString = false
-            continue
-        }
-        if (char === '"') {
-            inString = true
-            out += char
-            continue
-        }
-        if (char === ' ' || char === '\t' || char === '\n' || char === '\r') continue
-        out += char
-    }
-
-    return out
 }
 
 export function parseCsv(text: string): { headers: string[]; rows: string[][] } {
--- a/src/lib/file-content-helpers.ts
+++ b/src/lib/file-content-helpers.ts
@@ -43,6 +43,20 @@ export function parseLogMessages(text: string): LogEntry[] | null {
     }
 }
 
+// Results are frequently written as minified JSON, which renders as one enormous line. Re-indent it
+// for display. Returns null when the text isn't valid JSON, so callers fall back to the raw text
+// rather than showing nothing.
+export function formatJson(text: string): string | null {
+    const trimmed = text.trim()
+    if (!trimmed.startsWith('{') && !trimmed.startsWith('[')) return null
+
+    try {
+        return JSON.stringify(JSON.parse(trimmed), null, 2)
+    } catch {
+        return null
+    }
+}
+
 export function parseCsv(text: string): { headers: string[]; rows: string[][] } {
     const result = Papa.parse<string[]>(text, {
         header: false,
--- a/src/lib/file-content-helpers.ts
+++ b/src/lib/file-content-helpers.ts
@@ -35,6 +35,11 @@ export function parseLogMessages(text: string): LogEntry[] | null {
         for (const item of parsed) {
             if (typeof item !== 'object' || item === null) return null
             if (typeof item.timestamp !== 'number' || typeof item.message !== 'string') return null
+            // Exactly these two keys. Since this viewer is chosen by sniffing content rather than by
+            // file identity, a result file that happens to be log-shaped would otherwise render as a
+            // timestamp/message table with its other fields silently dropped, in a view whose whole
+            // purpose is deciding what is safe to disclose.
+            if (Object.keys(item).length !== 2) return null
             entries.push({ timestamp: item.timestamp, message: item.message })
         }
         return entries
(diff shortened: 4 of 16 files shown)


src/components/study/use-security-key-form.ts
```diff
@@ -20,10 +20,12 @@ export function useSecurityKeyForm({ job }: { job: LatestJobForStudy }) {
...
 
     const { data: encryptedFiles, isLoading: isLoadingFiles } = useQuery({
-        queryKey: ['encrypted-files', job.id],
+        // Researcher-only flow: the study author entering their key to read shared outputs. The role
+        // is part of the key so this never collides with the reviewer view's cache.
+        queryKey: ['encrypted-files', job.id, 'researcher'],
         queryFn: async () => {
             try {
-                return await fetchEncryptedJobFilesAction({ jobId: job.id })
+                return await fetchEncryptedJobFilesAction({ jobId: job.id, type: 'researcher' })
```
therealmarv: **Severity: blocker (P1).** This one breaks reviewer error-log decryption, and I think it's the e2e failure.

The comment says "researcher-only flow", but this hook has exactly one production caller, and it's the reviewer:

- [`reviewer-outputs-errored-screen.tsx:52`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/_screens/reviewer-outputs-errored-screen.tsx#L52) mounts `<SecurityKeyForm job={job} />`
- [`reviewer-screen-rules.ts:11`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/lib/study-screen/reviewer-screen-rules.ts#L11) selects that screen on `awaitingFilesDecisionOnError(s)`, so it's reviewer-side and strictly pre-decision

There is no researcher errored screen yet. [`researcher-screen-rules.ts:11`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/lib/study-screen/researcher-screen-rules.ts#L11) explicitly excludes that state (`s.hasResults && !awaitingFilesDecisionOnError(s)`), so the researcher holds on `code-decision-screen` until a `FILES-*` decision exists. When the researcher view does land in OTTER-675, it goes through `EncryptedFilesPanel`, so it will pick up your `fetchAs` derivation rather than this hook.

So with `type: 'researcher'` hardcoded, a reviewer here falls into the wrapped-keys branch and bails at [`study-job.actions.ts:280`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/server/actions/study-job.actions.ts#L280) (`if (!wrappedKeys.length) return []`), which surfaces as "No encrypted outputs available to decrypt."

Worth stressing that this isn't only a pre-approval problem: `studyJobFileRecipientKey` rows are written solely by `insertSharedFileKeys` and only ever wrap to lab recipients, so a reviewer's fingerprint is never in that table at any point in the lifecycle. The researcher path can never succeed from this screen.

Before this change the action inferred `isEnclaveReviewer` from `session.orgs`, so the reviewer got the manifest path and this worked. You updated `use-encrypted-files-panel.ts` correctly in the same commit, so this reads like a missed second call site rather than a design disagreement.

Suggested fix: take the role as a prop, `useSecurityKeyForm({ job, type })`, and pass `'reviewer'` from the errored screen. Keeping it in the query key as you've already done stays correct.

On the test gap: [`security-key-form.test.tsx:20-22`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/components/study/security-key-form.test.tsx#L20-L22) `vi.mock`s the whole actions module, so no unit test ever reaches this branch, which is why unit went green and only e2e caught it. The fixture comment in that same file at [line 30](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/components/study/security-key-form.test.tsx#L29-L31) even says "The reviewer is a manifest recipient (empty recipientKeys)", which is the intent spelled out. A test that lets the real action run would pin this down.

src/lib/file-content-helpers.ts
```diff
@@ -43,6 +43,20 @@ export function parseLogMessages(text: string): LogEntry[] | null {
...
 
+// Results are frequently written as minified JSON, which renders as one enormous line. Re-indent it
+// for display. Returns null when the text isn't valid JSON, so callers fall back to the raw text
+// rather than showing nothing.
+export function formatJson(text: string): string | null {
+    const trimmed = text.trim()
+    if (!trimmed.startsWith('{') && !trimmed.startsWith('[')) return null
+
+    try {
+        return JSON.stringify(JSON.parse(trimmed), null, 2)
```
therealmarv: **Severity: medium (P2).** Round-tripping through `JSON.parse` means the rendered numbers can differ from what's actually in the file:

```
in : {"n":9007199254740993,"m":0.1234567890123456789,"big":12345678901234567890}
out: {"n": 9007199254740992, "m": 0.12345678901234568, "big": 12345678901234567000}
```

It's broader than just integers past `Number.MAX_SAFE_INTEGER`: high-precision floats get rounded, `1.50` renders as `1.5`, `1e5` as `100000`. Display-only, and nothing is corrupted at rest, so I'm not calling this a blocker. But this is the surface a reviewer uses to decide whether output is safe to disclose, and a number that doesn't match the file is a bad thing to show them there.

Cheapest safe version is to keep the pretty-print but verify fidelity first, something like bailing out when `JSON.stringify(parsed)` isn't numerically equivalent to the minified input, and falling back to raw text. Re-indenting without reparsing would also work if you'd rather not pay for the round trip check.

src/components/file-viewers/index.tsx
```diff
@@ -7,7 +7,10 @@ import { textViewer } from './text-viewer'
 export { CodeViewer } from './code-viewer'
 export { ImageViewer } from './image-viewer'
 
-const viewers = [codeViewer, csvViewer, logViewer, textViewer]
+// Most specific first. logViewer inspects content rather than extension, so it must precede
+// codeViewer: run logs are written as .json, and matching on the extension alone would render a
+// structured log as a single minified line instead of the timestamp/message table.
+const viewers = [logViewer, codeViewer, csvViewer, textViewer]
```
therealmarv, line 13: **Severity: low (P3).** The reasoning in the comment is right and the reorder does fix the real problem. My worry is the blast radius: `logViewer` now sniffs content for *every* file, and [`parseLogMessages`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/lib/file-content-helpers.ts#L34-L39) keeps only `timestamp` and `message` and drops everything else on the object.

So a results file shaped like this:

```json
[{"timestamp":1,"message":"ok","estimate":42}]
```

renders as a two-column log table with `estimate` silently gone. Not a crash, just quietly missing data in a disclosure-review view, which is the flavor I'd rather not leave to chance.

Two things keep this low rather than medium: `parseLogMessages` returns `null` if *any* element fails the check, so it takes a uniformly log-shaped array to trip, and that's an unlikely result-file shape.

Since `isEncryptedLogType` already distinguishes these artifacts server side, dispatching on file identity instead of sniffing content would sidestep it entirely. Failing that, requiring an exact two-key schema in `parseLogMessages` would be enough.

src/server/clerk.ts
```diff
@@ -69,7 +69,10 @@ export const updateClerkUserMetadata = async (userId: string) => {
 
     logger.info('Updating user metadata for clerkId:', clerkId, 'with metadata:', metadata)
 
-    await client.users.updateUserMetadata(clerkId, {
+    // updateUser replaces publicMetadata wholesale; updateUserMetadata deep-merges it. orgs is a
+    // slug-keyed map, so merging left a removed org's key in place forever — the JWT claim kept
+    // granting access long after the membership was revoked.
+    await client.users.updateUser(clerkId, {
```
therealmarv, line 75: This is a genuinely good catch, and it's the part of the PR I'd most want merged. `updateUserMetadata` deep-merging a slug-keyed `orgs` map means a revoked membership could never be expressed as an absence, only as a key that keeps existing. That's the kind of bug that hides for a long time because every individual write looks correct.

Calling out in the description that this is a general authorization fix and not just a decryption fix was the right framing, as was being upfront in the follow-ups that `marshalSession` trusting `format: v3` leaves already-issued claims stale until something forces a refresh. That honesty is more useful to the next reader than a clean-looking writeup would have been.

The three-line comment explaining merge versus replace is exactly the "why" that belongs here.

src/server/session.ts
```diff
@@ -37,30 +36,7 @@ async function syncAndUpdateUserMetadata(clerkUserId: string): Promise<UserInfo
...
-    }
-
-    logger.info('Updating user metadata for clerkId:', clerkUserId, 'with metadata:', metadata)
-
-    await client.users.updateUserMetadata(clerkUserId, {
-        publicMetadata: metadata as unknown as UserPublicMetadata,
-    })
-
-    return metadata
+    return await updateClerkUserMetadata(userId)
```
therealmarv, line 39: Collapsing the duplicated inline metadata write down to the shared helper is a nice win: 25 lines out, and more importantly the sign-in path can no longer drift away from `updateClerkUserMetadata`. Having two copies of that write is precisely how the `updateUserMetadata` bug would have survived a partial fix.

src/hooks/use-encrypted-files-panel.ts
```diff
@@ -43,11 +43,15 @@ export function useEncryptedFilesPanel({ job, onFilesApproved, isReviewer }: Opt
         enabled: isJobApproved,
     })
 
+    // The role is part of the key: the action returns a different key set per role, so a dual-role
+    // user switching between the review and researcher views must not be served the other's cache.
+    const fetchAs = isReviewer ? 'reviewer' : 'researcher'
```
therealmarv, line 48: This is the correct shape, and folding the role into the query key is a good catch on its own. Two roles returning different key sets under one cache key would have been a nasty intermittent bug for dual-membership users, and it's the sort of thing that only reproduces for the handful of people who have both memberships.

Reference point for the blocker I left on `use-security-key-form.ts`: this call site derives the role, that one hardcodes it, and the difference is what's failing e2e.

src/server/actions/study-job.actions.ts
```diff
@@ -220,6 +220,12 @@ export const fetchEncryptedJobFilesAction = new Action('fetchEncryptedJobFilesAc
     .params(
         z.object({
             jobId: z.string(),
+            // The caller states which role it is acting as. This used to be inferred from
+            // session.orgs, which cannot answer the question: the claim survives removal from an
+            // org, and a dual-role lab+enclave user is legitimately both. Either way they were
+            // treated as a reviewer and handed recipientKeys:{}, and since their fingerprint is
+            // absent from the zip's manifest, decrypt failed as "private key is not valid".
+            type: z.enum(['researcher', 'reviewer']),
```
therealmarv, line 407: Agreed with the direction here. `session.orgs` genuinely cannot answer "which role is this person acting as", and the dual-membership case makes that structural rather than a bug you can patch. Making the caller declare its intent is the right call, and the comment explaining why the inference was unfixable is worth keeping.

**Severity: low (P3), observation rather than a change request.** You called out in the description that `'reviewer'` isn't separately entitlement-checked because the manifest is encrypted to the enclave's keys, and that reasoning holds for confidentiality. One consequence I'd want recorded somewhere more durable than the PR body: any caller who passes the `view StudyJob` gate can now request the reviewer path and receive the raw ciphertext of every encrypted artifact on the job, where the old `session.orgs` check would have refused. Still safe, since they can't open it, but it's a widening of what reaches the client and it now rests entirely on the encryption rather than on two independent gates.

A sentence in the handler comment saying so explicitly would keep the next person from having to re-derive it. Not blocking.

therealmarv (changes requested):
Hi Nathan, I went through this carefully and traced the decryption paths end to end. The Clerk metadata fix is the real prize here and I want it merged. I'm requesting changes only because one call site was missed in the role refactor, and it breaks reviewer error-log decryption.

## What I liked

The `updateUserMetadata` to `updateUser` fix is an excellent find. A deep-merge over a slug-keyed `orgs` map means a revoked membership can only ever be expressed as a stale key, never as an absence, so no amount of rewriting could clear it. Framing that as a general authorization fix rather than just a decryption fix was right, and I appreciated that the follow-ups section says out loud that `marshalSession` trusting `format: v3` leaves existing claims stale. That is more useful than a tidier-sounding writeup would have been.

Deleting the second inline copy of the metadata write in `session.ts` is the right instinct too. Two copies of that write is exactly how this bug would have survived a partial fix.

Killing the `session.orgs` role inference is also correct in principle. That claim cannot answer "which role is this person acting as", and a dual-membership lab plus enclave user is legitimately both, so no amount of care in the inference would have fixed it.

## Findings, by severity

**1. Blocker (P1): reviewer error-log decryption is broken.** `use-security-key-form.ts` hardcodes `type: 'researcher'`, but its only production caller is [`reviewer-outputs-errored-screen.tsx:52`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/_screens/reviewer-outputs-errored-screen.tsx#L52), selected reviewer-side by [`reviewer-screen-rules.ts:11`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/lib/study-screen/reviewer-screen-rules.ts#L11). A reviewer never has `studyJobFileRecipientKey` rows, since `insertSharedFileKeys` only wraps to lab recipients, so the researcher branch returns `[]` at [`study-job.actions.ts:280`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/server/actions/study-job.actions.ts#L280) and the form shows "No encrypted outputs available to decrypt." Details and a suggested fix inline.

**2. Medium (P2): JSON pretty-printing can change the values shown.** `JSON.parse` then `JSON.stringify` in `formatJson` rounds large integers and high-precision floats, and normalizes things like `1.50` to `1.5`. Display-only, but this is the surface a reviewer uses to judge whether output is safe to disclose. Inline.

**3. Low (P3): valid result JSON can be misclassified as a log.** With `logViewer` first, it now sniffs content for every file, and `parseLogMessages` keeps only `timestamp` and `message`. A uniformly log-shaped array with extra fields renders as a table with those fields silently dropped. Narrow, but silent omission in a disclosure view. Inline.

**4. Low (P3), observation only: the reviewer path is no longer entitlement-gated.** Your reasoning in the description is sound, but the consequence is that any `view StudyJob` caller can now pull raw ciphertext for every artifact on the job. Worth a line in the handler comment. Not blocking.

## On the e2e failure

CI ran synthetic merge `2931dc7`, which picks up main's post-OTTER-675 spec:

```
✘ 21 [chromium] › tests/study-flow.spec.ts:516:1 › Error log review (37.0s)
   Locator: getByTestId('outputs-files-section')
   Error: element(s) not found
   > 330 | await expect(page.getByTestId('outputs-files-section')).toBeVisible()
```

This isn't merge staleness. The branch's own copy of `reviewerDecryptsErrorLogs` asserts `'Security key accepted.'`, which fails for the same reason. Everything else was green: lint, all three unit shards, CodeQL, and infosec.

The reason unit tests didn't catch it is that [`security-key-form.test.tsx:20-22`](https://github.com/safeinsights/management-app/blob/34cd03b20b4eadfa24fe7c884fd408afeb6d6efa/src/components/study/security-key-form.test.tsx#L20-L22) mocks the whole actions module, so the researcher versus reviewer branch never executes through that component. Given the role is now caller-supplied, a test that lets the real action run for both roles seems worth adding alongside the fix.

Fix the blocker and I'm happy to approve. The other three are worth addressing but I won't hold the PR on them if you'd rather split them out, especially since the Clerk fix has value independent of the viewer changes.

nathanstitt:
Thanks — the P1 diagnosis was exactly right, including the mechanism. All four addressed.

**1. Blocker — reviewer error-log decryption.** Fixed in `9535b2b7`, which was pushed before I read this; we converged on the same fix independently. `type` is now a prop on `useSecurityKeyForm` and `SecurityKeyForm`, and `OutputsReviewPanel` passes `'reviewer'`. Your call on `insertSharedFileKeys` was the key point: a reviewer's fingerprint is never in `studyJobFileRecipientKey` at any point in the lifecycle, so the researcher path could never have succeeded from that screen.

I confirmed causality rather than inferring it — setting `type="researcher"` reproduces the CI failure verbatim (same locator, same line 330), `'reviewer'` passes. e2e and all unit shards are green on that commit.

One correction to the writeup: this is a semantic merge conflict, not purely a missed call site. When my commit touched that hook it genuinely was researcher-only, and its `SecurityKeyForm` had no reviewer caller. #924 then made it the decrypt gate for `OutputsReviewPanel`. Both branches merged cleanly with no textual conflict, which is why it only appeared in CI's merge commit and never on either branch alone. The end state you described is right; the sequence just means it wasn't visible at the time the hardcode was written.

**2. Medium — JSON number fidelity.** Fixed in `ed6dab7e`. Reproduced your exact case. Rather than dropping the pretty-print, `formatJson` now re-serializes and compares against the minified input, returning `null` on any disagreement so the caller shows raw text. The minifier skips string contents, since whitespace is significant there. Covered for all five drift classes you listed plus the string-whitespace edge.

**3. Low — log misclassification.** Fixed in `ed6dab7e` via the exact-two-key check. I took that over dispatching on file identity because `FileViewer` receives only `path` and `text`, so routing on `isEncryptedLogType` means threading `fileType` through the viewer layer — worth doing, but wider than this PR.

**4. Low — reviewer path no longer double-gated.** Recorded in the handler comment, stating plainly that any caller past `view StudyJob` can now pull raw ciphertext for every artifact, and that confidentiality rests on the encryption alone rather than two independent gates.

**On the test gap:** agreed, and that was the real failure here — the module-level `vi.mock` meant no unit test could reach the branch. Added `it.each` over both roles asserting the role actually reaches the action, which fails against the old hardcode. Unit suite is 2408 passing.

**One thing I want to flag rather than leave buried:** running the full e2e suite locally, I get 1-3 failures per run, scattered across unrelated tests and locators — `Proposal rejection`, `Code change request`, `org-admin` starter code, and bare assertions like `getByText('Review Studies')` and `STEP 1`. I verified this reproduces on the branch with my viewer changes stashed, so it is not from this PR, and CI is green. It looks like local-environment contention rather than a code defect, but given the zero-tolerance policy on e2e flakiness I would rather surface it than assume it is only my machine. Happy to open a separate ticket if you have not seen it.

src/lib/file-content-helpers.ts
nathanstitt: while true, I don't think this will be an issue for us because:
*  the source is JSON already so it would have numbers expressed correctly to start with.  
* And even if does truncate large numbers, I think it's mostly harmless because anyone who _really_ is operating on the log information will download them which will be the original source

therealmarv (approved):
thanks!

PR #935, opened 2026-08-04
OTTER-647 follow-up: fix the QA defects and retract the out-of-spec footer hint

therealmarv:
see [OTTER-647](https://openstax.atlassian.net/browse/OTTER-647)

The card went back to Development after Iris&#39; summary of her own notes plus Greg&#39;s two validation passes. The acceptance criterion itself held (Greg: VALIDATED, twice), so this PR does not touch it. It fixes the seven reported defects and removes the footer line UX rejected.

## Where the defects sit

The card had one acceptance criterion and [#913](https://github.com/safeinsights/management-app/pull/913) delivered it. Most of the reported defects are in the surrounding changes that PR also made: the `FormField` component, the ARIA work across sign-in / MFA / OTP / password-reset, the schema trimming, and the outstanding-fields footer hint. Six of the seven fall outside the criterion itself:

| Defect | In the AC? | From #913? |
|---|---|---|
| Footer hint out of spec (Iris) | no | yes |
| Case 51, footer wraps to three lines | no | yes, caused by the hint |
| Case 25, toolbar click raises the error | yes, as a false positive | yes |
| Case 4, radio circles stay grey | partly | pre-existing, newly visible |
| Case 49, label weight mismatch | no | yes, partial conversion |
| Case 38, raw Zod message | no | yes, admin surface pulled in |
| Case 40, sign-in email message | no | pre-existing + #913 ARIA work |
| `aria-required` on an optional field | no | yes |

The approach here is correspondingly narrow: remove the hint UX rejected, fix each defect with the smallest change that holds, and record what is deliberately left out.

## What changed

**1. The outstanding-fields footer hint is removed (also fixes case 51).** Iris: *&#34;The line of text displayed in footer isn&#39;t part of the requirements and UX has confirmed that should not be included.&#34;* `IncompleteFieldsHint` and `missing-fields.ts` are deleted along with the `missingFields` plumbing in all four footers and the e2e assertion. Removing the `Stack` that wrapped each submit button is what fixes case 51: that stack measured 900px inside a 924px `Group`, which is why `Previous` / `View as reviewer` / `Submit` landed on three separate lines. `isLinkedPiUserId` stays, it is still used by the proposal schema. `submitDisabledReason` in the code-upload footer is untouched, it predates this card.

**2. Case 25, the review-feedback toolbar tripped its own required error.** `focusout` fires on `#review-feedback` with `relatedTarget: null` while `document.hasFocus()` is still true, so the blur guard took its &#34;clicked neutral space&#34; branch and validated. The fix is to stop reading that off the focus event at all and read it off the press instead; see the `useWidgetBlur` section below for the shape it landed in. The toolbar additionally restores focus to the editor after each command, which is the root cause and worth fixing on its own. Confirmed against Lexical&#39;s own issue tracker (facebook/lexical#5689); further research on this class of blur bug lands on the same recommendation, which is to flag the internal interaction rather than try to read it back off the focus event.

**3. Case 4, radio circles stayed grey.** `Radio.Group`&#39;s context carries `value` / `onChange` / `size` / `name` / `disabled` to its children but not `error`, while `Radio` drives its own red circle from its own `error` prop. Per the Mantine docs (via context7) a boolean `error` applies error styling without adding a second message, so each child now gets the group&#39;s error as a flag. Applied to the programming-language group and both decision groups.

**4. Case 38, starter code leaked a Zod message.** The form seeded `starterCodes` as `undefined`, so the create schema&#39;s array type check failed before `.min(1, &#39;At least one starter code file is required&#39;)` could run and Save showed `Invalid input: expected array, received undefined`. Seeded as `[]` instead. The edit payload still omits `starterCodeFileNames` when nothing new was picked, because an empty array there would read as &#34;the admin cleared the list&#34; rather than &#34;left the existing files alone&#34;. Greg also reported no message on dropzone blur; that path turned out to be wired correctly already (see the verification note below), so no code change was needed there.

**5. Case 49, label weight mismatch.** `Principal Investigator` had moved to `FormField` (label 14px / 600) while `Researcher` stayed on `FormFieldLabel`, which renders a `Title order={5}` (16px) and passes `fw=&#34;semibold&#34;`, not a valid CSS `font-weight` keyword, so the weight was dropped too. Both fields in the panel now use `FormField`, on the proposal page and on edit-and-resubmit.

The first attempt here fixed `FormFieldLabel`&#39;s dead `fw` value instead. Visual inspection showed that closed the weight gap but left a *size* gap (16px vs 14px), so the two labels still did not match, and it changed the label weight at six other `FormFieldLabel` call sites outside this card&#39;s scope. That commit is reverted in favor of converting the one field, which is what the defect describes.

**6. Case 40, sign-in messages.** A blank email fails both `min(1)` and `email()`, and `mantine-form-zod-resolver` keeps the last issue unless told otherwise, so an untouched field read &#34;Invalid email&#34;. With `errorPriority: &#39;first&#39;` each case keeps its own message: blank, malformed, too long. Separately, Mantine renders `PasswordInput`&#39;s inner `<input>` under an `Input` with `withAria: false`, so its `error` lands `data-invalid` and the describedby but never `aria-invalid`; that is now passed explicitly at every `PasswordInput` with an unmarked error.

**7. `aria-required` on an optional field.** `ariaRequired` was passed bare while the asterisk correctly followed `field.required`, so &#34;Additional notes or requests&#34; announced as required. Now `ariaRequired={field.required}`, which covers the edit-and-resubmit surface too since it renders the same component. The bare `ariaRequired` on the review feedback and resubmission note editors is correct; those fields really are required.

## Review round: `useWidgetBlur`

The blur guard was first written as a `widgetBlurHandler` wrapper over module-level `pointerdown` / `keydown` listeners on `document`. Per review it is now a `useWidgetBlur` hook built on Mantine&#39;s `useClickOutside`, which deletes the global listeners, the `__resetLastPointerDownForTests` escape hatch, the `document.hasFocus()` probe, and gains real cleanup on unmount. Ten call sites converted.

Two pieces sit alongside `useClickOutside` rather than being replaced by it, because dropping them would have regressed the criterion this card is about:

- **Keyboard exit.** `useClickOutside` listens on `mousedown` / `touchstart` only, so a pure swap would lose Tab, one of the two routes Greg validated. The hook keeps a small `onBlur` that acts only on a non-null `relatedTarget` outside the widget. A null `relatedTarget` is left entirely to the pointer signal, which is what retires the `document.hasFocus()` probe: a tab or window switch produces no press, so it cannot validate.
- **An `onFocus` gate.** `useClickOutside` fires whether or not the widget was ever entered, so without this the first click anywhere on the page would flag every required field at once, including ones the user has not reached.

Two call sites needed more than a mechanical swap. `PillsInput` forwards its ref to the inner field rather than the root, so scoping the guard there would put the pills and their remove buttons *outside* the widget and make removing a pill commit the pending draft; it gets a wrapper element instead. `invite-form-view` chains Mantine&#39;s own `onFocus` from `getInputProps` rather than replacing it, so the field still marks touched.

Also from review: the toolbar&#39;s unlink branch now goes through `runOnEditor` like every other command (the decision is read out of the editor state, the dispatch happens after the read closes), the dead optional chain in the code-env edit payload is gone, `renderWithProviders` takes a `singleUserEditing` option instead of tests hand-rolling a partial provider stack, and the sign-in `errorPriority` comment now records that the scoping is deliberate rather than an oversight.

### Second round: the two signals&#39; edges

A follow-up review named two behaviors the rewrite changed that the doc comment did not account for. Both are documentation rather than code, because in each case the alternative reintroduces the ambiguity the rewrite removed.

- **Escape does not validate.** `EscapeFocusPlugin` blurs the editor root without moving focus anywhere, so it reaches the hook as a null `relatedTarget` with no press behind it and both signals decline it. The pre-rewrite handler cleared the recorded press on keydown and so did validate here; that was lost silently along with the test pinning it. Acting on it would mean either validating every null `relatedTarget` (the original defect) or probing `document.activeElement` a tick later. A bare Escape handler is worse than both: only the two editors blur on Escape, so at the other eight call sites it would raise a required error with the caret still in the field. The error still surfaces on the next outside press or on submit.
- **Any outside press counts, including presses that never move focus.** Because the press alone decides, a touch-scroll of the page or a drag-select starting on neutral space validates a visited widget, and in `research-interests-input` the same press commits the pending pill. Corroborating the press with a focus check is what would put the null `relatedTarget` interpretation back.

The `useWidgetBlur` doc comment now names both, so neither reads as an oversight.

## Deliberately not in this PR

Recorded here so the card can close instead of collecting another round.

- **Cases 45, 47, 48, 54** (OTP entry, rejected MFA code, Clerk-rejected phone, spent recovery code). No code defect was identified. These are blocked on a dedicated Clerk test instance: running them against shared QA accounts risks rate-limiting accounts CI depends on. A QA-environment request rather than engineering work on this card.
- **Case 18** (stale collaborative error) and **case 52 steps 4-7** (tab or application switch): need a second signed-in member or focus outside the browser.
- **Case 19** (code resubmission path): not reached, approved instead of requesting changes.
- **Case 20** (files present, no main file): the precondition looks unreachable because uploading auto-designates a main file and the star cannot be un-toggled. Reads as a test-plan correction rather than a defect.
- **Case 57** (legacy draft with an unlinked PI): the plan itself says to skip absent one.
- **The radio styling in `outputs-decision-section.tsx`**: same one-line pattern, but that file sits on the active OTTER-667/675/676 surface and QA did not flag it for this card.
- **The dead `fw=&#34;semibold&#34;` values** in `form-field-label.tsx`, `panel.tsx` and `invite-form-view.tsx`: a real latent bug of one class, found while fixing case 49, but none of them is a label sitting beside a converted field, and `panel.tsx` would restyle every panel in the app. Worth its own card.
- **Making `errorPriority: &#39;first&#39;` the default** for every `zodResolver` call site: any schema stacking `min(1)` before a format check has the same latent problem, but flipping it in the `@/common` re-export would rewrite the messages of ~15 other forms, none of which this card covers, and each wants its messages re-read. Worth its own card.
- **Making Escape validate**, and **suppressing validation for outside presses that do not move focus**: both are described in the second review round above. Each needs an asynchronous focus probe to tell the wanted case from the unwanted one, which is the timing guesswork the `useClickOutside` rewrite exists to avoid. Documented in the hook instead.
- **The rest of #913&#39;s work outside the criterion** (the sign-in / MFA / OTP / password ARIA, `FormField`, schema trimming) stays. It is merged and validated, and reverting it would create more churn than it removes. Only the footer line UX rejected is removed.

## Validation

- `pnpm run checks` clean.
- Full unit suite green: 249 files, 2391 tests.
- New coverage: the guard&#39;s six cases including Tab out of a widget and the never-entered gate; radio `data-error` appearing and clearing on both flagged groups; the plain-language starter code message, the dropzone blur path, and an edit-path save that leaves the existing starter code alone; the blank vs malformed email messages and the password `aria-invalid`; required vs optional `aria-required` on the proposal editors. Deleted `incomplete-fields-hint.test.tsx` and `missing-fields.test.ts` with the code they covered.

One note for anyone extending the programming-language tests: `TestingProviders` shares a single query client across a file with `staleTime: 60s`, so a second test reusing `test-org` replays the first test&#39;s cached language list. The new case uses its own slug.

### Visual verification

Each defect was checked in a browser against throwaway local preview pages rendering the real components (not committed), at a 1440px viewport. The guard was re-checked the same way after the `useWidgetBlur` rewrite, driving the real `useReviewFeedback` / `useReviewDecision` hooks.

| Reported | Measured now |
|---|---|
| Footer hint, out of spec | Absent from all four footers; no &#34;required before submitting&#34; text anywhere |
| Case 51, buttons at tops 644 / 702 / 760 | `Previous` / `View as reviewer` / `Submit initial request` all at top 389; no horizontal page scroll |
| Case 25, Bold and Bullet list raise &#34;Feedback is required.&#34; | Bold, Bullet list and Italic each leave no error and keep `document.activeElement` on `#review-feedback`; clicking neutral space still raises it with `aria-invalid=&#34;true&#34;` and `aria-describedby=&#34;review-feedback-error&#34;`. The Link flow still hands focus to the URL input, cancelling returns it to the editor, and unlinking now keeps it there too |
| Case 4, circles stuck at `rgb(206,212,218)` | Clean state `rgb(206,212,218)`, error state `rgb(255,41,41)` with `data-error=&#34;true&#34;` on all three radios, cleared on every circle once a decision is picked |
| Case 49, 600 vs 500 at 14px | Both labels 14px / 600 |
| Case 38, `Invalid input: expected array, received undefined` | &#34;At least one starter code file is required&#34; inline and in the save summary; no Zod text on the page |
| Case 38, no message on dropzone blur | Focusing the dropzone and leaving it does raise the message and sets `aria-invalid=&#34;true&#34;`. This path was already correct; the original observation was most likely a click that opened the file dialog rather than blurring in-page |
| Case 40, blank email reads &#34;Invalid email&#34; | Blank reads &#34;Email is required&#34;, malformed still reads &#34;Invalid email&#34;, a valid address clears both; password carries `aria-invalid=&#34;true&#34;` plus its describedby |

Two behaviors worth recording from that pass, neither changed by this PR. Moving between the radios of a group still validates nothing, and tabbing out of the group does validate. Inside the Lexical editor, both Tab and Shift+Tab are consumed by the editor for indent/outdent, so keyboard focus cannot leave it and no error is raised; that is Lexical&#39;s own rich-text handling, and with no `focusout` there was nothing for the previous handler to act on either. Leaving that editor by click validates as expected.

`aria-required` is not visible, so it is covered by a unit test over the real component in single-user mode.

[OTTER-647]: https://openstax.atlassian.net/browse/OTTER-647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/form-field.tsx
```diff
@@ -100,33 +100,73 @@ export function revalidateOnBlur(form: ValidatableForm, path: string) {
...
+ *
+ * {@link widgetBlurHandler} needs it because `relatedTarget` is null for two opposite cases and
+ * focus alone cannot tell them apart. The press target can: see the pointer branch there.
+ * Registered in the capture phase so a handler that stops propagation can't hide the press, and
+ * cleared on keydown so a keyboard-driven blur (Tab out of a widget, Escape) is never mistaken
+ * for a leftover click.
+ */
+let lastPointerDownTarget: Node | null = null
+
+if (typeof document !== 'undefined') {
```
nathanstitt: Could we replace this with `useClickOutside` from `@mantine/hooks`? It's already a dependency, and it's built for exactly this.

The reason it fits: it listens on `mousedown`/`touchstart` rather than focus, so the null-`relatedTarget` ambiguity this whole function is working around never arises — there is no focus event to disambiguate. It scopes to a ref, registers on mount, and cleans up on unmount:

```ts
const ref = useClickOutside(() => form.validateField('language'))

<Radio.Group ref={ref} /* no onBlur wrapper */>
```

Clicking the Lexical toolbar is inside the ref, so nothing fires. Clicking neutral page space is outside, so it validates. The hook's own `ref.current.contains(target)` check is doing what `lastPointerDownTarget` is hand-rolling here.

That would let us delete `lastPointerDownTarget`, both `addEventListener` calls, the `__reset` export, the test `beforeEach`, and `widgetBlurHandler` itself — and we'd get real cleanup on unmount, which the current module-level listeners don't have.

Mechanically the call sites look ready for it: `Radio.Group` forwards a ref to `HTMLDivElement` (`RadioGroupFactory`), and the `PinInput` site already wraps in a plain `<div>` (`otp-input.tsx:28`), so the ref lands on the div rather than on `PinInput`'s own `HTMLInputElement` ref. Six sites to convert (`otp-input.tsx:28`, `invite-form-view.tsx:36`, `programming-language-section.tsx:84`, `code-evaluation-section.tsx:59`, and the two review sections), all mechanical.

**One hook to avoid:** `useFocusWithin` looks like the natural candidate and is a trap. Its `containsRelatedTarget` helper returns `false` for a null `relatedTarget`, so it calls `onBlur` — the identical bug this PR is fixing. Adopting it would reintroduce case 25. Worth a line in the doc comment noting the focus-based hook was considered and rejected, so nobody re-treads it.

src/components/form-field.tsx
```diff
@@ -100,33 +100,73 @@ export function revalidateOnBlur(form: ValidatableForm, path: string) {
...
-            if (typeof document !== 'undefined' && !document.hasFocus()) return
+        if (next) {
+            if (widget.contains(next)) return
             onLeave(event)
             return
         }
 
-        if (event.currentTarget.contains(next)) return
+        if (typeof document !== 'undefined' && !document.hasFocus()) return
+        if (lastPointerDownTarget && widget.contains(lastPointerDownTarget)) return
```
nathanstitt: If the `useClickOutside` switch above is more than this PR should take on, the ambiguous branch can at least be resolved from the event itself rather than from module state:

```ts
if (typeof document !== 'undefined' && !document.hasFocus()) return
// `event.target` on a bubbled focusout is the element that LOST focus — a widget descendant
// when the toolbar steals and returns the caret, and the widget itself when the user leaves.
if (widget.contains(event.target as Node) && widget.matches(':focus-within')) return
onLeave(event)
```

The toolbar renders inside the editor (`single-user-editor.tsx:154`, `collaborative-editor.tsx:419`), which is inside the element carrying this handler — so the press target `lastPointerDownTarget` is reconstructing is already on the event as `event.target`. This keeps the current prop-based signature at all six call sites and still deletes the global listeners.

Worth verifying if we go this way: `focusout` fires before focus lands, so if Lexical restores the caret asynchronously `:focus-within` reads false here and won't suppress. `runOnEditor` calls `editor.focus()` synchronously in the same tick so I'd expect it to hold, but I read that rather than ran it; deferring a frame works if not.

Separately, on the current implementation as written: this also changes behaviour for in-widget *non-focusable* space — clicking the padding around the editor now suppresses validation, since the press landed inside the widget. Under either alternative that click validates instead, which seems more defensible given the user has genuinely left the control.

src/components/form-field.tsx
```diff
@@ -100,33 +100,73 @@ export function revalidateOnBlur(form: ValidatableForm, path: string) {
...
+        'keydown',
+        () => {
+            lastPointerDownTarget = null
+        },
+        true,
+    )
+}
+
+/** Test-only: drops the recorded press so one case can't leak into the next. */
+export function __resetLastPointerDownForTests(): void {
```
nathanstitt: This export and the `beforeEach` that calls it both disappear under the `useClickOutside` change suggested above, since there'd be no module state left to reset.

If we do keep the current design: `__` plus `ForTests` signals intent, but it's still a public export from a component module — nothing stops production code importing it, and it won't be tree-shaken since the module has side effects. Resetting via a real event would avoid the export, as `document.dispatchEvent(new KeyboardEvent('keydown'))` already clears the state and the existing "fires when a key is pressed after an in-widget press" case proves that path works.

src/components/editable-text/toolbar.tsx
```diff
@@ -127,8 +129,16 @@ export const Toolbar = () => {
...
+        dispatch()
+        editor.focus()
+    }
+
     const formatText = (format: TextFormatType) => {
-        editor.dispatchCommand(FORMAT_TEXT_COMMAND, format)
+        runOnEditor(() => editor.dispatchCommand(FORMAT_TEXT_COMMAND, format))
     }
 
     const toggleLink = () => {
```
nathanstitt, line 149: `toggleLink` is the one command path that doesn't go through `runOnEditor`. The `openLinkEditor` branch is fine — focus is meant to move into the URL input — but the `existingUrl` branch dispatches `TOGGLE_LINK_COMMAND` to unlink and leaves focus wherever Lexical dropped it, which is exactly the shape of the bug this PR is fixing everywhere else.

The unlink path can't just be wrapped because the dispatch happens inside `editor.getEditorState().read()`. Something like hoisting the decision out of the read and then calling `runOnEditor` on the unlink branch would keep it consistent:

```ts
const toggleLink = () => {
    let existingUrl: string | null = null
    editor.getEditorState().read(() => { /* ...compute existingUrl... */ })
    if (existingUrl) {
        runOnEditor(() => editor.dispatchCommand(TOGGLE_LINK_COMMAND, null))
    } else {
        linkEditor.openLinkEditor(null)
    }
}
```

Worth checking whether unlink actually reproduces the grey-circle/false-error symptom before restructuring — if it doesn't, a comment saying so would save the next person the same double-take.

src/app/[orgSlug]/admin/settings/use-code-env-form.ts
```diff
@@ -184,7 +187,10 @@ export function useCodeEnvForm(image: CodeEnv | undefined, onCompleteAction: ()
             orgSlug,
             codeEnvId: image!.id,
             ...rest,
-            starterCodeFileNames: starterCodes?.map((f) => f.name),
+            // Omitted rather than sent empty. The form seeds `starterCodes` as `[]` so the
+            // create schema can report its own requirement, and on edit an empty array would
+            // read as "the admin cleared the list" instead of "left the existing files alone".
+            starterCodeFileNames: starterCodeUploaded ? starterCodes?.map((f) => f.name) : undefined,
```
Comment (nathanstitt) [inline src/app/[orgSlug]/admin/settings/use-code-env-form.ts]: `starterCodeUploaded` is `!!starterCodes?.length` on the line above, so `starterCodeUploaded ? starterCodes?.map(...) : undefined` can only ever take the truthy branch when the array is non-empty — the `?.` is dead once we're inside the ternary. Reads a bit like the optionality still matters when it doesn't:

```ts
starterCodeFileNames: starterCodeUploaded ? starterCodes!.map((f) => f.name) : undefined,
```

or, avoiding the bang, derive both from one narrowed local. Minor, but the current shape made me re-read it twice to confirm the empty-array case really does become `undefined`.

src/app/[orgSlug]/admin/settings/use-code-env-form.ts
```diff
@@ -63,7 +63,10 @@ export function useCodeEnvForm(image: CodeEnv | undefined, onCompleteAction: ()
             language: (image?.language || 'R') as Language,
             url: image?.url || '',
             isTesting: image?.isTesting || false,
-            starterCodes: undefined,
+            // `[]`, not undefined: the create schema's array type check fails before `.min(1)`
+            // runs, so an untouched dropzone reported Zod's "expected array, received undefined"
+            // instead of "At least one starter code file is required" (OTTER-647).
+            starterCodes: [],
```
Comment (nathanstitt) [inline src/app/[orgSlug]/admin/settings/use-code-env-form.ts:69]: Seeding `[]` fixes the create-path message cleanly. One thing worth confirming for the *edit* path though: `editOrgCodeEnvSchema` has `starterCodes` as `.optional()` with a `.refine()` on file size. With `undefined` the refine was skipped entirely; with `[]` it now runs against an empty array — `[].every(...)` is `true`, so it passes, and `requireCommandLineForEachExtension` already does `data.starterCodes ?? []` and falls back to `existingStarterCodeFileNames`. So I think this is genuinely inert on edit.

But that's four interacting pieces holding it up, and none of them are covered by a test that opens the *edit* modal and saves without touching the dropzone. The two new tests here are both create-path. Could we add an edit-path case asserting Save still works untouched and that `starterCodeFileNames` is omitted from the update payload? That's the regression this seeding change could plausibly cause, and right now nothing would catch it.

src/app/[orgSlug]/admin/settings/code-envs.test.tsx
```diff
@@ -121,6 +121,39 @@ describe('CodeEnvs', async () => {
...
+
+        fireEvent.click(screen.getByRole('button', { name: /Add Code Environment/i }))
+        await waitFor(() => {
+            expect(screen.getByRole('heading', { name: /Add Code Environment/i })).toBeInTheDocument()
+        })
+
+        expect(screen.queryByText(/At least one starter code file is required/i)).not.toBeInTheDocument()
+
+        // The dropzone has no input to blur, so "left incomplete" is visited-then-left-empty.
+        fireEvent.blur(screen.getByText(/Drop files or click to browse/i))
```
Comment (nathanstitt) [inline src/app/[orgSlug]/admin/settings/code-envs.test.tsx]: This blurs the inner `<Text>` (`Drop files or click to browse`), not the `Dropzone` that actually carries `onBlur={() => setTouched(true)}`. It passes because React's `onBlur` is `focusout` and bubbles — so the assertion holds, but the test is coupled to a piece of copy nested two levels inside the element under test. If someone restructures the dropzone's children the test breaks for a reason unrelated to what it's checking.

Could we anchor to the dropzone itself instead? It has `aria-invalid` wired on it, so something like a `data-testid` on the `Dropzone` (or `getByRole('presentation')` if Mantine gives it one) would express "blur the dropzone" directly. The comment above the line already explains the visited-then-left-empty model nicely — this would just make the selector match the prose.

src/app/[orgSlug]/admin/settings/code-envs.test.tsx
```diff
@@ -121,6 +121,39 @@ describe('CodeEnvs', async () => {
         expect(nameInput).toHaveValue('1BAD')
     })
 
+    // OTTER-647: the form seeded starterCodes as undefined, so the create schema's array type
+    // check failed before `.min(1)` could run and Save surfaced Zod's internal
+    // "expected array, received undefined" instead of naming the requirement.
+    it('names the starter code requirement in plain language', { timeout: 15000 }, async () => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/admin/settings/code-envs.test.tsx]: `{ timeout: 15000 }` matches the two cases above it so it's consistent with the file, but it's worth a sanity check on whether this one actually needs it — a 15s ceiling on a test that clicks Save and reads an error message suggests something slow in the render path rather than the assertion. If it passes comfortably without the override, dropping it keeps the suite honest about where the real cost is.

src/components/study/programming-language-section.test.tsx
```diff
@@ -28,6 +29,27 @@ const FormWrapper: FC<FormWrapperProps> = ({ orgSlug = '' }) => {
     )
 }
 
+const LANGUAGE_ERROR = 'Programming language is required'
+// Its own slug, so the languages query misses the cache `TestingProviders` shares across this
+// file (staleTime 60s) and actually reads the two-language response below. Reusing 'test-org'
+// replays the single-language result, which the section auto-selects, leaving no error state.
+const MULTI_LANGUAGE_ORG = 'multi-language-org'
```
nathanstitt, line 36: The `MULTI_LANGUAGE_ORG` comment is genuinely useful — it names a cache-sharing trap that would otherwise cost someone an afternoon. Thank you for writing that down rather than just picking a different slug silently.

One thought on robustness: this relies on `TestingProviders` creating a query client whose cache is shared across the file with `staleTime: 60s`. That's an implicit contract with a component in another file. If `TestingProviders` ever moves to per-render clients, this test starts passing for the wrong reason (fresh cache, correct response, still works) — which is fine — but the *other* tests in the file that depend on the shared cache would change behaviour with no signal. Not blocking, just noting the coupling is load-bearing in both directions.

src/app/account/signin/sign-in-form.tsx
```diff
@@ -54,7 +54,11 @@ export const SignInForm: FC<{
             email: '',
             password: '',
         },
-        validate: zodResolver(signInSchema),
+        // errorPriority: the resolver assigns issues in order, so the LAST one wins by default.
+        // An empty email fails both `min(1)` and `email()`, and reported "Invalid email" when the
+        // field was simply blank. First-issue priority keeps each message on its own case
+        // (blank / malformed / too long) (OTTER-647).
+        validate: zodResolver(signInSchema, { errorPriority: 'first' }),
```
nathanstitt, line 66: Nice, surgical fix — `errorPriority: 'first'` is exactly the right lever here and the comment explains the last-wins default well.

Worth considering whether this should be the default everywhere rather than a one-off. There are ~15 other `zodResolver(...)` call sites in the app, and any schema that stacks `min(1)` before a format check has the same blank-reads-as-malformed problem waiting in it. Since we already re-export the resolver through `@/common`, wrapping it there with `errorPriority: 'first'` baked in would fix the class rather than the instance:

```ts
// src/common.ts
export const zodResolver: typeof zod4Resolver = (schema, options) =>
    zod4Resolver(schema, { errorPriority: 'first', ...options })
```

Probably out of scope for a defect-fix PR — but if it's deliberately scoped to sign-in, saying so in the comment would stop the next person wondering whether the other forms were missed or considered.

src/app/[orgSlug]/study/[studyId]/proposal/collaborative-proposal-text-field.test.tsx
```diff
@@ -0,0 +1,50 @@
+import { MantineProvider } from '@mantine/core'
+import { describe, expect, faker, it, render, screen, vi } from '@/tests/unit.helpers'
+import { YjsWebsocketProvider } from '@/lib/realtime/yjs-websocket-context'
+import { theme } from '@/theme'
+import type { ProposalTextFieldKey } from '@/lib/collaboration-documents'
+import { CollaborativeProposalTextField } from './collaborative-proposal-text-field'
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/collaborative-proposal-text-field.test.tsx:3]: This file builds its own provider stack (`MantineProvider` + `YjsWebsocketProvider`) and calls bare `render` rather than `renderWithProviders`. Given the `singleUserEditing` requirement that's understandable — the shared helper probably doesn't wire Yjs — but it does mean this test opts out of whatever else `renderWithProviders` sets up (QueryClient, Modals), and it'll silently drift if that helper gains anything.

Would it be worth teaching `renderWithProviders` an optional Yjs wrapper, or at least leaving a line here saying which providers were deliberately skipped and why? The `singleUserEditing` comment at the top explains the *Yjs* choice well; it's the absence of the usual helper that isn't accounted for.

src/app/[orgSlug]/study/[studyId]/proposal/collaborative-proposal-text-field.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+                    error={undefined}
+                    onChange={vi.fn()}
+                    onBlur={vi.fn()}
+                    websocketProvider={null}
+                />
+            </YjsWebsocketProvider>
+        </MantineProvider>,
+    )
+
+const fieldNamed = (label: string) => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/proposal/collaborative-proposal-text-field.test.tsx]: `fieldNamed` looks up fields by their display label, so these two cases are pinned to the exact strings `'Research question(s)'` and `'Additional notes or requests'`. The `throw` gives a clear failure if someone renames a label, which is good — but it fails the test for a copy change rather than a behaviour change.

Since what we actually care about is "a required field" and "an optional field", could we select on that property instead? Something like `editableTextFields.find((f) => f.required)` / `.find((f) => !f.required)` expresses the invariant directly and survives copy edits. The current form is fine if the labels are considered stable, just noting the tradeoff.

src/app/[orgSlug]/study/[studyId]/review/review-decision-section.test.tsx
```diff
@@ -14,7 +14,14 @@ import { ReviewDecisionSection } from './review-decision-section'
 
 function Wrapper({ study, labName = 'Rice University' }: { study: Submitted<SelectedStudy>; labName?: string }) {
     const decision = useReviewDecision()
-    return <ReviewDecisionSection decision={decision} study={study} labName={labName} />
+    return (
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/review-decision-section.test.tsx:27]: The `leave the group` button is a neat way to drive the error state without reaching into the form internals, and it keeps the test honest about going through `decision.onBlur()`. Worth noting it's now rendered in *every* case in this describe, not just the two new ones — so any existing assertion that counts buttons or queries broadly (`getAllByRole('button')`, snapshot-ish checks) now sees one extra element.

Skimming the file I don't think anything currently does that, so this is safe today. Just a heads-up that the shared `Wrapper` is a slightly wider blast radius than adding a local wrapper for the two new cases would be.

nathanstitt (changes requested):
looks good overall but think we should use the suggested `useClickOutside` vs hand rolling dom listeners.

src/components/form-field.tsx
therealmarv: Done — [`useWidgetBlur`](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/components/form-field.tsx#L103-L155) in `39a30f9`. The global listeners, `lastPointerDownTarget`, `__resetLastPointerDownForTests`, the test `beforeEach` and `document.hasFocus()` are all gone, and we get cleanup on unmount. Ten call sites, not six — `outputs-decision-section`, `research-interests-input` and both editors also used it.

Two pieces I had to keep alongside `useClickOutside`, because dropping them regressed the thing this card is about:

**Tab.** `useClickOutside` is `mousedown`/`touchstart` only, so a straight swap loses keyboard exit — one of the two routes Greg explicitly validated ("Tab to the next field *and* clicking neutral page whitespace each fire the error"). So a small `onBlur` stays, but it now acts *only* on a non-null `relatedTarget` outside the widget. Null is handed entirely to the pointer signal, which is what lets `document.hasFocus()` go: a tab/window switch produces no press, so it can't validate. The ambiguity you were pointing at is genuinely gone — nothing reads a null `relatedTarget` any more.

**An `onFocus` gate.** This one surprised me. `useClickOutside` fires whether or not the widget was ever entered, so without a gate the first click anywhere on the page validates *every* required field at once. Confirmed in a browser before adding it.

Two call sites weren't mechanical. `PillsInput` forwards its ref to the inner field, not the root (`RefAttributes<HTMLInputElement>` — tsc caught it), so scoping there puts the pills and their remove buttons *outside* the widget and removing a pill would commit the pending draft. It gets a [wrapper element](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/components/researcher-profile/research-interests-input.tsx#L42-L52) instead. And `invite-form-view` [chains Mantine's own `onFocus`](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/app/%5BorgSlug%5D/admin/team/invite-form-view.tsx#L29-L45) from `getInputProps` rather than replacing it, or the field stops marking touched.

Good call on `useFocusWithin` — [noted in the doc comment](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/components/form-field.tsx#L127-L128) so nobody re-treads it.

Re-verified in a browser after the rewrite against the real `useReviewFeedback` / `useReviewDecision` hooks: Bold, Bullet list and Italic each leave no error with focus still on `#review-feedback`; neutral-space click raises it with `aria-invalid="true"`; all three decision circles go `rgb(255,41,41)` and clear on select; tab-out of the radio group validates; an untouched field stays clean.

src/components/form-field.tsx
therealmarv: Went with the `useClickOutside` route above, so this branch no longer exists — there's no `document.hasFocus()` and no null-`relatedTarget` handling left to disambiguate.

Your reading of `event.target` was right, and it's what convinced me the *press* is the honest signal rather than anything on the focus event. `useClickOutside` gets at the same fact from the other side (it never fires for a press inside the ref), with the bonus that it doesn't depend on when Lexical restores the caret — so the `:focus-within` timing question you flagged doesn't arise either.

On the in-widget non-focusable space: `useClickOutside` preserves the current behavior rather than your alternative, since a click on the widget's own padding is inside the ref and so suppresses validation. I left it that way deliberately — the ref boundary is now the single definition of "inside", and having padding-clicks validate while toolbar-clicks don't would reintroduce a second, fuzzier notion of it. Happy to revisit if you think the padding case matters in practice; it wasn't in any of the QA reports.

src/components/form-field.tsx
therealmarv: Gone, along with the module state it existed to reset and the `beforeEach` that called it. You were right that it had no business being a public export from a component module.

The [rewritten tests](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/components/form-field.test.tsx#L50-L143) now drive real interactions throughout (`user.click`, `user.tab`) rather than dispatching synthetic events at module state, so there's nothing to leak between cases. Six cases: intra-widget focus moves, click on an outside control, click on neutral space, Tab out, the drop-to-body case that shipped broken, and the never-entered gate.

src/components/editable-text/toolbar.tsx
therealmarv: Good catch, fixed in [`toolbar.tsx`](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/components/editable-text/toolbar.tsx#L140-L163). Close to your sketch, with one simplification: `EditorState.read()` returns its callback's value, so the decision comes straight out of the read with no mutable holder:

```ts
const selected = editor.getEditorState().read(() => {
    const selection = $getSelection()
    if (!$isRangeSelection(selection)) return null
    ...
    return { existingUrl: ... }
})
if (!selected) return
if (selected.existingUrl) runOnEditor(() => editor.dispatchCommand(TOGGLE_LINK_COMMAND, null))
else linkEditor.openLinkEditor(null)
```

The `if (!selected) return` matters — my first pass hoisted a bare `existingUrl` and accidentally turned "no range selection" into "open the link editor", which the original `return` inside the read had prevented.

I did check whether unlink reproduced the symptom before restructuring, in a browser: on the pre-fix code the unlink left `document.activeElement` on `<body>`, same shape as Bold. So it was a real instance, not a theoretical one. Verified after: unlinking removes the anchor (`#review-feedback a` goes 1 → 0), keeps focus on `#review-feedback`, and raises no error. `openLinkEditor` still hands focus to the URL input, and cancelling returns it to the editor.

src/app/[orgSlug]/admin/settings/use-code-env-form.ts
Comment (therealmarv) [reply src/app/[orgSlug]/admin/settings/use-code-env-form.ts]: Right, the `?.` was dead. Fixed in [`use-code-env-form.ts`](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/app/%5BorgSlug%5D/admin/settings/use-code-env-form.ts#L180-L205) by deriving one narrowed local instead of reaching for the bang:

```ts
const newStarterCodes = starterCodes?.length ? starterCodes : null
...
starterCodeFileNames: newStarterCodes?.map((f) => f.name),
starterCodeUploaded: !!newStarterCodes,
```

Now the `?.` is load-bearing (it's what produces the `undefined`), and `starterCodeUploaded` and the later upload guard both read off the same local rather than re-deriving `starterCodes?.length` three times.

src/app/[orgSlug]/admin/settings/use-code-env-form.ts
Comment (therealmarv) [reply src/app/[orgSlug]/admin/settings/use-code-env-form.ts:69]: Your analysis is right — it is inert on edit, and you're also right that nothing was holding those four pieces together. Added [an edit-path case](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/app/%5BorgSlug%5D/admin/settings/code-envs.test.tsx#L157-L191): seeds a code env with `starterCodeFileNames: ['existing.R']`, opens the edit modal, renames it, saves without touching the dropzone.

Rather than spy on the payload I assert the outcome in the DB, since that's where sending `[]` with `starterCodeUploaded: true` would actually bite:

```ts
expect(saved.name).toBe('Renamed Env')
expect(saved.starterCodeFileNames).toEqual(['existing.R'])
```

The rename is there to prove the save really went through rather than silently no-op'ing. It fails against the payload shape you were worried about, so it's a real guard rather than a passing formality.

src/app/[orgSlug]/admin/settings/code-envs.test.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/admin/settings/code-envs.test.tsx]: Agreed — it passed for the wrong reason. Put a `data-testid` on the [`Dropzone` itself](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/app/%5BorgSlug%5D/admin/settings/code-env-form.tsx#L112-L119) and the [test now blurs that](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/app/%5BorgSlug%5D/admin/settings/code-envs.test.tsx#L141-L155):

```ts
fireEvent.blur(screen.getByTestId('starter-code-dropzone'))
```

I checked `getByRole('presentation')` first, as you suggested — Mantine's `Dropzone` doesn't give it one, so the testid was the honest option. Now the element under test is the one carrying both `onBlur` and `aria-invalid`, and restructuring the children can't break it.

src/app/[orgSlug]/admin/settings/code-envs.test.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/admin/settings/code-envs.test.tsx]: You were right to poke at it — dropped, and it passes comfortably. It runs in **93ms**; I'd copied the 15s from its neighbours without checking whether this one earned it. It doesn't: it opens the modal and clicks Save with an empty form, so there's no upload or mutation round-trip on the path.

The two above it do genuinely need theirs (the create test uploads a file and runs the full save, ~855ms locally, and slower under load), so I left those alone.

src/app/account/signin/sign-in-form.tsx
therealmarv: Taking your "probably out of scope" read and [saying so in the comment](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/app/account/signin/sign-in-form.tsx#L57-L66), so the next person knows the other forms were considered rather than missed:

> Scoped here rather than baked into the `@/common` re-export deliberately: any schema stacking `min(1)` before a format check has the same latent problem, but flipping the default would rewrite the messages of every other form in the app, none of which this card covers. Worth its own card, with the messages re-read form by form.

The class fix is the right end state and your wrapper is the right shape for it. What stops me doing it here is that `errorPriority: 'first'` doesn't just fix blank-reads-as-malformed — it changes which message wins wherever *any* schema stacks two failing rules, and some of those may currently be showing the more useful of the two. That's a per-form read, not a flag flip, and it'd land unreviewed in a PR QA is re-validating against a specific defect list. Also added to the "not in this PR" section of the description so it doesn't get lost.

src/app/[orgSlug]/study/[studyId]/proposal/collaborative-proposal-text-field.test.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/proposal/collaborative-proposal-text-field.test.tsx:3]: Fixed properly rather than documented — the fork wasn't necessary. `renderWithProviders` already wraps `YjsWebsocketProvider`; the only thing this file needed was `singleUserEditing`, so [that's now an option](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/tests/unit.helpers.tsx#L123-L146):

```tsx
renderWithProviders(<CollaborativeProposalTextField ... />, { singleUserEditing: true })
```

The [test file](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/proposal/collaborative-proposal-text-field.test.tsx) drops its `MantineProvider` / `YjsWebsocketProvider` / bare `render` entirely and now gets the query client, modals and spy mode like everywhere else, so the drift you were pointing at can't happen. The "why" moved onto the helper, where it applies to every future caller rather than just this file.

src/app/[orgSlug]/study/[studyId]/proposal/collaborative-proposal-text-field.test.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/proposal/collaborative-proposal-text-field.test.tsx]: Agreed, the invariant is the property not the copy. [Selecting on it now](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/proposal/collaborative-proposal-text-field.test.tsx#L20-L41):

```ts
const field = fieldWhere((f) => !!f.required, 'required')
const field = fieldWhere((f) => !f.required, 'optional')
```

Kept the throw, since "the config no longer has an optional field" should still fail loudly — but now it fails on the config actually losing that property rather than on someone rewording a label. The assertions still use `field.label`, so they follow whatever the found field is called.

src/app/[orgSlug]/study/[studyId]/review/review-decision-section.test.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/review-decision-section.test.tsx:27]: Fair — narrowed it rather than leaving it as a known-safe-today. The button is [opt-in now](https://github.com/safeinsights/management-app/blob/39a30f9cdfad251368cd035b044d33e6d304d50c/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/review/review-decision-section.test.tsx#L15-L43), so only the two new cases render it and every pre-existing case keeps exactly the DOM it had:

```tsx
renderWithProviders(<Wrapper study={study} withLeaveButton />)
```

Extracted as a small `LeaveGroupButton` with `isVisible` rather than a `{cond && ...}` inline, to match the project's conditional-visibility rule. Blast radius is back to the two cases that need it.

therealmarv:
Thanks — all 13 comments addressed in `39a30f9`, replies in thread. Summary of the round:

**The `useClickOutside` switch is done.** Gone: the module-level `pointerdown`/`keydown` listeners, `lastPointerDownTarget`, `__resetLastPointerDownForTests`, the test `beforeEach`, and the `document.hasFocus()` probe. Ten call sites converted (six was an undercount — `outputs-decision-section`, `research-interests-input` and both editors also used it). We now get cleanup on unmount, and nothing reads a null `relatedTarget` any more.

Two things I kept alongside it, both because dropping them regressed the criterion this card is about:

- **Tab.** `useClickOutside` is `mousedown`/`touchstart` only, so a straight swap loses keyboard exit — one of the two routes Greg validated. A small `onBlur` remains, acting only on a non-null `relatedTarget` outside the widget.
- **An `onFocus` gate.** `useClickOutside` fires whether or not the widget was ever entered, so without it the first click anywhere validates every required field on the page at once.

Two call sites weren't mechanical: `PillsInput` forwards its ref to the inner field rather than the root (scoping there would make removing a pill commit the draft), and `invite-form-view` has to chain Mantine's own `onFocus` rather than replace it.

**Also fixed:** the toolbar unlink now goes through `runOnEditor`, and it did reproduce the symptom — I checked before restructuring. The dead `?.` is gone. New edit-path test asserting the starter code survives an untouched dropzone. Dropzone test anchored to the element under test. The 15s timeout dropped (93ms). `renderWithProviders` learned `singleUserEditing` so the forked provider stack could go entirely. Fields selected by `required` rather than label. Leave-button made opt-in. Sign-in `errorPriority` scoping now recorded as deliberate, with a note in the description so the class fix isn't lost.

**Verification:** `pnpm run checks` clean, full unit suite green (249 files, 2391 tests). Since the guard is the load-bearing piece for the QA defects, I also re-checked it in a browser after the rewrite, driving the real `useReviewFeedback`/`useReviewDecision` hooks — 14/14: Bold/Bullet list/Italic leave no error with focus still on `#review-feedback`, neutral-space click raises it with `aria-invalid="true"` and a reachable describedby, all three decision circles go `rgb(255,41,41)` and clear on select, tab-out of the radio group validates, an untouched field stays clean, and unlink keeps focus in the editor.

One behavior worth recording, unchanged by this PR: inside the Lexical editor, Tab and Shift+Tab are both consumed for indent/outdent, so keyboard focus cannot leave it and no error is raised. That's Lexical's own handling — with no `focusout`, the previous handler had nothing to act on either. Leaving that editor by click validates as expected.

src/components/form-field.tsx
```diff
@@ -100,34 +101,60 @@ export function revalidateOnBlur(form: ValidatableForm, path: string) {
...
+    const ref = useClickOutside<T>(leave)
+
+    return {
+        ref,
+        onFocus: () => {
+            visited.current = true
+        },
+        onBlur: (event: FocusEvent<T>) => {
+            const next = event.relatedTarget
+            if (!next || event.currentTarget.contains(next)) return
```
nathanstitt, line 168: One keyboard route quietly changed behavior in the rewrite: **Escape**. `EscapeFocusPlugin` blurs the editor root on Escape, which lands here as a focusout with a null `relatedTarget` — dropped by design — and there's no press for the pointer signal to act on. The old implementation cleared the recorded press on keydown precisely so this case would validate, and the test that pinned it (`fires when a key is pressed after an in-widget press`) was deleted in the rewrite without a replacement.

Severity is low — the error still surfaces on the next outside press, and submit validation is the backstop — but the doc comment's Keyboard bullet reads as "keyboard exit is covered" when Tab is covered and Escape isn't. If deferring Escape to the next signal is the intended trade-off (it may well be — there's no clean way to see it from these two signals alone), worth a sentence in the Keyboard bullet saying so, so the deleted test doesn't read as an accidental loss.

src/components/form-field.tsx
```diff
@@ -100,34 +101,60 @@ export function revalidateOnBlur(form: ValidatableForm, path: string) {
...
+export function useWidgetBlur<T extends HTMLElement = HTMLDivElement>(onLeave?: () => void): WidgetBlurProps<T> {
+    const visited = useRef(false)
+
+    const leave = () => {
+        if (!visited.current) return
+        visited.current = false
+        onLeave?.()
+    }
+
+    const ref = useClickOutside<T>(leave)
```
nathanstitt, line 159: The pointer signal is now decoupled from focus in the other direction too: `useClickOutside` fires on any outside `mousedown`/`touchstart` once the widget has been visited, including presses that never move focus — grabbing the scrollbar, a touch-scroll of the page, a drag-select starting on neutral space. The old design only ever validated from a `focusout`, so the error could not appear while the caret was still in the field; now an empty required editor can flag mid-edit when the user touch-scrolls to reread something, and in `research-interests-input` the same press commits the draft pill.

On a desktop-first admin app this is probably fine, and every alternative has its own warts — just flagging it as a behavior the Pointer bullet could name, since "the press target decides" currently implies the press also implies leaving.

nathanstitt (approved):
thanks Marvin!

src/components/form-field.tsx
therealmarv: You're right on both counts, and it was a silent loss rather than a considered trade-off at the time. I verified it: `EscapeFocusPlugin` calls `rootElement?.blur()` with nothing to receive focus, so it lands as a null `relatedTarget` with no press behind it, and both signals decline it.

Deferring is the right answer, but I went looking for a real fix first and the reason it isn't one is worth recording, so [the Keyboard bullet now says it](https://github.com/safeinsights/management-app/blob/09e6ce58c908831189a97ae3a10ac15ddbb1b2bb/src/components/form-field.tsx#L131-L138):

> Escape is the one keyboard exit deliberately *not* covered. [...] Acting on it here would mean either validating every null `relatedTarget` (the OTTER-647 bug) or probing `document.activeElement` a tick later, and a bare Escape handler would be worse still: the widgets that do not blur on Escape (radio groups, `PinInput`, pills) would flash a required error with the caret still inside them.

That last part is what settled it for me. An `onKeyDown` on the widget root looks like the cheap fix and propagates free through the existing `{...widgetBlur}` spread, but only the two editors blur on Escape. At the other eight call sites Escape moves nothing, so the handler would raise a required error while the user is still typing in the field, which is precisely the mid-interaction flash the hook exists to prevent. Distinguishing them needs an async focus probe, i.e. the timing guesswork this round deleted.

So: error still surfaces on the next outside press or on submit, and the doc no longer implies Tab and Escape are both handled.

src/components/form-field.tsx
therealmarv: Agreed, and the `research-interests-input` half is the part I'd underweighted: there the press doesn't just surface an error, it commits the pending pill via `onAdd`. A touch-scroll mid-typing turns a half-typed interest into a pill. Same trade-off, but with a side effect rather than just a message.

[Named in the Pointer bullet now](https://github.com/safeinsights/management-app/blob/09e6ce58c908831189a97ae3a10ac15ddbb1b2bb/src/components/form-field.tsx#L124-L129):

> The converse is that the press alone decides, so any outside press counts once the widget has been visited, including ones that never move focus: a touch-scroll of the page, or a drag-select starting on neutral space. [...] Accepted deliberately, because the alternative is to re-couple the press to a focus check, which is where the null `relatedTarget` ambiguity came from.

Keeping the behavior as-is: the moment the press has to be corroborated by a focus check we're back to interpreting a null `relatedTarget`, which is the bug. Desktop-first admin app, so the touch-scroll case is the least likely of the three in practice, and drag-select from neutral space is arguably a real "moved on" anyway.

I trimmed the scrollbar example from the doc, incidentally: the other two I'm confident of, but native scrollbar presses are inconsistent about emitting `mousedown` across browsers and I didn't want to assert it without checking.

src/components/study/programming-language-section.test.tsx
therealmarv: Thanks. And you're right that the coupling cuts both ways, which hadn't occurred to me: my comment warns the next person off the trap, but it doesn't protect the *other* tests in the file that quietly depend on the shared cache. If `TestingProviders` moves to per-render clients this test keeps passing on a fresh cache while those change behavior with no signal.

Leaving it here since it's a property of the shared helper rather than anything this card touches, but it's a real gap worth its own look at some point.

PR #936, opened 2026-08-04
OTTER-667: Fix security key focus and study heading hierarchy

stetradis:
## Summary
- Return focus to the security key input after an invalid key submission (was falling to `<body>` because `requestAnimationFrame` raced React's re-enable of the disabled field). Focus on mount instead of hardcoding a blue focus border.
- Correct study heading levels: `ProposalStepHeader` is `h2`, `FormSectionHeader` is `h3`, under the page `h1`. 

  On study screens the structure is:
  
    StudyPageHeader → h1 — page title (e.g. “Study proposal”, “Secondary analysis study”)
    ProposalStepHeader → h2 — main step section (e.g. “Study code”, “Verify outputs”, “Review study code”)
    FormSectionHeader → h3 — subsection inside that step (e.g. “Security key”)
    There is nothing between the page title and the step header in these flows — no intermediate section that would own h2 or h3. So the step heading is the next outline level under the page title, which makes it h2.
    
    It was previously order={4} (h4), which skipped h2 and h3. That looked fine visually (Mantine’s fz controls size independently of order), but it broke the outline: AT users jumping by heading would go from h1 straight to h4 and miss the expected hierarchy.


src/components/study/proposal-step-header.tsx
```diff
@@ -26,7 +26,7 @@ export function ProposalStepHeader({
             <Text fz={10} fw={700} c="charcoal.7" pb={4}>
                 {stepLabel}
             </Text>
-            <Title order={4} fz={20} c="charcoal.9" pb={4}>
+            <Title order={2} fz="xl" c="charcoal.9" pb={4}>
```
nathanstitt, line 29: The reasoning in the PR description is right, but I think this only fixes half the outline — the h4 skip moves rather than goes away.

On the code-review page (`review/code-review.tsx`), `FeedbackAndNotesSection` renders as a *sibling* of this step header, and its "Feedback and notes" title is `Title order={4} fz={20}` (`feedback-and-notes.tsx:168`) — the exact same styling this line had before the change. So the two were peers by design, and after this PR that page reads **h1 → h2 → h4**.

Same story for the other `order={4} fz={20}` peers that sit alongside step headers: `programming-language-section.tsx:108`, `study-org-selector.tsx:34`, `proposal/form.tsx:106`, `edit-initial-request-section.tsx:74`, plus the `order={4} size="xl"` ones in `study-overview-screen.tsx` and `study-results.tsx`.

Could we bump the sibling section titles to `order={3}` in the same pass? Otherwise an AT user jumping by heading still hits a skipped level, just one page over.

src/components/study/use-security-key-form.ts
```diff
@@ -38,11 +38,12 @@ export function useSecurityKeyForm({ job, onDecrypted }: UseSecurityKeyFormOptio
         },
     })
 
-    const failInvalid = useCallback(() => {
-        setError(ERRORS.invalid)
-        requestAnimationFrame(() => inputRef.current?.focus())
+    useEffect(() => {
```
nathanstitt: Focusing on mount is a bigger behavior change than the `rAF` fix it's bundled with — this grabs focus as soon as the form renders, which on the outputs step means the page scrolls the textarea into view and skips past the `StudyPageHeader` / `ProposalStepHeader` / banner above it.

That's a bit at odds with the rest of the PR: a screen reader user landing on the page would normally want to read the h1 and the locked-outputs banner before being dropped into an input. WCAG doesn't forbid autofocus, but 2.4.3 (focus order) generally wants the initial focus at the top of the document.

The error-recovery focus below (line 64) is the part that fixes the reported bug. Is the mount focus something design actually asked for, or did it come along to replace the blue border? If it's the latter, we could drop it and keep just the recovery effect.

**If we do keep it, it doesn't need an effect.** Mantine's `Textarea` forwards `autoFocus` straight to the underlying `<textarea>`, so this whole hook is just a prop on the input:

```tsx
<SecurityKeyInput autoFocus ... />
```

I checked that it actually passes through (rendered `SecurityKeyInput` with `autoFocus` and asserted `toHaveFocus()` — passes), and that the field renders enabled on first paint, since `security-key-form.tsx` sets only `disabled={isDecrypting}` and that's false on mount. `autoFocus` on a disabled field would be a no-op, but that's not the case here. We already use this prop in `otp-input.tsx:30` and `organization-settings-edit.tsx:106`. (There's also `data-autofocus`, as in `edit-org-form.tsx:75`, but that's the Modal/`FocusTrap` convention and only means something inside a focus-trapped container — not this.)

Worth noting Mantine has nothing for the *recovery* case: `useFocusTrap` is containment for modals/popovers, not "pull focus to this control when state changes," and `@mantine/form` doesn't cover it either. So the `useEffect([error, isPending])` below is the right shape and should stay — dropping the mount effect in favor of `autoFocus` would also make it unambiguously *the* error-recovery effect, which makes my note on the test easier to act on.

src/components/study/use-security-key-form.ts
```diff
@@ -60,6 +61,12 @@ export function useSecurityKeyForm({ job, onDecrypted }: UseSecurityKeyFormOptio
         onError: failInvalid,
     })
 
+    useEffect(() => {
```
nathanstitt, line 68: Nice fix on the root cause — keying off `error && !isPending` runs after React has committed the re-enable, so there's no race with the disabled field the way `requestAnimationFrame` had.

One edge worth thinking about: because the dep array is `[error, isPending]` and `failInvalid` sets the *same* `ERRORS.invalid` string every time, a second bad key won't change `error`, but it will flip `isPending` true→false, so the effect does re-run. Good — that case works.

The one that doesn't: if the user clicks away to the "lost your key" popover and then submits the same invalid key again, focus gets yanked back mid-interaction. Probably fine in practice, just flagging since it's the kind of thing that reads as a jump.

src/components/study/security-key-input.tsx
```diff
@@ -4,23 +4,18 @@ import { forwardRef } from 'react'
...
-        )
-    },
+    ({ error, disabled, ...props }, ref) => (
+        <Textarea
+            ref={ref}
+            autoComplete="off"
+            aria-required
+            disabled={disabled}
+            error={error ? <span role="alert">{error}</span> : undefined}
+            styles={{ input: { minHeight: 72 } }}
```
nathanstitt, line 15: Dropping the always-on blue border is a clear improvement — a non-focused field painted with the focus color is misleading for sighted keyboard users, since it makes focus position ambiguous once the real focus ring shows up elsewhere.

Worth confirming with design that they're okay losing it entirely, though. The old comment claimed it was a spec ("Design specifies the focus-state border (blue-7) on page load"), and the PR swaps it for real focus-on-mount. If design's actual intent was "this field should look active on load," the mount focus delivers that; if it was purely decorative, we may now want no autofocus at all (see my note on the mount effect).

src/components/study/security-key-form.test.tsx
```diff
@@ -138,6 +138,7 @@ describe('SecurityKeyForm', () => {
         expect(await screen.findByText(INVALID_ERROR)).toBeInTheDocument()
         expect(screen.getByRole('textbox')).toBeEnabled()
         expect(screen.getByRole('button', { name: 'View' })).toBeEnabled()
+        await waitFor(() => expect(screen.getByRole('textbox')).toHaveFocus())
```
nathanstitt, line 166: Since the hook now focuses on mount too, this assertion would pass even if the error-recovery effect were deleted — focus never left the input in the first place. That makes it a weak guard on the actual regression.

Could we blur first so the test proves the recovery path specifically? Something like `screen.getByRole('textbox').blur()` (or clicking another control) after the error appears, then asserting focus comes back. Same applies to the blank-submit test above.

src/components/study/form-section-header.tsx
```diff
@@ -9,7 +9,7 @@ interface FormSectionHeaderProps {
 
 export const FormSectionHeader: FC<FormSectionHeaderProps> = ({ title, description, required }) => (
     <Stack gap="md">
-        <Title fz="xl" fw={700} c="charcoal.9">
+        <Title fz="xl" fw={700} c="charcoal.9" order={3}>
```
nathanstitt, line 13: `order={3}` is right for the security-key case, where the chain is `StudyPageHeader` h1 → `ProposalStepHeader` h2 → this h3.

Since `SecurityKeyForm` is currently the only consumer, that holds today. If this ever gets reused on a screen without a step header in between it'd become the skip case — might be worth a short comment noting the assumed nesting, so the next caller knows the level isn't arbitrary.

nathanstitt (approved):
Looks good Stella,   I'm fine to merge if you think the auto-fucus things are needed, but might check to see about it.  In particular I'd prefer plain `autoFocus` prop

src/components/study/proposal-step-header.tsx
stetradis: Good catch - updated the heading levels throughout the study flow.

One thing to flag: the step-header copies (`study-org-selector`, `programming-language-section`, `proposal/form`, `edit-initial-request-section`, `code-post-submission-view`) went to h2 rather than h3 - they render the same `STEP N` + heading pattern as `ProposalStepHeader`, so they're the step header
itself rather than siblings of it. The genuinely subordinate sections are h3 as suggested

src/components/study/use-security-key-form.ts
stetradis: I think I initially misread the Figma on the border color - this [comment](https://www.figma.com/design/84pQzoYf7rKqOjo4qsDOeG/Card-59-69-71---Outputs-step?node-id=12287-26064&m=dev) makes it clear the field should be focused on page load. Moved it to `autoFocus` on the input as suggested.

Fair point on 2.4.3 - flagging it back to design rather than deciding it here.

PR #937, opened 2026-08-05
OTTER-674: render the note error in the editor footer slot, directly under the input

Rod0352:
Follow-up to #920, addressing the QA rejection: on the Resubmission Note the empty-field error rendered a row below the word counter (~25px under the field) instead of directly under the input.

**Cause** — as diagnosed in QA: `footerRight` gives the word counter the editor's footer row, while `InputError` rendered in a sibling `Group` below it. `isVisible={!error}` could only blank the slot, not move the error into it.

**Fix** — `Editor` / `CollaborativeEditor` / `SingleUserEditor` gain a `footerLeft` slot: the left half of the footer row, directly under the input. The note section passes the error box and the single-user save indicator into it and drops the sibling row. The error now takes exactly the slot "All changes saved" vacates, sharing the row with the counter — same geometry as the Research question(s) surface QA passed.

Notes:

- Expectation 1 (error and save label never co-exist) is preserved and now structural — they compete for the same slot.
- `aria-describedby` wiring unchanged: the `resubmissionNote-error` id moves with the node.
- Single-user mode: the saved indicator now sits beside the counter, matching the collaborative layout.
- Other `footerRight` callers are unaffected; `footerLeft` is optional and renders nothing when absent.
- Out of scope: `review-feedback-section` / `code-review-feedback-section` have the same sibling-row structure and need the same one-move fix — separate ticket.

**Tests** — new: error shares the footer row with the counter; single-user saved indicator shares the row; `SingleUserEditor` renders `footerLeft` in the footer row. Existing OTTER-674 exclusivity tests unchanged. Touched + neighboring unit suites green; lint/format clean.


src/components/editable-text/single-user-editor.tsx
```diff
@@ -153,10 +156,11 @@ export function SingleUserEditor({
                 {children}
                 <Toolbar />
             </Paper>
-            {footerRight && (
+            {(footerLeft || footerRight) && (
                 <Stack gap={4} mt={4}>
                     <Group align="center" wrap="nowrap">
-                        <Box ml="auto">{footerRight}</Box>
+                        {footerLeft}
```
nathanstitt: With `footerLeft` added, this footer row is now byte-identical to the one in `collaborative-editor.tsx` (`Group align="center" wrap="nowrap"` → `footerLeft` → optional `Box ml="auto"` for `footerRight`). Worth extracting a tiny shared `EditorFooter` so the two slots can't drift — the next slot change (or a styling tweak to the row) would otherwise need to be made in both places again.

src/components/study/collaborative-resubmission-note-section.tsx
```diff
@@ -107,19 +125,10 @@ export const CollaborativeResubmissionNoteSection: FC<CollaborativeResubmissionN
                             hasError: !!error,
                             hasDescription: false,
                         })}
+                        footerLeft={footerLeft}
```
nathanstitt, line 128: The error now lives in a `wrap="nowrap"` row shared with the word counter, where the old sibling `Group` could wrap freely. The error text itself should wrap inside its flex child, but on a narrow viewport a long message will squeeze against the counter — worth a quick narrow-width sanity check that the counter doesn't get crowded or clipped.

src/components/study/collaborative-resubmission-note-section.tsx
```diff
@@ -78,6 +88,14 @@ export const CollaborativeResubmissionNoteSection: FC<CollaborativeResubmissionN
 
     const onNoteChange = (json: string) => noteForm.setFieldValue('resubmissionNote', json)
 
+    // The error takes exactly the slot 'All changes saved' vacates, so the two can never co-exist (OTTER-674).
+    const footerLeft = (
```
nathanstitt, line 92: Nice — making the error and the save indicator compete for the same slot turns the "never co-exist" QA expectation from a coordination of two `isVisible` flags into a structural guarantee. Also checked that `fieldDescribedBy(..., { hasError })` only emits the error id when `NoteFieldError` actually mounts the `Box` carrying it, so the `aria-describedby` wiring stays consistent through the move.

nathanstitt (approved):
looks good, but probably should extract a EditorFooter as suggested then good to merge

PR #938, opened 2026-08-05
Scope load IDE to the submitting lab (OTTER-719)

nathanstitt:
[OTTER-719](https://openstax.atlassian.net/browse/OTTER-719)

## Problem

`permit('load', 'IDE')` inside the lab-membership block was unconditioned. CASL OR-combines rules, so every member of any lab org held it on every study. Eight workspace server actions are gated solely by this check and do no further ownership testing in their handlers, so a researcher could supply another study's id and read, overwrite, or delete that lab's in-progress code on the shared `CODER_FILES` volume.

## Fix

- Condition the `load IDE` grant on `submittedByOrgId`, matching the sibling `update`/`delete`/`create StudyJob` rules. A `$in` denies when the field is absent, so a subject missing it fails closed.
- Widen the `IDE` ability arm — a condition may only reference fields declared on it, so the ticket's one-line change would not have compiled. No middleware change was needed for the IDE actions; `getInfoForStudyId` already supplies `submittedByOrgId`.
- Scope `create Study` the same way. `submittingOrgSlug` arrives as a client param and `getOrgIdFromSlug` resolves any slug, so a caller could otherwise stamp another lab's id onto a new study and gain IDE access to it under the new rule. `onSaveDraftStudyAction`'s middleware now resolves that slug into `submittedByOrgId`, putting it in the ability subject so the CASL rule enforces lab membership — rather than a hand-rolled check in the handler, which would sit where someone auditing `permissions.ts` would never look.

## Note

The `researcherId` grant is deliberately left in place, so a study's author keeps IDE access regardless of current lab membership. Narrower than the cross-lab hole, but it is a second unscoped path worth tracking separately.

## Tests

Full suite passes (2410 tests, 251 files), typecheck and lint clean. New coverage:

- The ticket's required lab-B denial, plus fail-closed on a missing field and the enclave-only case.
- Cross-lab denial against the RPC endpoints themselves, asserting the file bytes survive on disk rather than only that an error returns.
- Regression guard that the submitting lab still reaches its own workspace.
- `create Study` scoped to the caller's own labs and fail-closed with no submitting lab in the subject, plus an end-to-end denial for `onSaveDraftStudyAction` with a foreign `submittingOrgSlug` asserting no row is written.

[OTTER-719]: https://openstax.atlassian.net/browse/OTTER-719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/permission-types.ts
+++ b/src/lib/permission-types.ts
@@ -60,7 +60,10 @@ type Abilities =
     | Ability<'OrgMembers', 'view', { orgId?: UUID; orgSlug?: string }>
     | Ability<'Orgs', 'view', object>
     | Ability<'MFA', 'reset', object>
-    | Ability<'IDE', 'load', { researcherId: UUID }>
+    // Both fields optional: `load IDE` is granted by two OR-combined rules — the study's own
+    // researcher (researcherId) and any member of the submitting lab (submittedByOrgId, OTTER-719).
+    // Every field used in a condition must appear on the arm for the CASL subject union to accept it.
+    | Ability<'IDE', 'load', { researcherId?: UUID; submittedByOrgId?: UUID }>
     | Ability<
           'AgentContext',
           'create' | 'update' | 'view',
--- a/src/lib/permissions.test.ts
+++ b/src/lib/permissions.test.ts
@@ -93,17 +93,21 @@ test('researcher role', () => {
         ability.can('approve', toRecord('Study', { orgId: ownLabId })),
     ).toBe(false)
 
-    expect(ability.can('create', 'Study')).toBe(true)
-
-    // update/delete are scoped to studies the researcher's own lab submitted
+    // create/update/delete are scoped to studies the researcher's own lab submitted
+    expect(ability.can('create', toRecord('Study', { submittedByOrgId: ownLabId }))).toBe(true)
     expect(ability.can('update', toRecord('Study', { submittedByOrgId: ownLabId }))).toBe(true)
     expect(ability.can('delete', toRecord('Study', { submittedByOrgId: ownLabId }))).toBe(true)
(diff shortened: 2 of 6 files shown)


therealmarv (approved):
Approving, no blocker.

The fix checks out. All eight `load IDE` sites resolve `researcherId` and `submittedByOrgId` through `getInfoForStudyId` before authorization, so the scoped rule has the fields it needs; `create Study` has exactly one authorization site, so conditioning it strands no other caller.

Two optional follow-ups, neither worth holding this up:

1. `src/server/actions/study-request.test.ts:100` still says `create Study` is unconditioned and the handler is the only enforcement point. The second commit made that false (rule at `permissions.ts:84`, enforced via middleware plus `requireAbilityTo`). A stale comment on an authz boundary is the kind of thing the next audit trips over.
2. `'the submitting lab still reaches its own study workspace'` currently passes through the `researcherId` grant rather than the new one, since `createTestProposalDraft` leaves the study author authenticated. Authenticating a second member of the same lab would make it guard what its name claims. The ability-level test already separates the two arms, so this is a weak guard rather than a coverage gap.

On the widened `IDE` arm: leaving both fields optional is the right call, and worth recording why. TypeScript's weak-type rule still makes `requireAbilityTo('load', 'IDE')` resolve to `never` when the merged args and context carry neither field, so "at least one" is enforced at compile time, and both single-field subject shapes fail closed at runtime. Requiring both would buy nothing and would break the fail-closed assertions in `permissions.test.ts`.

On size: 149 of the 170 added lines are tests, which is the right ratio for a fix like this. The rationale comments are dense in a couple of spots (the middleware block in `study-request.ts`, the arm comment in `permission-types.ts`); trimming is a matter of taste, though the one in item 1 needs to go regardless.

PR #939, opened 2026-08-05
OTTER-720: require a dedicated admin-only ability to change user roles

nathanstitt:
[OTTER-720](https://openstax.atlassian.net/browse/OTTER-720) — CRITICAL / launch blocker.

`updateUserRoleAction` was gated on `update User`, but every user holds that verb on their own id for self-profile edits. The middleware injected the client-supplied target `userId` into the ability subject, so passing your own id matched the self-update rule and the check passed regardless of admin status. Any authenticated member could call it with `isAdmin: true` and become org admin.

## Changes

- New `manageRole` verb on `User`, granted only for orgs the caller administers. The self-profile `update User` rule is untouched — `onUserResetPWAction` and the researcher-profile actions still need it (all derive `id` from the session, never from params).
- `updateUserRoleAction` gates on `manageRole` and no longer returns the target `id` from middleware. Self-edits are refused in the handler too: an admin *does* legitimately hold `manageRole` for their own org, so the ability check alone wouldn't stop it. Also prevents an org being left with zero admins.
- `onRevokeInviteAction` had no authorization at all — any caller could delete any invite by id. It serves both an org admin and the invitee declining, and the invitee may not be an org member, so no single rule covers both. Now authorized explicitly: the new admin-only `revoke PendingUser` ability, or a server-side match of the caller's Clerk emails against the invite address. That email check previously existed only in the browser.
- Role `Select` is disabled for your own row.

## Testing

9 new tests covering non-admin self-promotion, admin self-demotion, cross-org, and the four revoke paths — each asserting the db row is unchanged, not just the error response. Full suite green (251 files, 2418 tests); `pnpm run checks` passes.

One caveat: I tried to confirm the new tests fail against the pre-fix code by temporarily reverting it, but the sandbox blocked that test run partway. The fix was restored and re-verified, so the tests are green against the fix — but I haven't empirically proven they'd catch a regression.

[OTTER-720]: https://openstax.atlassian.net/browse/OTTER-720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/account/invitation/[inviteId]/create-account.action.ts
+++ b/src/app/account/invitation/[inviteId]/create-account.action.ts
@@ -5,6 +5,7 @@ import { updateClerkUserMetadata } from '@/server/clerk'
 import { getUserPublicKey } from '@/server/db/queries'
 import { onUserAcceptInvite } from '@/server/events'
 import { extractClerkCodeAndMessage, isClerkApiError } from '@/lib/errors'
+import { toRecord } from '@/lib/permissions'
 import { clerkClient } from '@clerk/nextjs/server'
 
 export const onPendingUserLoginAction = new Action('onPendingUserLoginAction')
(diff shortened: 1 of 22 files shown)


src/app/account/invitation/[inviteId]/create-account.action.ts
```diff
@@ -42,16 +43,49 @@ export const getOrgInfoForInviteAction = new Action('getOrgInfoForInviteAction')
...
+            .select(['id', 'orgId', 'email'])
+            .where('id', '=', inviteId)
+            .executeTakeFirstOrThrow(() => new ActionFailure({ invite: 'not found' }))
+
+        const isOrgAdmin = session.ability.can('revoke', toRecord('PendingUser', { orgId: invite.orgId }))
+
+        // The invitee is identified by email rather than claimedByUserId, which is only set on the
+        // signup path and stays null when an existing user opens the invite (join-team page).
+        // Checked against all Clerk addresses so merged/secondary emails match, as the UI does.
+        const isInvitee = await callerOwnsEmail(session.user.clerkUserId, invite.email)
```
Comment (therealmarv) [inline src/app/account/invitation/[inviteId]/create-account.action.ts:74]: Non-blocking nit: `isOrgAdmin` is already known, so this could short-circuit and skip the Clerk round trip on the admin path:

```ts
const isInvitee = !isOrgAdmin && (await callerOwnsEmail(session.user.clerkUserId, invite.email))
```

Reads closer to the intent too. Not a correctness issue either way, and `orgAdminInviteUserAction` on the same page already calls the Clerk backend, so this is just a saved request.

therealmarv (approved):
No blockers, approving.

Reviewed the fix: dedicated `manageRole` verb, middleware no longer returning the target `id`, the independent self-role refusal in the handler, and `revoke PendingUser` scoped by the db-derived `orgId`. Self-profile `update User` is correctly left alone. Found nothing else.

One non-blocking nit inline.

PR #940, opened 2026-08-05
Render legacy job results as text instead of writing them as HTML

nathanstitt:
[OTTER-721](https://openstax.atlassian.net/browse/OTTER-721)

Viewing a legacy `APPROVED-RESULT` or approved-log file opened an `about:blank` tab and called `document.write()` with the raw file bytes. Job output is researcher-controlled, so an inline script, an `img onerror`, or an `svg onload` executed in the app origin in the session of whoever opened the result.

The tab was also outside any CSP — an `about:blank` document has no HTTP response, so it inherits none of the headers set in `next.config.ts`. Removing the tab is the fix; tightening the policy would not have covered this.

## Change

Route the View link through `FileOrImagePreviewModal`, which is what the encrypted path already uses. It takes the `{name, contents}` shape this component already had and handles the image/text split itself, so the hand-rolled image branch goes away and legacy results pick up the JSON/CSV/log viewers.

Net −31/+14 lines. Also drops a stale `reportError()` call that was never imported in this file and so resolved to the DOM global rather than the app helper, meaning neither Sentry nor the toast ever fired (the Low finding noted on the ticket).

## Verification

New `legacy-job-results.test.tsx` asserts no tab is opened, and that an `` payload renders as literal text for both `.txt` and `.html` results. Confirmed these fail against the pre-fix code (3 of 5).

- Full unit suite: 252 files / 2408 tests passing
- `pnpm checks` green

A follow-up PR adds a nonce-based `script-src` as defense-in-depth.

[OTTER-721]: https://openstax.atlassian.net/browse/OTTER-721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/legacy-job-results.test.tsx
+++ b/src/components/legacy-job-results.test.tsx
@@ -1,0 +1,75 @@
+import { describe, it, expect, vi } from 'vitest'
+import { screen, userEvent, waitFor, renderWithProviders } from '@/tests/unit.helpers'
+import { ViewFile } from './legacy-job-results'
+import { JobFile } from '@/lib/types'
+
+const XSS_PAYLOAD = '<img src=x onerror="window.__pwned = true">'
+
+const encode = (text: string): ArrayBuffer => {
+    const bytes = new TextEncoder().encode(text)
+    return bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength) as ArrayBuffer
+}
+
+const jobFile = (path: string, contents: string): JobFile => ({
+    path,
+    fileType: 'APPROVED-RESULT',
+    contents: encode(contents),
+})
+
+const clickView = async () => {
+    await userEvent.click(screen.getByRole('button', { name: /view/i }))
+}
+
+describe('LegacyJobResults ViewFile', () => {
+    // Legacy results were written into an about:blank tab via document.write, which executed
+    // researcher-controlled job output as HTML in the app origin (OTTER-721).
+    it('never opens a tab when viewing a result', async () => {
+        const open = vi.spyOn(window, 'open').mockReturnValue(null)
+        renderWithProviders(<ViewFile file={jobFile('results.txt', XSS_PAYLOAD)} />)
+
+        await clickView()
+
+        await screen.findByRole('dialog')
+        expect(open).not.toHaveBeenCalled()
+    })
+
+    it('renders a script payload as inert text rather than markup', async () => {
+        renderWithProviders(<ViewFile file={jobFile('results.txt', XSS_PAYLOAD)} />)
+
+        await clickView()
+
+        const dialog = await screen.findByRole('dialog')
+        expect(dialog).toHaveTextContent(XSS_PAYLOAD)
+        expect(dialog.querySelector('img')).toBeNull()
+        expect(dialog.querySelector('script')).toBeNull()
+        expect((window as unknown as { __pwned?: boolean }).__pwned).toBeUndefined()
+    })
+
+    // .html routes to the highlight.js viewer rather than the plain text viewer, so it is the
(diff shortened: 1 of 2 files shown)


src/components/legacy-job-results.tsx
```diff
@@ -6,52 +6,35 @@ import { ArrowSquareOutIcon } from '@phosphor-icons/react/dist/ssr'
...
+            <Anchor role="button" onClick={() => setPreviewing(true)} style={{ display: 'flex', alignItems: 'center' }}>
                 View <ArrowSquareOutIcon size={16} style={{ marginLeft: 4 }} />
             </Anchor>
-            <ImagePreviewModal
-                isVisible={showImageModal}
-                name={path}
-                contents={content}
-                mime={mime}
-                onClose={() => setShowImageModal(false)}
+            <FileOrImagePreviewModal
```
therealmarv, line 26: Non-blocking, and probably a note in the PR body rather than a change here.

Legacy `APPROVED-RESULT` files are whole result sets, and they now open in a 500px modal instead of being rendered natively by the browser in its own tab. `csvViewer` builds a `DataTable` with one record per row and no virtualization, and `CodeViewer` highlights the entire file synchronously inside a `useMemo`. So opening a large legacy CSV is meaningfully heavier than it was, and the user also loses native find-in-page over the full document.

The encrypted path already behaves this way, so nothing about the shared viewers changes here. It is the legacy path becoming newly subject to it that seems worth recording, since these are the files most likely to be large. Fine to ship as is.

src/components/legacy-job-results.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+        await clickView()
+
+        const dialog = await screen.findByRole('dialog')
+        expect(dialog).toHaveTextContent(XSS_PAYLOAD)
+        expect(dialog.querySelector('img')).toBeNull()
+        expect((window as unknown as { __pwned?: boolean }).__pwned).toBeUndefined()
+    })
+
+    it('previews an image result as an image', async () => {
```
therealmarv, line 61: Non-blocking. The XSS cases above are the ones that matter and they cover the right paths, including `.html`, which is the extension that actually resolves to `xml` and reaches `CodeViewer`.

The one gap: the PR body notes that legacy results now pick up the JSON/CSV/log viewers, and that new routing is untested. A legacy `.csv` rendering through `csvViewer`, and one file with an `isApprovedLogType`/`isPlaintextLogType` type reaching the viewer, would cover the two capabilities this change adds beyond the security fix.

therealmarv (approved):
Approving.

This removes a real same-origin sink. Researcher-controlled bytes went through `document.write` into an `about:blank` tab, and that document has no HTTP response, so it inherited none of the headers from `next.config.ts` and there was no CSP in front of it. Agreed that removing the tab is the fix rather than tightening policy. Routing through `FileOrImagePreviewModal` also puts legacy results on the same escaped viewers the encrypted path already uses and drops the hand-rolled image branch, so the net is fewer lines and one less code path.

I checked the highlight path specifically, since `.html` now resolves to `xml` and reaches `CodeViewer`'s `dangerouslySetInnerHTML`. highlight.js 11.11.1 returns `escape(codeToHighlight)` on every internal failure under its default safe mode, and the only case that actually throws is an unregistered language, which cannot happen because the `HighlightLanguage` union and the `registerLanguage` calls in `code-viewer.tsx` are the same 30 names. Confirmed directly that the `img onerror` payload comes back escaped under both `xml` and `javascript`. The `.html` test covers the path that runs in practice.

Two non-blocking notes inline, neither of which should hold the merge.

PR #941, opened 2026-08-05
Add a nonce-based script-src CSP in production

nathanstitt:
[OTTER-721](https://openstax.atlassian.net/browse/OTTER-721) — defense-in-depth follow-up.

**Stacked on #940.** Targets `otter-721-legacy-results-xss` so the diff stays clean; retarget to `main` once that merges.

Adds the `script-src` the existing policy deliberately omitted, using a per-request nonce so Clerk and Sentry keep working. This does not fix a live hole on its own — the XSS in #940 wrote into an `about:blank` tab, which has no HTTP response and so never inherited any policy.

## Approach

A nonce cannot come from `next.config.ts`, which is request-independent, so `script-src` is emitted from the proxy under a **separate header name** — report-only while the rollout is measured, vs. the already-enforcing static policy (duplicate CSP headers under one name are enforced independently, so disjoint directives could safely share a name; see `src/lib/csp.ts`). Both terminal `NextResponse.next()` paths route through one helper, including the anon-route early return that serves `/account/signin`.

Per review: the proxy also forwards the policy as a `content-security-policy` **request** header — the documented way Next learns the nonce to stamp its inline scripts — and drops the `ClerkProvider` `nonce` prop, which Clerk discards without `dynamic`; `'strict-dynamic'` covers `clerk.browser.js`. Violations now report to Sentry's security endpoint (derived from the DSN) via `report-uri`/`report-to` plus a `Reporting-Endpoints` header, so the enforce flip can be gated on a clean feed.

Making the **root layout** dynamic is what lets Next stamp the nonce on its inline hydration scripts for the not-found entry. `export const dynamic` on `not-found.tsx` does not work — Next prerenders that internal entry regardless. Verified against a production build: the 404 page went from 8 un-nonced inline scripts to 7 of 7 carrying the nonce.

## Ships report-only

`Content-Security-Policy-Report-Only`, so violations are reported and nothing is blocked. Flipping to enforcing is a one-line change in `src/lib/csp.ts`.

Two things to settle before that flip:

1. **`global-error` is a known blocker.** It renders its own `<html>`, bypassing the root layout, and ignores a `dynamic` export, so it still ships ~6 un-nonced inline scripts. If those get blocked, `Sentry.captureException` never runs and error reporting dies silently on the error page. Recorded in the file.
2. **Real Clerk is unverified.** Clerk injects `clerk.browser.js` itself without a nonce and relies on `'strict-dynamic'` to inherit trust. e2e builds with Clerk faked, so only a deployed environment can confirm this.

## Verification

Against a production build (`NODE_ENV=production`):

- Exactly one static CSP header and one report-only header — no duplication
- Nonce differs per request; all of Next's inline scripts carry it
- SRI and nonce coexist: all 9 `integrity=` tags also carry the nonce
- API routes correctly excluded from the proxy
- Full e2e suite 28/28 passing **with the policy active** — `next start` sets `NODE_ENV=production`, so the suite does exercise it
- Full unit suite: 253 files / 2420 tests passing; `pnpm checks` green

[OTTER-721]: https://openstax.atlassian.net/browse/OTTER-721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/dashboard/layout.tsx
+++ b/src/app/dashboard/layout.tsx
@@ -6,7 +6,6 @@ import { currentUser } from '@clerk/nextjs/server'
 import { redirect } from 'next/navigation'
 import { ReactNode } from 'react'
 import { connection } from 'next/server'
-import { getCspNonce } from '@/lib/csp'
 
 // this page must be dynamically rendered
 export const dynamic = 'force-dynamic'
@@ -26,7 +25,7 @@ export default async function DashboardLayout({ children }: Readonly<{ children:
     }
 
     return (
-        <ClerkProvider publishableKey={clerkPublishableKey} nonce={await getCspNonce()}>
+        <ClerkProvider publishableKey={clerkPublishableKey}>
             <SentryUserProvider />
             <AppShell>{children}</AppShell>
         </ClerkProvider>
--- a/src/app/dashboard/layout.tsx
+++ b/src/app/dashboard/layout.tsx
@@ -6,6 +6,7 @@ import { currentUser } from '@clerk/nextjs/server'
 import { redirect } from 'next/navigation'
 import { ReactNode } from 'react'
 import { connection } from 'next/server'
+import { getCspNonce } from '@/lib/csp'
 
 // this page must be dynamically rendered
 export const dynamic = 'force-dynamic'
@@ -25,7 +26,7 @@ export default async function DashboardLayout({ children }: Readonly<{ children:
     }
 
     return (
-        <ClerkProvider publishableKey={clerkPublishableKey}>
+        <ClerkProvider publishableKey={clerkPublishableKey} nonce={await getCspNonce()}>
             <SentryUserProvider />
             <AppShell>{children}</AppShell>
         </ClerkProvider>
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -30,9 +30,11 @@ export async function generateMetadata(): Promise<Metadata> {
     }
 }
 
-// Rendered per-request so every document, including Next's own not-found entry, is generated at
-// request time and can carry the CSP nonce on its inline hydration scripts. A `dynamic` export on
-// not-found.tsx does not achieve this: Next prerenders that entry regardless (OTTER-721).
+// `connection()` in the ROOT layout opts the entire app out of static generation — every route,
+// previously-static ones like /about included, now renders per-request, not just Next's not-found
+// entry. That app-wide cost is deliberate: it is what lets every document carry the CSP nonce on
+// its inline hydration scripts. A `dynamic` export on not-found.tsx does not achieve this: Next
+// prerenders that entry regardless (OTTER-721).
 export default async function RootLayout({
     children,
 }: Readonly<{
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -15,6 +15,7 @@ import { SINGLE_USER_EDITING } from '@/server/config'
 import { Suspense, type ReactNode } from 'react'
 import { PiSymbol } from '../components/pi-symbol'
 import { GlobalLoading } from '@/components/layout/global-loading'
+import { connection } from 'next/server'
 
 export async function generateMetadata(): Promise<Metadata> {
     return {
@@ -29,11 +30,16 @@ export async function generateMetadata(): Promise<Metadata> {
     }
 }
 
+// Rendered per-request so every document, including Next's own not-found entry, is generated at
+// request time and can carry the CSP nonce on its inline hydration scripts. A `dynamic` export on
+// not-found.tsx does not achieve this: Next prerenders that entry regardless (OTTER-721).
 export default async function RootLayout({
     children,
 }: Readonly<{
     children: ReactNode
 }>) {
+    await connection()
+
     return (
         <html lang="en" translate="no" className={globalFont.className}>
             <body>
--- a/src/lib/csp.test.ts
+++ b/src/lib/csp.test.ts
@@ -1,0 +1,38 @@
+import { describe, it, expect } from 'vitest'
+import { CSP_HEADER, cspHeaderValue, generateNonce } from './csp'
+
+describe('csp', () => {
+    it('generates a distinct nonce each call', () => {
+        const nonces = new Set(Array.from({ length: 50 }, generateNonce))
+
+        expect(nonces.size).toBe(50)
+    })
+
+    it('generates base64 nonces with at least 128 bits of entropy', () => {
+        const nonce = generateNonce()
+
+        expect(nonce).toMatch(/^[A-Za-z0-9+/]+={0,2}$/)
+        expect(atob(nonce)).toHaveLength(16)
+    })
+
+    it('binds script-src to the supplied nonce', () => {
+        expect(cspHeaderValue('abc123')).toContain("script-src 'nonce-abc123'")
+    })
+
+    // Sentry Session Replay compresses in a blob-URL worker, which some browsers resolve through
+    // script-src; without this it would be blocked by 'strict-dynamic'.
+    it('allows blob workers so Sentry session replay keeps working', () => {
+        expect(cspHeaderValue('abc123')).toContain("worker-src 'self' blob:")
+    })
+
+    it('forbids plugin embeds', () => {
+        expect(cspHeaderValue('abc123')).toContain("object-src 'none'")
+    })
+
+    // The rollout ships report-only so a policy that is wrong about real Clerk cannot break signin.
+    // e2e runs under this policy but with Clerk faked, so real Clerk remains unverified until a
+    // deployed environment reports clean. This guards the flip to enforcing being deliberate.
+    it('is report-only until the policy is verified against real Clerk', () => {
+        expect(CSP_HEADER).toBe('Content-Security-Policy-Report-Only')
+    })
+})
--- a/src/lib/csp.test.ts
+++ b/src/lib/csp.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest'
-import { CSP_HEADER, cspHeaderValue, generateNonce } from './csp'
+import { CSP_HEADER, cspHeaderValue, cspReportUrl, generateNonce, reportingEndpointsValue } from './csp'
 
 describe('csp', () => {
     it('generates a distinct nonce each call', () => {
@@ -27,6 +27,29 @@ describe('csp', () => {
 
     it('forbids plugin embeds', () => {
         expect(cspHeaderValue('abc123')).toContain("object-src 'none'")
+    })
+
+    it('derives the Sentry security endpoint from the DSN', () => {
+        expect(cspReportUrl('https://key123@o484030.ingest.us.sentry.io/450700')).toBe(
+            'https://o484030.ingest.us.sentry.io/api/450700/security/?sentry_key=key123',
+        )
+    })
+
+    it('yields no report url for a missing or malformed DSN', () => {
+        expect(cspReportUrl('')).toBeUndefined()
+        expect(cspReportUrl('not-a-dsn')).toBeUndefined()
+        expect(cspReportUrl('https://o484030.ingest.us.sentry.io/450700')).toBeUndefined()
+    })
+
+    it('appends report directives only when given a destination', () => {
+        expect(cspHeaderValue('abc123', 'https://r.example/csp')).toContain(
+            'report-uri https://r.example/csp; report-to csp-report',
+        )
+        expect(cspHeaderValue('abc123')).not.toContain('report')
+    })
+
+    it('names the report endpoint the policy refers to', () => {
+        expect(reportingEndpointsValue('https://r.example/csp')).toBe('csp-report="https://r.example/csp"')
     })
 
     // The rollout ships report-only so a policy that is wrong about real Clerk cannot break signin.
--- a/src/lib/csp.ts
+++ b/src/lib/csp.ts
@@ -1,9 +1,11 @@
 // Nonce-based script-src, applied in production only.
 //
 // The static directives (frame-ancestors/form-action/base-uri) stay in next.config.ts because they
-// are request-independent. A nonce is per-request, so it has to be emitted from the proxy — and the
-// two must never emit the same header name, since browsers intersect duplicate CSP headers and the
-// combined policy fails closed.
+// are request-independent; a nonce is per-request, so script-src has to be emitted from the proxy.
+// The two use different header names because this policy ships report-only while the static one
+// already enforces — not because sharing a name would be unsafe. Duplicate CSP headers under one
+// name are enforced independently, which for disjoint directives like these changes nothing, so
+// after the flip to enforcing a shared name would also be fine.
 //
 // Gated on NODE_ENV=production because `next dev` compiles with eval-based source maps and would
 // need 'unsafe-eval'. `next start` sets NODE_ENV=production, so the e2e suite does run under this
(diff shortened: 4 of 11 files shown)


src/proxy.ts
```diff
@@ -53,6 +54,23 @@ function sanitizeRedirectParam(req: NextRequest): NextResponse | null {
...
+// the nonce cannot be missed on one branch. The anon-route return is the easy one to overlook, and
+// it serves /account/signin — the Clerk page that must not break.
+function continueWithNonce(req: NextRequest): NextResponse {
+    if (!isCspEnabled()) return NextResponse.next()
+
+    const nonce = generateNonce()
+    // Next reads the nonce off the request headers to stamp its own inline scripts; server
+    // components read the same header to hand it to ClerkProvider.
+    const headers = new Headers(req.headers)
+    headers.set(CSP_NONCE_HEADER, nonce)
```
therealmarv, line 86: This is the main blocker, and I think it's a small fix hiding behind a very reasonable assumption.

Next doesn't read the nonce from an arbitrary request header. In `next/dist/server/app-render/app-render.js`, `parseRequestHeaders()` (~line 167 on the 16.2.11 we have installed):

```js
const csp = headers['content-security-policy'] || headers['content-security-policy-report-only']
const nonce = typeof csp === 'string' ? getScriptNonceFromHeader(csp) : undefined
```

It parses `'nonce-{value}'` back out of the policy string itself. `x-csp-nonce` matches neither branch, so `nonce` comes out `undefined` and none of Next's inline scripts get stamped.

Next's CSP guide frames it as two headers doing two different jobs: the CSP header on the request is what Next parses, and the custom header exists purely so your own server components can read the value back. Here the custom header is being asked to do both.

Fix is to set both on the forwarded request:

```ts
headers.set(CSP_NONCE_HEADER, nonce)
headers.set('content-security-policy', cspHeaderValue(nonce))
```

Worth using the *enforcing* name on the request even while the response stays report-only. Request headers never reach the browser, so this is still fully report-only from the client's point of view. 16.2.11 does also accept the report-only name on the request, but that's undocumented and I wouldn't build on it. Same shape as the fix in vercel/next.js#49648.

Small thing while you're here: the comment on the two lines above says Next reads the nonce off the request headers and that server components read "the same header". After this fix those are genuinely two different headers, so it's worth rewording so the next person doesn't collapse them back into one.

src/app/dashboard/layout.tsx
```diff
@@ -25,7 +26,7 @@ export default async function DashboardLayout({ children }: Readonly<{ children:
     }
 
     return (
-        <ClerkProvider publishableKey={clerkPublishableKey}>
+        <ClerkProvider publishableKey={clerkPublishableKey} nonce={await getCspNonce()}>
```
therealmarv: Same applies to `focused-layout.tsx` and `user-layout.tsx`. Clerk throws this prop away.

In a server component, `ClerkProvider` from `@clerk/nextjs` resolves through the `react-server` condition in `dist/esm/package.json` (`#components` to `components.server.js`) to `app-router/server/ClerkProvider.js`. That component does:

```js
const { children, dynamic, ...rest } = props     // our nonce stays in `rest`
async function generateNonce() {
  if (!dynamic) return Promise.resolve('')
  ...
}
const propsWithEnvs = mergeNextClerkPropsWithEnv({ ...rest })   // nonce still intact here
React.createElement(ClientClerkProvider, {
  ...propsWithEnvs,
  nonce: await generateNonce(),    // overrides it, unconditionally
})
```

Because the explicit key comes after the spread, the caller's `nonce` never wins. With no `dynamic` prop at any of our three call sites, `generateNonce()` returns `''` every time, and `buildClerkJsScriptAttributes` skips the attribute on a falsy value. Clerk's docs match: `nonce` "requires the `dynamic` prop to also be set."

The part that surprised me: adding `dynamic` alone still won't deliver *our* value. With `dynamic` set, `getNonceHeaders()` reads the `X-Nonce` request header and falls back to parsing the enforcing `Content-Security-Policy` request header. Neither is `x-csp-nonce`, and the prop is still overridden. So the proxy would also need to emit `X-Nonce`.

Good news is the impact is smaller than it looks. `loadScript` builds the tag with `document.createElement('script')` and `document.body.appendChild`, which under `strict-dynamic` inherits trust from the injecting script with no nonce of its own. That's exactly what you predicted in the `csp.ts` header comment, and it's right. So this shouldn't break auth on the flip, provided the Next-side fix lands so the bootstrap is actually nonced.

Given that, my preference would be to just drop the `nonce` prop from all three layouts and let `strict-dynamic` do its job, rather than adding `dynamic` plus `X-Nonce` plumbing for something Clerk doesn't need. A prop that looks like it works but is silently discarded is worse than no prop. If you do drop it, `getCspNonce()` loses its only caller and can go too.

Also worth a look either way: the installed Clerk ships `clerkMiddleware({ contentSecurityPolicy: { strict: true } })`, which generates the nonce and emits `X-Nonce` itself. Might not fit given the static directives live in `next.config.ts`, but probably worth a few minutes before hand-rolling.

src/lib/csp.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function cspHeaderValue(nonce: string): string {
+    return [
+        // 'strict-dynamic' lets Next's nonce-carrying bootstrap scripts load the chunks they pull in
+        // (including Clerk's and Sentry's), which a host allowlist cannot express reliably.
+        `script-src 'nonce-${nonce}' 'strict-dynamic' https: 'unsafe-inline'`,
+        // Sentry Session Replay compresses in a worker created from a blob URL. Some browsers
+        // resolve worker-src through script-src, which 'strict-dynamic' would otherwise block.
+        "worker-src 'self' blob:",
+        "object-src 'none'",
```
therealmarv, line 66: There's no `report-uri` or `report-to` here, no `Reporting-Endpoints` response header anywhere, and no `securitypolicyviolation` listener in the app. So right now violations only show up in a console someone happens to have open.

That matters more than it sounds, because the whole point of the report-only phase is to answer "is this policy right about real Clerk and Sentry" and then flip. Without a collector there's no evidence to gate that decision on, and the flip ends up being a judgment call instead of a measurement.

I'd add a reporting destination in this PR rather than as a follow-up. Sentry can ingest CSP reports, so this is probably a `report-to` plus a `Reporting-Endpoints` header and not much else.

src/app/layout.tsx
```diff
@@ -29,11 +30,16 @@ export async function generateMetadata(): Promise<Metadata> {
...
 
+// Rendered per-request so every document, including Next's own not-found entry, is generated at
+// request time and can carry the CSP nonce on its inline hydration scripts. A `dynamic` export on
+// not-found.tsx does not achieve this: Next prerenders that entry regardless (OTTER-721).
 export default async function RootLayout({
     children,
 }: Readonly<{
     children: ReactNode
 }>) {
+    await connection()
```
therealmarv, line 43: Not blocking, and I think it's the right call, but the comment undersells the blast radius.

`connection()` in the *root* layout opts the entire app out of static generation, not just Next's not-found entry. Every route that was previously static, `/about` being the obvious one, now renders per request. Most of our routes are already dynamic via the proxy and Clerk so the real-world delta is probably small, but it's an app-wide change being made to fix one internal page.

Worth saying that explicitly in the comment so nobody later reads "rendered per-request" as scoped to not-found and gets surprised by the perf profile.

src/lib/csp.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    })
+
+    it('generates base64 nonces with at least 128 bits of entropy', () => {
+        const nonce = generateNonce()
+
+        expect(nonce).toMatch(/^[A-Za-z0-9+/]+={0,2}$/)
+        expect(atob(nonce)).toHaveLength(16)
+    })
+
+    it('binds script-src to the supplied nonce', () => {
```
therealmarv, line 18: These are good tests for what they cover, and the report-only guard at the bottom is a nice touch for making the flip deliberate.

The gap is that every assertion here is about the policy *string*, so none of them can fail when the nonce doesn't reach the renderer, which is the actual bug. The string was correct the whole time.

Once the proxy fix lands, something that asserts the wiring would be worth having. Even a small test over `continueWithNonce` checking that the forwarded request headers carry a policy whose `script-src` nonce matches the one in the response header would have caught this, and it's cheap.

therealmarv (changes requested):
First off, the security reasoning in here is genuinely good and I don't want that buried under the findings below.

Choosing `strict-dynamic` over a host allowlist for the right stated reason, catching that Sentry Replay's blob worker gets resolved through `script-src` in some browsers, noticing that two policies under one header name get intersected, shipping report-only instead of flipping straight to enforcing: that's a careful read of a topic most people get wrong. Routing both terminal `NextResponse.next()` paths through a single helper so the anon-route branch can't be quietly forgotten is exactly the right instinct, and it's the kind of thing that's obvious only after someone ships the bug. Writing the `global-error` gap into the file, with the specific consequence spelled out (Sentry stops reporting from the one page whose job is reporting), is better than a ticket nobody reads.

The `csp.ts` header comment also correctly predicts that Clerk's script inherits trust through `strict-dynamic` without a nonce of its own. I verified that against the Clerk source and it's right, which is a subtle call.

So the issues below are two integration details, not the thinking.

## I think the nonce never reaches the page

Two separate links in the chain look severed, and between them no script in the document ends up carrying the nonce:

1. **Next never sees a CSP header on the request**, so it doesn't stamp its own inline scripts. Details in the `src/proxy.ts` comment.
2. **Clerk overrides the `nonce` prop** with `''` when `dynamic` isn't set, so `getCspNonce()` is discarded at all three call sites. Details in the `dashboard/layout.tsx` comment.

**This contradicts your verification note**, which says the 404 page went from 8 un-nonced inline scripts to 7 of 7 carrying the nonce. I can't reconcile that with what I'm reading in `parseRequestHeaders`, so one of us is wrong and I'd rather establish which before you spend time on a fix that might not be needed.

Could you re-run that check and paste the actual `<script>` tags from the HTML? Something like a production build, load any page, and grep the response body for `nonce=`. If they're there, I'm wrong and I'd really like to understand the mechanism, because I couldn't find a second path by which the value gets in. If they're not, the fix in the proxy comment is a couple of lines.

Worth noting my read is source-reading only. I haven't run the build, and you have, which is why I'm asking rather than asserting.

## Blocking

- The request header name in `src/proxy.ts`
- The discarded `nonce` prop in the three layouts, either wired properly or removed
- A reporting destination, since report-only with nowhere to report gives us nothing to gate the flip on

## Non-blocking

- `await connection()` in the root layout is app-wide, not scoped to not-found. Fine by me, but the comment should say so.
- The proxy matcher excludes `.html`, so a request like `/missing.html` still gets an App Router HTML 404 without ever passing through `continueWithNonce()`. Minor, since no header means nothing to violate, but it does undercut the "every path that renders a document goes through here" comment.
- The `csp.ts` header comment says two policies under one header name "fail closed". Duplicate CSP headers are enforced independently and intersected per directive, so two policies with disjoint directives (which is our case) are fine. The separate header name is still the right design, just not for that reason. Worth fixing since it'll matter when someone reasons about the flip.

Happy to pair on the proxy bit if it's useful, and genuinely, nice work on the hard parts of this.

src/proxy.ts
nathanstitt: Re-ran the verification as requested (production build via `E2E_FAKE_CLERK` like the e2e suite — Clerk faked, but nonce stamping is Next's own path and unaffected). The result is the interesting part: **we were both right.** The nonce genuinely was reaching the page before any fix, contradicting your `parseRequestHeaders` read — but your read of `parseRequestHeaders` is also correct. Pre-fix, `GET /account/signin`:

```
content-security-policy-report-only: script-src 'nonce-aNaBTuDpMvTHH1t6jr4BaA==' 'strict-dynamic' https: 'unsafe-inline'; worker-src 'self' blob:; object-src 'none'
```

```html
<script nonce="aNaBTuDpMvTHH1t6jr4BaA==">            <!-- x8, every inline script -->
<script src="/assets/_next/static/chunks/turbopack-0sh3v01wx9crh.js" async="" integrity="sha256-L3+..." nonce="aNaBTuDpMvTHH1t6jr4BaA==">
```

The second path you couldn't find is in `next/dist/server/lib/router-utils/resolve-routes.js` (~line 455): after middleware runs, Next copies every middleware **response** header onto the request before routing —

```js
if (value) {
    resHeaders[key] = value;
    req.headers[key] = value;   // <-- response header lands on the request too
}
```

— so the `Content-Security-Policy-Report-Only` response header set in `continueWithNonce` arrives at `parseRequestHeaders` as a request header, and its report-only branch parses the nonce out. That's why the original verification note was accurate.

But I agree completely that this is nothing to build on: it's an undocumented side effect of the Node router process, even less contractual than the report-only-name-on-request behavior you flagged. So the fix landed exactly as you proposed — both headers set on the forwarded request, enforcing name on the CSP one, comment reworded so the two headers' two jobs can't be collapsed back into one. Post-fix, same page:

```
content-security-policy-report-only: script-src 'nonce-Fw+dAN/n2TUrV60M1rAWcw==' 'strict-dynamic' https: 'unsafe-inline'; worker-src 'self' blob:; object-src 'none'; report-uri https://o484030.ingest.us.sentry.io/api/.../security/?sentry_key=...; report-to csp-report
reporting-endpoints: csp-report="https://o484030.ingest.us.sentry.io/api/.../security/?sentry_key=..."
```

```html
<script nonce="Fw+dAN/n2TUrV60M1rAWcw==">            <!-- x8 inline on signin; 7/7 on the 404 page -->
```

Zero un-nonced script tags on either page, fresh nonce per request.

src/app/dashboard/layout.tsx
nathanstitt: Verified your read against the installed Clerk and it's right — the explicit `nonce: await generateNonce()` after the spread means the caller's prop can never win, and without `dynamic` it's always `''`. Went with your preference: the `nonce` prop is dropped from all three layouts and `getCspNonce()` is gone with its last caller. A prop that looks load-bearing but is silently discarded is exactly the kind of thing that misleads the next reader, and `strict-dynamic` covers `clerk.browser.js` as you confirmed.

On `clerkMiddleware({ contentSecurityPolicy: { strict: true } })`: looked at it, decided against. It moves policy authorship into Clerk (its own host allowlist plus its nonce), while we'd still be hand-rolling `worker-src` for Sentry Replay, the report-to wiring, and the report-only/enforcing flip on our side — two owners for one policy seems worse than one. Happy to revisit if we ever want Clerk to own the whole thing.

src/lib/csp.ts
nathanstitt: Agreed — report-only with no collector is a judgment call dressed up as a measurement. Added in this PR:

- `report-uri <sentry security endpoint>` (legacy fallback) and `report-to csp-report` directives on the policy
- a `Reporting-Endpoints: csp-report="<sentry security endpoint>"` response header

The endpoint is derived from the Sentry DSN at runtime (`cspReportUrl()` in `src/lib/csp.ts`): `https://<key>@<host>/<project>` → `https://<host>/api/<project>/security/?sentry_key=<key>`. No DSN (local builds) means the directives are simply omitted. Verified against a local production build with a DSN set:

```
reporting-endpoints: csp-report="https://o484030.ingest.us.sentry.io/api/<project>/security/?sentry_key=<key>"
content-security-policy-report-only: ...; report-uri https://o484030.ingest.us.sentry.io/api/<project>/security/?sentry_key=<key>; report-to csp-report
```

So the flip can now be gated on the Sentry security feed reporting clean in a deployed environment.

src/app/layout.tsx
nathanstitt: Reworded to state the blast radius outright:

```
// `connection()` in the ROOT layout opts the entire app out of static generation — every route,
// previously-static ones like /about included, now renders per-request, not just Next's not-found
// entry. That app-wide cost is deliberate: it is what lets every document carry the CSP nonce on
// its inline hydration scripts.
```

src/lib/csp.test.ts
nathanstitt: Added it in `src/proxy.test.ts` — `continueWithNonce` is now exported and the test asserts on the forwarded request headers (`x-middleware-request-content-security-policy`, which is exactly what Next's server consumes): the forwarded policy must carry a `script-src 'nonce-...'` matching the nonce in the response's report-only header, and `x-middleware-request-x-csp-nonce` must match too. Run against the pre-fix proxy it fails with the forwarded CSP header missing entirely — it's the test that would have caught this. A second test covers the new reporting wiring (`Reporting-Endpoints` + `report-to` when a DSN is configured), plus string-level tests in `csp.test.ts` for the DSN-to-security-endpoint derivation.

nathanstitt:
Thanks for the depth of this review — the `parseRequestHeaders` catch in particular. All three blockers are addressed in 6b510e95:

- **Request header name**: the proxy now forwards `content-security-policy` (enforcing name) on the request alongside `x-csp-nonce`, with the comment reworded so the two headers' distinct jobs stay distinct. Re-verification resolved the contradiction you flagged: the nonce *was* reaching the page, but only via an undocumented Next behavior — the router copies middleware response headers onto the request (`resolve-routes.js`), so `parseRequestHeaders` was finding our report-only *response* header on the request. Your source reading was right; the mechanism was just a third path neither of us would want to build on. Full before/after `<script>` evidence in the proxy thread.
- **Clerk nonce prop**: dropped from all three layouts, `getCspNonce()` removed with it. `strict-dynamic` carries `clerk.browser.js`, as you verified.
- **Reporting destination**: `report-uri`/`report-to` plus a `Reporting-Endpoints` header pointing at Sentry's security endpoint, derived from the DSN. The flip can now be gated on the Sentry feed.

Non-blocking items are in too: the root-layout comment now states the app-wide static-generation opt-out, the "fail closed" claim in `csp.ts` is corrected (duplicate headers enforce independently; the separate name is about report-only vs. enforcing, and the same correction landed in `next.config.ts`), the "every path" proxy comment now acknowledges the `.html` matcher exclusion, and there's a wiring test over `continueWithNonce` asserting the forwarded request headers — the one that fails against the pre-fix proxy.

Verified post-fix on a production build: 8/8 inline scripts nonced on `/account/signin`, 7/7 on the 404 page, zero un-nonced script tags, fresh nonce per request, reporting headers present. Full unit suite (253 files / 2420 tests) and `pnpm checks` green.

therealmarv (approved):
approved! Thanks.

PR #942, opened 2026-08-05
OTTER-724: await index creation in foundational migrations

nathanstitt:
Finding MA-13.

Nine `.execute()` calls across five early migrations were never awaited. Kysely runs each migration in its own transaction, so these floating promises raced the commit and were not guaranteed to apply:

- `1727370622501_user.ts` — unique index on `user.clerk_id`
- `1727379622503_member_users.ts` — both `member_user` indexes
- `1728503375785_study-jobs.ts` — `study_job` and the two `job_status_change` indexes
- `1750186286777_study_job_file.ts` — two `study_job` column drops
- `1758569988001_split_orgs_into_labs_and_enclaves.ts` — `orgUser` delete inside the loop

The affected indexes were verified present in the dev database, so this is latent risk for a fresh database build rather than active breakage. **No corrective migration is included.** An AST scan of the whole migrations directory confirms zero floating promises remain.

### smart_concat finding

`1742581222801_smart_concat.ts:26` has a `dropColumn(full_name)` with no `.execute()` at all. It is in `down()`, not `up()`, and is **deliberately left as-is** — a code comment now explains why. Adding `.execute()` would be wrong regardless: the preceding `DROP FUNCTION ... RETURNS text` is invalid Postgres and fails first, and the drop order is backwards since the generated column depends on `smart_concat()`. The forward migration is correct and `full_name` is relied on by `1774000000000_add_pi_user_id_to_study`.

### DIFF
--- a/src/database/migrations/1742581222801_smart_concat.ts
+++ b/src/database/migrations/1742581222801_smart_concat.ts
@@ -21,6 +21,13 @@ $$;
     )
 }
 
+// Left deliberately un-awaited and un-executed (OTTER-724 / MA-13). The builder below is never
+// run, so `user.full_name` survives a rollback. Adding `.execute()` would be wrong on two counts:
+// the DROP FUNCTION above uses invalid syntax (`RETURNS text` is not accepted by Postgres) and so
+// fails first, and the drop order is backwards anyway -- the generated column depends on
+// smart_concat(), so the column must go before the function. Fixing this rollback path is out of
+// scope here; the forward migration is correct and `full_name` is relied on by
+// 1774000000000_add_pi_user_id_to_study.
 export async function down(db: Kysely<unknown>): Promise<void> {
     await sql`DROP FUNCTION smart_concat(a text, b text) RETURNS text;`.execute(db)
     db.schema.alterTable('user').dropColumn('full_name')
--- a/src/database/migrations/1727370622501_user.ts
+++ b/src/database/migrations/1727370622501_user.ts
@@ -21,7 +21,7 @@ export async function up(db: Kysely<unknown>): Promise<void> {
         .addColumn('updated_at', 'timestamp', (col) => col.defaultTo(sql`now()`).notNull())
         .execute()
 
-    db.schema.createIndex('user_clerk_id_indx').on('user').column('clerk_id').unique().execute()
+    await db.schema.createIndex('user_clerk_id_indx').on('user').column('clerk_id').unique().execute()
 }
 
 export async function down(db: Kysely<unknown>): Promise<void> {
--- a/src/database/migrations/1727379622503_member_users.ts
+++ b/src/database/migrations/1727379622503_member_users.ts
@@ -12,14 +12,14 @@ export async function up(db: Kysely<unknown>): Promise<void> {
         .addColumn('is_admin', 'boolean', (col) => col.notNull())
         .execute()
 
-    db.schema
+    await db.schema
         .createIndex('member_user_mbr_usrid_indx')
         .on('member_user')
         .columns(['member_id', 'user_id'])
         .unique()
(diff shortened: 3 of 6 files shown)


src/database/migrations/1742581222801_smart_concat.ts
```diff
@@ -21,6 +21,13 @@ $$;
     )
 }
 
+// Left deliberately un-awaited and un-executed (OTTER-724 / MA-13). The builder below is never
+// run, so `user.full_name` survives a rollback. Adding `.execute()` would be wrong on two counts:
+// the DROP FUNCTION above uses invalid syntax (`RETURNS text` is not accepted by Postgres) and so
+// fails first, and the drop order is backwards anyway -- the generated column depends on
+// smart_concat(), so the column must go before the function. Fixing this rollback path is out of
+// scope here; the forward migration is correct and `full_name` is relied on by
+// 1774000000000_add_pi_user_id_to_study.
```
therealmarv, line 30: Non-blocking, but this attributes the behavior to the wrong cause.

`down()` throws on the next line first: `DROP FUNCTION smart_concat(a text, b text) RETURNS text` is not valid Postgres. Since `migrate:down` runs inside a transaction, that abort rolls the whole step back, so `user.full_name` survives because the DROP failed, not because the builder below was left unexecuted. Giving that builder an `.execute()` would not change the outcome at all.

The comment does note the DROP "fails first" two sentences later, so as written it contradicts its own opening claim. Suggest splitting it into two independent facts:

1. `down()` fails wholesale at the invalid `DROP FUNCTION`, so any rollback reaching this migration is broken and a no-op.
2. Separately, the `dropColumn` builder is dormant, and its ordering is backwards anyway since the generated column depends on `smart_concat()`.

Worth tightening because this comment is the only substantive change to this file in the PR, so it becomes the durable record of the "why".

therealmarv (approved):
Approved, nothing blocking here.

All nine added `await`s are correct. Each one anchors an `.execute()` that was already present, in an `up()` path whose statements were meant to run sequentially. Adding `await` cannot introduce a statement, reorder a working one, or change the SQL that gets sent, so the change class is strictly narrowing. CI is green.

The diff accounting is also clean and matches the description exactly: nine `await`s across five migrations plus a seven-line comment gives +16, and the nine rewritten lines give -9, across 6 files. No drive-by edits hiding in the diff.

Two non-blocking notes, both about the prose rather than the behavior.

### Transaction boundary in the description

The body says Kysely "runs each migration in its own transaction." With the installed Kysely 0.28.17, `Migrator.#runMigrations` wraps the *entire* run in a single transaction when `supportsTransactionalDdl && !disableTransactions` (`kysely/dist/esm/migration/migrator.js:435-440`). `PostgresAdapter.supportsTransactionalDdl` returns `true`, and `kysely.config.ts` sets only `migrationFolder` and `allowUnorderedMigrations`, so every pending migration shares one transaction per `kysely migrate:latest` invocation.

That's worth correcting because it also changes the stated failure mode. With one transaction on one pg connection, these queries were queued in order and in practice did apply, which is exactly why the indexes were found present in dev. The defect the awaits actually fix is narrower but still real: errors from those statements became unhandled rejections instead of failing the migration, and correct ordering was incidental to pg's client-side queueing rather than guaranteed. Suggest rewording "raced the commit and were not guaranteed to apply" along those lines.

Minor: the `1742581222801_smart_concat.ts:26` reference in the description is now stale. The dormant builder sits at line 33 after the comment was added.

### Follow-up card for the broken `down()`

Agreed that fixing it is out of scope. But as it stands, a hard-failing rollback path is documented only in a code comment, which is where this kind of thing goes to get rediscovered by whoever actually needs the rollback. Worth a ticket.

One inline note on the comment wording.

PR #943, opened 2026-08-05
OTTER-724: restrict data-source URLs to http(s) schemes

nathanstitt:
https://openstax.atlassian.net/browse/OTTER-724 (finding MA-12)

Org data-source URLs were validated with a bare `z.url()`, which accepts any parseable URI — `javascript:alert(1)`, `data:text/html,...` and `vbscript:` all passed. The stored value is rendered directly into an `<Anchor href>`, and `next.config.ts` sets only frame-ancestors/form-action/base-uri (no script-src or default-src), so there is no CSP backstop. That makes this admin-plantable stored XSS against anyone viewing the org settings page.

Changes:
- Lift the existing `httpUrl` / `httpUrlOptionalItem` helpers from `src/schema/researcher-profile.ts` into a shared `src/schema/url.ts` (new `isHttpUrl` predicate) rather than duplicating the logic.
- Apply the scheme check to both offending spots in `data-sources.schema.ts`: the URL row schema and the draft-URL superRefine. The same zod schema runs client and server side, so both paths are covered.
- Render-time guard in `data-sources-view.tsx`: rows persisted before this change still hold arbitrary schemes, so non-http(s) values render as plain text instead of a link.
- Schema-level tests covering the three payloads plus ordinary http/https, and an action-level test asserting nothing is written.

### DIFF
--- a/src/app/[orgSlug]/admin/settings/data-sources-view.tsx
+++ b/src/app/[orgSlug]/admin/settings/data-sources-view.tsx
@@ -3,6 +3,7 @@
 import type { ReactNode } from 'react'
 import { Anchor, Box, Button, Divider, Group, Paper, Stack, Text, Title } from '@mantine/core'
 import { PlusCircleIcon } from '@phosphor-icons/react/dist/ssr'
+import { isHttpUrl } from '@/schema/url'
 
 // Presentational pieces for the "Data Sources" settings card. They own the card chrome
 // and the visible row (name / code-env names / description / linked URLs) but NOT data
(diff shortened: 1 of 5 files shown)


src/schema/url.ts
```diff
@@ -0,0 +1,28 @@
+import { z } from 'zod'
+
+// `z.url()` accepts any parseable URI, including `javascript:`, `data:` and `vbscript:`.
+// Those reach an `href` unsanitized and execute, so every user-supplied URL that we later
+// render as a link must be restricted to the two schemes we actually intend to support.
+export const isHttpUrl = (value: string) => value.startsWith('http://') || value.startsWith('https://')
```
therealmarv, line 6: Non-blocking, but worth a follow-up: URI schemes are case-insensitive (RFC 3986) while `startsWith` is not, so `HTTPS://example.com/data` parses fine under `z.url()` but fails `isHttpUrl`.

This was a latent quirk in the researcher-profile helpers, but the branch newly applies the predicate to data-source create/update **and** to render of already-stored rows, so an existing legitimate row saved with an uppercase scheme would silently lose its link and fall through to the plain-text branch.

```ts
export const isHttpUrl = (value: string) => {
    const scheme = value.toLowerCase()
    return scheme.startsWith('http://') || scheme.startsWith('https://')
}
```

src/app/[orgSlug]/admin/settings/data-sources-view.tsx
```diff
@@ -25,7 +26,35 @@ export type DataSourceRowViewProps = {
     actions: ReactNode
 }
 
+// Defense in depth for rows persisted before the schema restricted schemes (OTTER-724): a
+// stored `javascript:` URL still executes when React renders it into an href, so refuse to
+// link anything that is not http(s) and show the raw text instead.
+function DataSourceUrlLink({ url, description }: { url: string; description: string | null }) {
+    const linkable = isHttpUrl(url)
```
Comment (therealmarv) [inline src/app/[orgSlug]/admin/settings/data-sources-view.tsx:33]: Also non-blocking: this render guard is the only protection for values already persisted before the schema was tightened, but the new tests only cover write-time rejection, so the non-linkable branch never executes in the suite. That makes it the piece most likely to get dropped by a future refactor without anything failing.

`data-sources.test.tsx` already renders through the container against real DB fixtures, so a case that inserts a data source with a `javascript:alert(1)` URL and asserts `queryByRole('link')` is null while the raw text is visible would slot in next to the existing tests.

Minor style point while you're here: the `linkable ? <Anchor> : <Text>` ternary inside the return brushes against the "no ternaries in JSX" rule in `CLAUDE.md`. An early `if (!isHttpUrl(url))` returning the plain-text row would sidestep it.

therealmarv (approved):
Approving, nothing blocking.

The fix does what it claims: `javascript:`, `data:` and `vbscript:` values can no longer be persisted through either create or update, the same zod schema enforces it on both the client and the server, and the render-time guard covers rows written before the tightening. Lifting the existing `httpUrl` / `httpUrlOptionalItem` helpers into `src/schema/url.ts` instead of duplicating them is the right call, since it leaves one definition of "http(s) only" rather than two that can drift apart.

Two low-severity notes inline, neither of which weakens the security fix:

1. `isHttpUrl` is case-sensitive, so a valid uppercase-scheme URL is rejected on write and de-linked on render.
2. The render-time guard, which is the only defense for already-stored values, has no regression test.

Both are cheap enough to fold in here, but a follow-up is fine too.

PR #944, opened 2026-08-05
OTTER-724: fix org route matchers and enforce org-admin authorization server-side

nathanstitt:
https://openstax.atlassian.net/browse/OTTER-724 (finding MA-11)

Two related defects.

**Dead route matchers.** `proxy.ts` declared the org matchers with the Next.js filesystem convention (`/[orgSlug]/admin/(.*)`), but `createRouteMatcher` compiles with path-to-regexp, which treats the brackets as **literal characters**. `/acme/admin/team` did not match. It failed silently — Clerk's bundled path-to-regexp fork does not throw where standalone v8 does — so there was no crash and no log, and both org guards were dead code while `isAdmin` was computed correctly and never consulted.

The org-admin matcher becomes `/:orgSlug/admin/(.*)`. The `/[orgSlug]` matcher is dropped rather than converted: a single `/:orgSlug` segment also matches `/dashboard`, `/about`, and every other top-level route, so it cannot gate membership. `extractOrgSlugFromPath` already returns null for non-org prefixes and is what actually distinguishes the two, so the membership check now gates on it directly.

**Middleware is not a sufficient authorization boundary.** The org-admin pages were guarded only by `RequireOrgAdmin`, a client component redirecting in `useLayoutEffect` — after the server rendered and streamed the admin payload. A non-admin member requesting `/{org}/admin/settings` received the content before the bounce. `[orgSlug]/layout.tsx` was a bare `// TODO: check user is a member of the org`.

Adds `src/app/[orgSlug]/admin/layout.tsx` so authorization resolves before rendering, and fills in the membership check in the parent layout. Both use the existing `sessionFromClerk` pattern and `Routes.dashboard`. SI admins retain cross-org access, matching the prior client guard and `reviewer-page-guard`.

`RequireOrgAdmin` is removed — it had no remaining callers, the server guard applies the same condition and target, and leaving it around invites reuse as if it were an authz boundary.

**Tests:** `src/proxy.matchers.test.ts` exercises the real (unmocked) `createRouteMatcher` and asserts the bracket form matches nothing, so it cannot silently return. Layout guard tests cover non-admin member, wrong-org admin, org admin, and SI admin. Verified the new guard tests fail when the redirect is removed.

No route definitions changed.


src/proxy.ts
```diff
@@ -111,12 +114,17 @@ export const proxy = clerkMiddleware(async (auth, req) => {
...
     }
 
-    if (isOrgAdminRoute(req) && !isAdmin && currentOrgSlug) {
-        return redirectToDashboard(req, 'org-admin', session)
-    }
+    // extractOrgSlugFromPath already returns null for every non-org top-level prefix, so it —
+    // not a route matcher — is what distinguishes `/acme/...` from `/dashboard`. A `/:orgSlug/(.*)`
+    // matcher would match those too and is therefore not a usable guard on its own.
+    if (currentOrgSlug) {
+        if (!session.orgs[currentOrgSlug] && !session.user.isSiAdmin) {
```
therealmarv, line 122: Blocking, one-line fix.

`extractOrgSlugFromPath('/404')` returns `'404'`, because `NON_ORG_PREFIXES` (`src/lib/paths.ts:74`) doesn't list it. So this check reads `session.orgs['404']`, finds nothing, and redirects to the dashboard.

`Routes.notFound` is `/404`, so the not-found page is now unreachable for anyone but SI admins. `[orgSlug]/dashboard/page.tsx:19` and `admin/settings/page.tsx:17` both redirect there and will land on the dashboard instead. It only surfaces now because the old `/[orgSlug]` matcher meant this check never ran.

Fix: add `'404'` to `NON_ORG_PREFIXES`. Two small things while you're in that file:

- the list has `error-demo`, but `ANON_ROUTES` and `Routes.editorDemo` say `editor-demo`, so that entry has never matched
- `extractOrgSlugFromPath` has no tests and is now the only org/non-org gate; a table test (`/404`, `/dashboard`, `/acme`, `/`) would have caught this

Side note: the same classification sends any unknown top-level path to the dashboard rather than `not-found.tsx`, which hides whether an org slug exists. Fine if intended, worth confirming.

src/proxy.ts
```diff
@@ -11,8 +11,11 @@ import { extractOrgSlugFromPath } from '@/lib/paths'
 import * as Sentry from '@sentry/nextjs'
 
 const isSIAdminRoute = createRouteMatcher(['/admin/safeinsights(.*)'])
-const isOrgAdminRoute = createRouteMatcher(['/[orgSlug]/admin/(.*)'])
-const isOrgRoute = createRouteMatcher(['/[orgSlug]'])
+// createRouteMatcher compiles with path-to-regexp, NOT the Next.js filesystem convention.
+// `/[orgSlug]/admin/(.*)` looks right but treats the brackets as literal characters, so it
+// matched nothing and failed silently — Clerk's bundled path-to-regexp fork does not throw
+// where the standalone v8 does. The colon form is the one that actually captures a segment.
+const isOrgAdminRoute = createRouteMatcher(['/:orgSlug/admin/(.*)'])
```
therealmarv: Non-blocking, two notes.

`/:orgSlug/admin/(.*)` needs a segment after `admin/`, so bare `/acme/admin` skips the org-admin branch below. Latent today (no admin index page, and the admin layout guards it), but it goes quiet again if someone adds one. `/:orgSlug/admin(.*)` covers both.

Also `src/lib/clerk-fake/route-matcher.ts` doesn't handle `:param`, so under the e2e fake this pattern compiles to a literal `:orgSlug` and the org-admin check never fires. Not a prod issue, but that file's comment and tests still assert the bracket form this PR removes. Fine as a follow-up.

src/app/[orgSlug]/layout.tsx
```diff
@@ -1,10 +1,22 @@
...
+// Middleware runs before rendering but is not the authorization boundary: it can be bypassed by
+// direct RSC/segment requests and it only sees the URL. Re-check membership here so a non-member
+// never gets a rendered payload for another org's pages.
+export default async function OrgLayout({
+    children,
+    params,
+}: Readonly<{ children: ReactNode; params: Promise<{ orgSlug: string }> }>) {
+    const { orgSlug } = await params
+    const session = await sessionFromClerk()
+
```
Comment (therealmarv) [inline src/app/[orgSlug]/layout.tsx:16]: Not blocking: `sessionFromClerk` isn't wrapped in `cache()`, so an admin page marshals the session three times per request (middleware, here, admin layout), each doing a `select from user`. Worth a `cache()` if it ever shows up in traces.

src/proxy.matchers.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    // Regression guard for MA-11: the Next.js filesystem `[param]` convention is NOT path-to-regexp
+    // syntax. Clerk's bundled fork treats the brackets as literal characters and silently matches
+    // nothing rather than throwing, so this dead guard is invisible without an explicit assertion.
+    it('confirms the bracket convention matches nothing, which is why it must never come back', async () => {
+        const createRouteMatcher = await realCreateRouteMatcher()
+        const bracketMatcher = createRouteMatcher(['/[orgSlug]/admin/(.*)'])
+
+        expect(bracketMatcher(req('/acme/admin/team'))).toBe(false)
+        expect(bracketMatcher(req('/acme/admin/settings'))).toBe(false)
+    })
```
therealmarv, line 43: Favorite part of the PR: asserting that the broken bracket form stays broken is stronger than only testing the fixed form, and using `importActual` for the real matcher instead of the global mock is the right call.

therealmarv (changes requested):
Nice catch, and the fix is in the right place: a server layout holds even for direct RSC requests that skip middleware, which a matcher fix alone wouldn't cover. Dropping `/[orgSlug]` rather than converting it is the right read, and removing `RequireOrgAdmin` beats leaving it around to be reused as an authz boundary.

One blocking item, one line. The membership guard is live for the first time, and `NON_ORG_PREFIXES` (`src/lib/paths.ts:74`) has no `404`, so `/404` now redirects to the dashboard for everyone except SI admins. Details on `proxy.ts`.

Non-blocking: bare `/:orgSlug/admin` isn't matched, the e2e Clerk fake can't compile `:param`, `extractOrgSlugFromPath` has no tests, `sessionFromClerk` isn't cached.

No authz bypass left as far as I can tell, so MA-11 looks handled. Just want the 404 fix before it lands.

src/proxy.ts
nathanstitt: Fixed in c382b8e6: added `404` to `NON_ORG_PREFIXES` (with a comment tying it to `Routes.notFound`), corrected the stale `error-demo` entry to `editor-demo`, and added a table test in `src/lib/paths.test.ts` covering `/404`, `/dashboard`, `/acme`, `/`, and the other reserved prefixes. While in there, `extractOrgSlugFromPath` now returns `null` instead of `""` for `/` (both callers only check truthiness, so no behavior change). On the side note: yes, sending unknown top-level paths to the dashboard is intended — it avoids revealing whether a given org slug exists.

src/proxy.ts
nathanstitt: Changed to `/:orgSlug/admin(.*)` and updated the matcher tests to assert bare `/acme/admin` is gated. Agreed on the clerk-fake `route-matcher.ts` `:param` limitation — leaving that (and its comment/tests still referencing the bracket form) as a follow-up rather than growing this PR.

src/app/[orgSlug]/layout.tsx
Comment (nathanstitt) [reply src/app/[orgSlug]/layout.tsx:16]: Done — `sessionFromClerk` is now wrapped in React `cache()`, so the org layout, admin layout, and page marshal the session once per render. The middleware is unaffected since it calls `marshalSession` directly, and the `forceUpdate` callers pass a fresh options object whose identity never hits the zero-arg cache entry, so a forced re-sync always executes.

src/proxy.matchers.test.ts
nathanstitt: Thanks! Asserting the broken form stays broken felt like the only way to make that failure mode visible — glad it reads that way. Those tests now also cover the widened `admin(.*)` pattern.

therealmarv (approved):
approved (if the CI says 🟢 again).

PR #945, opened 2026-08-05
rich text editor modifications

jbwilson8:
Issue: [OTTER-463](https://openstax.atlassian.net/browse/OTTER-463).

This PR closes out the two remaining open comments. The original format controls (bold/italic/underline/link/bullet/numbered) were verified on QA and staging back in March, and indent/outdent landed in 93de454 — the rest of the ticket was already done.

| Comment | Resolution |
| --- | --- |
| Malar(Apr 8): "Could we make the `hyperlink` open the linked url in a new tab?" | Fixed here |
| Greg (Aug 3): "Request: remove double spacing between paragraphs" | Fixed here |

### Root causes

**Links opened in the same tab.** All three `LinkPlugin` mounts passed only `validateUrl`. In `@lexical/link`, `rel` falls back to `'noreferrer'` but `target` stays `undefined`, so `LinkNode.__target` remained `null` and `createDOM` emitted no `target` attribute. Separately, `ClickableLinkPlugin` was not mounted anywhere, so read-only views (`editable: false`) fell through to raw anchor navigation — that's the "opened from a submitted proposal" half of the report.

**Double paragraph spacing.** `lexicalTheme` had no `paragraph` key and no CSS in the repo targets `p`. Mantine's `baseline.css` resets only `body` margin, and there's no `TypographyStylesProvider`, so every `<p>` Lexical emits inherited the UA default `margin-block: 1em` (~16px). Against Mantine's ~24.8px line height that gap reads as a full blank line.

### Changes

- **`config.ts`** — added a module-level `linkAttributes` (`target: '_blank'`, `rel: 'noopener noreferrer'`) and a `paragraph: 'editable-text-paragraph'` theme key. `linkAttributes` is module-level on purpose: `LinkPlugin` lists `attributes` in its effect deps, so an inline literal would re-register `TOGGLE_LINK_COMMAND` every render.
- **`globals.css`** — `.editable-text-paragraph { margin: 0 }`.
- **`LinkPlugin` mounts** (`editable-text.tsx`, `single-user-editor.tsx`, `collaborative-editor.tsx`) — pass `attributes={linkAttributes}`. No toolbar change was needed: `registerLink` forwards plugin attributes to `$toggleLink` on the string-payload branch the toolbar already dispatches. This also covers paste-to-link.
- **`ClickableLinkPlugin`** — `newTab` in `readonly-lexical-content.tsx`, and `newTab disabled={isEditable}` in `editable-text.tsx`.

Two things worth calling out for review:

1. **`newTab` is what fixes existing content.** Links already in the DB have `target: null` in their serialized JSON, so the attribute change alone would only help links created from now on. `ClickableLinkPlugin newTab` forces `_blank` at click time regardless of the stored target.
2. **In edit mode a click places the caret rather than opening a tab** (`disabled={isEditable}`, matching the Lexical playground). A literal reading of the comment would be `newTab` everywhere, but that makes link text nearly uneditable — every click would launch a tab. Read-only renders (including the reviewer preview) do open in a new tab. Happy to flip this if reviewers disagree.

Because all four render paths share `lexicalTheme`, the spacing fix covers the editor and every read-only view without per-component edits. No node-registry change, so `SEED_NODES` in `services/editor/lexical-seed.ts` doesn't need mirroring, and no data migration is required — spacing was never persisted, only rendered.

The reported edit-mode "save as draft modal" symptom is stale: that modal was removed in f19258d and 8fee14a, and there's no `beforeunload`/router guard in the repo. Preventing the default click also means no unload either way.

### Manual verification checklist

- [ ] Edit mode: select text → Link → enter URL → Apply. Clicking the link places the caret; no navigation, no dialog.
- [ ] Submitted proposal (read-only): the link opens in a new tab and the SI tab stays put.
- [ ] **Legacy content**: open a proposal whose link predates this change and confirm it also opens in a new tab.
- [ ] Paragraph spacing: several paragraphs separated by single Enters render single-spaced in both the editor and the submitted view, including pre-existing stored content.

[OTTER-463]: https://openstax.atlassian.net/browse/OTTER-463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -28,6 +28,10 @@
 }
 
 /* EditableText (Lexical) formatting classes */
+.editable-text-paragraph {
+    margin: 0;
+}
+
 .editable-text-bold {
     font-weight: bold;
 }
--- a/src/components/editable-text/toolbar.test.tsx
+++ b/src/components/editable-text/toolbar.test.tsx
@@ -1,0 +1,95 @@
+import { renderWithProviders, screen, act, waitFor, userEvent, describe, it, expect } from '@/tests/unit.helpers'
+import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext'
+import { $getRoot, $isElementNode, $isTextNode, type LexicalEditor } from 'lexical'
+import { SingleUserEditor } from './single-user-editor'
+
+function lexicalJson(text: string) {
+    return JSON.stringify({
+        root: {
+            children: [
+                {
+                    children: [{ detail: 0, format: 0, mode: 'normal', style: '', text, type: 'text', version: 1 }],
+                    direction: 'ltr',
+                    format: '',
+                    indent: 0,
+                    type: 'paragraph',
+                    version: 1,
+                },
+            ],
+            direction: 'ltr',
+            format: '',
+            indent: 0,
+            type: 'root',
+            version: 1,
+        },
+    })
+}
+
+// Captures the live Lexical editor so the test can drive real edits through the
+// editor's own API — jsdom can't dispatch the beforeinput events Lexical relies on.
+function CaptureEditor({ onReady }: { onReady: (editor: LexicalEditor) => void }) {
+    const [editor] = useLexicalComposerContext()
+    onReady(editor)
+    return null
+}
+
+async function renderEditorWithText(id: string, text: string) {
+    let editor: LexicalEditor | null = null
+    const { container } = renderWithProviders(
+        <SingleUserEditor id={id} initialValue={lexicalJson(text)} ariaLabel="Feedback">
+            <CaptureEditor onReady={(e) => (editor = e)} />
+        </SingleUserEditor>,
+    )
+
+    await waitFor(() => expect(editor).not.toBeNull())
+    return { container, editor: editor! }
+}
+
+// The toolbar reads the selection out of the editor state rather than the DOM, so
+// selecting programmatically is enough to stand in for a user drag.
+function selectAllText(editor: LexicalEditor) {
+    act(() => {
+        editor.update(() => {
+            const paragraph = $getRoot().getFirstChild()
+            if (!$isElementNode(paragraph)) return
+            const textNode = paragraph.getFirstChild()
+            if (!$isTextNode(textNode)) return
+            textNode.select(0, textNode.getTextContentSize())
+        })
+    })
+}
+
+async function insertLinkViaToolbar(url: string) {
+    const user = userEvent.setup()
+    await user.click(screen.getByLabelText('Link'))
+    const input = await screen.findByPlaceholderText('https://')
+    await user.clear(input)
+    await user.type(input, url)
+    await user.click(screen.getByLabelText('Apply link'))
+}
+
+describe('Toolbar link insertion', () => {
+    it('marks links to open in a new tab without leaking the referrer', async () => {
+        const { container, editor } = await renderEditorWithText('toolbar-1', 'hello world')
+        selectAllText(editor)
+
+        await insertLinkViaToolbar('https://example.com')
+
+        await waitFor(() => expect(container.querySelector('a')).not.toBeNull())
+        const anchor = container.querySelector('a')!
+        expect(anchor.getAttribute('href')).toBe('https://example.com')
+        expect(anchor.getAttribute('target')).toBe('_blank')
+        expect(anchor.getAttribute('rel')).toContain('noopener')
+    })
+
+    it('does not create a link for a URL that fails validation', async () => {
+        const { container, editor } = await renderEditorWithText('toolbar-2', 'hello world')
+        selectAllText(editor)
+
+        await insertLinkViaToolbar('javascript:alert(1)')
+
+        // The link editor closing is the signal that submitLink ran.
+        await waitFor(() => expect(screen.queryByPlaceholderText('https://')).toBeNull())
+        expect(container.querySelector('a')).toBeNull()
+    })
+})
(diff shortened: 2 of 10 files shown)


src/components/editable-text/toolbar.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+        await insertLinkViaToolbar('https://example.com')
+
+        await waitFor(() => expect(container.querySelector('a')).not.toBeNull())
+        const anchor = container.querySelector('a')!
+        expect(anchor.getAttribute('href')).toBe('https://example.com')
+        expect(anchor.getAttribute('target')).toBe('_blank')
+        expect(anchor.getAttribute('rel')).toContain('noopener')
+    })
+
+    it('does not create a link for a URL that fails validation', async () => {
```
nathanstitt, line 85: This one doesn't actually pin the behavior its name claims. I mutation-tested it: changing `submitLink` to `if (trimmed) {` (dropping the `isValidUrl` call entirely) leaves this test **still passing**. The thing that keeps the anchor out of the DOM is the second-layer `validateUrl` on `LinkPlugin` — I confirmed that by also removing `validateUrl` from the `LinkPlugin` mount, at which point the test fails with `expected <a href="about:blank"> to be null`.

So as written this is a regression test for `LinkPlugin validateUrl`, not for the toolbar's guard, even though it lives in `toolbar.test.tsx` and is worded as a toolbar assertion. Worth either renaming it to say what it covers, or tightening it so the toolbar guard is what's under test — e.g. spying that `TOGGLE_LINK_COMMAND` was dispatched with `null` rather than the raw string.

(The `about:blank` in that failure is also a nice incidental find: Lexical's `sanitizeUrl` neutralizes the `javascript:` payload, so there's defense in depth here. Just not the depth this test's name advertises.)

src/components/editable-text/toolbar.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+// The toolbar reads the selection out of the editor state rather than the DOM, so
+// selecting programmatically is enough to stand in for a user drag.
+function selectAllText(editor: LexicalEditor) {
+    act(() => {
+        editor.update(() => {
+            const paragraph = $getRoot().getFirstChild()
+            if (!$isElementNode(paragraph)) return
+            const textNode = paragraph.getFirstChild()
+            if (!$isTextNode(textNode)) return
+            textNode.select(0, textNode.getTextContentSize())
```
nathanstitt, line 57: Heads up that this `select()` makes each of these two tests dump a `TypeError: Cannot assign to read only property 'format' of object '#<RangeSelection>'` stack to stderr — three of them per full-file run. The tests still pass, so it's noise rather than a failure, but I checked and it's specific to this file (`readonly-lexical-content.test.tsx` runs clean).

Root cause is the programmatic `select()` reaching happy-dom's `Selection.setBaseAndExtent`, which synchronously fires `selectionchange`; Lexical's `onDocumentSelectionChange` then tries to write `format` on a frozen `RangeSelection`. Probably worth resolving so we're not training ourselves to scroll past red stacks in a green run — wrapping the select in `editor.update(..., { discrete: true })`, or setting the selection via `$setSelection($createRangeSelection())` instead of `textNode.select()`, are the two angles I'd try first.

src/app/globals.css
```diff
@@ -28,6 +28,10 @@
 }
 
 /* EditableText (Lexical) formatting classes */
+.editable-text-paragraph {
```
nathanstitt, line 31: Since this zeroes the margin outright rather than reducing it, paragraphs now sit at pure line-height with no separation at all. That's the literal reading of Greg's "remove double spacing," but it seems like we should have _some_ spacing between paras.  

how about a `margin: 0 0 0.25em` which would kill the blank-line effect while staying consistent with the list spacing right below.

src/components/editable-text/toolbar.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+            version: 1,
+        },
+    })
+}
+
+// Captures the live Lexical editor so the test can drive real edits through the
+// editor's own API — jsdom can't dispatch the beforeinput events Lexical relies on.
+function CaptureEditor({ onReady }: { onReady: (editor: LexicalEditor) => void }) {
+    const [editor] = useLexicalComposerContext()
+    onReady(editor)
```
nathanstitt, line 32: `@lexical/react` already ships a plugin for exactly this capture — `EditorRefPlugin` (`@lexical/react/LexicalEditorRefPlugin`). Using it would drop the custom component and also avoid calling `onReady` during the render phase (it fires on every re-render, and side effects in render are something React strict mode is increasingly unhappy about):

```tsx
const editorRef = { current: null as LexicalEditor | null }
// ...
<EditorRefPlugin editorRef={editorRef} />
```

Not a blocker — the current version works — but since the library provides it, we could lean on that.

src/components/readonly-lexical-content.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+        const { container } = renderWithProviders(
+            <ReadOnlyLexicalContent value={legacyLinkState('https://example.com')} />,
+        )
+
+        const anchor = await waitFor(() => {
+            const found = container.querySelector('a')
+            expect(found).not.toBeNull()
+            return found!
+        })
+        // Guards the premise: this is legacy content, so the fix cannot come from the attribute.
```
nathanstitt, line 48: Nice touch — asserting the premise (`target` is null) before asserting the behavior means this test cannot silently degrade into re-testing the attribute path if the fixture ever changes. The legacy-content case was the easy one to lose, and this pins it.

nathanstitt (approved):
Looks great @jbwilson8 - both are minor but take a look at my comment on para spacing and see if we should use the lexical provided helper.

PR #946, opened 2026-08-05
Require an explicit opt-in before seeding test data, and rotate the test key

nathanstitt:
This seed writes fixed test users, admin memberships on real org slugs, and a public key whose private half is committed to this repo. **It ran against production** — its only guard was an opt-out on `NO_TESTING_DATA`, and the infrastructure computed that variable from an unresolved CloudFormation token, so it was empty in every environment and never fired.

Paired with safeinsights/iac#198, **which must deploy first**: the migrator Lambda only gets `ALLOW_TESTING_DATA` from that change, so deploying this alone leaves non-prod environments unseeded.

## Changes

- **Guard inverted to opt-in.** `ALLOW_TESTING_DATA=TRUE` (the literal, so `false`/`0` cannot opt in) must be set, plus an independent refusal when `ENVIRONMENT_ID` names production. The seed keeps both checks inline (migrator Lambda bundling constraint) and logs when it skips; the tsx entry points — `bin/seed-environment.ts` and both Clerk test-user scripts — share one gate helper, `bin/lib/testing-data-gate.ts`. `bin/migrate-dev-db` sets the flag, so local dev and CI are unchanged.
- **Key rotated.** All four dependent artifacts regenerated together — `private_key.pem`, `public_key.pem`, `public_key.sig` (its `SHA2-256(stdin)= ` prefix preserved), and the embedded `TEST_PUBLIC_KEY_PEM`.
- **Migration removes existing exposure.** Rotation only prevents *future* use. `1780700000000_remove_exposed_test_public_keys` deletes `user_public_key` rows holding the old key (matched by hashing the stored bytes) **and** the `study_job_file_recipient_key` rows wrapped to it (both fingerprint spellings), logging affected counts. Expected to remove nothing in production — running it is how we establish that. Consequence in long-lived non-prod environments: artifacts whose only recipient keys are removed become undecryptable; users re-share or regenerate them.
- **Placeholder dropped.** Enclaves no longer get `settings.publicKey = 'BAD KEY, UPDATE ME'`; nothing reads it, and on a fresh insert it would leave a real enclave unable to receive encrypted results.

## The real org slugs deliberately stay

The ticket asked to stop seeding into real slugs. That isn't safe to do by renaming: `safe-insights` is the live SI-admin slug (`CLERK_ADMIN_ORG_SLUG`, used by `src/lib/session.ts`) and the openstax slugs drive feature gating (`src/lib/constants.ts`). Renaming would change production authorization for no security benefit, and cascades into ~34 files including hard-coded e2e URL paths. The gate is what keeps the seed out of deployed databases. Called out in a comment on the seed so this doesn't get re-litigated.

## Verification

Guard regression tests now delete a seeded fixture and assert the gated seed does not restore it (verified to fail with either guard removed), and the migration has its own test on pg_temp shadow tables. Against a real Postgres:

| case | result |
|---|---|
| seed without opt-in | 0 users |
| seed with opt-in + `ENVIRONMENT_ID=production` | 0 users |
| seed with opt-in, non-prod | 6 users, rotated key |
| env holding the old key, then migrate + seed | 0 exposed, 6 rotated |
| migration vs. both fingerprint spellings + wrapped recipient keys | all deleted, rotated rows untouched |

`pnpm run db:migrate` still seeds. Full unit suite: **253 files / 2410 tests passing**. Typecheck and lint clean.

https://openstax.atlassian.net/browse/OTTER-722


src/database/seeds/1743608138837_test_users.ts
```diff
@@ -69,7 +69,33 @@ const ORG_MEMBERSHIPS_BY_ROLE: Record<TestUserRole, { slug: string; isAdmin: boo
...
+    // =======================================================================
+
+    // Opt in explicitly. Set by local dev, CI, and the non-production migrator Lambdas only.
+    if (!process.env.ALLOW_TESTING_DATA) return
+
+    // Independent production backstop, in case something ever sets the opt-in there. NODE_ENV is
+    // deliberately not consulted: it is 'production' in every Next production build, including
+    // non-production deployed environments.
+    const envName = (process.env.ENVIRONMENT_ID || process.env.DEPLOY_ENV || '').toLowerCase()
+    if (envName === 'production' || envName === 'prod') return
```
therealmarv: **High.** Once prod skips this seed, nothing creates the `safe-insights` org row any more, and that row is what SI-admin authorization reads.

Line 27 here is the only automated writer of `{ slug: 'safe-insights', id: ...0101 }`. The bootstrap script can't stand in: `findOrCreateOrgMembership` does [`executeTakeFirstOrThrow(throwNotFound(...))`](https://github.com/safeinsights/management-app/blob/OTTER-722-gate-test-user-seed/src/server/mutations.ts#L58) on the org lookup and never inserts an org. `insertOrgAction` can't either, since the only grant for `create Org` is [`permit('manage', 'all')`](https://github.com/safeinsights/management-app/blob/OTTER-722-gate-test-user-seed/src/lib/permissions.ts#L117-L118) behind `isSiAdmin`, which [reads that same row](https://github.com/safeinsights/management-app/blob/OTTER-722-gate-test-user-seed/src/lib/session.ts#L31).

So a new AWS account or a prod restore into an empty DB now fails at `pnpm run create-admin-user` with `organization with slug safe-insights not found`, and nobody can be granted SI admin without hand-written SQL. Existing prod keeps its row, so this is latent rather than immediate, but it's a real regression.

Requested change: add a non-seed bootstrap path for the SI-admin org, either a migration that upserts it or an insert inside `create-admin-user`.

Separately, `DEPLOY_ENV` on line 97 is dead: nothing in either repo sets it (iac exports `DEPLOY_ID` and `ENVIRONMENT_ID`). Worth dropping so the backstop doesn't read as two variables when it's one.

src/database/seed-test-public-key.test.ts
```diff
@@ -10,4 +10,49 @@ describe('test users seed', () => {
...
+                    '00000000-0000-4000-8000-000000000105',
+                ])
+                .execute()
+            return rows.length
+        }
+
+        it('does nothing without ALLOW_TESTING_DATA', async () => {
+            const before = await seededOrgCount()
+            await seed(db)
+            expect(await seededOrgCount()).toEqual(before)
```
therealmarv: **High.** Both guard tests pass with the guard deleted, so the one bug this PR exists to prevent has no working regression test.

`seededOrgCount()` counts three fixed UUIDs that `db:migrate` has already inserted, which the comment on line 29 acknowledges. On a pre-seeded database (which is what CI is) `after === before` holds whether the gate fired or not. Remove `if (!process.env.ALLOW_TESTING_DATA) return` from the seed and this file still goes green.

Requested change: delete a fixture first, then assert the guarded call does not restore it. Something like:

```ts
await db.deleteFrom('org').where('id', '=', '00000000-0000-4000-8000-000000000101').execute()
await seed(db)
expect(await seededOrgCount()).toEqual(0)
```

Each test runs in a rolled-back transaction, so the delete is safe.

bin/seed-environment.ts
```diff
@@ -485,6 +485,20 @@ async function setupOrganizations() {
 }
 
 async function seedEnvironment() {
+    // Same opt-in and production backstop as the Kysely seed. This script only ever runs from
+    // bin/migrate-dev-db, never from the migrator Lambda, so it is not part of the production
+    // exposure path — but it writes the same shared test public key and touches the same real
+    // org slugs, so it gets the same guard rather than being left as a second footgun.
+    if (!process.env.ALLOW_TESTING_DATA) {
```
therealmarv: **High.** The comment above says this gets the guard "rather than being left as a second footgun", but there are two more footguns still ungated.

[`bin/find-or-update-clerk-test-users.ts`](https://github.com/safeinsights/management-app/blob/OTTER-722-gate-test-user-seed/bin/find-or-update-clerk-test-users.ts) does the same dangerous writes with no flag and no prod backstop: `ensurePublicKey` at L21 inserts the committed test key, and L46/L51/L77 grant `isAdmin: true` on the real org slugs, called from L118. [`bin/setup-clerk-test-users.ts`](https://github.com/safeinsights/management-app/blob/OTTER-722-gate-test-user-seed/bin/setup-clerk-test-users.ts) is in the same position.

That means an operator pointed at a deployed DB can silently re-create exactly the rows the new migration just deleted, and the gate reads as complete when it isn't.

Requested change: extract the opt-in plus prod check into one small helper and call it from all four entry points, rather than a third copy.

src/database/migrations/1780700000000_remove_exposed_test_public_keys.ts
```diff
@@ -0,0 +1,31 @@
...
+// Matched three ways because the two seed paths wrote different spellings of the same
+// fingerprint: the Kysely seed (1743608138837_test_users.ts) stores bare hex, while
+// bin/seed-environment.ts inserts tests/support/public_key.sig verbatim, including its
+// "SHA2-256(stdin)= " prefix. The public_key comparison is authoritative — it hashes the stored
+// key bytes, so it catches rows under any fingerprint spelling, including none at all.
+const EXPOSED_FINGERPRINT = '5b67466d712b406f54d7ec37b1ab1ac58687eb3a604a5ed8ccb1e4cca4dd06cc'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await sql`
+        DELETE FROM user_public_key
```
therealmarv: **Medium.** Deleting `user_public_key` removes the pointer, not the exposed ciphertext. The AES keys actually wrapped to this public key live in `study_job_file_recipient_key.crypt` ([1780100000000_study_job_file_keys.ts](https://github.com/safeinsights/management-app/blob/OTTER-722-gate-test-user-seed/src/database/migrations/1780100000000_study_job_file_keys.ts)), and in the manifests inside already-uploaded result archives in S3. Anyone with the old private key from git history can still unwrap those.

Second effect, and this one is a regression rather than an omission: rotating the keypair *and* deleting these rows leaves those recipient-key rows stranded on the old fingerprint in Dev/Staging/Demo, so results already shared there become permanently undecryptable with no documented recovery path.

Requested change: also delete `study_job_file_recipient_key` rows matching the exposed fingerprint, and say in the PR description what happens to existing artifacts in long-lived non-prod environments.

src/database/migrations/1780700000000_remove_exposed_test_public_keys.ts
```diff
@@ -0,0 +1,31 @@
...
+        WHERE fingerprint = ${EXPOSED_FINGERPRINT}
+           OR fingerprint = ${`SHA2-256(stdin)= ${EXPOSED_FINGERPRINT}`}
+           OR encode(sha256(public_key), 'hex') = ${EXPOSED_FINGERPRINT}
+    `.execute(db)
```
therealmarv, line 26: **Medium.** Line 10 says "running it is how we establish" the zero-row result, but this throws away the only evidence. Kysely's `RawBuilder.execute()` returns a `QueryResult` carrying `numAffectedRows`, and it's discarded with nothing logged. After the DELETE, the question "how exposed were we, and where" is unanswerable.

Requested change:

```ts
const result = await sql`...`.execute(db)
console.log(`removed ${result.numAffectedRows ?? 0n} exposed public key row(s)`)
```

Also worth flagging: a destructive migration with no test at all, while the repo already has migration tests (`backfill_code_phase_study_status.test.ts`, `study_job_file_artifact_slot.test.ts`). If one predicate is subtly wrong this deletes nothing and the deploy still reports success.

src/database/seed-test-public-key.test.ts
```diff
@@ -10,4 +10,49 @@ describe('test users seed', () => {
...
+        const savedEnv = { ...process.env }
+
+        beforeEach(() => {
+            delete process.env.ALLOW_TESTING_DATA
+            delete process.env.ENVIRONMENT_ID
+            delete process.env.DEPLOY_ENV
+        })
+
+        afterEach(() => {
+            process.env = { ...savedEnv }
```
therealmarv: **Medium.** This leaks into every later test in the worker.

`savedEnv` on line 17 is captured in the `describe` body, i.e. during collection. [`tests/vitest.setup.ts#L252-L256`](https://github.com/safeinsights/management-app/blob/OTTER-722-gate-test-user-seed/tests/vitest.setup.ts#L252-L256) deletes `CLERK_SECRET_KEY` / `CLERK_PUBLISHABLE_KEY` / `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` inside a `beforeAll`, and Vitest runs `beforeAll` *after* collection. So the snapshot still has those keys, and the first `afterEach` puts them back.

Wholesale `process.env = {...}` is also an anti-pattern in Node: it swaps the real environment object for a plain object and loses its automatic string coercion.

Requested change: delete the three keys individually instead.

```ts
afterEach(() => {
    delete process.env.ALLOW_TESTING_DATA
    delete process.env.ENVIRONMENT_ID
    delete process.env.DEPLOY_ENV
})
```

src/database/seeds/1743608138837_test_users.ts
```diff
@@ -69,7 +69,33 @@ const ORG_MEMBERSHIPS_BY_ROLE: Record<TestUserRole, { slug: string; isAdmin: boo
...
+    // src/lib/session.ts) and the openstax slugs drive feature gating (src/lib/constants.ts).
+    // The e2e specs hard-code them as URL paths. Safety comes from this gate, not from the names.
+    //
+    // Keep both checks as inline process.env reads with no new imports: importing anything that
+    // pulls in a CJS dependency (@/server/config -> debug -> require('tty')) crashes the esbuild
+    // ESM bundle the migrator Lambda runs seeds from. See the note near TEST_PUBLIC_KEY_PEM.
+    // =======================================================================
+
+    // Opt in explicitly. Set by local dev, CI, and the non-production migrator Lambdas only.
+    if (!process.env.ALLOW_TESTING_DATA) return
```
therealmarv: **Medium.** Two things on this line.

Silent return: `kysely seed:run` now exits 0 having done nothing, from every caller that doesn't hardcode the flag. `bin/migrate-dev-db` is the only one that does. A developer running `seed:run` directly gets no output and then debugs missing orgs with nothing pointing at the cause. A one-line log would make the gate self-explaining.

Truthiness: this is a bare check on a string, so `ALLOW_TESTING_DATA=false`, `0`, `no` and `off` all enable seeding. iac writes the literal `'TRUE'` while `bin/migrate-dev-db` writes `1`, so an explicit comparison needs the two producers to agree first.

Requested change: log on the skip, and pick one accepted value across the seed, `bin/seed-environment.ts`, `bin/migrate-dev-db` and iac's `app-stack.ts`.

src/database/migrations/1780700000000_remove_exposed_test_public_keys.ts
```diff
@@ -0,0 +1,31 @@
...
+// "SHA2-256(stdin)= " prefix. The public_key comparison is authoritative — it hashes the stored
+// key bytes, so it catches rows under any fingerprint spelling, including none at all.
+const EXPOSED_FINGERPRINT = '5b67466d712b406f54d7ec37b1ab1ac58687eb3a604a5ed8ccb1e4cca4dd06cc'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+    await sql`
+        DELETE FROM user_public_key
+        WHERE fingerprint = ${EXPOSED_FINGERPRINT}
+           OR fingerprint = ${`SHA2-256(stdin)= ${EXPOSED_FINGERPRINT}`}
+           OR encode(sha256(public_key), 'hex') = ${EXPOSED_FINGERPRINT}
```
therealmarv: **Low.** The first two predicates are dead code, and the comment on line 15 already implies it.

I checked: `sha256` over `origin/main:tests/support/public_key.pem` decoded to DER is `5b67466d...dd06cc`, which is both `EXPOSED_FINGERPRINT` and the old `public_key.sig` hex. Both writers store DER in `public_key` (this seed at line 320, `bin/seed-environment.ts` at L188 via `pemToArrayBuffer`), so clause 3 already matches every row clauses 1 and 2 could.

Requested change: keep only the `encode(sha256(public_key), 'hex')` comparison and trim the comment to match. Fewer moving parts on a destructive statement.

src/database/migrations/1780700000000_remove_exposed_test_public_keys.ts
```diff
@@ -0,0 +1,31 @@
...
+    `.execute(db)
+}
+
+export async function down(): Promise<void> {
```
therealmarv, line 37: **Low.** An empty `down()` reports success while restoring nothing. Someone running `migrate:down` to back this out will believe the rows are back, and in prod nothing will ever re-create them since seeds are gated off there.

Requested change: `throw new Error('irreversible: ...')` and keep the existing comment as the explanation. That surfaces the irreversibility instead of hiding it behind a green result.

src/database/seeds/1743608138837_test_users.ts
```diff
@@ -69,7 +69,33 @@ const ORG_MEMBERSHIPS_BY_ROLE: Record<TestUserRole, { slug: string; isAdmin: boo
...
+    // variable the infrastructure computed from an unresolved CloudFormation token and so never
+    // actually set. Two independent opt-IN checks now apply, and both must pass.
+    //
+    // The real slugs below (safe-insights, openstax, openstax-lab) are deliberate and must stay:
+    // safe-insights is the live SI-admin slug (CLERK_ADMIN_ORG_SLUG in src/lib/types.ts, used by
+    // src/lib/session.ts) and the openstax slugs drive feature gating (src/lib/constants.ts).
+    // The e2e specs hard-code them as URL paths. Safety comes from this gate, not from the names.
+    //
+    // Keep both checks as inline process.env reads with no new imports: importing anything that
+    // pulls in a CJS dependency (@/server/config -> debug -> require('tty')) crashes the esbuild
```
therealmarv: **Low.** The stated cause is wrong. `@/server/config` doesn't import `debug` at all, its only import is `@aws-sdk/client-secrets-manager`.

The constraint may well be real, but a load-bearing "don't touch this" comment with a false reason is worse than none: the next person verifies it, finds it doesn't hold, and either removes the constraint or stops trusting the whole block.

Requested change: replace with the actual failure you hit, or soften to "importing `@/server/config` broke the esbuild ESM bundle, cause not fully diagnosed".

bin/migrate-dev-db
```diff
@@ -13,8 +13,11 @@ pnpm exec kysely-codegen \
   --overrides='{"columns":{"org_code_env.settings":"Generated<OrgCodeEnvSettings>","org_code_env.command_lines":"CommandLines"}}' \
   --custom-imports='{"OrgCodeEnvSettings":"./types-manual#OrgCodeEnvSettings","CommandLines":"./types-manual#CommandLines"}'
 
-# run seeds after generating types above so the seed files see the updated DB state
-pnpm exec kysely seed:run
+# run seeds after generating types above so the seed files see the updated DB state.
+# ALLOW_TESTING_DATA is the seed's opt-in: seeds no-op without it, so that a deployed
+# environment that never sets it cannot be seeded with test fixtures. Scoped to this
+# command rather than exported, to keep it off the migration and seed-environment steps.
```
therealmarv: **Low.** This says the flag is kept off the seed-environment step, but line 45 sets it there too. One of the two is wrong, and the diff doesn't say which is intended.

Requested change: drop "and seed-environment" from the comment, since keeping the flag on line 45 is clearly deliberate.

bin/seed-environment.ts
```diff
@@ -485,6 +485,20 @@ async function setupOrganizations() {
...
+        console.log('ALLOW_TESTING_DATA is not set - skipping environment seeding.')
+        return
+    }
+    const envName = (process.env.ENVIRONMENT_ID || process.env.DEPLOY_ENV || '').toLowerCase()
```
therealmarv: **Low.** This duplicates `PROD_ENV` / `ENVIRONMENT_ID` from `@/server/config`. The seed has a documented bundling reason for its inline copy; this script has none, it's plain `tsx`.

Requested change: import the existing helper here so there's one definition of "is production" to keep in sync.

src/database/seeds/1743608138837_test_users.ts
```diff
@@ -103,7 +128,12 @@ export async function seed(db: Kysely<DB>): Promise<void> {
                   ? 'Enclave where the reviewer is an admin'
                   : null
 
-        const settings = isEnclave ? { publicKey: 'BAD KEY, UPDATE ME' } : {}
+        // Enclaves previously got `publicKey: 'BAD KEY, UPDATE ME'` here. Nothing reads that
+        // placeholder, and writing it on a fresh insert would leave a real enclave unable to
+        // receive encrypted results, so enclaves now start with empty settings like labs and
+        // get a real key through the normal admin flow. Existing rows were never affected:
+        // the onConflict below updates only slug and name.
+        const settings = {}
```
therealmarv, line 153: **Nit.** Freshly seeded enclave orgs now fail [`enclaveSettingsSchema`](https://github.com/safeinsights/management-app/blob/OTTER-722-gate-test-user-seed/src/schema/org.ts#L6-L8) (`publicKey: z.string().min(1)`), so an SI admin editing just the name or email on one of those orgs can't submit until they supply a key.

I think the tradeoff is right, a garbage placeholder is worse than none, and your comment reasons about it explicitly. Flagging only so it's a known consequence and nobody debugs it cold. Non-prod only.

therealmarv (changes requested):
**Edited to correct a mistake.** My original review led with a Critical about the deploy dependency being unhandled. That was wrong: iac#198's description already has a "Deploy ordering" section prescribing Dev first, then Sandbox/Staging/Demo, then Production, which is the safe sequence. I've corrected that over there and downgraded it. Sorry for the false alarm. The findings below are unaffected.

Reviewed together with safeinsights/iac#198, since the two only work as a pair.

The direction is right and the write-up is genuinely good. Three things need to change before this lands.

**High:**
- Both new guard tests pass with the guard deleted, so the bug this PR fixes has no regression test.
- Two sibling scripts (`bin/find-or-update-clerk-test-users.ts`, `bin/setup-clerk-test-users.ts`) still write the committed key and real-slug admin memberships ungated, and can undo the migration's cleanup.
- Gating seeds off in prod removes the only automated creator of the `safe-insights` org, so a new or restored prod DB can't bootstrap an SI admin.

**Medium:** the cleanup stops at `user_public_key` and leaves the wrapped keys in `study_job_file_recipient_key`; the migration discards its row count, which makes the audit OTTER-722 asks for impossible after deploy; the test file's `afterEach` re-arms Clerk credentials for the rest of the worker.

**Low:** this PR's description doesn't mention that it requires iac#198 deployed first. It is the half that breaks things if it lands first: the Lambda would have no `ALLOW_TESTING_DATA`, the new seed self-returns in every environment, and this migration has already deleted the old key rows. One line in the body would close it.

Rest is Low/Nit and inline.

Two notes on things I checked and cleared, so nobody chases them: the `encode(sha256(public_key),'hex')` predicate is correct (I recomputed the fingerprint from `origin/main`, it matches, and both writers store DER), and `sha256()` is fine on Aurora PostgreSQL 16.10.

One judgment call I agree with: keeping the real `safe-insights` / `openstax` slugs. Renaming them would move production authorization, and the gate is the right place for safety. Worth stating in the PR body that this deliberately declines that part of the ticket, so it reads as a decision rather than an oversight.

src/database/seeds/1743608138837_test_users.ts
nathanstitt: Went with the owner call to not add a bootstrap code path: the gate now carries a comment documenting the concern and the recovery flow (valid Clerk accounts are auto-created on first login, so a fresh prod DB is bootstrapped by creating the admin user in Clerk and having them log in). The dead DEPLOY_ENV check is dropped here and in bin/seed-environment.ts.

src/database/seed-test-public-key.test.ts
nathanstitt: Rewritten as suggested: each guard test now deletes a seeded fixture, runs the gated seed, and asserts it is not restored, plus a positive control that proves the harness observes a real seed run. The fixture is the admin user public key row rather than the org row, because org has NO ACTION inbound FKs (study, org_user, pending_user) that would make the delete fragile. Verified both guard tests fail when their respective check is removed from the seed.

bin/seed-environment.ts
nathanstitt: Extracted the opt-in plus prod backstop into bin/lib/testing-data-gate.ts and wired it into seed-environment.ts and both Clerk scripts (the Clerk scripts exit non-zero when refused). The Kysely seed keeps a minimal inline copy with a comment pointing at the helper, since the migrator Lambda bundling constraint keeps imports out of it.

src/database/migrations/1780700000000_remove_exposed_test_public_keys.ts
nathanstitt: The migration now also deletes study_job_file_recipient_key rows wrapped to the exposed key, matching both fingerprint spellings the two writers produced (recipient rows copy the user_public_key fingerprint verbatim, so those are the only spellings possible). The PR description now states the non-prod consequence: artifacts whose only recipient keys are removed become undecryptable and must be re-shared or regenerated. S3-side manifests are out of reach for a DB migration.

src/database/migrations/1780700000000_remove_exposed_test_public_keys.ts
nathanstitt: Both DELETEs now capture the QueryResult and log removed row counts via console.warn (the channel eslint allows). Added src/database/remove_exposed_test_public_keys.test.ts following the existing migration-test patterns: it runs up() against pg_temp shadow tables, asserts rows under both fingerprint spellings are removed and non-matching rows survive, and ties the embedded pre-rotation PEM to the hardcoded fingerprint.

src/database/seed-test-public-key.test.ts
nathanstitt: Fixed — the snapshot is gone; beforeEach/afterEach now delete the three variables individually (DEPLOY_ENV no longer among them since it was removed from the seed), so the collection-time Clerk keys can no longer be re-armed and process.env stays the real environment object.

src/database/seeds/1743608138837_test_users.ts
nathanstitt: The seed now logs when it skips, and the accepted value is standardized on the literal TRUE across the seed, the new shared gate helper, and bin/migrate-dev-db (iac app-stack.ts already writes TRUE). A new guard test pins that ALLOW_TESTING_DATA=1 (or false/0) no longer opts in.

src/database/migrations/1780700000000_remove_exposed_test_public_keys.ts
nathanstitt: Trimmed as suggested — the user_public_key DELETE now carries only the encode(sha256(public_key), hex) predicate, and the comment explains the spellings in the context of the recipient-key DELETE, where matching the two literals is genuinely needed.

src/database/migrations/1780700000000_remove_exposed_test_public_keys.ts
nathanstitt: down() now throws with the irreversibility spelled out, keeping the original comment as the explanation, and the new migration test pins that behavior.

src/database/seeds/1743608138837_test_users.ts
nathanstitt: You are right, config never imports debug — its only import is the CJS @aws-sdk/client-secrets-manager. The comment now attributes the verified failure to si-encryption -> debug -> require(tty) and softens the config claim to what we actually know: importing it broke the bundle when tried, cause not fully diagnosed.

bin/migrate-dev-db
nathanstitt: Dropped "and seed-environment" from the comment — keeping the flag on the seed-environment invocation below is deliberate.

bin/seed-environment.ts
nathanstitt: Merged into the shared gate: bin/lib/testing-data-gate.ts imports PROD_ENV/ENVIRONMENT_ID from @/server/config, so there is now one definition of "is production" for every tsx entry point, and seed-environment.ts just calls the helper.

src/database/seeds/1743608138837_test_users.ts
nathanstitt: Agreed — a known, accepted tradeoff: an SI admin editing name/email on a freshly seeded enclave org must supply a key before submitting, and a garbage placeholder would be worse. Non-prod only, and the seed comment reasons about it explicitly.

bin/migrate-dev-db
```diff
@@ -13,8 +13,12 @@ pnpm exec kysely-codegen \
   --overrides='{"columns":{"org_code_env.settings":"Generated<OrgCodeEnvSettings>","org_code_env.command_lines":"CommandLines"}}' \
   --custom-imports='{"OrgCodeEnvSettings":"./types-manual#OrgCodeEnvSettings","CommandLines":"./types-manual#CommandLines"}'
 
-# run seeds after generating types above so the seed files see the updated DB state
-pnpm exec kysely seed:run
+# run seeds after generating types above so the seed files see the updated DB state.
+# ALLOW_TESTING_DATA=TRUE is the seed's opt-in (the literal iac's app-stack.ts also writes):
```
therealmarv, line 17: Missing a word: `(the literal value iac's app-stack.ts also writes)`.

bin/lib/testing-data-gate.ts
```diff
@@ -0,0 +1,23 @@
...
+// cannot import this file — the migrator Lambda's esbuild ESM bundle chokes on @/server/config —
+// so it duplicates these checks inline; keep the two in sync.
+// PROD_ENV is fixed at config-import time; that is fine here because every caller is a one-shot
+// tsx process whose environment cannot change after startup.
+export function testingDataAllowed(script: string): boolean {
+    if (process.env.ALLOW_TESTING_DATA !== 'TRUE') {
+        console.log(`${script}: skipping — writing test data requires ALLOW_TESTING_DATA=TRUE.`)
+        return false
+    }
+    if (PROD_ENV) {
```
therealmarv, line 18: This is narrower than the inline copy the comment above says to keep in sync.

`PROD_ENV` is `ENVIRONMENT_ID == 'production'`, exact and case-sensitive ([config.ts:32](https://github.com/safeinsights/management-app/blob/3e5ff5a765f53945a4753f827729cdaed23ea43e/src/server/config.ts#L32)). The seed lowercases and also matches `prod` ([test_users.ts:111-112](https://github.com/safeinsights/management-app/blob/3e5ff5a765f53945a4753f827729cdaed23ea43e/src/database/seeds/1743608138837_test_users.ts#L111-L112)).

So `ENVIRONMENT_ID=PRODUCTION` or `=prod` blocks the seed but not this gate. Low impact since iac writes `production` exactly, but it is the drift the comment warns about. Suggest normalizing here the same way instead of using `PROD_ENV`.

bin/migrate-dev-db
```diff
@@ -39,7 +43,7 @@ fi
 pnpm exec prettier --write $TYPES
 pnpm exec eslint --fix $TYPES
 
-pnpm exec tsx bin/seed-environment.ts
+ALLOW_TESTING_DATA=TRUE pnpm exec tsx bin/seed-environment.ts
 
 if [ -e "$CI" ]; then
```
therealmarv, line 48: Pre-existing, not from this PR, just noticed while reading: `-e` tests for a file named by `$CI`, so with `CI=true` it looks for a file called `true`. The `types.ts` drift check below it likely never fires in CI. Should be `[ -n "$CI" ]`. Fine to fix separately.

therealmarv (approved):
Approved.

Gate placement is right: both seed steps take the opt-in, `migrate:latest` correctly does not, and every entry point (`db:migrate`, `bin/docker-dev-init`, `bin/test-e2e-up`, `checks.yml`) funnels through `bin/migrate-dev-db`, so nothing loses its seeds.

Three non-blocking notes inline: a missing word in the gate comment, a real divergence between the shared gate and the inline copy it claims to mirror, and one pre-existing shell bug.

One residual risk worth knowing, by design: the script hardcodes the opt-in, so `PROD_ENV` is the only remaining backstop. Pointing `DATABASE_URL` at a deployed DB not labeled `production` (staging, say) and running `db:migrate` still seeds fixtures. Not a regression, and widening to `DEPLOYED_ENV` would break the non-production migrator Lambdas this intentionally supports.

PR #947, opened 2026-08-05
OTTER-724: scope org configuration reads and whitelist org updates

nathanstitt:

Stacks on #939 — review that one first; this branch is based on it, not main.

**MA-6 — cross-tenant org reads.** `permit('view','Org')` is unconditioned, so every authenticated user passed it for any org. It has to stay that way for the study-proposal dataset picker (97c118b1), so the narrowing moves into the actions:

- New `view OrgConfig` subject scoped to admins of that org. `fetchOrgCodeEnvsAction` (its `selectAll` carries `settings.environment` — plaintext env-var pairs, commonly credentials) and `fetchStarterCodeAction` move onto it.
- `getPendingUsersAction` → `invite User`, matching its two siblings in the same file. Its rows expose `pendingUser.id`, which is the live invite token.
- `getOrgFromSlugAction` read the whole row in *middleware*, which both handed `settings.publicKey` and `email` to any authenticated caller and echoed them back inside the `permission_denied` message returned to a refused one. Now uses `orgIdFromSlug` and selects a narrow `PublicOrg` in the handler.
- `fetchAdminOrgsWithStatsAction` returns every org's email and settings; its SI-admin-console callers have no layout gate, so it now requires `manage all`.
- `getOrgFromIdAction` deleted — no call sites, no middleware, `selectAll` on a client-supplied orgId.

**MA-5 — org mass-assignment.** `updateOrgAction` spread the full `updateOrgSchema` into `.set()`. That union extends the CREATE schemas, so it carries `type`, `slug`, `email` and `settings` — including `settings.publicKey`, the RS256 key verifying an enclave's M2M API bearer tokens. An org admin could overwrite it and forge API JWTs as that enclave, or flip `type` lab→enclave to gain reviewer abilities. Its only caller is the SI-admin console's `EditOrgForm`, so it now requires `manage all`; org admins keep `updateOrgSettingsAction`, which already whitelists name + description (and now runs in a transaction like its sibling).

The deliberately cross-org reads are commented so they aren't "fixed" later, and there are regression tests pinning them — the dataset-picker flow is preserved, as are cross-org enclave browsing and starter-code download for lab researchers.

### DIFF
--- a/src/app/[orgSlug]/admin/settings/code-envs.actions.ts
+++ b/src/app/[orgSlug]/admin/settings/code-envs.actions.ts
@@ -355,10 +355,12 @@ const fetchOrgCodeEnvsSchema = z.object({
     orgSlug: z.string(),
 })
 
+// `selectAll('orgCodeEnv')` includes settings.environment — plaintext env-var name/value pairs,
+// commonly credentials — plus scan results, so this is admin-console data, not public catalog data.
 export const fetchOrgCodeEnvsAction = new Action('fetchOrgCodeEnvsAction')
     .params(fetchOrgCodeEnvsSchema)
     .middleware(orgIdFromSlug)
-    .requireAbilityTo('view', 'Org')
+    .requireAbilityTo('view', 'OrgConfig')
     .handler(async ({ orgId, db }) => {
         return await db
             .selectFrom('orgCodeEnv')
@@ -549,10 +551,12 @@ const fetchStarterCodeSchema = z.object({
     fileName: z.string(),
 })
 
+// Returns S3 file contents. Note this is the admin console's editor view; the researcher-facing
+// starter-code download is getStarterCodeUrlAction, which stays cross-org on `view Orgs`.
 export const fetchStarterCodeAction = new Action('fetchStarterCodeAction')
     .params(fetchStarterCodeSchema)
     .middleware(codeEnvFromOrgAndId)
-    .requireAbilityTo('view', 'Org')
+    .requireAbilityTo('view', 'OrgConfig')
     .handler(async ({ codeEnv, params: { fileName } }) => {
         if (!codeEnv.starterCodeFileNames.includes(fileName)) {
             throw new Error(`Starter code file "${fileName}" not found`)
--- a/src/app/[orgSlug]/admin/settings/data-sources.actions.ts
+++ b/src/app/[orgSlug]/admin/settings/data-sources.actions.ts
@@ -19,6 +19,10 @@ const fetchOrgDataSourcesSchema = z.object({
     orgSlug: z.string(),
 })
 
+// Stays on the unconditioned `view Org` on purpose: the study-proposal dataset picker needs a lab
+// researcher to list an enclave org's datasets (97c118b1). The selected columns are therefore
+// catalog-level only — keep it that way, and route any new field carrying configuration or secrets
+// through `view OrgConfig` instead (OTTER-724 / MA-6).
 export const fetchOrgDataSourcesAction = new Action('fetchOrgDataSourcesAction')
     .params(fetchOrgDataSourcesSchema)
     .middleware(orgIdFromSlug)
--- a/src/server/actions/org.actions.test.ts
+++ b/src/server/actions/org.actions.test.ts
@@ -12,7 +12,9 @@ import {
 import { type Org } from '@/schema/org'
 import {
     deleteOrgAction,
+    fetchAdminOrgsWithStatsAction,
     getOrgFromSlugAction,
+    updateOrgAction,
     getUsersForOrgAction,
     fetchUsersOrgsAction,
     insertOrgAction,
@@ -21,6 +23,7 @@ import {
     getLanguagesForOrgAction,
 } from './org.actions'
 import logger from '@/lib/logger'
+import { isActionError } from '@/lib/errors'
 
 vi.mock('@/server/aws', async (importOriginal) => {
     const actual = await importOriginal<typeof import('@/server/aws')>()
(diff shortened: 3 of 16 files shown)


src/app/[orgSlug]/admin/settings/data-sources.actions.ts
```diff
@@ -19,6 +19,10 @@ const fetchOrgDataSourcesSchema = z.object({
     orgSlug: z.string(),
 })
 
+// Stays on the unconditioned `view Org` on purpose: the study-proposal dataset picker needs a lab
+// researcher to list an enclave org's datasets (97c118b1). The selected columns are therefore
+// catalog-level only — keep it that way, and route any new field carrying configuration or secrets
+// through `view OrgConfig` instead (OTTER-724 / MA-6).
```
Comment (therealmarv) [inline src/app/[orgSlug]/admin/settings/data-sources.actions.ts:25]: **Severity: Medium.** The comment says the selected columns are catalog-level only, but the `urls` sub-select below returns `orgDataSourceUrl.url` for every data source, and `codeEnvs` returns code-env ids and names. Since the action stays on the unconditioned `view Org`, any authenticated user can read those for any org given a guessable slug. That is exactly the leak MA-6 lists by name for this action ("`fetchOrgDataSourcesAction` ... data-source URLs").

The dataset-picker justification does not require those columns. The cross-org consumer is `useOrgDataSources`, which reduces the payload to `{ value: ds.id, label: ds.name }` (`src/hooks/use-org-data-sources.ts:19`). Only the admin console pages (`data-sources.tsx`, `data-source-form.tsx`) need `urls`.

Not blocking. Either fix the comment now so it does not assert a boundary the code does not enforce and open a follow-up to split a narrow catalog read from the full read on `view OrgConfig`, or do the split here.

src/server/actions/org.actions.ts
```diff
@@ -1,16 +1,22 @@
...
+// gated on ('manage','all') rather than the org-admin-scoped ('update','Org'). Its only caller is
+// the SI-admin console (EditOrgForm), which already required SI staff. Org admins edit their own
+// org through updateOrgSettingsAction below, which whitelists name + description (OTTER-724 / MA-5).
 export const updateOrgAction = new Action('updateOrgAction', { performsMutations: true })
     .params(updateOrgSchema)
-    .middleware(async ({ params: { id } }) => ({ orgId: id })) // translate params for requireAbility below
-    .requireAbilityTo('update', 'Org')
+    .requireAbilityTo('manage', 'all')
     .handler(async ({ params: { id, ...update }, db }) => {
         return await db.updateTable('org').set(update).where('id', '=', id).returningAll().executeTakeFirstOrThrow()
```
therealmarv, line 21: **Severity: Medium** (follow-up, not blocking). The `manage all` gate closes the org-admin half of MA-5. The rest of the card's fix text is still open: "route those to an SI-admin-only, audited path, and log or alert on public-key changes."

`.set(update)` writes `settings.publicKey`, the RS256 key verifying that enclave's M2M bearer tokens, with no record of who changed it, when, or for which org. I checked the whole audit surface: `src/server/events.ts` has no org emitter, and `AuditRecordType` is `'CODE_ENV' | 'STUDY' | 'USER'`, so nothing logs a successful org mutation (`logger.info` only fires inside `audit()`). A compromised or mistaken SI-admin account can swap an enclave's verification key with no application trail.

Worth saying this is not introduced here: `insertOrgAction` and `deleteOrgAction` are equally silent, and the fix needs a migration to add `ORG` to the `audit_record_type` enum (precedent: `1780400000000_add_code_env_audit_record_type.ts`), so a follow-up card seems right rather than growing this PR. `auditCodeEnv` in `events.ts` is the model to copy, and this action already runs in a transaction via `performsMutations`, so the audit row can commit atomically with the change.

Related nit while you are in here: `settings.publicKey` is `z.string().min(1)` (`src/schema/org.ts:7`) with no PEM validation, unlike the user-key path at `src/server/qa-provision.ts:107`. A malformed key silently breaks M2M auth for that enclave at `org-from-auth-token.ts:45`, and with no audit row there is no breadcrumb to diagnose it.

src/app/[orgSlug]/admin/settings/code-envs.actions.ts
```diff
@@ -549,10 +551,12 @@ const fetchStarterCodeSchema = z.object({
     fileName: z.string(),
 })
 
+// Returns S3 file contents. Note this is the admin console's editor view; the researcher-facing
+// starter-code download is getStarterCodeUrlAction, which stays cross-org on `view Orgs`.
 export const fetchStarterCodeAction = new Action('fetchStarterCodeAction')
     .params(fetchStarterCodeSchema)
     .middleware(codeEnvFromOrgAndId)
-    .requireAbilityTo('view', 'Org')
+    .requireAbilityTo('view', 'OrgConfig')
```
Comment (therealmarv) [inline src/app/[orgSlug]/admin/settings/code-envs.actions.ts:559]: **Severity: Low.** Flagging it only because this PR is what documents the rule at `action.ts:102-107`. The `codeEnvFromOrgAndId` middleware returns the whole `codeEnv` object, which lands in the CASL subject and is stringified into the `permission_denied` message returned to the caller you just refused.

While this sat on the unconditioned `view Org` the denial branch was unreachable. Now that it is conditioned, a cross-org caller gets refused and receives `url` (the Docker image ref, often an ECR URI carrying an account id), `identifier`, `sampleDataPath` and `starterCodeFileNames`. No secrets, since the middleware does not select `settings` or scan results, and it is a large net improvement over the previous behavior of handing over the file contents.

If you want it to match the rule exactly, have the middleware return `{ orgId }` and read the code-env row in the handler.

src/server/actions/org.actions.test.ts
```diff
@@ -291,5 +378,21 @@ describe('Org Actions', () => {
             const result = await getLanguagesForOrgAction({ orgSlug: '' })
             expect(result).toEqual({ error: expect.stringContaining('Validation error') })
         })
+
+        // Deliberately cross-org: a lab researcher starting a proposal browses enclaves they do
+        // not belong to and downloads their starter code. OTTER-724 narrowed the neighbouring
```
therealmarv, line 383: Nit (severity: trivial): "neighbouring" -> "neighboring", the repo uses US spelling.

therealmarv (approved):
Approving, nothing blocking.

This is a clear net improvement on both findings. Things I verified rather than assumed:

- No org-admin detour around the MA-5 gate: `create Org` and `delete Org` are reachable only through `manage all`, so an org admin cannot stand up an enclave with an attacker-supplied key via `insertOrgAction` instead.
- The fail-closed argument in the new `queries.ts` comment holds. With `executeTakeFirst`, an absent `orgId` makes the mongo `$in` conditions deny, and I confirmed that for `view OrgConfig`, `update Org` and `invite User`. `getOrgFromSlugAction` passes the unconditioned check and then throws in the handler, which the new tests pin for both the error shape and the response body.
- No researcher flow regresses: every caller of the two actions moved onto `view OrgConfig`, and of `getPendingUsersAction` on `invite User`, is an admin console page. The cross-org catalog pair has explicit regression tests.

Four comments inline, worst first: one Medium where the data-source read still returns URLs cross-org while the new comment asserts the columns are catalog-level only, one Medium for the unaudited half of MA-5 (follow-up, needs an enum migration), one Low on the denial-path echo in the code-env middleware, and one spelling nit.

PR #948, opened 2026-08-05
OTTER-724: bind researcher-profile lookup to the study

nathanstitt:
Security finding MA-10: `getResearcherProfileByUserIdAction` accepted an arbitrary `userId` next to a `studyId` and only checked the caller's `view` ability on the study. Nothing linked the two, so anyone who could view any single study could read any user's email and profile PII (education, research interests, publication URLs) by editing the `userId` in the researcher-profile URL. Returning `null` for unknown ids also made it a userId-to-account oracle.

## Changes

- The action's middleware now rejects any `userId` the study does not name — `researcherId` or `piUserId` — before the ability check runs. Those are the only two profiles the popover and the full-profile page ever request, so no real use case changes.
- Keyed the popover's React Query cache on `studyId` as well as `userId`, so two studies asking about the same user can't share a cache entry and skip the per-study recheck.
- Tests for the hand-edited-URL denial, an unrelated org member, a same-org user not named on the study, and the legitimate researcher and PI paths.

### DIFF
--- a/src/server/actions/researcher-profile.actions.ts
+++ b/src/server/actions/researcher-profile.actions.ts
@@ -1,9 +1,10 @@
 'use server'
 
-import { Action, z } from '@/server/actions/action'
+import { Action, ActionFailure, z } from '@/server/actions/action'
 import { updateClerkUserName } from '@/server/clerk'
 import { positionSchema, educationSchema, personalInfoSchema, researchDetailsSchema } from '@/schema/researcher-profile'
 import { throwNotFound } from '@/lib/errors'
+import logger from '@/lib/logger'
 
 export const getResearcherProfileAction = new Action('getResearcherProfileAction')
     .middleware(async ({ session }) => ({ id: session?.user.id }))
@@ -157,12 +158,22 @@ export const updateResearchDetailsAction = new Action('updateResearchDetailsActi
 
 export const getResearcherProfileByUserIdAction = new Action('getResearcherProfileByUserIdAction')
     .params(z.object({ userId: z.string(), studyId: z.string() }))
-    .middleware(async ({ params: { studyId }, db }) => {
+    .middleware(async ({ params: { studyId, userId }, db }) => {
         const study = await db
             .selectFrom('study')
-            .select(['orgId', 'submittedByOrgId', 'status'])
+            .select(['orgId', 'submittedByOrgId', 'status', 'researcherId', 'piUserId'])
             .where('id', '=', studyId)
             .executeTakeFirstOrThrow(throwNotFound('Study'))
+
+        // The study's view ability says nothing about whose profile is being requested, so without
+        // this anyone able to view any single study could read every user's email and PII by
+        // editing the userId in the URL. Only the two people the study actually names are exposed.
+        if (userId !== study.researcherId && userId !== study.piUserId) {
+            const msg = `getResearcherProfileByUserIdAction: user ${userId} is not associated with study ${studyId}`
+            logger.error(msg)
+            throw new ActionFailure({ permission_denied: msg })
+        }
+
         return { orgId: study.orgId, submittedByOrgId: study.submittedByOrgId, status: study.status }
     })
     .requireAbilityTo('view', 'Study')
--- a/src/server/actions/researcher-profile.actions.ts
+++ b/src/server/actions/researcher-profile.actions.ts
@@ -1,6 +1,6 @@
 'use server'
 
-import { Action, ActionFailure, z } from '@/server/actions/action'
+import { Action, z } from '@/server/actions/action'
 import { updateClerkUserName } from '@/server/clerk'
 import { positionSchema, educationSchema, personalInfoSchema, researchDetailsSchema } from '@/schema/researcher-profile'
 import { throwNotFound } from '@/lib/errors'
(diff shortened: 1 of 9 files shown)


src/server/actions/researcher-profile.actions.ts
```diff
@@ -157,12 +158,22 @@ export const updateResearchDetailsAction = new Action('updateResearchDetailsActi
...
             .selectFrom('study')
-            .select(['orgId', 'submittedByOrgId', 'status'])
+            .select(['orgId', 'submittedByOrgId', 'status', 'researcherId', 'piUserId'])
             .where('id', '=', studyId)
             .executeTakeFirstOrThrow(throwNotFound('Study'))
+
+        // The study's view ability says nothing about whose profile is being requested, so without
+        // this anyone able to view any single study could read every user's email and PII by
+        // editing the userId in the URL. Only the two people the study actually names are exposed.
+        if (userId !== study.researcherId && userId !== study.piUserId) {
```
therealmarv, line 189: Blocking: this denies a legitimate flow.

`reviewer-preview.tsx:88` passes `values.piUserId`, live Mantine form state, not the persisted row. `study.piUserId` is only written by `onUpdateDraftStudyAction` (`study-request.ts:180`) and `finalizeStudySubmissionAction` (`study-request.ts:342`).

Repro: on a draft proposal, select a PI, click "View as reviewer" (`proposal/footer.tsx:73`, `onClick={openReviewer}` with no save), hover the PI info icon. The DB row still has `piUserId = null`, so this check throws and the popover renders "Profile not available". Same on edit-and-resubmit after changing the PI before resubmitting.

Yjs does not cover this: `use-yjs-form-map.ts:169` syncs `piUserId` into the Y.Doc, but that is Hocuspocus state, not the `study` row.

Suggested fix, which keeps this check strict: have "View as reviewer" await `saveDraft()` before opening the modal, the way `handlePrevious` already does. Widening the check to accept unsaved ids would partly undo the fix.

src/server/actions/researcher-profile.actions.ts
```diff
@@ -157,12 +158,22 @@ export const updateResearchDetailsAction = new Action('updateResearchDetailsActi
...
+            const msg = `getResearcherProfileByUserIdAction: user ${userId} is not associated with study ${studyId}`
+            logger.error(msg)
+            throw new ActionFailure({ permission_denied: msg })
+        }
+
         return { orgId: study.orgId, submittedByOrgId: study.submittedByOrgId, status: study.status }
     })
     .requireAbilityTo('view', 'Study')
```
therealmarv: The association check runs before this ability check, which leaves a smaller oracle.

Middleware executes in registration order (`action.ts:174-178`), so any authenticated user, including one with no view rights on the study, gets two distinguishable errors: `permission_denied: "... is not associated with study ..."` for a wrong guess, versus the generic CASL `permission_denied: "cannot view Study..."` when the guessed `userId` really is the study's `researcherId` or `piUserId`. That still confirms "is user X the researcher/PI on study Y" for an arbitrary study id.

OWASP BOLA and the IDOR Prevention Cheat Sheet both call for the general object-access check first and a single indistinguishable denial for callers without object access.

Fix: chain the association check as a second `.middleware()` after `.requireAbilityTo('view', 'Study')`. `requireAbilityTo` returns `this`, so it chains.

src/server/actions/researcher-profile.actions.ts
```diff
@@ -157,12 +158,22 @@ export const updateResearchDetailsAction = new Action('updateResearchDetailsActi
...
+            .select(['orgId', 'submittedByOrgId', 'status', 'researcherId', 'piUserId'])
             .where('id', '=', studyId)
             .executeTakeFirstOrThrow(throwNotFound('Study'))
+
+        // The study's view ability says nothing about whose profile is being requested, so without
+        // this anyone able to view any single study could read every user's email and PII by
+        // editing the userId in the URL. Only the two people the study actually names are exposed.
+        if (userId !== study.researcherId && userId !== study.piUserId) {
+            const msg = `getResearcherProfileByUserIdAction: user ${userId} is not associated with study ${studyId}`
+            logger.error(msg)
```
therealmarv: Non-blocking: every denial now becomes a Sentry issue.

`action.ts:200` calls `Sentry.captureException(error)` before the `instanceof ActionFailure` branch, and `scrubSentryEvent` only redacts fields, it never returns `null`. So the previously silent `return null` path is now one error log plus one Sentry event per request. A user iterating user ids, exactly the attack this blocks, floods Sentry.

Note that `logger.warn` alone is not enough: `sentry.server.config.ts` registers `consoleLoggingIntegration({ levels: ['error', 'warn'] })`. To actually drop these, filter the event in `beforeSend` or use a denial path that is not captured.

therealmarv (changes requested):
The core fix is right: rejecting any `userId` the study does not name closes the enumeration hole, and the tests exercise it against a real DB.

One blocking item and two follow-ups, all commented inline:

1. Blocking, `researcher-profile.actions.ts:171`. The reviewer preview passes unsaved form state, so selecting a PI on a draft and hovering the info icon now shows "Profile not available". The PR description's claim that the popover only ever requests persisted profiles does not hold for that path.
2. `researcher-profile.actions.ts:179`. The association check runs before `requireAbilityTo`, so a user with no access to the study can still distinguish the two denials and confirm whether user X is on study Y. Two-line reorder.
3. `researcher-profile.actions.ts:173`. Every denial now creates a Sentry issue.

Checked and clear: no `invalidateQueries` call in the repo targets the `researcher-profile` key, so adding `studyId` to it breaks nothing; both consumers route the new `{ error }` shape through `isActionError`.

src/server/actions/researcher-profile.actions.ts
nathanstitt: Fixed as suggested, in 42c382ad: "View as reviewer" now awaits saveDraft() before opening the modal, in both the proposal footer and the edit-and-resubmit footer, so the persisted row names the PI the popover requests — the server check stays strict. Both footer test files gained a test asserting the study row holds the form's title and piUserId before the dialog opens.

src/server/actions/researcher-profile.actions.ts
nathanstitt: Reordered in 42c382ad: the association check now runs strictly after requireAbilityTo — it lives at the top of the handler rather than as a second middleware, because the Sentry thread required the denial to be a returned value and middleware can only deny by throwing (its return merges into ctx and execution continues). The pre-auth middleware also no longer selects researcherId/piUserId, since anything it returns is merged into the ability args and echoed in the CASL denial message, which would have leaked the very ids this hides; a new test asserts a caller without study access gets the same generic "cannot view Study" denial for a correct guess and a wrong one, with no id echo.

src/server/actions/researcher-profile.actions.ts
nathanstitt: Went with the non-captured path in 42c382ad: the denial is now returned as an { error } response from the handler instead of thrown, so it never enters the catch in action.ts where Sentry.captureException fires, and it logs at info so consoleLoggingIntegration's warn/error forwarding never sees it. I preferred this over a beforeSend filter because the filter would need brittle message matching, applies globally, and the event would still be built per request; the unit test pins the returned/info branch and fails if the denial is reverted to the throwing one.

therealmarv (approved):
approved now. Thanks!

PR #949, opened 2026-08-05
OTTER-724: invite acceptance — bearer-token invites, no privilege escalation

nathanstitt:
Fixes MA-9 from the OTTER-724 security review: https://openstax.atlassian.net/browse/OTTER-724

Stacks on #939 — please merge that first.

Scope changed after review — design decision recorded in [this discussion](https://github.com/safeinsights/management-app/pull/949#issuecomment-5207647379): invites are **bearer tokens by design**. Any authenticated user holding the link may accept it, and the membership attaches to *their own* account — including an invite addressed to a different email. The verified-email binding originally implemented here was deliberately reverted. The invariant this PR enforces is **no privilege escalation**:

- the acting identity is `session.user.id`, never a client-supplied `loggedInEmail` (the original bug let an unauthenticated caller grant any account membership — and admin — in any org from an invite id)
- the granted role comes only from the invite row; no caller-supplied input can influence it
- accepting never writes or verifies an email on any Clerk account (the old flow force-verified the invited address on the acceptor — an account-takeover primitive)

Also from review:

- invite claiming is atomic and transactional in `onJoinTeamAccountAction` and `onCreateAccountAction` (unclaimed filter + `returning()`), so the claimed-invite guard is self-enforcing; `onPendingUserLoginAction` is idempotent per-user
- revoke-by-invitee keys off invite-id possession (consistent with the bearer design, and strictly weaker than accepting) instead of unverified Clerk email match; claimed invites are admin-only to remove
- an admin re-invite of an existing contributor promotes them rather than consuming the invite with the role silently dropped
- claimed invites render an "invitation is no longer valid" panel on the join-team and signup pages instead of an infinite spinner; the MFA sign-in path distinguishes "no longer valid" from "join failed", reports via `reportError`, and lands on the join-team page

### Deliberately deferred

The unauthenticated half of `onCreateAccountAction` (review summary item 4): it still creates and force-verifies the invitee's Clerk account from the invite id alone, so a leaked id lets an attacker pre-create that account. The right fix is Clerk's ticket-strategy invitations; deferred as a follow-up rather than changed silently here.

Invite expiry is also still unenforced (`pendingUser` has no `expiresAt`) — needs a migration and a product decision.

### DIFF
--- a/src/app/account/invitation/[inviteId]/create-account.action.ts
+++ b/src/app/account/invitation/[inviteId]/create-account.action.ts
@@ -8,9 +8,11 @@ import { extractClerkCodeAndMessage, isClerkApiError } from '@/lib/errors'
 import { toRecord } from '@/lib/permissions'
 import { clerkClient } from '@clerk/nextjs/server'
 
-// `claim PendingUser` is unconditioned in permissions.ts, so it grants every signed-in user the
-// verb on every invite. The `claimedByUserId is null` guard is what stops one user burning an
-// invite that belongs to (or was already accepted by) somebody else.
+// Invites are bearer credentials by design: holding the invite id is what authorizes acting on
+// it, so `claim PendingUser` is unconditioned in permissions.ts. The `claimedByUserId` guard is
+// what stops one user burning an invite somebody else already accepted; re-claiming by the same
+// user is a no-op success because onCreateAccountAction claims in-transaction and this action
+// runs again from the signup page after sign-in.
 export const onPendingUserLoginAction = new Action('onPendingUserLoginAction')
     .params(z.object({ inviteId: z.string() }))
     .requireAbilityTo('claim', 'PendingUser')
@@ -19,7 +21,7 @@ export const onPendingUserLoginAction = new Action('onPendingUserLoginAction')
             .updateTable('pendingUser')
             .set({ claimedByUserId: session.user.id })
             .where('id', '=', inviteId)
-            .where('claimedByUserId', 'is', null)
+            .where((eb) => eb.or([eb('claimedByUserId', 'is', null), eb('claimedByUserId', '=', session.user.id)]))
             // returning() so an update that matched nothing yields no row and raises, rather than
             // an UpdateResult that looks like success regardless of how many rows were touched.
             .returning('id')
@@ -56,9 +58,12 @@ export const getOrgInfoForInviteAction = new Action('getOrgInfoForInviteAction')
     })
 
 // Two legitimate callers with different rights: an org admin revoking an invite, and the invitee
-// declining it. The invitee may not be a member of the org at all, so no single ability rule covers
-// both — authorization is an explicit either/or below. Previously this action had no check of any
-// kind, letting any caller delete any invite by id.
+// declining it. Invites are bearer credentials — holding the id is the same credential that
+// authorizes accepting one, and declining is strictly weaker, so possession authorizes declining
+// an unclaimed invite. (The previous email-match check was weaker than it looked: Clerk reports
+// unverified addresses, so anyone could add the invitee's address to their own account unverified
+// and delete their invites.) Claimed invites are spent bearer tokens; only an org admin may still
+// remove those rows.
 export const onRevokeInviteAction = new Action('onRevokeInviteAction')
     .params(
         z.object({
@@ -74,52 +79,27 @@ export const onRevokeInviteAction = new Action('onRevokeInviteAction')
 
         const invite = await db
             .selectFrom('pendingUser')
-            .select(['id', 'orgId', 'email'])
+            .select(['id', 'orgId', 'claimedByUserId'])
             .where('id', '=', inviteId)
             .executeTakeFirstOrThrow(() => new ActionFailure({ invite: 'not found' }))
 
         const isOrgAdmin = session.ability.can('revoke', toRecord('PendingUser', { orgId: invite.orgId }))
+        const isBearerOfUnclaimedInvite = invite.claimedByUserId === null
 
-        // The invitee is identified by email rather than claimedByUserId, which is only set on the
-        // signup path and stays null when an existing user opens the invite (join-team page).
-        // Checked against all Clerk addresses so merged/secondary emails match, as the UI does.
-        const isInvitee = await callerOwnsEmail(session.user.clerkUserId, invite.email)
-
-        if (!isOrgAdmin && !isInvitee) {
+        if (!isOrgAdmin && !isBearerOfUnclaimedInvite) {
             throw new ActionFailure({ permission_denied: 'cannot revoke this invite' })
         }
 
         await db.deleteFrom('pendingUser').where('id', '=', invite.id).executeTakeFirstOrThrow()
     })
 
-const callerEmails = async (clerkUserId: string) => {
-    const clerk = await clerkClient()
-    const clerkUser = await clerk.users.getUser(clerkUserId)
-    return clerkUser.emailAddresses
-}
-
-const callerOwnsEmail = async (clerkUserId: string, email: string) => {
-    const emails = await callerEmails(clerkUserId)
-    return emails.some((ea) => ea.emailAddress.toLowerCase() === email.toLowerCase())
-}
-
-// Clerk returns unverified addresses too, and anyone can add an arbitrary address to their own
-// account. Only a verified address proves the caller controls that mailbox, which is the whole
-// basis for accepting an invite addressed to it.
-const callerOwnsVerifiedEmail = async (clerkUserId: string, email: string) => {
-    const emails = await callerEmails(clerkUserId)
-    return emails.some(
-        (ea) => ea.emailAddress.toLowerCase() === email.toLowerCase() && ea.verification?.status === 'verified',
-    )
-}
-
-// Accepting an invite grants org membership (possibly as admin), so the acting identity must come
-// from the session, never from a parameter. The previous version took a `loggedInEmail` param and
-// granted membership to whichever account owned it, then created and force-verified the invite
-// email on that account — an unauthenticated account-takeover primitive (OTTER-724 / MA-9).
-//
-// There is no ability rule to gate this on: the invitee is by definition not yet a member of the
-// inviting org, so authorization is "the session already owns the invited address", checked below.
+// Invites are bearer credentials by design: any authenticated user holding the link may accept it,
+// and the membership attaches to the accepting session's account — a user opening an invite
+// addressed to another email joins under their own identity. The invariant enforced here is no
+// privilege escalation: the acting identity comes from the session, never from a parameter (the
+// pre-fix version took a `loggedInEmail` param and granted membership to whichever account owned
+// it, unauthenticated — OTTER-724 / MA-9), the granted role comes only from the invite row, and
+// nothing writes or verifies email addresses on any Clerk account.
 export const onJoinTeamAccountAction = new Action('onJoinTeamAccountAction')
     .params(
         z.object({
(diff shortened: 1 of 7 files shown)


src/app/account/invitation/[inviteId]/create-account.action.ts
```diff
@@ -16,9 +19,17 @@ export const onPendingUserLoginAction = new Action('onPendingUserLoginAction')
             .updateTable('pendingUser')
             .set({ claimedByUserId: session.user.id })
             .where('id', '=', inviteId)
-            .executeTakeFirstOrThrow()
+            .where('claimedByUserId', 'is', null)
```
Comment (therealmarv) [inline src/app/account/invitation/[inviteId]/create-account.action.ts]: This filter is new here, and combined with `permit('claim', 'PendingUser')` at `permissions.ts:52` (unconditioned, so every signed-in user holds `claim` on every invite) it changes the impact of a hostile claim.

Before this branch, burning an invite id through this action set `claimedByUserId` without granting membership, and the real invitee could still accept. Now the same call blocks them permanently: `getOrgInfoForInviteAction` and `onJoinTeamAccountAction` both gained `claimedByUserId is null`, so the invitee gets an indefinite spinner on the join-team page and the invite can never be used.

Suggest gating this action on `callerOwnsVerifiedEmail` as well, so all three invite-consuming actions apply the same rule.

src/app/account/invitation/[inviteId]/create-account.action.ts
```diff
@@ -80,48 +92,64 @@ export const onRevokeInviteAction = new Action('onRevokeInviteAction')
...
     const clerk = await clerkClient()
     const clerkUser = await clerk.users.getUser(clerkUserId)
-    return clerkUser.emailAddresses.some((ea) => ea.emailAddress.toLowerCase() === email.toLowerCase())
+    return clerkUser.emailAddresses
+}
+
+const callerOwnsEmail = async (clerkUserId: string, email: string) => {
+    const emails = await callerEmails(clerkUserId)
+    return emails.some((ea) => ea.emailAddress.toLowerCase() === email.toLowerCase())
 }
```
Comment (therealmarv) [inline src/app/account/invitation/[inviteId]/create-account.action.ts]: `onRevokeInviteAction` still authorizes the invitee through this unverified variant, while `callerOwnsVerifiedEmail` directly below spells out why that is not sufficient: anyone can add an arbitrary address to their own Clerk account, and Clerk returns unverified addresses here.

Concretely, an attacker who adds the victim's address to their own account without verifying it can delete any pending invite to that address. Destructive, cross-user, no undo.

The revoke check came in with #939 so the fix probably belongs there, but after this PR the stack applies three different standards to the same question: verified for join, unverified for revoke, nothing at all for claim.

src/app/account/invitation/[inviteId]/create-account.action.ts
```diff
@@ -154,19 +181,6 @@ export const onJoinTeamAccountAction = new Action('onJoinTeamAccountAction')
...
-                userId: user.clerkId,
-                emailAddress: invite.email,
-            })
-
-            // auto-verify email (the user has already followed the email invite link)
-            await clerk.emailAddresses.updateEmailAddress(emailAddress.id, { verified: true })
-        }
-
         await updateClerkUserMetadata(siUser.id)
         onUserAcceptInvite(siUser.id)
```
Comment (therealmarv) [inline src/app/account/invitation/[inviteId]/create-account.action.ts:175]: Minor: the claim write just below uses `executeTakeFirst()` with the result unchecked, while the unclaimed check happened back at the invite lookup. Two concurrent accepts both pass that check, one claim update matches zero rows, and the action still reports success.

Row locking prevents an actual double-claim so the impact is low, but doing the claim first with `.returning('id')` (the pattern `onPendingUserLoginAction` now uses) would make the guard atomic.

src/app/account/invitation/[inviteId]/create-account.action.ts
```diff
@@ -193,16 +207,19 @@ export const onCreateAccountAction = new Action('onCreateAccountAction')
...
 
     .handler(async function ({ params: { inviteId, form }, db }) {
+        // Unauthenticated by necessity — this is what creates the account the invite is for — so
+        // the invite id is the only credential. A claimed invite is spent and must not create a
+        // second account or re-grant membership.
         const invite = await db
             .selectFrom('pendingUser')
             .selectAll('pendingUser')
             .where('id', '=', inviteId)
+            .where('claimedByUserId', 'is', null)
```
Comment (therealmarv) [inline src/app/account/invitation/[inviteId]/create-account.action.ts:204]: The "a claimed invite is spent" guard is not self-enforcing on this path: `onCreateAccountAction` never sets `claimedByUserId`.

The only thing that claims the invite on signup is the client-side `onPendingUserLoginAction({ inviteId })` at `signup/page.tsx:93`, whose result is unchecked and which is skipped entirely if `signIn.create` throws, the `needs_second_factor` branch is hit, or the user closes the tab. In those cases the invite stays unclaimed forever, the link remains live, and `getOrgInfoForInviteAction` keeps disclosing the invitee's email and the inviter's name.

Setting `claimedByUserId: siUser.id` inside the same transaction that inserts the `orgUser` row would make the guard enforce itself.

src/app/account/signin/mfa.tsx
```diff
@@ -92,16 +89,15 @@ export const RequestMFA: FC<{ mfa: MFAState }> = ({ mfa }) => {
...
                                 }
 
-                                const email = signInAttempt?.identifier || 'your account'
                                 notifications.show({
                                     color: 'green',
-                                    message: `You've successfully linked your SafeInsights accounts under ${email}.`,
+                                    message: `You've successfully joined ${name}.`,
                                 })
                                 await auth.getToken({ skipCache: true })
                             } catch {
```
therealmarv: After the reorder this also swallows `getOrgInfoForInviteAction` failures, so a claimed or missing invite shows "Failed to accept your invitation. Please try again." when retry can never succeed, with no `reportError` or Sentry breadcrumb.

Given the dead "Login with existing account" path noted in the review summary, this is the toast most affected users will actually hit. Worth distinguishing "invite no longer valid" from "join failed" and reporting the error.

therealmarv (changes requested):
The core fix is right. The acting identity now comes from `session.user.id`, the invited address must be verified on the caller's own Clerk account, and nothing writes emails onto a foreign account any more. That closes the takeover primitive, and the comments explaining the reasoning made this easy to follow.

Requesting changes on two themes: the new unclaimed filters turn a pre-existing weakness into a permanent invite-kill, and three UI paths now dead-end. Inline comments cover the server side. The UI ones are in files this PR does not touch, so they are listed here.

### 1. `invitation/[inviteId]/page.tsx:101-108`: the "Login with existing account" button is now a guaranteed failure

It renders only when no account holds the invited address, and the copy addresses "you already have a SafeInsights account", meaning under a different email. It links to `/account/signin?invite_id=`, whose only handler is `mfa.tsx`, which now calls `onJoinTeamAccountAction` with no `loggedInEmail`. `callerOwnsVerifiedEmail` therefore always rejects, because an account under a different address never has the invited one verified. The user gets a red failure toast and lands on the dashboard with no membership, on a path the page labels "Strongly recommended", and retrying can never work.

Either remove the button or change the copy to something achievable, such as verifying the address on the existing account first, or requesting a new invite.

### 2. `join-team/page.tsx:44`: the client check is weaker than the server rule

`userEmails` is built from `user.emailAddresses` with no verification filter, while `callerOwnsVerifiedEmail` requires `verification.status === 'verified'`. Clerk's `getUserList({ emailAddress })` matches unverified secondary addresses, so someone in that state gets redirected here by `page.tsx:53`, passes `emailMismatch`, clicks Accept, and receives a generic "Unable to join team" with no hint that verification is what is missing. Filtering `userEmails` on verified status routes them to the mismatch panel that already has actionable copy.

### 3. `join-team/page.tsx:81`: claimed invites hang on a spinner

`getOrgInfoForInviteAction` now throws for claimed invites, the wrapped `useQuery` rethrows, so `data` stays undefined and `if (isLoading || !org || !user)` renders "Loading account invitation" forever. Reachable by ordinary means: browser Back after a successful accept, a second tab, or a refetch-on-focus during the 500 ms delay before `router.push`. Rendering the "This invitation is no longer valid" panel from `page.tsx:127` on `isError` would fix it. Same shape at `signup/page.tsx:252`.

### 4. `create-account.action.ts:257-260` (outside the diff): the other half of MA-9 is still open

`onCreateAccountAction` remains unauthenticated, creates a Clerk account for `invite.email` with a caller-supplied password, and force-verifies the address with no proof of mailbox control. With a leaked invite id an attacker still pre-creates and owns the invitee's account. Clerk's native Invitations plus `signUp.create({ strategy: 'ticket', ticket })` auto-verify precisely because the token was mailed to that address, which is the proof step this hand-rolled flow is missing. Fine to defer, but worth recording on the card rather than dropping silently.

### Checked and fine

The `getOrgInfoForInviteAction` / `onJoinTeamAccountAction` reorder in `mfa.tsx` is required by the new unclaimed filter and is correct. Dropping the Clerk `getUserList` fallback in `onJoinTeamAccountAction` is safe now that the user resolves from `session.user.id`. Removing `confirmPassword` from the action schema loses no validation, since it was never checked server-side and the `superRefine` match stays in the form. The zero-column `select from "user"` in the new test is valid Postgres, so that assertion holds.

src/app/account/invitation/[inviteId]/create-account.action.ts
Comment (nathanstitt) [reply src/app/account/invitation/[inviteId]/create-account.action.ts]: Design decision went the other way: invites are bearer tokens, so possession of the id is the credential and the email checks are gone everywhere. A hostile claim now requires the same credential that would let the caller accept the invite outright, and a same-user re-claim is a no-op success since `onCreateAccountAction` claims in-transaction.

src/app/account/invitation/[inviteId]/create-account.action.ts
Comment (nathanstitt) [reply src/app/account/invitation/[inviteId]/create-account.action.ts]: Agreed the unverified email match was the weakest of the three standards — it is gone. Revoke now keys off invite-id possession to match the bearer design: any authenticated holder may decline an unclaimed invite (strictly weaker than accepting it, which the same possession authorizes), while claimed invites can only be removed by an org admin.

src/app/account/invitation/[inviteId]/create-account.action.ts
Comment (nathanstitt) [reply src/app/account/invitation/[inviteId]/create-account.action.ts:175]: Done — the claim now runs first, inside the transaction, with the unclaimed filter plus `.returning('id')`, so the loser of a concurrent accept fails with "invite not found" instead of reporting success, and a failed membership insert rolls the claim back.

src/app/account/invitation/[inviteId]/create-account.action.ts
Comment (nathanstitt) [reply src/app/account/invitation/[inviteId]/create-account.action.ts:204]: Done — `claimedByUserId` is now set in the same transaction as the `orgUser` insert, so the claimed-invite guard is self-enforcing. The signup page still calls `onPendingUserLoginAction`, which became a same-user no-op success rather than the only thing doing the claiming.

src/app/account/signin/mfa.tsx
nathanstitt: Done — the two failures are distinguished ("This invitation is no longer valid" vs "Failed to accept your invitation"), both report via `reportError`, and both now land on the join-team page — which shows a persistent invalid-invite panel or a retryable Accept — instead of silently following `redirect_url`.

nathanstitt:
Thanks for the thorough review — it prompted a design decision that changed the PR's scope.

**Items 1 and 2 (dead "Login with existing account" button, client check weaker than server rule):** resolved by reverting the rule rather than the UI. Invites are bearer tokens by design: any authenticated user holding the link may accept it, and the membership silently attaches to their own account, including when the invite was addressed to a different email. The verified-email requirement is gone, the signin-with-existing-account path works again, and the join-team email-mismatch panel is removed to match the server. What stays from the original fix is the no-escalation invariant: acting identity from the session (never a parameter), the granted role read only from the invite row, and no email writes or verification on any Clerk account during acceptance — each pinned by a test.

**Item 3 (claimed-invite spinner hang):** fixed. A failing `getOrgInfoForInviteAction` now renders the "This invitation is no longer valid" panel on both the join-team and signup pages instead of loading forever.

**Item 4 (unauthenticated `onCreateAccountAction`):** deliberately deferred and now recorded in the PR description — agreed the proper fix is Clerk's ticket-strategy invitations, which is a larger change than this PR should carry.

One behavior fix that fell out of re-review under the bearer model: an admin re-invite of an existing contributor now promotes them; previously the invite was consumed with the role change silently dropped.

therealmarv (approved):
approved! thx

PR #950, opened 2026-08-05
OTTER-686: Add researcher outputs-pending screen for code processing

stetradis:
## Summary

Researchers whose code is approved and actively executing in the enclave previously landed on the
`code-approved` screen (`CodePostDecisionView`), which showed a "Code approved" decision banner and
the submitted code, describing the run only in the future tense ("Your code will now proceed to run
in the secure enclave"). This PR adds a dedicated **STEP 4 "Verify outputs"** screen that tells the
researcher their code is processing and that they'll be notified when outputs are ready.

Labeled STEP 4 per the upcoming agreements-page removal (1 request, 2 proposal, 3 code, 4 outputs).
The code screens' existing `STEP 4` labels and the "Proceed to step 5" CTA get renumbered with that
change, not this one.

## Changes

**New screen** — `outputs-pending-screen.tsx`
- STEP 4 "Verify outputs" header with the study title
- Informative status alert: `Outputs not ready, code processing started • <date>`, dated from the
  job's `CODE-APPROVED` status change, falling back to the current execution stage's `startedAt`
- Copy explaining the code is running in the secure enclave and the researcher will be notified
- Nav: "Previous step" → `/[orgSlug]/study/[studyId]/view/code` (preserving `returnTo`), plus
  "Back to my studies"
- Deliberately shows the same banner for every execution stage (`JOB-PROVISIONING`, `JOB-PACKAGING`,
  `JOB-READY`, `JOB-RUNNING`) — the researcher doesn't see enclave internals
- Not-found alerts when the study isn't submitted, has no submitted job, or has no execution stage yet

**Routing** — `researcher-screen-rules.ts`
- Replaced the old `codeDecision === 'CODE-APPROVED' || isExecuting` rule with two narrower ones:
  `outputs-pending` when approved **and** executing, `code-approved` when approved but not yet
  executing
- Both now require a *live* approval, so an executing job with no live code decision (a duplicate
  `CODE-SUBMITTED` tipping decisions behind submissions) resolves to `code-under-review` rather than
  `code-approved`. That matches the pill for the same state — `isStaleCodeDecision` drops the stale
  approval and researchers have no `JOB-*` labels — where the old rule contradicted it.
- Registered `outputs-pending` in `screens.ts`

**Errored-run behavior preserved (OTTER-598, 43898)**
- When a run errors before the reviewer records `FILES-APPROVED`/`FILES-REJECTED`, the error stays
  hidden from the researcher. `isExecuting` stays true, so routing now holds on `outputs-pending`
  instead of `code-approved` — still no jump to the results screen, still no disclosure of the failure.
- Updated the comments documenting this invariant (`researcher-screen-rules.ts`,
  `awaitingFilesDecisionOnError` in `state.ts`, the OTTER-598 fixture in `consistency.test.ts`) —
  all three still named `code-approved` as the destination.

**Small cleanup** — extracted the bullet character used in status-alert titles as
`STATUS_ALERT_SEPARATOR` in `status-alert.tsx` and used it in
both the new banner and the existing reviewer errored banner.

## Testing

- New unit tests in
  `outputs-pending-screen.test.tsx`:
  header/step content, nav hrefs (incl. `returnTo` passthrough), identical banner across all four
  execution stages, `CODE-APPROVED` date vs. `startedAt` fallback, and the no-submitted-job case
- Updated `resolve.test.ts` and
  `view/page.test.tsx` for the new destination,
  including the hidden-error and late-`CODE-SCANNED` race cases. The `expectSubmittedCodeCanExpand()`
  assertions were dropped from the two executing-window tests — the submitted code now lives on
  `/view/code` rather than `/view` (still asserted in the packaging-error test).


src/app/[orgSlug]/study/[studyId]/_screens/outputs-pending-screen.tsx
```diff
@@ -0,0 +1,75 @@
...
+    dashboardHref,
+    returnTo,
+}: Pick<ScreenComponentProps, 'study' | 'orgSlug' | 'dashboardHref' | 'returnTo'>) {
+    if (!isSubmittedStudy(study)) {
+        return <AlertNotFound title="Study was not found" message="No such study exists" />
+    }
+
+    const job = await latestSubmittedJobForStudy(study.id)
+    if (!job) {
+        return <AlertNotFound title="No submission found" message="This study has no submitted code to review." />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/outputs-pending-screen.tsx]: These three guards (not-submitted, no job, no stage — including the exact alert copy) are a verbatim copy of `ReviewerOutputsPendingScreen`. Two thoughts: (1) we could extract a small shared helper so the guards can't drift between the two screens; (2) the copied message "This study has no submitted code to review." is reviewer-voiced — on the researcher's own screen something like "This study has no submitted code yet." might read better.

src/app/[orgSlug]/study/[studyId]/_screens/outputs-pending-screen.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+        const { org, study, job } = await setupExecuting('JOB-READY')
+        await db
+            .insertInto('jobStatusChange')
+            .values({ studyJobId: job.id, status: 'CODE-APPROVED', createdAt: approvedDate })
+            .execute()
+        await renderScreen(study, org.slug)
+        expect(screen.getByTestId('status-alert')).toHaveTextContent(dayjs(approvedDate).format('MMM DD, YYYY'))
+    })
+
+    it('falls back to stage startedAt when CODE-APPROVED is missing', async () => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/outputs-pending-screen.test.tsx:95]: This asserts "some date rendered, not 'undefined'", which would pass no matter which branch produced the date. Since the helper only inserts the JOB-READY row, we could make the fallback observable by reading that row's `createdAt` back from the db and asserting `dayjs(createdAt).format('MMM DD, YYYY')` appears — then the test actually fails if the fallback stops using `stage.startedAt`.

nathanstitt (approved):
looks good, up to you if you want to add a shared helper as suggested

src/lib/study-screen/researcher-screen-rules.ts
```diff
@@ -6,12 +6,14 @@ import { awaitingFilesDecisionOnError } from './state'
...
+    // records a FILES-* decision (awaitingFilesDecisionOnError) — until then isExecuting stays true and
+    // the researcher holds on the outputs-pending page below, which discloses no error (OTTER-598, 43898).
     ['study-results', { when: (s) => s.hasResults && !awaitingFilesDecisionOnError(s) }],
 
-    // Code approved (or actively running): the approved/executing code screen.
-    ['code-approved', { when: (s) => s.codeDecision === 'CODE-APPROVED' || s.isExecuting }],
+    // Code approved and executing in the enclave: researcher outputs-pending screen (OTTER-686).
+    ['outputs-pending', { when: (s) => s.codeDecision === 'CODE-APPROVED' && s.isExecuting }],
+    // Code approved but not yet executing: the approved code screen.
+    ['code-approved', { when: (s) => s.codeDecision === 'CODE-APPROVED' }],
```
therealmarv, line 18: The tracked spec still documents the rule this replaces. [`docs/study-screens-logic.md` L135](https://github.com/safeinsights/management-app/blob/5e5e79faba009dbf02bc3e1949e3c471271dfb1e/docs/study-screens-logic.md#L135) reads:

```
| 2   | `codeDecision === 'CODE-APPROVED'` or `isExecuting`   | `code-approved` |
```

and `outputs-pending` appears nowhere in the document. Line 4 of this file points readers at "spec §6" as the contract for this table, so whoever extends it from the doc re-adds `|| isExecuting` to the `code-approved` rule, which then matches before `outputs-pending` and silently takes the new screen out of service for every executing study.

Worth adding the row and splitting row 2 in this PR, along with the screen-to-component note at [L190](https://github.com/safeinsights/management-app/blob/5e5e79faba009dbf02bc3e1949e3c471271dfb1e/docs/study-screens-logic.md#L190).

src/lib/study-screen/resolve.test.ts
```diff
@@ -77,8 +77,10 @@ describe('resolveScreen (researcher)', () => {
     it('approved decision → code-approved', () => {
         expect(resolveScreen('researcher', state({ codeDecision: 'CODE-APPROVED' }), ctx).screen).toBe('code-approved')
     })
-    it('executing window → code-approved', () => {
-        expect(resolveScreen('researcher', state({ isExecuting: true }), ctx).screen).toBe('code-approved')
+    it('executing window → outputs-pending', () => {
+        expect(
+            resolveScreen('researcher', state({ codeDecision: 'CODE-APPROVED', isExecuting: true }), ctx).screen,
+        ).toBe('outputs-pending')
```
therealmarv, line 83: This was the only assertion covering `isExecuting` with no live code decision, and it now carries a `codeDecision`, so the `code-under-review` outcome the PR description documents has no test. That outcome is the load-bearing part of dropping `|| isExecuting`, so it is worth pinning:

```ts
it('executing with no live code decision → code-under-review', () => {
    expect(resolveScreen('researcher', state({ isExecuting: true }), ctx).screen).toBe('code-under-review')
})
```

src/app/[orgSlug]/study/[studyId]/_screens/outputs-pending-screen.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    banner={<ProcessingBanner approvedAt={approvedAt} />}
+                />
+                <Group justify="space-between">
+                    <ButtonLink href={previousHref} variant="subtle" leftSection={<CaretLeftIcon />}>
+                        Previous step
+                    </ButtonLink>
+                    <ButtonLink href={dashboardHref as Route} variant="filled" size="md">
+                        Back to my studies
+                    </ButtonLink>
+                </Group>
```
Comment (therealmarv) [inline src/app/[orgSlug]/study/[studyId]/_screens/outputs-pending-screen.tsx:53]: Not a change request for this PR. The card's AC asks for focus to move to the destination's main heading after either navigation, and for the document title to update on client-side routes. Neither `ButtonLink` does that.

No sibling screen does either (`ReviewerOutputsPendingScreen`, `SecondaryAnalysisView`, `CodePostDecisionView`), and there is no route-change focus or title handler anywhere in `src/app`, `src/components`, or `src/hooks`, so this reads as an app-wide gap rather than something to solve here. Flagging it only so it does not bounce at QA review: it likely wants its own card for a router-level handler applied to every step screen, plus confirmation from the card author that per-screen was never the intent.

therealmarv (approved):
Approving.

Re-verified the routing split and the OTTER-598 invariant: `outputs-pending` sits above `code-approved` so the narrower rule wins first, and a bare `JOB-ERRORED` keeps `isExecuting` true, so a run-stage failure holds on the new screen without disclosing the error while a packaging-stage failure still lands on `code-approved`. Dropping `|| isExecuting` is correct, it makes the destination agree with the pill where the old rule contradicted it.

Three threads. Only the first is something I would fix before merge, and it is a doc file that is not in the diff. The other two are a test gap and a scope question.

Also +1 to @nathanstitt's guard-extraction thread. His point that "This study has no submitted code to review." is reviewer-voiced on the researcher's own screen is worth fixing either way.

src/lib/study-screen/researcher-screen-rules.ts
therealmarv: Wonder how [`docs/study-screens-logic.md`](https://github.com/safeinsights/management-app/blob/5e5e79faba009dbf02bc3e1949e3c471271dfb1e/docs/study-screens-logic.md) should be handled when it needs updating along the process?!

stetradis: Updated docs/study-screens-logic.md - split the researcher row into outputs-pending (approved && executing) and code-approved (approved only), added !awaitingFilesDecisionOnError to row 1, and brought the reviewer table up to date with the OTTER-667/668 screens. Also added the component mapping note for OutputsPendingScreen/ReviewerOutputsPendingScreen sharing guardExecutionStage.

To prevent future drift, added a rule to CLAUDE.md that triggers when the screen rules files are modified - it reminds the agent to update the doc tables.

src/app/[orgSlug]/study/[studyId]/_screens/outputs-pending-screen.tsx
Comment (stetradis) [reply src/app/[orgSlug]/study/[studyId]/_screens/outputs-pending-screen.tsx:53]: Added a comment to the ticket for follow-up, ty!

PR #951, opened 2026-08-05
OTTER-675 follow-up: deliver the first submit click and let Tab leave the editor

therealmarv:
see [OTTER-675](https://openstax.atlassian.net/browse/OTTER-675), specifically the two failing rows in the Aug 05 QA verdict: a failed "Submit decision" flagged only one of two unresolved fields and moved no focus, and the Decision feedback field was a keyboard trap.

## 1. The first submit click was never delivered

Not a validation bug. `attemptSubmit` already flagged both fields and moved focus in one pass; the click never reached it.

Measured in Chromium against the real components, splitting `mousedown` from `mouseup` and re-measuring in between:

| measurement | value |
| --- | --- |
| Submit button `top` before `mousedown` | 881.17px |
| Submit button `top` after `mousedown`, before `mouseup` | 902.47px |
| shift | 21.30px |
| error slot height, before to after | 0px to 20.30px |
| pointer still over the button | false |
| `elementFromPoint` at the press coordinates | the page `DIV`, not the button |

`mousedown` on the button blurs the editor, the blur handler raises the "Enter your feedback" message synchronously, React flushes it before `mouseup`, and the button moves out from under the pointer. `mouseup` lands elsewhere, so the browser fires `click` on an ancestor. The one message on screen came from the blur, not from validation, which is why the radio group stayed unflagged and focus stayed on the button. The second click worked because the message was already in place.

Three controls were affected: "Submit decision", "Previous step" (same row, also confirmed swallowed, with the second click navigating fine), and the radio group's own blur, which raises its message above the options and moves the same row.

**Fix:** emptiness is flagged on a submit attempt only, never on blur. `showFeedbackError` and `showDecisionError` collapse into one `hasAttemptedSubmit` flag, and `onFeedbackBlur` / `onDecisionBlur` are gone from the hook and the section. Nothing can now be inserted during the mousedown-to-mouseup window, so all three paths are fixed by one change, and the hook holds less state than before.

This is also what the AC asks for: "Every empty required field is visually flagged ... at the moment of the failed click (not only on blur or initial typing)". The blur trigger was our own addition, not a review request. Further research on the pattern agrees: prefer validating on submit and avoid reflowing interactive targets, and W3C ARIA guidance is explicit that a required field should not be marked invalid before the user has tried to submit. `preventDefault` on the button's `mousedown`, or activating on `pointerdown`, were both considered and rejected as workarounds that change standard button behavior.

Unchanged deliberately: the over-limit message still appears live while typing, well before any click, so it is not a shift source. Reserving vertical space for the message would also have worked but changes spacing that has not been signed off, and it still shifts when a message wraps on a narrow viewport.

## 2. Tab was trapped in every Lexical editor

`TabIndentationPlugin` cancels the Tab keydown as soon as there is a range selection, then indents in a list or inserts a literal tab anywhere else. Measured: six Tabs and three Shift+Tabs never moved focus off `#outputs-decision-feedback`, the content gained nine tab characters, every Tab keydown reached document with `defaultPrevented: true`, and a 30-key traversal starting in the editor visited one element. Escape did release it, but nothing on screen says so, and WCAG 2.1.2 allows that escape hatch only if the user is advised of the method.

The plugin's own props do not help: `preventDefault()` runs before `$canIndent` is consulted. Lexical's docs discourage the plugin for exactly this reason and state that by default Tab follows the browser's focus order, which is what removing it restores. Nothing else in our stack cancels Tab (`@lexical/rich-text` registers the command only to reset capitalization and returns false).

**Fix:** removed from all three editor surfaces (single-user, collaborative, and the legacy `EditableText`). Nothing is lost: the plugin only indented inside a list, and in a paragraph it inserted a literal tab, which nobody wants in a feedback field. The list case still works through the toolbar's Indent / Outdent buttons, which enable (and so become focusable) exactly when the selection is in a list, and Tab can now reach them. `EscapeFocusPlugin` stays.

The plugin has been copied into a new editor twice along with the rest of the plugin list, so it is now banned via `no-restricted-imports` with a message pointing at the trap.

## Tests

- The unit test named for the trap row could not fail: Lexical's Tab handler returns early unless `$getSelection()` is a `RangeSelection`, and calling `focus()` on the contenteditable never establishes one in jsdom, so the keydown was never cancelled and `userEvent.tab()` always moved focus. It passed throughout the period the real browser was trapping. Removed, with a comment recording why and where the row is now covered.
- `tests/study-flow.spec.ts`: the blank-submit helper now clicks into the empty editor first, which is the precondition that used to break it, and asserts both messages, `aria-invalid="true"` on both radios, focus on the feedback field, and no modal, all from one click. It then asserts Tab moves focus off the editor without inserting a tab character and keeps going until it reaches the radios.
- New `src/hooks/use-outputs-decision.test.tsx`: pristine form unflagged, both fields flagged by a single `attemptSubmit`, over-limit reported without a submit attempt, each message clearing with its own field, and the modal confirming once everything is resolved.

## Not touched

The code review screen uses the same blur-validation pattern on its radio group, but its submit button is disabled while the form is invalid, so there is no click to lose, and OTTER-647 asked for that blur validation deliberately. Left alone.

Both copy deviations from the QA verdict ("Output files" versus "Outputs files", and the limit counted in words rather than characters) are unchanged and still waiting on PM sign-off.

## Verified in the browser

Same measurements as the diagnosis, re-run against these changes:

| check | before | after |
| --- | --- | --- |
| shift across `mousedown` on Submit | 21.30px | 1.00px, Mantine's own button press transform, and the pointer stays over the button |
| first click, caret in empty editor | feedback message only, `aria-invalid [null, null]`, focus stuck on the button | both messages, `["true", "true"]`, focus on the feedback field |
| first click on "Previous step" | URL unchanged | navigates |
| radio focused, nothing selected | focus stuck on the button | focus moves to the radio group |
| 12 Tabs from the editor | 1 stop, 9 tab characters inserted, every keydown cancelled | Bold, Italic, Underline, Link, Bullet list, Numbered list, radio, Previous step, Submit decision; 0 tab characters, no keydown cancelled |

Regressions checked and clean: pristine submit still flags both and focuses the editor, the over-limit message still appears live while typing with the radios untouched, the invalid to valid to modal cycle opens the modal with the right body copy, and toolbar Indent still produces nested `<ul>` markup.

CI is green, including e2e at `retries: 0` with 0 flaky, and the two tests that exercise the extended helper (`Successful results review`, `Error log review`) both pass. Those run the collaborative editor, so both editor implementations are covered.

## Regression audit

- `OutputsDecisionSection` / `useOutputsDecision` have two consumers, the errored screen and the available-outputs screen, both through `outputs-review-panel`. Typecheck confirms none was missed.
- `Editor`'s `onBlur` prop is still used by five other surfaces, so dropping this call site orphans nothing.
- Both `no-restricted-imports` entries were confirmed to fire independently, so the new one does not shadow the existing `@tanstack/react-query` ban.
- `TabNode` is registered by Lexical core rather than by the plugin, so feedback already saved containing tab characters still hydrates.
- No existing unit or e2e test asserted tab indentation.

## The same pattern elsewhere, deliberately out of scope

Expanding on "Not touched" above, since the audit went further than the code review screen. Both the code review and the proposal review screens validate their editor on blur, render the message in a zero-height slot, and sit above an enabled "Back" button, which is the same lost-click configuration. Their Submit is disabled while the form is invalid, so the reviewer is never blocked from learning what is missing, and the exposure is limited to a swallowed first click on "Back". It predates this branch, since clicking "Back" always blurred the editor, and freeing Tab does not worsen it because keyboard activation does not depend on hit testing. Worth its own card rather than widening this one, especially as OTTER-647 asked for that blur validation deliberately.

## Review follow-ups

- The new hook test builds its Lexical states with `lexicalJson` from `@/lib/lexical` rather than a local helper, so it runs on the same root / paragraph / text shape the editor actually produces instead of a bare text node at the root.
- The rationale for leaving `TabIndentationPlugin` out now lives only in the `no-restricted-imports` entry, which is what the person re-adding it will read. The three editor surfaces carry a one-line pointer to it instead of three copies of the same paragraph to keep in sync.

Re-verified in the browser after these changes, against the single-user editor surface:

| check | result |
| --- | --- |
| first click on Submit, caret in the empty editor | button shifts 1.00px (Mantine's press transform) and the pointer stays on it, both messages appear, `aria-invalid` is `["true", "true"]`, focus lands on the feedback field, no modal |
| first click on "Previous step", caret in the empty editor | navigates |
| Tab from the editor | Bold, Italic, Underline, Link, Bullet list, Numbered list, radio; 0 tab characters inserted |

Regressions clean: a pristine submit still flags both fields and focuses the editor, the over-limit message still appears live while typing ("Feedback exceeds the 300 word limit.") with the radios untouched, a valid form still opens the modal with the feedback-only copy, and toolbar Indent still produces nested `<ul>` markup.

Side effects checked: `widgetBlurHandler` and the `Editor` `onBlur` prop both keep other consumers, so neither is orphaned by the call sites this branch drops, and no test or code path referenced tab indentation.

### DIFF
--- a/src/components/editable-text/single-user-editor.tsx
+++ b/src/components/editable-text/single-user-editor.tsx
@@ -145,10 +145,7 @@ export function SingleUserEditor({
                 />
                 <HistoryPlugin />
                 <ListPlugin />
-                {/* No TabIndentationPlugin: it cancels the Tab keydown, so focus never leaves the
-                    editor and Tab types a literal tab instead. Nothing is lost, because that plugin
-                    only indented inside a list, which the toolbar's Indent / Outdent buttons still
-                    do (they enable exactly there, and Tab can now reach them). */}
+                {/* No TabIndentationPlugin: banned in eslint.config.mjs, which carries the why. */}
                 <EscapeFocusPlugin />
                 <LinkPlugin validateUrl={isValidUrl} />
                 {onChange && <EditorChangePlugin onChange={onChange} />}
--- a/src/components/editable-text/single-user-editor.tsx
+++ b/src/components/editable-text/single-user-editor.tsx
@@ -9,7 +9,6 @@ import { HistoryPlugin } from '@lexical/react/LexicalHistoryPlugin'
 import { LexicalErrorBoundary } from '@lexical/react/LexicalErrorBoundary'
 import { ListPlugin } from '@lexical/react/LexicalListPlugin'
 import { LinkPlugin } from '@lexical/react/LexicalLinkPlugin'
-import { TabIndentationPlugin } from '@lexical/react/LexicalTabIndentationPlugin'
 import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext'
 import type { EditorState } from 'lexical'
 
@@ -146,7 +145,9 @@ export function SingleUserEditor({
                 />
                 <HistoryPlugin />
                 <ListPlugin />
-                <TabIndentationPlugin />
+                {/* No TabIndentationPlugin: it cancels the Tab keydown, so focus never leaves the
+                    editor and Tab types a literal tab instead. Indentation lives on the toolbar's
+                    Indent / Outdent buttons, which are themselves reachable by Tab. */}
                 <EscapeFocusPlugin />
                 <LinkPlugin validateUrl={isValidUrl} />
                 {onChange && <EditorChangePlugin onChange={onChange} />}
--- a/src/components/editable-text/single-user-editor.tsx
+++ b/src/components/editable-text/single-user-editor.tsx
@@ -146,8 +146,9 @@ export function SingleUserEditor({
                 <HistoryPlugin />
                 <ListPlugin />
                 {/* No TabIndentationPlugin: it cancels the Tab keydown, so focus never leaves the
-                    editor and Tab types a literal tab instead. Indentation lives on the toolbar's
-                    Indent / Outdent buttons, which are themselves reachable by Tab. */}
+                    editor and Tab types a literal tab instead. Nothing is lost, because that plugin
+                    only indented inside a list, which the toolbar's Indent / Outdent buttons still
+                    do (they enable exactly there, and Tab can now reach them). */}
                 <EscapeFocusPlugin />
                 <LinkPlugin validateUrl={isValidUrl} />
                 {onChange && <EditorChangePlugin onChange={onChange} />}
--- a/src/hooks/use-outputs-decision.test.tsx
+++ b/src/hooks/use-outputs-decision.test.tsx
@@ -1,10 +1,9 @@
 import { act, createTestQueryWrapper, describe, expect, faker, it, renderHook } from '@/tests/unit.helpers'
+import { lexicalJson } from '@/lib/lexical'
 import { ERRORED_OUTPUTS_FEEDBACK_MAX_WORDS, OUTPUTS_DECISION_ERRORS } from '@/lib/outputs-review'
 import { useOutputsDecision } from './use-outputs-decision'
 
 const LAB = 'Rice Lab'
-
-const lexicalText = (text: string) => JSON.stringify({ root: { type: 'text', text } })
 
 const renderDecision = () =>
     renderHook(
@@ -34,7 +33,7 @@ describe('useOutputsDecision', () => {
     it('leaves an untouched empty form unflagged until a submit attempt', () => {
         const { result } = renderDecision()
 
-        act(() => result.current.onFeedbackChange(lexicalText('')))
+        act(() => result.current.onFeedbackChange(lexicalJson('')))
 
         expect(result.current.feedbackError).toBeUndefined()
         expect(result.current.decisionError).toBeUndefined()
@@ -55,7 +54,7 @@ describe('useOutputsDecision', () => {
 
         act(() =>
             result.current.onFeedbackChange(
-                lexicalText(
+                lexicalJson(
                     Array.from({ length: ERRORED_OUTPUTS_FEEDBACK_MAX_WORDS + 1 }, (_, i) => `word${i}`).join(' '),
                 ),
             ),
@@ -76,7 +75,7 @@ describe('useOutputsDecision', () => {
         expect(result.current.decisionError).toBeUndefined()
         expect(result.current.feedbackError).toBe(OUTPUTS_DECISION_ERRORS.feedbackEmpty(LAB))
 
-        act(() => result.current.onFeedbackChange(lexicalText('Outputs look clean, no PII observed.')))
+        act(() => result.current.onFeedbackChange(lexicalJson('Outputs look clean, no PII observed.')))
 
         expect(result.current.feedbackError).toBeUndefined()
     })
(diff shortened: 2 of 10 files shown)


src/hooks/use-outputs-decision.test.tsx
```diff
@@ -0,0 +1,94 @@
+import { act, createTestQueryWrapper, describe, expect, faker, it, renderHook } from '@/tests/unit.helpers'
+import { ERRORED_OUTPUTS_FEEDBACK_MAX_WORDS, OUTPUTS_DECISION_ERRORS } from '@/lib/outputs-review'
+import { useOutputsDecision } from './use-outputs-decision'
+
+const LAB = 'Rice Lab'
+
+const lexicalText = (text: string) => JSON.stringify({ root: { type: 'text', text } })
```
nathanstitt: We already have `lexicalJson()` in `@/lib/lexical` (the proposal/resubmit hook tests use it) that builds the real root → paragraph → text shape. This local helper works because `extractTextFromLexicalNode` happens to accept a bare text node at the root, but that's a shape the editor never actually produces — reusing `lexicalJson` would keep these tests on the same wire format as production and drop the duplicate.

src/components/editable-text/single-user-editor.tsx
```diff
@@ -146,7 +145,10 @@ export function SingleUserEditor({
                 />
                 <HistoryPlugin />
                 <ListPlugin />
-                <TabIndentationPlugin />
+                {/* No TabIndentationPlugin: it cancels the Tab keydown, so focus never leaves the
```
nathanstitt: Minor: this same four-line comment now lives in all three editor surfaces, and the `no-restricted-imports` message already delivers the rationale to exactly the person who'd re-add the plugin. A one-liner pointing at the eslint ban would be less to keep in sync across the three copies.

nathanstitt (approved):
looks good but should switch to `lexicalJson` as pointed out in review

src/hooks/use-outputs-decision.test.tsx
therealmarv: Switched, the local helper is gone: [use-outputs-decision.test.tsx#L2](https://github.com/safeinsights/management-app/blob/7f75261f4cd329e085386c51e994dcd085ddb513/src/hooks/use-outputs-decision.test.tsx#L2) in 7f75261.

You are right that the shape was the giveaway. `extractTextFromLexicalNode` reads a bare text node at the root only because it recurses on `type === 'text'` before it ever looks for `children`, so the helper was passing for a reason that had nothing to do with what the editor serializes. With `lexicalJson` the empty case now goes through root to paragraph to text and still counts 0 words, and the over-limit case still counts N+1, so every assertion holds on the real wire format.

src/components/editable-text/single-user-editor.tsx
therealmarv: Agreed, three copies to keep in sync was the wrong trade. All three are now a single line pointing at the ban, and the paragraph lives only in the `no-restricted-imports` entry where the person re-adding the plugin will actually meet it:

- [single-user-editor.tsx#L148](https://github.com/safeinsights/management-app/blob/7f75261f4cd329e085386c51e994dcd085ddb513/src/components/editable-text/single-user-editor.tsx#L148)
- [collaborative-editor.tsx#L415](https://github.com/safeinsights/management-app/blob/7f75261f4cd329e085386c51e994dcd085ddb513/src/components/editable-text/collaborative-editor.tsx#L415)
- [editable-text.tsx#L188](https://github.com/safeinsights/management-app/blob/7f75261f4cd329e085386c51e994dcd085ddb513/src/components/editable-text.tsx#L188)

I kept a marker line rather than deleting the comment outright, so the gap in the plugin list next to `ListPlugin` reads as deliberate instead of looking like something that got dropped.

PR #952, opened 2026-08-06
OTTER-668: address #922 review 

Rod0352:
## Summary

#922 was auto-closed as merged when #928 (stacked on the OTTER-668 branch) landed, which mechanically bypassed the pending changes-requested review. This picks up the review items that weren't already addressed on the OTTER-676 branch:

- **One routing fact.** Rule 1b now keys on `resultsDisplayStatus === 'RUN-COMPLETE'` instead of re-deriving the result-status priority from four booleans, and both outputs screens guard on the same projected state (`projectStudyState(raw)`) their rules route on. Routing and rendering read one fact, so the "stranded on Outputs not found" divergence can't occur; `latestSubmittedJobForStudy` now supplies only the panel's job payload. The errored screen's anchoring comment is gone because the code no longer needs it.
- **Shared, typed timestamp helper.** The per-screen timestamp finders are replaced by `latestStatusAt(statusChanges, status)` in `study-job-status.ts`, typed to `StudyJobStatus` (a `'RUN_COMPLETE'` typo now fails to compile) and selected by timestamp rather than array position, so it carries no hidden ordering contract. The banner date degrades to an undated banner rather than a dead end if the payload job lacks the status row.
- **1c comment.** Reworded to say what the rule actually catches now that 1a/1b claim the undecided results states.

Review notes:

- `resolve.test.ts`'s undecided-results fixture now sets `resultsDisplayStatus: 'RUN-COMPLETE'` — hand-built states must carry the fact the rule keys on, the same coherence argument as the review.
- The remaining #922 item (the silent #921 revert) never reached main: merging main into OTTER-676 re-resolved it, and rule 1a still uses `awaitingFilesDecisionOnError`. No change needed.
- The e2e selector scoping and the researcher errored-view e2e leg were already fixed on OTTER-676 before it merged.

Validation: `pnpm run checks` clean (typecheck, eslint, prettier, validate-actions); 166 unit tests pass across the touched areas (study-screen suite, study-job-status, both outputs screen suites) against a real Postgres 16.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
@@ -25,6 +25,7 @@ import type { FileType, StudyJobStatus } from '@/database/types'
 import { getStudyAction } from '@/server/actions/study.actions'
 import { fetchEncryptedJobFilesAction } from '@/server/actions/study-job.actions'
 import { latestJobForStudy } from '@/server/db/queries'
+import { rawStudyStateForStudy } from '@/server/db/study-state-query'
 import { setupStudyAction } from '@/tests/db-action.helpers'
 import { ReviewerOutputsAvailableScreen } from './reviewer-outputs-available-screen'
 import type { ScreenComponentProps } from './types'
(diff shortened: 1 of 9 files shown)


src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
```diff
@@ -180,14 +190,15 @@ describe('ReviewerOutputsAvailableScreen before decryption', () => {
...
     it('shows a not-found alert when the study has no submitted job', async () => {
         const { org, study } = await setupStudyAction({ orgSlug: 'openstax', orgType: 'enclave', createJob: false })
+        const raw = await requireRawState(study.id)
         ;(useParams as Mock).mockReturnValue({ orgSlug: org.slug, studyId: study.id })
-        await renderScreen(study, org.slug)
+        await renderScreen({ study, raw }, org.slug)
         expect(screen.getByText('No submission found')).toBeInTheDocument()
     })
 
     it('shows a not-found alert when the job has not completed a run', async () => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx:191]: The behavioral change this PR makes to the guard isn't pinned by a test yet. With the old `availableTimestamp` guard, a job carrying RUN-COMPLETE *plus* a FILES-APPROVED/REJECTED would still render the review panel; the new `resultsDisplayStatus !== 'RUN-COMPLETE'` guard correctly 404s it in step with routing. Since `insertTestStudyJobData` only writes one status row, none of the current cases hit that path — worth adding a case that inserts both statuses (e.g. an extra `jobStatusChange` row) and asserts "Outputs not found", so a regression back to timestamp-sniffing can't slip in. Same story for the errored screen's guard (JOB-ERRORED + FILES-APPROVED → "No error found").

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
```diff
@@ -73,26 +74,35 @@ async function seedArtifact(jobId: string, files: { name: string; content: strin
     }
 }
 
+const requireRawState = async (studyId: string) => {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx]: `requireRawState` (and the `ScreenInputs` pick below) are duplicated verbatim in reviewer-outputs-errored-screen.test.tsx. As more screens take `raw`, every screen test will grow the same copy — we could hoist this into `@/tests/unit.helpers` (or have `insertTestStudyJobData`/`setupStudyAction` return `raw` alongside `study`/`job`) and keep the screen tests to just the render call.

nathanstitt (approved):
Thanks for picking this back up!  a few minor things i'd missed earlier in the review

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx:191]: Added in 5d8bbf94 — both screens pin the decided-run path via an extra FILES-APPROVED row.

src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/_screens/reviewer-outputs-available-screen.test.tsx]: Done in 5d8bbf94 — moved to unit.helpers.

PR #955, opened 2026-08-09
Trivy: update `nanoid` to `3.3.17` to address security vulnerability

stetradis:


nathanstitt (approved):
Straightforward override pin matching the existing pattern. Checks pass.

PR #956, opened 2026-08-10
allow manual dismissal of error toasts

jbwilson8:
Issue: [OTTER-665](https://openstax.atlassian.net/browse/OTTER-665) (Error toasts should stay visible until manually dismissed)

### Problem

Every notification inherits an 8-second auto-close from the two global `<Notifications>` mounts (`NOTIFICATION_DISPLAY_MS`). Error toasts inherit it too, which is a problem: `reportError` appends a Sentry event id (`\nReference: ${eventId}`) that users are asked to quote when reporting a bug, and 8 seconds isn't enough time to read a failure message and copy a 32-character id before it vanishes.

### Solution

Added a shared constant in `src/lib/constants.ts`:

```ts
// Error toasts persist until dismissed so users can read the message and copy the
// Sentry reference id before it disappears.
export const ERROR_NOTIFICATION_OPTIONS = {
    color: 'red',
    autoClose: false,
    withCloseButton: true,
} as const
```

and spread it into every red toast in the app — one source of truth instead of per-site literals.

**`reportError`** (`src/components/errors.tsx`) covers the whole `reportError` / `reportMutationError` / `handleMutationErrorsWithForm` family (~35 call sites) with no signature change.

**13 direct call sites** bypassed `reportError` and would otherwise still auto-dismiss:

| Area | File |
| --- | --- |
| Researcher profile "Save failed" | `use-personal-info-section.ts`, `use-positions-section.ts`, `use-education-section.ts`, `use-research-details-section.ts` |
| Study request / resubmit | `use-save-draft.ts`, `use-resubmit-proposal.ts`, `use-ide-files.ts` |
| Files | `use-encrypted-files-panel.ts`, `file-upload.tsx`, `file-drop-overlay.tsx` |
| Dashboard / admin / auth | `delete-draft-button.tsx`, `agent-context.tsx`, `add-sms-mfa.tsx` |

`agent-context.tsx` already had a one-off `autoClose: false`; it now uses the shared constant.

### Scope notes

- Success/info toasts are unchanged — they keep the 8s `NOTIFICATION_DISPLAY_MS` default. `NOTIFICATION_DISPLAY_MS` itself was not touched.
- `withCloseButton: true` is redundant with Mantine 8's per-notification default, but is set explicitly so the sticky-toast contract is self-documenting and survives a future theme default change.
- Left alone: the yellow "Popup blocked" toast (`use-workspace-launcher.tsx`) and the session-expired toasts (`activity-context.tsx`), which already opt out of auto-close for their own reasons and shouldn't adopt red styling.
- Excluded: three `color: 'red'` hits in `code-envs-view.tsx`, `code-env-history-view.tsx`, and `proposal-submitted.tsx` — these are Badge/status config maps, not notifications.

### Testing

- Added a regression test in `errors.test.tsx` asserting error toasts are shown with `autoClose: false` and `withCloseButton: true`, so a future refactor can't silently restore auto-dismiss.
- Loosened exact-object `toHaveBeenCalledWith` assertions to `expect.objectContaining` in `errors.test.tsx` and `upload-study-job-code.test.ts` — they asserted on the full notification payload and would break on any added default.

[OTTER-665]: https://openstax.atlassian.net/browse/OTTER-665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/errors.test.tsx
+++ b/src/components/errors.test.tsx
@@ -108,11 +108,13 @@ describe('reportError', () => {
     it('calls notifications.show with the default title and error message', () => {
         const errorMsg = 'Test error'
         reportError(errorMsg)
-        expect(notificationsShowSpy).toHaveBeenCalledWith({
-            color: 'red',
-            title: 'An error occurred',
-            message: expect.stringMatching(new RegExp(`${errorMsg}\\nReference: [a-f0-9]{32}`)),
-        })
+        expect(notificationsShowSpy).toHaveBeenCalledWith(
+            expect.objectContaining({
+                color: 'red',
+                title: 'An error occurred',
+                message: expect.stringMatching(new RegExp(`${errorMsg}\\nReference: [a-f0-9]{32}`)),
+            }),
+        )
     })
 
     it('calls notifications.show with a custom title if provided', () => {
@@ -120,11 +122,20 @@ describe('reportError', () => {
         const customTitle = 'Custom Title'
         reportError(errorInstance, customTitle)
         const errorString = errorInstance.toString()
-        expect(notificationsShowSpy).toHaveBeenCalledWith({
-            color: 'red',
-            title: customTitle,
-            message: expect.stringMatching(new RegExp(`${errorString}\\nReference: [a-f0-9]{32}`)),
-        })
+        expect(notificationsShowSpy).toHaveBeenCalledWith(
+            expect.objectContaining({
+                color: 'red',
+                title: customTitle,
+                message: expect.stringMatching(new RegExp(`${errorString}\\nReference: [a-f0-9]{32}`)),
+            }),
+        )
+    })
+
+    it('shows error toasts that persist until manually dismissed', () => {
+        reportError('Test error')
+        expect(notificationsShowSpy).toHaveBeenCalledWith(
+            expect.objectContaining({ autoClose: false, withCloseButton: true }),
+        )
     })
 })
 
--- a/src/contexts/edit-resubmit/hooks/use-resubmit-proposal.ts
+++ b/src/contexts/edit-resubmit/hooks/use-resubmit-proposal.ts
@@ -5,6 +5,7 @@ import { type UseFormReturnType } from '@mantine/form'
 import { useMutation } from '@/common'
 import { resubmitProposalAction } from '@/server/actions/study-request'
 import { actionResult } from '@/lib/utils'
+import { ERROR_NOTIFICATION_OPTIONS } from '@/lib/constants'
 import { Routes } from '@/lib/routes'
 import { type ProposalFormValues } from '@/app/[orgSlug]/study/[studyId]/proposal/schema'
 import { type ResubmitNoteValue } from '@/app/[orgSlug]/study/[studyId]/edit-and-resubmit/schema'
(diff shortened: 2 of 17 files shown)


src/lib/constants.ts
```diff
@@ -29,3 +29,11 @@ export const APP_SHELL = {
 
 // 8 seconds
 export const NOTIFICATION_DISPLAY_MS = 8000
+
+// Error toasts persist until dismissed so users can read the message and copy the
+// Sentry reference id before it disappears.
+export const ERROR_NOTIFICATION_OPTIONS = {
```
nathanstitt, line 35: Now that error toasts never self-dismiss, nothing bounds how many can pile up — neither `<Notifications>` mount sets a `limit`, and the only `notifications.hide` calls in the app are the id-scoped session/inactivity ones in `activity-context.tsx`.

The realistic case is a retry loop or a burst of failed mutations (e.g. autosave failing repeatedly while offline) stacking N identical red toasts in the top-right, each needing its own click. Two cheap options if we want to bound it:

- add `limit={...}` to the two `<Notifications>` mounts, or
- give the constant a stable `id` so repeat errors replace rather than stack — though that would collapse genuinely distinct errors too, so the `limit` route is probably the safer default.

Not a blocker, just the natural consequence of removing the thing that used to clean up after itself.

src/lib/constants.ts
```diff
@@ -29,3 +29,11 @@ export const APP_SHELL = {
...
+    color: 'red',
+    autoClose: false,
+    withCloseButton: true,
```
nathanstitt, line 38: Agreed with the reasoning in the PR description for setting this explicitly rather than leaning on Mantine 8's default — the sticky-toast contract is only meaningful if the close affordance is guaranteed, so pinning it here is the right call even though it's currently redundant.

src/components/errors.test.tsx
```diff
@@ -108,23 +108,34 @@ describe('reportError', () => {
...
+        expect(notificationsShowSpy).toHaveBeenCalledWith(
+            expect.objectContaining({
+                color: 'red',
+                title: customTitle,
+                message: expect.stringMatching(new RegExp(`${errorString}\\nReference: [a-f0-9]{32}`)),
+            }),
+        )
+    })
+
+    it('shows error toasts that persist until manually dismissed', () => {
```
nathanstitt, line 134: Nice — this is the assertion that actually protects the behavior change. Worth noting the loosening to `objectContaining` in the two tests above slightly widens what they accept, so having this one pin `autoClose`/`withCloseButton` explicitly keeps the net coverage from regressing.

src/lib/constants.ts
nathanstitt: we def need to set a limit here, it can be high even say 10, but probably 3 is better?  Otherwise if we happen to have a loop we could potentially litter the screen with thousands

nathanstitt (approved):
looks good @jbwilson8 but probably do need a limit as pointed out below. thanks!

src/contexts/edit-resubmit/hooks/use-resubmit-proposal.ts
```diff
@@ -53,9 +54,9 @@ export function useResubmitProposal({ studyId, form, noteForm, yjsForm, tabSessi
         },
         onError: (error) => {
             notifications.show({
+                ...ERROR_NOTIFICATION_OPTIONS,
                 title: 'Failed to resubmit proposal',
                 message: error instanceof Error ? error.message : 'An unexpected error occurred. Please try again.',
```
nathanstitt, line 59: not your change but we should use `errorToString` like the `use-save-draft` does instead of instanceOf

PR #957, opened 2026-08-10
OTTER-649 follow-up: add an indeterminate scan status so an unscanned result is not a finding

therealmarv:
see [OTTER-649](https://openstax.atlassian.net/browse/OTTER-649)

Pairs with safeinsights/iac#215. **This PR must reach an environment before that one** (see Deploy order below).

## Why

QA showed `Trivy Filesystem Scan: (!) Vulnerabilities found` on a scan that had never run. The downloaded log said:

```
Trivy Filesystem Scan: no results

SonarQube Quality Gate: OK
  new_violations: OK
```

The root cause is in the CodeBuild scanner and is fixed in the iac PR. This side is the other half: the reviewer UI could only express two outcomes, so a log that reported no verdict had to become either a pass or a finding. `parseTrivyStatus` accepted one exact clean phrase and mapped everything else, including `no results` and any unrecognized text, to `FAILED`.

That answers the question left open in the card's comments. There is a third outcome, and it is not rare. Trivy has no analyzer for R, so a *successful* scan of a typical submission examines nothing: the last good QA scan (job `019eadb3-6cc5-7b1c-8da1-5f925b4a50e6`) produced a Trivy report with no `Results` key, an SBOM with zero components, and a summary reporting its only target as not scanned. A scan that clears nothing is not the same as a scan that finds nothing.

## Changes

- `ScanToolStatus` gains `INDETERMINATE`: the scan reported, but not a verdict. `null` keeps its existing meaning of "has not reported yet" and still drives "Scan in progress…".
- `parseTrivyStatus` reads the single status phrase the scanner now leads its section with, rather than inferring a verdict from the shape of the detail lines. Only `no vulnerabilities found` and `vulnerabilities found` are honored. `nothing scanned`, `scan did not complete`, the legacy `no results`, and anything unrecognized are indeterminate. A fallback still reads the older `Trivy Filesystem Scan Results` findings header as a finding so logs already stored do not lose their verdict.
- Both patterns are matched against whole trimmed lines, and the legacy header is only consulted when there is no status line at all, so nothing appearing in the detail lines (a scanned file path, an upstream CVE title) can promote a result to a finding.
- `ScanRowValue` takes a per-status label map instead of a passed/failed pair. Copy lives in `TRIVY_LABELS` and `SONARQUBE_LABELS`.
- The indeterminate icon is amber (`yellow-10`) rather than the red used for a real finding, reusing the "action needed" pairing the design system already applies to `WarningCircle` in `StatusAlert`. A passed row still carries no icon at all.
- Tests cover all four Trivy markers, the legacy findings header, line-scoping against detail-line text, the new SonarQube `not available` marker, and the indeterminate row's text and icon color. Two fixtures are pinned verbatim to real QA logs: the 82-byte aborted-scan log behind the reported defect, and the 96-byte log from the last successful scan.

## Copy is provisional

The indeterminate state reads **"Needs review"**, taken from the card's own overview ("we should indicate whether it passed or whether it needs human review"). There is no Figma frame for this state, so both the label and the amber icon treatment are placeholders pending UX. The label is a single string in `TRIVY_LABELS` and the color a single entry in `SCAN_ICON_COLORS`; changing either touches no logic.

Two Trivy cases collapse into it: Trivy examined nothing, and Trivy produced no report. Neither is a finding and neither is a clean result. Whether a Data Partner should be able to tell those apart is a UX question; a separate label would need a second Figma frame for a distinction they probably cannot act on differently.

## Not changed, deliberately

- `parseSonarqubeStatus`. Its existing "anything but OK needs review" behavior is already correct for the scanner's new `not available` marker, and the card is explicit that we can confirm a SonarQube pass but never a SonarQube failure. So SonarQube keeps two states, and `SONARQUBE_LABELS` maps failed and indeterminate to the same text.
- The download rule. `logFile != null` shows Download, which already matches the logs-present and logs-absent Figma pairs, and a log explaining that a scan did not complete is still worth downloading.

## Deploy order

This PR is safe in either order on its own: new markers map to the new state and existing phrases keep their meaning, so it reads both old and new logs correctly. The iac PR is not, because its new Trivy markers would fall through this repo's old parser and render as "Vulnerabilities found", which is the defect being fixed. So this ships first.

## Verification

- `pnpm run checks` passes (typecheck, eslint, prettier, action validation).
- Full unit suite passes: 264 files, 2578 tests.
- All eleven scan states were rendered locally and checked against the acceptance criteria and the Figma frames: no icon in the panel title, the two tool rows always present and in order, Trivy pass/fail wording with the exclamation icon only on the fail, SonarQube pass/needs-review wording likewise, and Download shown only when a log file exists. States 1 to 8 match Figma nodes 4905-6027, 5199-94, 4905-14275, 5109-6072, 4905-14412, 5199-106, 4905-14349 and 5199-130 respectively. The two indeterminate states and the not-yet-reported state have no frame yet.

Full end-to-end verification against a real scan still needs a dev/QA stack deploy of the iac side, since the scanner scripts only re-sync when the stack changeset executes and a PR preview skips it.

[OTTER-649]: https://openstax.atlassian.net/browse/OTTER-649

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
@@ -50,11 +50,29 @@ function DatasetPills({ names }: { names: string[] }) {
     )
 }
 
+type ScanStatusLabels = Record<ScanToolStatus, string>
+
+// "Needs review" is the card's own phrasing for the case where we cannot state an outcome with
+// confidence. Trivy reaches it two ways: it examined nothing (no analyzer for R, no lockfile to
+// read), or it produced no report at all. Neither is a finding and neither is a clean bill of
+// health. Pending UX sign-off on whether those two should read differently to a Data Partner.
+const TRIVY_LABELS: ScanStatusLabels = {
+    PASSED: 'No vulnerabilities found',
+    FAILED: 'Vulnerabilities found',
+    INDETERMINATE: 'Needs review',
+}
+
+// SonarQube has no third label: a failing gate and an unresolvable one both need the same human look.
+const SONARQUBE_LABELS: ScanStatusLabels = {
+    PASSED: 'Passed',
+    FAILED: 'Needs review',
+    INDETERMINATE: 'Needs review',
+}
+
 type ScanRowProps = {
     label: string
     status: ScanToolStatus | null
-    passedLabel: string
-    failedLabel: string
+    labels: ScanStatusLabels
     testId: string
 }
 
@@ -63,19 +81,11 @@ function ScanWarningIcon({ isVisible }: { isVisible: boolean }) {
     return <WarningCircle size={20} color="var(--mantine-color-red-9)" data-icon="warning" aria-hidden="true" />
 }
 
-// A tool's result: plain text when it passed, a warning icon plus the failure
-// phrasing when it failed, and a neutral pending note while the scan has not
-// reported (status null). Deliberately no "pass" icon, and never a fabricated
-// pass/fail when the status is unknown; we only flag what needs a human (OTTER-649).
-function ScanRowValue({
-    status,
-    passedLabel,
-    failedLabel,
-}: {
-    status: ScanToolStatus | null
-    passedLabel: string
-    failedLabel: string
-}) {
+// A tool's result: plain text when it passed, a warning icon plus the relevant phrasing when it did
+// not, and a neutral pending note while the scan has not reported (status null). Deliberately no
+// "pass" icon, and never a fabricated pass/fail when the status is unknown; we only flag what needs
+// a human (OTTER-649).
+function ScanRowValue({ status, labels }: { status: ScanToolStatus | null; labels: ScanStatusLabels }) {
     if (status === null) {
         return (
             <Text size="sm" c="dimmed">
@@ -88,17 +98,17 @@ function ScanRowValue({
         <Group gap={4} wrap="nowrap" align="center">
             <ScanWarningIcon isVisible={!passed} />
             <Text size="sm" fw={600}>
-                {passed ? passedLabel : failedLabel}
+                {labels[status]}
             </Text>
         </Group>
     )
 }
 
-function ScanRow({ label, status, passedLabel, failedLabel, testId }: ScanRowProps) {
+function ScanRow({ label, status, labels, testId }: ScanRowProps) {
     return (
         <Group gap="xs" wrap="nowrap" align="center" data-testid={testId}>
             <Text size="sm">{label}</Text>
-            <ScanRowValue status={status} passedLabel={passedLabel} failedLabel={failedLabel} />
+            <ScanRowValue status={status} labels={labels} />
         </Group>
     )
 }
@@ -130,15 +140,13 @@ function ScanLogBody({ scan }: { scan: JobScanResult }) {
             <ScanRow
                 label="Trivy Filesystem Scan:"
                 status={scan.trivy}
-                passedLabel="No vulnerabilities found"
-                failedLabel="Vulnerabilities found"
+                labels={TRIVY_LABELS}
                 testId="security-scan-trivy"
             />
             <ScanRow
                 label="SonarQube Quality Gate:"
                 status={scan.sonarqube}
-                passedLabel="Passed"
-                failedLabel="Needs review"
+                labels={SONARQUBE_LABELS}
                 testId="security-scan-sonarqube"
             />
         </Stack>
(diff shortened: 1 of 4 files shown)


src/server/db/queries.ts
```diff
@@ -573,14 +578,36 @@ export type JobScanResult = {
...
+    const phrase = log
+        .match(/trivy (?:filesystem|image) scan:[ \t]*(.*)/i)?.[1]
+        ?.trim()
+        .toLowerCase()
+    const verdict = phrase && TRIVY_VERDICTS.get(phrase)
+    if (verdict) return verdict
+    // Logs written before the scanner emitted a status phrase headed their findings with
+    // "<label> Results" and listed vulnerabilities beneath it. Keep reading those as findings so
+    // already-stored logs do not lose their verdict.
+    if (/trivy (?:filesystem|image) scan results/i.test(log)) return 'FAILED'
```
nathanstitt: This fallback tests anywhere in the log, so a new-format log whose status phrase is unrecognized (e.g. `nothing scanned`) but whose detail lines happen to contain the words "scan results" would flip to FAILED instead of indeterminate. Since the recognized-phrase check runs first this only bites on the indeterminate paths, but worth confirming the new scanner output on the iac side can never echo that legacy header in its detail section — if that's guaranteed, we're fine as-is.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
```diff
@@ -88,17 +98,17 @@ function ScanRowValue({
         <Group gap={4} wrap="nowrap" align="center">
             <ScanWarningIcon isVisible={!passed} />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx]: The indeterminate row gets the same red warning icon as a real finding, so the only thing distinguishing "Needs review" from "Vulnerabilities found" is the text. Given the whole point of the PR is that this state is *not* a finding, an amber/neutral icon treatment might signal that better — probably one to bundle into the pending UX sign-off already flagged in the label comment rather than change now.

src/server/db/scan-log.test.ts
```diff
@@ -9,13 +9,39 @@ import { jobScanResultForJob, parseTrivyStatus, parseSonarqubeStatus } from './q
...
     'Target: package-lock.json',
     '  HIGH CVE-2024-1234 lodash 4.17.0 (fix: 4.17.21) - Prototype pollution',
 ].join('\n')
 const SONAR_OK = 'SonarQube Quality Gate: OK'
 const SONAR_ERROR = ['SonarQube Quality Gate: ERROR', '  new_coverage: ERROR'].join('\n')
 
+// Verbatim from QA (job 019fd838-f3da-73c8-afe2-1c72c9142161, 2026-08-06): the scanner aborted before
+// Trivy ran, and its failure handler still posted this as a completed scan. The old parser read it as
+// a vulnerability finding, which is the defect this card was reopened for.
+const QA_ABORTED_SCAN_LOG = [
```
nathanstitt, line 28: Really like these verbatim QA fixtures with the job IDs and dates pinned in the comments — the defect log and the last-known-good log as regression anchors make the intent of the parser change self-documenting.

nathanstitt (approved):
looks good when paired with iac pr

src/server/db/queries.ts
therealmarv: Good catch, and I'd rather not rest this on a guarantee about iac output. Fixed in 98be201b so the parser doesn't depend on one.

Both patterns are now matched against whole trimmed lines, and the legacy header is only consulted when there is **no** status line at all rather than merely an unrecognized one. So a `nothing scanned` log can no longer be flipped by its detail lines regardless of what they contain:

https://github.com/safeinsights/management-app/blob/98be201b/src/server/db/queries.ts#L586-L607

For the record, the case was reachable. A `nothing scanned` log now carries Trivy's appended Report Summary, whose rows are scanned target paths, so a submitted file named `Trivy Filesystem Scan Results.txt` would have done it. Unlikely, but this is a reviewer-facing security status, so line-scoping is worth the few extra lines.

Two tests pin it: one with a `Target:` line containing the legacy header text alongside a `nothing scanned` status, and one with the header embedded mid-line.

src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx
Comment (therealmarv) [reply src/app/[orgSlug]/study/[studyId]/review/submitted-code-section.tsx]: Agreed, and I went ahead and changed it rather than deferring, because it turned out not to need a design decision. The codebase already pairs `WarningCircle` with `yellow-10` for "action needed" in `StatusAlert`'s `action` variant, so amber for the indeterminate row reuses an existing token and icon pairing instead of inventing a treatment:

https://github.com/safeinsights/management-app/blob/98be201b/src/app/%5BorgSlug%5D/study/%5BstudyId%5D/review/submitted-code-section.tsx#L72-L79

Reasoning for not waiting: red for "we don't know" is itself an unreviewed design choice, and it is the more misleading of the two. Leaving it red until UX weighs in means shipping the wrong signal in the meantime.

It stays flagged as provisional in the PR description alongside the label, and it is one entry in `SCAN_ICON_COLORS` if UX wants something else. Verified visually across all eleven states: in the state where an amber Trivy row sits directly above a red SonarQube row the difference is clear, though subtler than I'd like at 20px, which is worth raising with UX.

PR #958, opened 2026-08-10
OTTER-639: fix invite banner not showing on new user signups

stetradis:
## Summary

Fixes [OTTER-639](https://openstax.atlassian.net/browse/OTTER-639) — the "You have been added to {org}" confirmation banner was not shown to new users joining an organization.

Follow-up to #899, which QA rejected on Staging: the banner worked for an existing user joining a new org, but never appeared for a newly-invited user creating an account. This PR fixes that race and a second path that never set the flag at all.

## Problem 1 — the reveal raced a redirect and lost

`JoinedOrgBanner` revealed itself on a 500ms `setTimeout`, and that same timer also cleared the one-shot `si:joined-org` flag. The delay existed so that a keyless user's *transient* dashboard mount — the one that happens before `RequireUserKey` redirects them to key setup — would unmount and cancel the timer, preserving the flag for the real landing.

That transient mount lasts exactly as long as an async `userKeyExistsAction` round-trip, so its duration is network-bound: under 500ms locally, over it on Staging. When the timer won, the flag was spent on a dashboard the user was already being redirected away from, and the banner never appeared on the one they landed on.

Raising the delay would only widen the window, so the timer is gone. The banner now waits on the key check itself — the same signal that decides the redirect, unwrapped with `actionResult` exactly as `RequireUserKey` does when making the identical call — and reveals only once that check confirms a key. Both failure modes are now deterministic rather than probabilistic:

- not yet keyed (`!hasKey`) → never consumes the flag
- a check that throws (an `ActionError`, e.g. a session that has not resolved yet) → never consumes the flag

This also removes the 500ms lag for users who already have a key.

**Why only new users hit this:** the existing-user path (`join-team`) routes to `/account/keys?redirect_url=<orgDashboard>` when a key is needed, so the dashboard is never mounted while keyless. The new-user path goes signup → MFA → `/dashboard` → `RequireUserKey` → key setup, which is what creates the transient mount.

## Problem 2 — a third join path never set the flag

`signin/mfa.tsx` also accepts invitations (`onJoinTeamAccountAction`) and pushes the user to the joined org's dashboard, but it still showed the old toast instead of calling `markOrgJoined`. #899 replaced that toast with the banner everywhere else and missed this call site.

The path is reached when the invited email matches no existing account: the invite page offers "Login with existing account", which is the only thing that sets `?invite_id=`, and the join runs in the MFA success handler. That's the account-linking case — invited at one address, signing in with an account under another.

[OTTER-639]: https://openstax.atlassian.net/browse/OTTER-639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/dashboard/joined-org-banner.tsx
+++ b/src/components/dashboard/joined-org-banner.tsx
@@ -18,9 +18,10 @@ export function JoinedOrgBanner() {
         const joined = sessionStorage.getItem(JOINED_ORG_STORAGE_KEY)
         if (!joined) return
 
+        let cancelled = false
         const revealOnceKeyed = async () => {
             const hasKey = actionResult(await userKeyExistsAction())
-            if (!hasKey) return
+            if (!hasKey || cancelled) return
 
             sessionStorage.removeItem(JOINED_ORG_STORAGE_KEY)
             setOrgName(joined)
@@ -28,6 +29,9 @@ export function JoinedOrgBanner() {
         // Anything short of a definite key — not yet keyed, or a check that threw — leaves the flag
         // in place for the dashboard the user actually lands on.
         revealOnceKeyed().catch(() => {})
+        return () => {
+            cancelled = true
+        }
     }, [])
 
     if (!orgName) return null
--- a/src/components/dashboard/joined-org-banner.tsx
+++ b/src/components/dashboard/joined-org-banner.tsx
@@ -1,6 +1,7 @@
 'use client'
 
 import { JOINED_ORG_STORAGE_KEY } from '@/lib/joined-org'
+import { actionResult } from '@/lib/utils'
 import { userKeyExistsAction } from '@/server/actions/user-keys.actions'
 import { Alert, Text, useMantineTheme } from '@mantine/core'
 import { CheckCircleIcon } from '@phosphor-icons/react'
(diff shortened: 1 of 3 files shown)


src/components/dashboard/joined-org-banner.tsx
```diff
@@ -1,28 +1,33 @@
...
+
             sessionStorage.removeItem(JOINED_ORG_STORAGE_KEY)
             setOrgName(joined)
-        }, REVEAL_DELAY_MS)
-
-        return () => clearTimeout(timer)
+        }
+        // Anything short of a definite key — not yet keyed, or a check that threw — leaves the flag
+        // in place for the dashboard the user actually lands on.
+        revealOnceKeyed().catch(() => {})
```
nathanstitt, line 31: The old timer version cancelled on unmount, but this async check can't be cancelled — so a *keyed* user whose dashboard unmounts while the check is in flight (quick navigation away) still spends the one-shot flag, and the banner is never seen. The keyless transient mount is safe (`hasKey` is false there), but we lose the "only spend the flag on a mount the user actually sees" guarantee for keyed mounts. A `cancelled` flag in the effect cleanup would restore it:

```ts
let cancelled = false
const revealOnceKeyed = async () => {
    const hasKey = actionResult(await userKeyExistsAction())
    if (!hasKey || cancelled) return
    ...
}
revealOnceKeyed().catch(() => {})
return () => { cancelled = true }
```

Edge-casey, so fine to punt — but it's a two-line guard.

nathanstitt (approved):
looks good, ty!

nathanstitt:
Heads-up: this PR and #959 both rewrite the same invite-accept block in the MFA success handler (`src/app/account/signin/mfa.tsx`). #959 restructures the block so invite acceptance runs before the key-generation redirect, but it branched before this PR and still shows the old success toast; this PR replaces that toast with `markOrgJoined` for the OTTER-639 banner. Whichever merges second will hit a conflict there — when resolving, keep `markOrgJoined` inside #959's restructured block, otherwise the MFA join path regresses OTTER-639.

PR #959, opened 2026-08-10
OTTER-655 follow-up: resolve the first-key landing from account state instead of a query param

therealmarv:
see [OTTER-655](https://openstax.atlassian.net/browse/OTTER-655), specifically the [QA report of 2026-07-26](https://openstax.atlassian.net/browse/OTTER-655?focusedCommentId=246437).

## The reported failure

A freshly invited Research Lab user finished signup, stored their first key, and landed on `/dashboard` ("My dashboard") instead of `/openstax-lab/dashboard`. The AC asks for the inviting organization's dashboard on a first key, and "My dashboard" only on a reset.

The confirmation modal picked its destination from `searchParams.get('redirect_url')`. Two of the three entry points to `/account/keys` set that parameter; the third, the `RequireUserKey` guard, pushes a bare url, and that is the one the real signup flow goes through:

```
invite -> signup -> /account/mfa -> recovery codes -> router.push('/dashboard')
       -> AppShell mounts RequireUserKey -> push('/account/keys')   <- no redirect_url
       -> safeRedirectUrl(null, Routes.dashboard) -> /dashboard
```

So the branch was keyed on "was a parameter passed" while the AC is keyed on account history. The existing test passed only because it set `?redirect_url=` by hand, which is the condition that does not hold in production.

## What changed

**The destination now comes from account state, with `redirect_url` as an override.**

- New `getKeyPageStateAction` answers the landing question server-side, which the client guard cannot, and reports whether the account already holds a key in the same round trip. It returns the org dashboard when the account belongs to **exactly one** org, and `/dashboard` otherwise. This is deliberately a fallback rather than a claim about invitation provenance: `org_user` records membership, not who invited whom, so where the answer could be ambiguous the action declines instead of guessing. A fresh invited signup always has exactly one membership, which is the failing case.
- `page.tsx` resolves it and passes it down; `postKeyRedirect` branches on `isRegenerating`, so a reset returns to "My dashboard" and ignores any stray parameter.

**Key generation no longer eats the invite on the way through.** `mfa.tsx` returned early on `redirectToKeyGeneration`, before the block that accepts a pending invite, so a keyless user signing in from an invite link generated a key and arrived as a non-member with their destination discarded. The key redirect now runs last, after the invite is accepted.

This one is a regression from #881 rather than an unrelated bug: the early return predates it, but #881 changed `onUserSignInAction` from "only if some org needs a key" to "every user needs a key", which is what routes lab users into it.

```diff
-    if (Object.values(session.orgs).some(orgNeedsKey)) {
-        const publicKey = await getUserPublicKey(session.user.id)
+    // Every user needs a key.
+    const publicKey = await getUserPublicKey(session.user.id)
```

**A shared `keyGenerationUrl` helper** builds the detour url and treats `Routes.dashboard` as "no destination". That matters because `safeRedirectUrl` returns the dashboard for a rejected `redirect_url`, and forwarding it would pin the key page and defeat the resolution above. Keeping the rule inside the helper means a call site cannot get it wrong by forgetting it.

**One copy indicator at a time.** Mantine's `useClipboard` never resets `error` on a later success, so denying the clipboard prompt and then allowing it showed the green check and the red failure at once, against the AC. Resetting before each attempt fixes the sequential case but not overlapping ones, because `reset()` cannot cancel a `writeText` that is already in flight: a prompt the user leaves open rejects seconds later and lights the failure again behind a successful retry, which reproduces in a browser. Each attempt now aborts the one before it through an `AbortController`, and an aborted attempt writes no state, since a copy the user has already superseded says nothing about what is on their clipboard. That replaces `useClipboard` with a small local hook, keeping the same 2 s auto-hide on "Copied!" and the same treatment of a browser with no clipboard API.

**Nothing in the post-sign-in sequence abandons the rest of it.** `onUserSignInAction` throws when Clerk has not propagated the session yet, and that used to skip invite acceptance and the key detour outright; a `getToken` refresh failing after a successful join used to send the user back to the join-team page as though the join had failed. Both steps now resolve to a destination and the sequence carries on, which is the invariant the change above claims. The invite handling moved into a named helper as part of that.

## Merged with main (OTTER-639)

`origin/main` is merged in, including #958. The one conflict was the predicted one, in the invite-success block of `mfa.tsx`: #958 swaps the "You've successfully joined X" toast for `markOrgJoined(org.name)` so the join surfaces as the dashboard banner instead. Resolved by keeping this branch's structure and taking that line, which is also what `tsc` requires, since #958 removes the `notifications` import.

The two changes meet at exactly that point and compose: #958's banner now waits on `userKeyExistsAction` before spending its one-shot flag, and this branch is what routes a keyless invited user through key generation and on to the inviting org's dashboard, where the flag is still unspent. `mfa.test.tsx` asserts the flag survives the detour.

## Testing

- `page.test.tsx` is new and mocks nothing but the clipboard: it drives the real page through copy, Next and confirm, and asserts the landing. The existing component tests inject `firstKeyRedirect` as a prop, which proves the component obeys it but not that the page computes and passes it, and that is the same shape of gap that let the original bug ship green.
- `mfa.test.tsx` is new and covers the invite regression by asserting the `org_user` row exists afterwards, not just the url.
- Clipboard transitions are covered in both directions, and the key generation timestamps are pinned so OTTER-654 can rely on `createdAt` surviving a rotation. The rotation assertion compares for inequality rather than ordering: `>=` would also have passed if a rotation stopped touching `updatedAt`, and the insert timestamp comes from the database default while the rotation writes one from the app, so a clock difference must not decide the result.
- `sign-in-form.test.tsx` covers the no-MFA key detour, which had no test at all: an explicit `redirect_url` survives key generation, and no destination yields a bare `/account/keys`.
- `page.test.tsx` mocks `generateKeyPair` with a real SPKI fixture rather than generating a fresh 4096-bit pair on every run. Locally that generation ranged from 41 ms to 670 ms against testing-library's 1000 ms default, which is a flake waiting for a slower CI runner. Everything the test is actually about, `setUserPublicKeyAction` and the landing resolver, stays real.

- A regression test covers the overlapping-copy case: a first attempt that rejects only after a second one has succeeded must leave the success on screen and nothing else. It fails against the previous implementation.

Beyond the unit suite, the flow was driven through the running app end to end: a keyless single-org account signing in reaches a bare `/account/keys` and lands on that org's dashboard after storing the key, a reset lands on "My dashboard", a proxy-captured deep link survives sign-in and the key page, a keyless account in three orgs lands on "My dashboard", and a keyless sign-in carrying `invite_id` writes the membership before the key detour and reaches the inviting org with the joined-org banner intact.

Full suite passes at 2627 tests, plus `pnpm run checks`.

## Review follow-up

- **`AbortController` replaces the attempt counter.** Same semantics, a standard idiom rather than a one-off, and the auto-hide timer is guarded by the same signal, so every state write in the hook follows one rule instead of two.
- **One indicator component.** The three-state union and the two `isVisible` components collapse into a nullable result and a single `CopyIndicator`. The `none` state existed only to say "render neither", which `null` already says. Colors and weights are unchanged: verified in a browser as `#C70000` on the failure text and X icon, `#2F9844` at 14px/500 on the success.
- **The two key-page actions merged.** `userKeyExistsAction` plus `getFirstKeyRedirectAction` were two sequential round trips, each resolving its own session, where the second was only needed when the first returned false. `getKeyPageStateAction` returns `{ hasKey, firstKeyRedirect }` from one session and skips the org lookup entirely on the reset path. `userKeyExistsAction` stays for its four other callers.
- **The `redirect_url` fallback is documented.** That a rejected parameter falls through to `firstKeyRedirect` rather than to `Routes.dashboard` is load-bearing and reads as incidental, so the comment now says so, to survive a later "simplification" to `safeRedirectUrl(redirectParam, Routes.dashboard)`.
- **Enclave members are deliberately in scope.** The card's audience is "DP & RL" and it asks for both to be routed through this flow, so `org.type` stays out of the condition. Recorded in the comment and covered by a test.
- **Invite precedence is pinned.** When `invite_id` and `redirect_url` both arrive, the invite's landing wins, unchanged from `main`. Joining is what just happened, its dashboard is what confirms it and carries the joined-org banner, and a deep link captured before the join may not be reachable yet. A test now holds that order.
- **`refreshSessionToken` names its caller in the log**, so a stale token after sign-in (nothing committed yet) is distinguishable from one after an invite accept (membership already written).

The overlapping-copy race was kept rather than reduced to a `reset()` one-liner. The AC is explicit that a failure must not show the green check, the case reproduces in a browser, and the rewrite above brings its cost down to a small hook using a standard mechanism.

Beyond the unit suite, the copy states were re-driven in a browser against a clipboard harness that settles each attempt on command: a denied attempt shows only the red message, a retry shows only the green check, a stale rejection arriving after a later success leaves the success alone, and a superseded success's auto-hide timer does not clear a newer failure. `postKeyRedirect` was exercised over its full input matrix in the same run, including the malformed-parameter row, which lands on the resolved org dashboard rather than "My dashboard".

## Not in this PR

- **Recovery-code sign-in** (`recovery-code-signin.tsx`) hard-codes `/dashboard` and runs no key check, no metadata refresh, and no invite handling, so it drops invites for keyed and keyless users alike. Real bug, but `git blame` shows #881 never touched that path and it breaks no criterion on this card. Worth its own ticket, along with `sign-in-form`'s missing `invite_id` handling.
- **Server-side key enforcement.** `RequireUserKey` is a client effect that awaits a server action while the page is already rendered, so protected content is briefly visible. That does not literally satisfy "No user must be able to land or see the BMA without having a security key". QA passed it and the codebase already works around the flash, but it should be an explicit decision rather than a silent one.
- Guarding `/account/keys` against an accidental rotation by a user who already holds a key (it is reachable directly, and only `/user-key` carries the warning that a new key cannot decrypt existing outputs).
- `onJoinTeamAccountAction` does its Clerk metadata update and key lookup after its transaction commits, so a failure there reports a failed join on an invite that is already claimed, and the user is sent to the join-team page's "no longer valid" panel while holding the membership. Pre-existing and unchanged here; fixing it means moving where that action does its post-commit work.
- A keyless account belonging to more than one org still loses a deep link: `RequireUserKey` pushes a bare `/account/keys` and the resolver declines to guess between the orgs, so the user lands on "My dashboard". One line with `keyGenerationUrl` if that turns out to matter.
- **Carrying the invite destination instead of inferring it.** At invite time the org is known exactly; it is lost only because `RequireUserKey` pushes a bare `Routes.accountKeys`, after which the action reconstructs a guess from org membership. Persisting the destination where it is known (the guard reading it, or off the `pendingUser` row) would remove the action, the single-org heuristic, the ambiguity fallback and the provenance caveat, and would answer for multi-org accounts too, which membership count cannot. Raised in review and worth its own card; the heuristic coincides with the correct answer in the failing case, so it is a reasonable thing to ship in the meantime.

## QA

1. Fresh RL user invited to an org, full signup through MFA and recovery codes. Expect to be held at `/account/keys` with no query parameter, and to land on that org's dashboard after storing the key.
2. Same user, `/user-key` -> "Generate new key" -> expect `/dashboard`.
3. Existing keyless account opening an invite link -> "Login with existing account" -> expect membership in the inviting org and its dashboard after the key step.
4. Deny the clipboard permission, click Copy key (red error), grant it, click again -> only the green check.

Recovery-code sign-in still lands on `/dashboard` and still drops invites; that is expected here and belongs to the follow-up ticket.

[OTTER-655]: https://openstax.atlassian.net/browse/OTTER-655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/account/keys/generate-keys.tsx
+++ b/src/app/account/keys/generate-keys.tsx
@@ -5,11 +5,11 @@ import { reportMutationError } from '@/components/errors'
 import { AppModal } from '@/components/modals/app-modal'
 import { setUserPublicKeyAction, updateUserPublicKeyAction } from '@/server/actions/user-keys.actions'
 import { Button, Code, Group, Paper, Stack, Text, Title, useMantineTheme } from '@mantine/core'
-import { useClipboard, useDisclosure } from '@mantine/hooks'
+import { useDisclosure } from '@mantine/hooks'
 import { CheckIcon, XIcon } from '@phosphor-icons/react/dist/ssr'
 import type { Route } from 'next'
 import { useRouter, useSearchParams } from 'next/navigation'
-import { FC, useEffect, useState } from 'react'
+import { FC, useEffect, useRef, useState } from 'react'
 import { generateKeyPair } from 'si-encryption/util/keypair'
 import { Routes } from '@/lib/routes'
 import { safeRedirectUrl } from '@/lib/utils'
@@ -36,6 +36,44 @@ type GenerateKeysProps = {
     firstKeyRedirect?: Route
 }
 
+const COPIED_VISIBLE_MS = 2000
+
+type CopyOutcome = 'none' | 'copied' | 'failed'
+
+// The AC allows exactly one indicator at a time, which rules out Mantine's useClipboard: it folds
+// every attempt into one pair of flags, so a slow rejection (a permission prompt left open) can
+// land after a later success and light the green check and the red failure together. Attempts are
+// numbered here and a result that is no longer the latest is dropped, because a copy the user has
+// already superseded says nothing about what is on their clipboard now (OTTER-655).
+function useCopyOutcome() {
+    const [outcome, setOutcome] = useState<CopyOutcome>('none')
+    const latestAttempt = useRef(0)
+    const hideCopied = useRef<ReturnType<typeof setTimeout>>(undefined)
+
+    useEffect(() => () => clearTimeout(hideCopied.current), [])
+
+    const copy = async (value: string) => {
+        const attempt = ++latestAttempt.current
+        clearTimeout(hideCopied.current)
+        setOutcome('none')
+
+        try {
+            if (!navigator.clipboard) throw new Error('clipboard is not available in this browser')
+            await navigator.clipboard.writeText(value)
+            if (attempt !== latestAttempt.current) return
+            setOutcome('copied')
+            hideCopied.current = setTimeout(() => setOutcome('none'), COPIED_VISIBLE_MS)
+        } catch {
+            if (attempt !== latestAttempt.current) return
+            // The message tells the user to select the key manually, so the reason does not matter
+            // and there is nothing actionable to report.
+            setOutcome('failed')
+        }
+    }
+
+    return { outcome, copy }
+}
+
 export const GenerateKeys: FC<GenerateKeysProps> = ({
     isRegenerating = false,
     firstKeyRedirect = Routes.dashboard,
@@ -45,7 +83,7 @@ export const GenerateKeys: FC<GenerateKeysProps> = ({
     const [confirmationOpened, { open: openConfirm, close: closeConfirm }] = useDisclosure(false)
     // Reveal Next after any copy attempt so a blocked clipboard doesn't trap the user.
     const [hasAttemptedCopy, setHasAttemptedCopy] = useState(false)
-    const clipboard = useClipboard({ timeout: 2000 })
+    const { outcome: copyOutcome, copy } = useCopyOutcome()
 
     const onGenerateKeys = async () => {
         const { privateKeyString, fingerprint, exportedPublicKey } = await generateKeyPair()
@@ -63,10 +101,7 @@ export const GenerateKeys: FC<GenerateKeysProps> = ({
     if (!keys) return null
 
     const onCopy = () => {
-        // useClipboard clears `error` only on reset, never on a later success, so without this a
-        // retry after a blocked copy shows the green check and the red failure at once.
-        clipboard.reset()
-        clipboard.copy(keys.privateKey)
+        void copy(keys.privateKey)
         setHasAttemptedCopy(true)
     }
 
@@ -108,8 +143,8 @@ export const GenerateKeys: FC<GenerateKeysProps> = ({
                         <Button onClick={onCopy}>Copy key</Button>
                         <NextButton isVisible={hasAttemptedCopy} onClick={openConfirm} />
                     </Group>
-                    <CopySucceededIndicator isVisible={clipboard.copied} />
-                    <CopyFailedIndicator isVisible={clipboard.error != null} />
+                    <CopySucceededIndicator isVisible={copyOutcome === 'copied'} />
+                    <CopyFailedIndicator isVisible={copyOutcome === 'failed'} />
                 </Stack>
             </Stack>
 
--- a/src/app/account/keys/generate-keys.tsx
+++ b/src/app/account/keys/generate-keys.tsx
@@ -24,6 +24,11 @@ interface Keys {
 // page resolved for this account. An explicit redirect_url (invite flows, deep links) overrides it.
 // Keyed off account state rather than off the presence of that parameter: the RequireUserKey guard
 // is the entry point most first keys arrive through, and it passes none (OTTER-655).
+//
+// firstKeyRedirect is the fallback argument on purpose, so a redirect_url that fails validation
+// falls back to the resolved landing rather than to Routes.dashboard. Passing Routes.dashboard here
+// instead would look equivalent and would quietly demote a first key to "My dashboard" whenever the
+// parameter is malformed, which is the bug this function exists to fix.
 export function postKeyRedirect(isRegenerating: boolean, redirectParam: string | null, firstKeyRedirect: Route): Route {
     if (isRegenerating) return Routes.dashboard
 
@@ -38,40 +43,51 @@ type GenerateKeysProps = {
 
 const COPIED_VISIBLE_MS = 2000
 
-type CopyOutcome = 'none' | 'copied' | 'failed'
+type CopyIndication = { hue: 'green' | 'red'; Icon: typeof CheckIcon; fw?: number; text: string }
+
+const COPY_SUCCEEDED: CopyIndication = { hue: 'green', Icon: CheckIcon, fw: 500, text: 'Copied!' }
+
+const COPY_FAILED: CopyIndication = {
+    hue: 'red',
+    Icon: XIcon,
+    text: 'Copy did not work. Select the key above and copy it manually.',
+}
 
 // The AC allows exactly one indicator at a time, which rules out Mantine's useClipboard: it folds
 // every attempt into one pair of flags, so a slow rejection (a permission prompt left open) can
-// land after a later success and light the green check and the red failure together. Attempts are
-// numbered here and a result that is no longer the latest is dropped, because a copy the user has
+// land after a later success and light the green check and the red failure together. Each attempt
+// aborts the one before it and an aborted attempt writes no state, because a copy the user has
 // already superseded says nothing about what is on their clipboard now (OTTER-655).
-function useCopyOutcome() {
-    const [outcome, setOutcome] = useState<CopyOutcome>('none')
-    const latestAttempt = useRef(0)
+function useCopyIndication() {
+    const [indication, setIndication] = useState<CopyIndication | null>(null)
+    const pending = useRef<AbortController>(undefined)
     const hideCopied = useRef<ReturnType<typeof setTimeout>>(undefined)
 
     useEffect(() => () => clearTimeout(hideCopied.current), [])
 
     const copy = async (value: string) => {
-        const attempt = ++latestAttempt.current
+        pending.current?.abort()
+        const attempt = (pending.current = new AbortController())
         clearTimeout(hideCopied.current)
-        setOutcome('none')
+        setIndication(null)
 
         try {
             if (!navigator.clipboard) throw new Error('clipboard is not available in this browser')
             await navigator.clipboard.writeText(value)
-            if (attempt !== latestAttempt.current) return
-            setOutcome('copied')
-            hideCopied.current = setTimeout(() => setOutcome('none'), COPIED_VISIBLE_MS)
+            if (attempt.signal.aborted) return
+            setIndication(COPY_SUCCEEDED)
+            hideCopied.current = setTimeout(() => {
+                if (!attempt.signal.aborted) setIndication(null)
+            }, COPIED_VISIBLE_MS)
         } catch {
-            if (attempt !== latestAttempt.current) return
+            if (attempt.signal.aborted) return
             // The message tells the user to select the key manually, so the reason does not matter
             // and there is nothing actionable to report.
-            setOutcome('failed')
+            setIndication(COPY_FAILED)
         }
     }
 
-    return { outcome, copy }
+    return { indication, copy }
 }
 
 export const GenerateKeys: FC<GenerateKeysProps> = ({
@@ -83,7 +99,7 @@ export const GenerateKeys: FC<GenerateKeysProps> = ({
     const [confirmationOpened, { open: openConfirm, close: closeConfirm }] = useDisclosure(false)
     // Reveal Next after any copy attempt so a blocked clipboard doesn't trap the user.
     const [hasAttemptedCopy, setHasAttemptedCopy] = useState(false)
-    const { outcome: copyOutcome, copy } = useCopyOutcome()
+    const { indication: copyIndication, copy } = useCopyIndication()
 
     const onGenerateKeys = async () => {
         const { privateKeyString, fingerprint, exportedPublicKey } = await generateKeyPair()
@@ -143,8 +159,7 @@ export const GenerateKeys: FC<GenerateKeysProps> = ({
                         <Button onClick={onCopy}>Copy key</Button>
                         <NextButton isVisible={hasAttemptedCopy} onClick={openConfirm} />
                     </Group>
-                    <CopySucceededIndicator isVisible={copyOutcome === 'copied'} />
-                    <CopyFailedIndicator isVisible={copyOutcome === 'failed'} />
+                    <CopyIndicator indication={copyIndication} />
                 </Stack>
             </Stack>
 
@@ -168,27 +183,16 @@ const NextButton: FC<{ isVisible: boolean; onClick: () => void }> = ({ isVisible
     )
 }
 
-const CopySucceededIndicator: FC<{ isVisible: boolean }> = ({ isVisible }) => {
+const CopyIndicator: FC<{ indication: CopyIndication | null }> = ({ indication }) => {
     const theme = useMantineTheme()
-    if (!isVisible) return null
-    return (
-        <Group gap="xs">
-            <CheckIcon size={16} color={theme.colors.green[9]} />
-            <Text c="green.9" fz={14} fw={500}>
-                Copied!
-            </Text>
-        </Group>
-    )
-}
+    if (!indication) return null
 
-const CopyFailedIndicator: FC<{ isVisible: boolean }> = ({ isVisible }) => {
-    const theme = useMantineTheme()
-    if (!isVisible) return null
+    const { hue, Icon, fw, text } = indication
     return (
         <Group gap="xs">
-            <XIcon size={16} color={theme.colors.red[9]} />
-            <Text c="red.9" fz={14}>
-                Copy did not work. Select the key above and copy it manually.
+            <Icon size={16} color={theme.colors[hue][9]} />
+            <Text c={`${hue}.9`} fz={14} fw={fw}>
+                {text}
             </Text>
         </Group>
     )
--- a/src/app/account/keys/generate-keys.tsx
+++ b/src/app/account/keys/generate-keys.tsx
@@ -7,6 +7,7 @@ import { setUserPublicKeyAction, updateUserPublicKeyAction } from '@/server/acti
 import { Button, Code, Group, Paper, Stack, Text, Title, useMantineTheme } from '@mantine/core'
 import { useClipboard, useDisclosure } from '@mantine/hooks'
 import { CheckIcon, XIcon } from '@phosphor-icons/react/dist/ssr'
+import type { Route } from 'next'
 import { useRouter, useSearchParams } from 'next/navigation'
 import { FC, useEffect, useState } from 'react'
 import { generateKeyPair } from 'si-encryption/util/keypair'
@@ -19,11 +20,26 @@ interface Keys {
     fingerprint: string
 }
 
-type GenerateKeysProps = {
-    isRegenerating?: boolean
+// AC: a reset returns to "My dashboard"; the first key ever generated returns to the landing the
+// page resolved for this account. An explicit redirect_url (invite flows, deep links) overrides it.
+// Keyed off account state rather than off the presence of that parameter: the RequireUserKey guard
+// is the entry point most first keys arrive through, and it passes none (OTTER-655).
+export function postKeyRedirect(isRegenerating: boolean, redirectParam: string | null, firstKeyRedirect: Route): Route {
+    if (isRegenerating) return Routes.dashboard
+
+    return safeRedirectUrl(redirectParam, firstKeyRedirect)
 }
 
-export const GenerateKeys: FC<GenerateKeysProps> = ({ isRegenerating = false }) => {
+type GenerateKeysProps = {
+    isRegenerating?: boolean
+    /** Where a first key lands when no redirect_url is supplied; ignored for a reset. */
+    firstKeyRedirect?: Route
+}
+
+export const GenerateKeys: FC<GenerateKeysProps> = ({
+    isRegenerating = false,
+    firstKeyRedirect = Routes.dashboard,
+}) => {
     const theme = useMantineTheme()
     const [keys, setKeys] = useState<Keys>()
     const [confirmationOpened, { open: openConfirm, close: closeConfirm }] = useDisclosure(false)
@@ -47,6 +63,9 @@ export const GenerateKeys: FC<GenerateKeysProps> = ({ isRegenerating = false }) 
     if (!keys) return null
 
     const onCopy = () => {
+        // useClipboard clears `error` only on reset, never on a later success, so without this a
+        // retry after a blocked copy shows the green check and the red failure at once.
+        clipboard.reset()
         clipboard.copy(keys.privateKey)
         setHasAttemptedCopy(true)
     }
@@ -99,6 +118,7 @@ export const GenerateKeys: FC<GenerateKeysProps> = ({ isRegenerating = false }) 
                 isOpen={confirmationOpened}
                 keys={keys}
                 isRegenerating={isRegenerating}
+                firstKeyRedirect={firstKeyRedirect}
             />
         </Paper>
     )
@@ -139,12 +159,13 @@ const CopyFailedIndicator: FC<{ isVisible: boolean }> = ({ isVisible }) => {
     )
 }
 
-const ConfirmationModal: FC<{ onClose: () => void; isOpen: boolean; keys: Keys; isRegenerating: boolean }> = ({
-    onClose,
-    isOpen,
-    keys,
-    isRegenerating,
-}) => {
+const ConfirmationModal: FC<{
+    onClose: () => void
+    isOpen: boolean
+    keys: Keys
+    isRegenerating: boolean
+    firstKeyRedirect: Route
+}> = ({ onClose, isOpen, keys, isRegenerating, firstKeyRedirect }) => {
     const router = useRouter()
     const searchParams = useSearchParams()
 
@@ -157,8 +178,7 @@ const ConfirmationModal: FC<{ onClose: () => void; isOpen: boolean; keys: Keys; 
         },
         onError: reportMutationError('Failed to save security key'),
         onSuccess() {
-            // First-time onboarding carries an org-dashboard redirect_url; a reset carries none.
-            router.push(safeRedirectUrl(searchParams.get('redirect_url'), Routes.dashboard))
+            router.push(postKeyRedirect(isRegenerating, searchParams.get('redirect_url'), firstKeyRedirect))
         },
     })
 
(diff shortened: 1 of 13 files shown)


nathanstitt:
Heads-up: this PR and #958 both rewrite the same invite-accept block in the MFA success handler (`src/app/account/signin/mfa.tsx`). #958 replaces the success toast with `markOrgJoined` so this path shows the OTTER-639 dashboard banner; this branch predates that and still calls `notifications.show`. Whichever merges second will hit a conflict there — when resolving, keep the restructured control flow from this PR but with `markOrgJoined` in place of the toast, otherwise the MFA join path regresses OTTER-639.

src/app/account/signin/mfa.tsx
```diff
@@ -67,60 +135,27 @@ export const RequestMFA: FC<{ mfa: MFAState }> = ({ mfa }) => {
...
-                            }
-                        }
-                        router.push(redirectUrl)
+                    const result = await completeServerSignIn(auth.getToken)
+
+                    const rawRedirect = searchParams.get('redirect_url')
+                    let redirectUrl = rawRedirect ? safeRedirectUrl(rawRedirect, Routes.dashboard) : null
+                    const inviteId = searchParams.get('invite_id')
+                    if (inviteId) {
+                        redirectUrl = await acceptInviteAndResolveLanding(inviteId, auth.getToken)
```
nathanstitt, line 151: The invite branch overwrites `redirectUrl` outright, so when a user arrives with **both** `invite_id` and `redirect_url` the deep link they were originally headed for is discarded in favour of the inviting org's dashboard.

Worth noting this precedence is unchanged from `main` (the old code assigned `redirectUrl = orgDashboard` in the same spot), so it isn't a regression from this PR. But the restructure is what makes it legible for the first time, and the new tests cover invite-only and redirect-only but never the pair, so nothing pins the intended precedence. Since a proxy-captured deep link is exactly what `redirect_url` exists for, this feels like the more surprising of the two orders.

Could go either way — but whichever is intended, a one-line test asserting it would stop the next refactor from flipping it silently.

src/server/actions/user-keys.actions.ts
```diff
@@ -30,6 +30,23 @@ export const userKeyExistsAction = new Action('userKeyExistsAction')
...
+// though it proves nothing about provenance. Anything ambiguous (no orgs, or several) returns "My
+// dashboard" rather than guessing.
+export const getFirstKeyRedirectAction = new Action('getFirstKeyRedirectAction')
+    .requireAbilityTo('view', 'UserKey')
+    .handler(async ({ session }) => {
+        // Read from the database, not from session.orgs: that map comes from Clerk metadata, which
+        // is stale exactly when this runs, right after a signup or an invite accept.
+        const orgs = await getOrgInfoForUserId(session.user.id)
+
+        if (orgs.length !== 1) return Routes.dashboard
```
nathanstitt: The `orgs.length !== 1` rule is a nice conservative call, and the comment about it being a landing choice rather than a provenance claim is exactly right. One case worth a thought though: this fires for **enclave** members too, not just lab users.

`getOrgInfoForUserId` selects `org.type` but the check ignores it, so a freshly invited single-org enclave (Data Partner) user landing on their first key gets `/{slug}/dashboard`. That page does render for enclave orgs (`isEnclaveOrg` branches the copy), so nothing breaks — but the AC in OTTER-655 is written about the Research Lab flow, and it's not obvious whether reviewers were meant to be included.

If the answer is "yes, both, deliberately" then the comment above is the place to say so, since the `type` column is already sitting right there and its absence from the condition reads like an oversight.

src/app/account/keys/page.tsx
```diff
@@ -1,11 +1,15 @@
...
+import { getFirstKeyRedirectAction, userKeyExistsAction } from '@/server/actions/user-keys.actions'
 import { GenerateKeys } from './generate-keys'
 
 export const dynamic = 'force-dynamic'
 
 export default async function KeysPage() {
     const hasKey = actionResult(await userKeyExistsAction())
+    // A reset always returns to "My dashboard"; a first key gets the account's resolved landing,
+    // which is an org dashboard only when that org is unambiguous.
+    const firstKeyRedirect = hasKey ? Routes.dashboard : actionResult(await getFirstKeyRedirectAction())
```
nathanstitt: Small thing: `userKeyExistsAction` and `getFirstKeyRedirectAction` are two sequential round trips (each with its own `requireAbilityTo` + session resolution) where the second is only ever needed when the first returns false.

The short-circuit already avoids the cost on the reset path, so this is only the first-key path paying for it, and that path is once-per-account — so it may well not be worth touching. Mentioning it mostly because both actions ultimately want the same session, and a single action returning `{ hasKey, firstKeyRedirect }` would drop a round trip and remove the chance of the two resolving against different session states.

src/app/account/keys/generate-keys.tsx
```diff
@@ -19,17 +20,70 @@ interface Keys {
...
 }
 
+// AC: a reset returns to "My dashboard"; the first key ever generated returns to the landing the
+// page resolved for this account. An explicit redirect_url (invite flows, deep links) overrides it.
+// Keyed off account state rather than off the presence of that parameter: the RequireUserKey guard
+// is the entry point most first keys arrive through, and it passes none (OTTER-655).
+export function postKeyRedirect(isRegenerating: boolean, redirectParam: string | null, firstKeyRedirect: Route): Route {
+    if (isRegenerating) return Routes.dashboard
+
+    return safeRedirectUrl(redirectParam, firstKeyRedirect)
```
nathanstitt, line 35: Nice touch that a rejected `redirect_url` falls through to `firstKeyRedirect` rather than to `Routes.dashboard` — the resolved landing stays in play instead of a malformed parameter quietly demoting the user to "My dashboard".

That's a genuinely non-obvious property and it's doing real work here, but it reads as incidental to `safeRedirectUrl`'s fallback argument. Worth a line in the comment above so it survives someone later "simplifying" this to `safeRedirectUrl(redirectParam, Routes.dashboard)`, which would look equivalent and silently undo the fix this PR is about.

src/app/account/keys/generate-keys.tsx
```diff
@@ -19,17 +20,70 @@ interface Keys {
...
+        const attempt = ++latestAttempt.current
+        clearTimeout(hideCopied.current)
+        setOutcome('none')
+
+        try {
+            if (!navigator.clipboard) throw new Error('clipboard is not available in this browser')
+            await navigator.clipboard.writeText(value)
+            if (attempt !== latestAttempt.current) return
+            setOutcome('copied')
+            hideCopied.current = setTimeout(() => setOutcome('none'), COPIED_VISIBLE_MS)
```
nathanstitt: The attempt-numbering approach is the right call, and the reasoning in the comment above (a superseded copy says nothing about the current clipboard) is convincing.

One gap: the `setTimeout` that clears `'copied'` isn't guarded by the attempt number the way the `await` resumption points are. If a copy succeeds and then the user clicks again within the 2 s window, `copy` clears the pending timer on line 57, so the common case is fine. But the timer callback itself closes over no attempt id, so it unconditionally does `setOutcome('none')` — meaning it can only ever clear, never resurrect a stale state. So this is safe today; it's more that the invariant "every state write is attempt-guarded" has one exception, and the next person adding a branch here won't know which rule applies.

Capturing `attempt` in the callback (`if (attempt === latestAttempt.current) setOutcome('none')`) would make the rule uniform for the cost of one comparison.

src/app/account/signin/mfa.tsx
```diff
@@ -21,6 +22,73 @@ import { VerifyCode } from './verify-code'
...
 type Method = 'sms' | 'totp'
 
+// The session token is what carries fresh org metadata to the next page, but a stale token is a
+// far smaller problem than losing the invite or the key detour that follows it, so a refresh
+// failure is logged rather than thrown.
+async function refreshSessionToken(getToken: GetToken) {
+    try {
+        await getToken({ skipCache: true })
+    } catch (error) {
+        console.error('session token refresh failed:', error)
```
nathanstitt: `refreshSessionToken` swallowing the failure is well argued, and I agree a stale token beats losing the invite. Two callers though, with different stakes: the one inside `completeServerSignIn` runs before anything has been committed, while the one at the end of `acceptInviteAndResolveLanding` runs after the membership row exists, which is the case the PR description calls out as previously sending users back to join-team as though the join had failed.

Since the second is the one that motivated the change, it might be worth `console.error`-ing with enough context to tell them apart in a session replay — right now both emit the identical `'session token refresh failed:'` string, so the log can't distinguish "token stale, user has no membership yet" from "token stale, user is already a member". A caller label as a second argument would be enough.

src/app/account/keys/generate-keys.tsx
```diff
@@ -19,17 +20,70 @@ interface Keys {
...
+type CopyOutcome = 'none' | 'copied' | 'failed'
+
+// The AC allows exactly one indicator at a time, which rules out Mantine's useClipboard: it folds
+// every attempt into one pair of flags, so a slow rejection (a permission prompt left open) can
+// land after a later success and light the green check and the red failure together. Attempts are
+// numbered here and a result that is no longer the latest is dropped, because a copy the user has
+// already superseded says nothing about what is on their clipboard now (OTTER-655).
+function useCopyOutcome() {
+    const [outcome, setOutcome] = useState<CopyOutcome>('none')
+    const latestAttempt = useRef(0)
```
nathanstitt: Stepping back from the specific bug for a second: this is hand-rolling "ignore a superseded async result", which is what `AbortController` is for. Nothing else in the repo uses an attempt-counter for this, so it introduces a one-off idiom that the next person has to read carefully to trust.

The equivalent with a standard mechanism:

```ts
const pending = useRef<AbortController>()

const copy = async (value: string) => {
    pending.current?.abort()
    const ctl = (pending.current = new AbortController())
    try {
        await navigator.clipboard.writeText(value)
        if (!ctl.signal.aborted) setResult(COPIED)
    } catch {
        if (!ctl.signal.aborted) setResult(FAILED)
    }
}
```

Same semantics, and `aborted` reads as intent in a way `attempt !== latestAttempt.current` does not. It also makes my separate note about the un-guarded `setTimeout` moot rather than needing a comment to explain why it is safe.

src/app/account/keys/generate-keys.tsx
```diff
@@ -89,8 +143,8 @@ export const GenerateKeys: FC<GenerateKeysProps> = ({ isRegenerating = false })
                         <Button onClick={onCopy}>Copy key</Button>
                         <NextButton isVisible={hasAttemptedCopy} onClick={openConfirm} />
                     </Group>
-                    <CopySucceededIndicator isVisible={clipboard.copied} />
-                    <CopyFailedIndicator isVisible={clipboard.error != null} />
+                    <CopySucceededIndicator isVisible={copyOutcome === 'copied'} />
+                    <CopyFailedIndicator isVisible={copyOutcome === 'failed'} />
```
nathanstitt: The three-state union plus two boolean-prop components is a heavier model than the UI needs. `CopySucceededIndicator` and `CopyFailedIndicator` are the same shape — an icon and a line of text — differing only in icon, colour and copy, and they are mutually exclusive by construction.

A single nullable result collapses the `CopyOutcome` union, both components, and both props into one render path:

```tsx
const COPIED = { color: "green.9", Icon: CheckIcon, text: "Copied!" }
const FAILED = { color: "red.9", Icon: XIcon, text: "Copy did not work. Select the key above and copy it manually." }

{result && <CopyIndicator {...result} />}
```

Not a defect, and if you would rather keep the two named components for grep-ability that is a fair trade — but the `none` state currently exists only to say "neither component renders", which `null` already says.

src/app/account/keys/generate-keys.tsx
```diff
@@ -19,17 +20,70 @@ interface Keys {
...
 type GenerateKeysProps = {
     isRegenerating?: boolean
+    /** Where a first key lands when no redirect_url is supplied; ignored for a reset. */
+    firstKeyRedirect?: Route
 }
 
-export const GenerateKeys: FC<GenerateKeysProps> = ({ isRegenerating = false }) => {
+const COPIED_VISIBLE_MS = 2000
+
+type CopyOutcome = 'none' | 'copied' | 'failed'
```
nathanstitt: Worth asking whether the overlapping-attempt race needs solving at all, since it is the thing driving most of the new code here.

The sequential case — deny, see the failure, click again, succeed — is what QA filed, and a `reset()` before each attempt fixes it in one line. The overlapping case needs the user to deny the prompt, leave it open, click again, succeed, and only then have the original prompt reject. Real, and you clearly reproduced it, but rare enough that it may not be worth a custom hook plus three tests to cover.

Genuinely a product call rather than a code one, so treat this as a question and not a request: if the AC is read strictly as "never two indicators", the current code is the honest answer. If it is read as "the indicator should reflect the last thing the user did", the one-liner gets there.

src/server/actions/user-keys.actions.ts
```diff
@@ -30,6 +30,23 @@ export const userKeyExistsAction = new Action('userKeyExistsAction')
         return Boolean(key)
     })
 
+// Landing for the key page when no destination was carried in: the RequireUserKey guard pushes a
+// bare url, so the page has to answer this itself. NOT a claim about which org invited the account:
+// a single-org account simply has one dashboard it could land on, so the answer cannot be wrong even
+// though it proves nothing about provenance. Anything ambiguous (no orgs, or several) returns "My
+// dashboard" rather than guessing.
+export const getFirstKeyRedirectAction = new Action('getFirstKeyRedirectAction')
```
nathanstitt: The larger design question, separate from the inline notes: this action exists to *infer* something the invite flow already knew for certain.

At invite time the org is known exactly. It gets lost only because `RequireUserKey` pushes a bare `Routes.accountKeys`, and this action then reconstructs a guess from org membership. Carrying the destination at the point it is known — the guard reading it, or off the `pendingUser` row — would remove this action, the single-org heuristic, the ambiguity fallback, and the "NOT a claim about which org invited the account" caveat that the comment has to make.

It would also be *more* correct: a multi-org user currently degrades to "My dashboard" because membership cannot answer the question, whereas a persisted destination answers it for everyone. The AC asks for the inviting org, and membership count is a proxy for that which happens to coincide in the failing case.

Bigger than a review comment and probably its own card — the current version is a reasonable thing to ship, and the comment is admirably honest that it is a fallback rather than provenance. Raising it because the heuristic and its caveats are a standing cost, and the information to avoid them exists upstream.

nathanstitt (approved):
seems good but pointed out some simplifications we should use

src/app/account/keys/generate-keys.tsx
therealmarv: Adopted in 8b27913. The hook is now [`useCopyIndication`](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/app/account/keys/generate-keys.tsx#L61-L92): each attempt aborts the one before it, and every state write is gated on `attempt.signal.aborted`.

You were right that it also moots the timer note. Rather than leave the `setTimeout` as the one unguarded write, [it checks the same signal](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/app/account/keys/generate-keys.tsx#L79-L81), so the rule is now uniform with no comment needed to explain an exception. Belt and braces, since `clearTimeout` at the top of each attempt already covers it, but the invariant is what makes the next branch here safe to add.

src/app/account/keys/generate-keys.tsx
therealmarv: Agreed, collapsed in 8b27913. `CopyOutcome`, both components and both props are gone, replaced by two data records and [one `CopyIndicator`](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/app/account/keys/generate-keys.tsx#L186-L199) that returns null on a null result.

The one thing I kept off the record is the hue: the icon needs `theme.colors[hue][9]` while the text takes `c={hue.9}`, so it stores `'green' | 'red'` and both derive from it. Checked in a browser that nothing drifted in the process: failure is `#C70000` on the text and the X icon, success is `#2F9844` at 14px/500, and exactly one indicator is in the DOM at a time.

src/app/account/keys/generate-keys.tsx
therealmarv: Kept it, but the cost is a lot lower than it was. Taking the AC literally on this one: it says "IF copying to clipboard fails THEN do not show the 'green check' AND show the following error message instead", which reads as a statement about what may be on screen, not about which click caused it. The overlapping case does reproduce in a browser, so the one-liner would leave a real path to two indicators.

What changed in response to this and your `AbortController` note is the price. The hook is now a standard mechanism rather than a bespoke counter, and it is down to a handful of lines with one rule for every state write. Re-drove all four cases in a browser against a clipboard harness that settles each attempt on command: denied shows only red, retry shows only green, a stale rejection arriving after a later success leaves the success alone, and a superseded success's auto-hide does not clear a newer failure.

Happy to reduce it to `reset()` if the AC is meant as "reflect the last thing the user did", but that is Iris's call rather than mine.

src/app/account/keys/page.tsx
therealmarv: Merged in 8b27913. [`getKeyPageStateAction`](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/server/actions/user-keys.actions.ts#L40-L61) returns `{ hasKey, firstKeyRedirect }`, so the [page destructures one call](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/app/account/keys/page.tsx#L10).

The short-circuit survives inside it: a `hasKey` account returns before the org lookup, so the reset path still does not pay for it. `userKeyExistsAction` stays as it was, since `RequireUserKey`, the joined-org banner and the `/user-key` layout all still want just the boolean. Agreed that the round trip was the smaller half of this; the two resolving against different session states was the part worth closing.

src/server/actions/user-keys.actions.ts
therealmarv: Yes, both, deliberately, and you are right that its absence read as an oversight. Recorded [where you suggested](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/server/actions/user-keys.actions.ts#L52-L54): the card's audience line is "DP & RL" and its first criterion is "Ensure to redirect both RL and DP to this flow", so an enclave member resolves exactly as a lab member does and `org.type` stays out of the condition.

Added [a test that pins it](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/server/actions/user-keys.actions.test.ts#L157-L165) rather than leaving the claim to the comment alone, since the previous cases all ran on the default enclave org and would have passed either way.

src/app/account/keys/generate-keys.tsx
therealmarv: Good catch that it reads as incidental. [Written down](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/app/account/keys/generate-keys.tsx#L28-L31), naming the exact rewrite it has to survive and what that rewrite would cost.

Also covered it in the browser rather than only in prose: ran `postKeyRedirect` over its full input matrix in the running app, and the malformed-parameter row lands on the resolved org dashboard, not "My dashboard".

src/app/account/keys/generate-keys.tsx
therealmarv: Handled by taking the `AbortController` suggestion from your other note, which is the version you predicted would make this moot. Rather than leave the timer as the single exception, [it checks the same signal](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/app/account/keys/generate-keys.tsx#L79-L81) as every other write, so "every state write is attempt-guarded" now holds without qualification.

Your read that it was safe today was right, and I checked the behavior it protects rather than assuming: a success followed by a failure inside the 2 s window still shows the failure once the original timer fires.

src/app/account/signin/mfa.tsx
therealmarv: Done, [as a second argument](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/app/account/signin/mfa.tsx#L31-L36) exactly as you suggested. The two now log `after sign-in` and `after invite-accepted`, and the comment says why they are worth telling apart: nothing is committed at the first, while the second runs with the membership row already written, so a stale token there means a joined user whose session cannot see the org.

src/app/account/signin/mfa.tsx
therealmarv: Kept the existing order and pinned it, since you are right that nothing held it.

Reasoning for invite-wins, now [in the code](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/app/account/signin/mfa.tsx#L146-L149): joining is the thing that just changed state, and the org dashboard is the only landing that reflects it, including the `markOrgJoined` banner which would otherwise sit unspent on a page that does not explain what happened. A deep link captured before the join can also point somewhere this account could not reach until the membership existed, so honoring it first is the order more likely to dead-end.

Your point about `redirect_url` being exactly the proxy's deep-link channel is the fair counter, and I do not think it is clear-cut. [The test](https://github.com/safeinsights/management-app/blob/8b279136ae0a27f2fd03c254d4530f367f79fbff/src/app/account/signin/mfa.test.tsx#L108-L120) at least means flipping it is now a deliberate act with a red test attached, rather than a silent side effect.

src/server/actions/user-keys.actions.ts
therealmarv: Agreed on all of it, and agreed it is its own card rather than something to fold in here. Recorded in "Not in this PR" so it does not evaporate with the review thread.

The part I think makes the strongest case for the card is the one you raised last: this is not merely a tidier way to reach the same answer, it is a strictly better answer. A multi-org invited account degrades to "My dashboard" today because membership genuinely cannot distinguish the inviting org, and a persisted destination answers for everyone. Also worth noting it would subsume the other known gap here, where a keyless multi-org account loses a deep link for the same underlying reason.

Leaving the heuristic in place for now on the grounds you named: it coincides with the correct answer in the failing case, and the comment is explicit that it is a landing choice rather than provenance.

PR #960, opened 2026-08-10
OTTER-674: move the reviewer feedback errors into the editor footer slot

Rod0352:
Follow-up to #937, closing out the two surfaces its description deferred — and the one Greg's re-validation failed the card on: `review-feedback-section` and `code-review-feedback-section` still had the sibling-row structure, so the empty-field error rendered a row below the word counter (measured 52.8px under the input on QA) instead of directly underneath it.

## What changed

Same one-move fix #937 applied to the resubmission note, on both reviewer surfaces:

- The empty-field error now renders in the editor's `footerLeft` slot — directly under the input, sharing the footer row with the word counter, in exactly the slot the save indicator vacates. The error and "All changes saved" still can never co-exist.
- The old sibling error row is removed.
- `fieldErrorId` / `aria-describedby` wiring is unchanged (`review-feedback-error`, `code-review-feedback-error` ids are the same, just relocated).

## Testing

- New/extended unit tests assert the error renders in the same footer row as the word counter on both sections, plus a no-error-box-when-clean guard. All 33 tests across the five touched/adjacent suites pass (`review-feedback-section`, `code-review-feedback-section`, `single-user-editor`, both resubmission-note suites); prettier, eslint, and `typecheck` are clean.
- The code review surface is not reachable on QA (needs an approved study with submitted code and a completed enclave run), so — as #937 did for `SingleUserEditor` — its unit test is the verification of record for that surface.

## QA validation path (reviewer surface)

Initial request review → type into the feedback field → delete all content → select a decision radio (the error is submit-time validation on this form). Expected: "Feedback is required." sits in the counter row, ~4px under the editor box, left-aligned, with no autosave label on the field.

With this merged, expectation 2 of OTTER-674 holds on every surface QA can reach: proposal fields (4–8px), resubmission note (4px, per #937), and reviewer feedback (now in the footer row).


src/app/[orgSlug]/study/[studyId]/review/review-feedback-section.tsx
```diff
@@ -34,6 +34,17 @@ function feedbackHeading(reviewVersion: number) {
     return reviewVersion <= 1 ? 'Initial request review' : `Round ${reviewVersion} review`
 }
 
+// Carries the id `aria-describedby` points at; null when clean so the save indicator keeps the
+// row's left edge (same shape as NoteFieldError in collaborative-resubmission-note-section).
+function FeedbackFieldError({ error }: { error?: string | null }) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/review-feedback-section.tsx]: With this PR we now have three identical copies of this component — this one, the twin in `code-review-feedback-section.tsx`, and `NoteFieldError` in `collaborative-resubmission-note-section.tsx` — differing only in the field id string. Worth extracting a shared one next to `fieldErrorId` in `@/components/form-field`:

```tsx
export function FieldErrorBox({ fieldId, error }: { fieldId: string; error?: string | null }) {
    if (!error) return null
    return (
        <Box id={fieldErrorId(fieldId)}>
            <InputError error={error} />
        </Box>
    )
}
```

Then all three call sites become `footerLeft={<FieldErrorBox fieldId="review-feedback" error={feedback.error} />}` and the next surface that needs this (there seems to be one per ticket 😄) gets it for free.

nathanstitt (approved):
Looks good but let's consolidate the field error as suggested.  We also have a ErrorType I think and should use errrorToString in it

src/app/[orgSlug]/study/[studyId]/review/review-feedback-section.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/review/review-feedback-section.tsx]: Extracted as suggested

PR #962, opened 2026-08-11
Show the deployed infrastructure version and editor release on /about

nathanstitt:
Adds two rows to `/about`: which infrastructure commit deployed this environment, and which release built the editor image it is actually serving.

Pairs with safeinsights/iac#217, which supplies both as Lambda environment variables. Until that merges and an environment is deployed, both render as `not deployed` — this PR is safe to merge first.

## `Infrastructure` (`IAC_VERSION`)

The commit of the iac repo behind the running infrastructure. It advances only on an infra deploy, so it intentionally lags the application release.

The value comes from `git describe --always --dirty`, so it is usually a bare short SHA but can carry a `-N-g<sha>` or `-dirty` suffix, or the literal `unknown` when git was unavailable at synth. Only a plain SHA is linked — the decorated forms are not valid refs and the link would 404. It points at the **iac** repo, not this one.

## `Editor Release` (`EDITOR_RELEASE_SHA`)

Which release built the editor image being served. This diverges from the application release whenever the editor was unchanged: the editor image is tagged by a content hash of its build inputs, so an untouched editor keeps serving the image an earlier release built. That divergence is the reason this is worth displaying separately.

## Notes

- The repeated anchor markup is extracted into `GithubLink` rather than copied a third time.
- Adds `src/app/about/page.test.tsx` — the page had no tests. Five cases, including that each value links to the correct repo.
- `/about` is in `ANON_ROUTES`, so this is publicly visible. Both values are public git SHAs, so nothing sensitive is exposed.

### DIFF
--- a/src/app/about/page.test.tsx
+++ b/src/app/about/page.test.tsx
@@ -1,0 +1,58 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import { screen } from '@testing-library/react'
+import { renderWithProviders } from '@/tests/unit.helpers'
+import AboutPage from './page'
+
+const linkFor = (label: string) => screen.getByText(label).closest('a')
+
+describe('AboutPage deployed versions', () => {
+    const ORIGINAL_ENV = process.env
+
+    beforeEach(() => {
+        process.env = { ...ORIGINAL_ENV }
+    })
+
+    afterEach(() => {
+        process.env = ORIGINAL_ENV
+    })
+
+    it('links the infrastructure commit to the iac repo, not the app repo', () => {
+        process.env.IAC_VERSION = '37cfdea'
+        renderWithProviders(<AboutPage />)
+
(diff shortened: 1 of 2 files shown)


therealmarv (approved):
Reviewed the diff for correctness and reuse. No findings.

Notes on what was checked:

- `TagLink` refactor onto the new `GithubLink` keeps the `tag || sha` fallback and the existing early return, so behavior is unchanged.
- `IacVersionLink` and `EditorReleaseLink` both guard the missing-env case before rendering a link, and the SHA regex correctly leaves `-dirty` / `-N-g<sha>` / `unknown` as plain text rather than producing a 404 link.
- No leftover dead code from the extraction.

Not covered by this pass: the iac repo side (whether `IAC_VERSION` and `EDITOR_RELEASE_SHA` are actually plumbed into the deployed task definition).

PR #963, opened 2026-08-11
OTTER-572 follow-up: resume Step 2 drafts whose edits only reached the collaborative document

therealmarv:
see [OTTER-572](https://openstax.atlassian.net/browse/OTTER-572)

A proposal draft left on Step 2 still reopened on the Step 1 data-partner picker, so the card regressed again after the June fix.

## Why the previous fix stopped holding

The dashboard resume rule is right, but the fact it reads was sourced from the wrong storage layer.

`hasStep2Progress` was derived only from the Step 2 study columns (`piUserId`, `datasets`, `researchQuestions`, `projectSummary`, `impact`, `additionalNotes`). In collaborative mode those columns are not what Step 2 writes: every edit goes into the `proposal-<studyId>-fields` Yjs document or a per-field lexical document, and the columns are flushed only by `Previous`, `View as reviewer`, and `Submit`. There is no interval autosave and no unload flush, and the editor service mirrors exactly one value back into the study row (the title, in `title-mirror.ts`).

So the reported flow: Step 1, Proceed, fill Step 2, see "All changes saved", then leave through the nav or close the tab. The columns are still null, `hasStep2Progress` is false, and the dashboard sends the researcher to Step 1. The work is not lost, it reloads from Yjs on the way back, but the landing step is wrong.

This only shows up in collaborative environments, which is why CI stayed green: the existing e2e leaves Step 2 through `Previous` (which flushes), and CI runs with `SINGLE_USER_EDITING=t`, where the columns and Yjs cannot diverge.

A second, smaller case had the same root cause: after reopening Step 2 the form is seeded from Yjs and reset to pristine, so `Previous` skips its flush and the columns stay empty even on the one path meant to write them.

## The fix

Read the resume signal from Step 2's actual persistence layer, and keep the rule and the state machine as they are.

- New `hasStep2CollabDoc` fact on `RawStudyState`, projected into `hasStep2Progress` alongside the existing column check.
- `server/db/step2-collab-doc.ts` holds the shared `EXISTS` fragment over `yjs_document`, used by both producers of `RawStudyState` (the dashboard listing query and `rawStudyStateForStudy`).
- The fragment matches the five document names Step 2 can write for a study (the fields map plus one per lexical field) by **exact name equality**, so each probe resolves through `yjs_document`'s `name` primary key. A `name LIKE 'proposal-' || id || '-%'` variant would be built per outer row rather than being a planner-time constant, so it could not drive a prefix range scan, and a non-C collation would additionally need `text_pattern_ops`.

No migration, no new write path, no extra mirroring: the change is read-side only.

The signal means "reached Step 2", so a researcher who opened Step 2, walked back with `Previous`, and then left from Step 1 now resumes on Step 2. That was equally true of the column-based version once any Step 2 field had been written, and `Previous` remains one click away.

## Alternatives considered

- **Extend the editor-service mirror** to write `datasets` / `piUserId` / `piName` back like the title. Follows existing precedent, but it does not cover a researcher who only types into the lexical fields, so the bug stays reproducible. Mirroring those would mean headless lexical serialization on every debounced store flush.
- **Link every DRAFT to Step 2** and delete the fact. Smallest diff, but it depends on "a draft row implies Step 2 was reached", which the in-flight draft-architecture work breaks by lazy-creating a draft on the first Step 1 edit.
- **A stored `last_draft_step` marker.** The only way to track the last step literally, but it needs a migration and writes on both steps. Worth raising with PM only if per-step fidelity is wanted over "reached Step 2".

## Tests

Full unit suite passes (266 files, 2612 tests), along with `pnpm run checks`.

- `state.test.ts`: Step 2 progress comes from the collaborative document with every column empty.
- `dashboard-raw-state.test.ts`: the row field reaches the projection.
- `study-action-link.test.tsx`: a DRAFT with only a collaborative document links to `/proposal`.
- `study-state-query.test.ts`: real rows for the fields document, a lexical document alone, another study's document, and a non-Step-2 document.
- `study.actions.test.ts`: the dashboard listing query reports the fact and the row projects to Step 2 progress.

E2E coverage note: the existing `Researcher resumes a Step 2 draft on Step 2` spec covers the flush path and still passes. The "leave through the nav" path cannot be asserted in CI, because in single-user mode nothing is persisted on that exit at all, so there is nothing to resume. The collaborative path is covered by the server-side tests above plus manual QA.

## Manual verification

Needs a collaborative environment (local docker qualifies, `editor-service` runs and `SINGLE_USER_EDITING` is unset).

1. Researcher, Propose New Study, pick Data Partner and language, Proceed to Step 2.
2. Fill only the research question and wait for "All changes saved".
3. Leave through the top nav or close the tab, not `Previous`.
4. Dashboard, the draft row, `Edit`: lands on `/study/{id}/proposal` on STEP 2 with the typed content.
5. Repeat with only datasets, then only the PI, to cover the fields document.
6. A draft abandoned before Step 2 ever opened still lands on Step 1.

[OTTER-572]: https://openstax.atlassian.net/browse/OTTER-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/study-screen/state.ts
+++ b/src/lib/study-screen/state.ts
@@ -118,7 +118,7 @@ export function projectStudyState(raw: RawStudyState): StudyState {
     return {
         status: raw.status,
         isDraft: raw.status === 'DRAFT',
-        hasStep2Progress: draftHasStep2Progress(raw),
+        hasStep2Progress: draftHasStep2Progress(raw) || raw.hasStep2CollabDoc,
         researcherAgreementsAcked: !!raw.researcherAgreementsAckedAt,
         reviewerAgreementsAcked: !!raw.reviewerAgreementsAckedAt,
         hasAnyJob: raw.jobs.length > 0,
--- a/src/server/db/step2-collab-doc.ts
+++ b/src/server/db/step2-collab-doc.ts
@@ -1,0 +1,26 @@
+import { sql } from 'kysely'
+import { PROPOSAL_FIELDS_SUFFIX, PROPOSAL_PREFIX, PROPOSAL_TEXT_SLUGS } from '@/lib/collaboration-documents'
+
+// Step 2 of the proposal wizard persists into Yjs, not into the study columns: every edit goes to the
+// `proposal-<studyId>-fields` document or to one per lexical field, and the columns are written only by
+// Previous / View as reviewer / Submit. A draft edited on Step 2 and left by any other exit (nav link,
+// closed tab, back button) therefore has empty Step 2 columns, which used to send the researcher back to
+// the Step 1 picker on reopen (OTTER-572).
+//
+// Every document name Step 2 can write for a study. The fields document alone would very nearly do, since
+// it is created the first time Step 2 mounts (use-yjs-form-map seeds title/datasets/piName/piUserId into a
+// document that does not exist yet), but listing the lexical documents too keeps the signal correct if that
+// seeding ever changes.
+const STEP2_DOC_SUFFIXES = [PROPOSAL_FIELDS_SUFFIX, ...PROPOSAL_TEXT_SLUGS.map((slug) => `-${slug}`)]
+
+const docNameForStudyRow = (suffix: string) => sql`${PROPOSAL_PREFIX} || "study"."id"::text || ${suffix}`
+
+// Exact name equality, never `name like 'proposal-' || id || '-%'`: a pattern built per outer row is not a
+// planner-time constant, so it could not drive a prefix range scan (and a non-C collation would need
+// text_pattern_ops anyway), whereas equality against the outer row resolves through yjs_document's
+// name primary key.
+export const hasStep2CollabDocSql = sql<boolean>`exists (
+    select 1
+      from "yjs_document"
+     where "yjs_document"."name" = any(array[${sql.join(STEP2_DOC_SUFFIXES.map(docNameForStudyRow))}])
+)`
--- a/docs/study-screens-logic.md
+++ b/docs/study-screens-logic.md
@@ -116,6 +116,11 @@ correctly reads "under review again." Counting handles both shapes and is permut
 - `displayStatus` — highest-priority **present** status (see `DISPLAY_STATUS_PRIORITY`), with
   stale code decisions dropped on a fresh resubmission; falls back to the study status.
 - `submissionRound` — count of jobs that ever carried a `CODE-SUBMITTED` (the one cross-job fact).
+- `hasStep2Progress` = the DRAFT reached Step 2 of the proposal wizard, from **either** persistence
+  layer: a written Step 2 column (`draftHasStep2Progress`) or an existing Step 2 collaborative
+  document (`hasStep2CollabDoc`, see `server/db/step2-collab-doc.ts`). Both are needed because
+  collaborative Step 2 autosaves into Yjs and flushes the columns only on Previous / View as
+  reviewer / Submit, so the columns alone under-report progress (OTTER-572).
 
 ---
 
@@ -207,12 +212,17 @@ share a `guardExecutionStage` helper for their common precondition checks.
 
 | #   | When                                                                                       | Link             | Label                     |
 | --- | ------------------------------------------------------------------------------------------ | ---------------- | ------------------------- |
-| 1   | `isDraft`                                                                                  | `studyEdit`      | `Edit` (+ `delete-draft`) |
-| 2   | `APPROVED && hasAnyJob && !hasSubmittedCode`                                               | `studyCode`      | `View`                    |
-| 3   | `hasAnyJob`                                                                                | `studyView`      | `View`                    |
-| 4   | `APPROVED && researcherAgreementsAcked`                                                    | `studyCode`      | `View`                    |
-| 5   | post-submission status, no job (`PENDING-REVIEW`/`APPROVED`/`REJECTED`/`CHANGE-REQUESTED`) | `studySubmitted` | `View`                    |
-| 6   | fallback                                                                                   | `studyView`      | `View`                    |
+| 1   | `isDraft && hasStep2Progress`                                                              | `studyProposal`  | `Edit` (+ `delete-draft`) |
+| 2   | `isDraft`                                                                                  | `studyEdit`      | `Edit` (+ `delete-draft`) |
+| 3   | `APPROVED && hasAnyJob && !hasSubmittedCode`                                               | `studyCode`      | `View`                    |
+| 4   | `hasAnyJob`                                                                                | `studyView`      | `View`                    |
+| 5   | `APPROVED && researcherAgreementsAcked`                                                    | `studyCode`      | `View`                    |
+| 6   | post-submission status, no job (`PENDING-REVIEW`/`APPROVED`/`REJECTED`/`CHANGE-REQUESTED`) | `studySubmitted` | `View`                    |
+| 7   | fallback                                                                                   | `studyView`      | `View`                    |
+
+Rule 1 is the draft-resume rule (OTTER-572): a draft left on Step 2 reopens on Step 2, and only a
+draft that never got there lands on the Step 1 data-partner picker. `/edit` itself stays
+redirect-free so Step 2's Previous button remains a working escape hatch.
 
 ---
 
--- a/src/components/dashboard/studies-table/dashboard-raw-state.test.ts
+++ b/src/components/dashboard/studies-table/dashboard-raw-state.test.ts
@@ -22,6 +22,7 @@ const row = (overrides: Partial<StudyRow>): StudyRow => ({
     projectSummary: null,
     impact: null,
     additionalNotes: null,
+    hasStep2CollabDoc: false,
     ...overrides,
 })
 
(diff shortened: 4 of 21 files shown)


src/server/db/step2-collab-doc.ts
```diff
@@ -0,0 +1,26 @@
...
+const docNameForStudyRow = (suffix: string) => sql`${PROPOSAL_PREFIX} || "study"."id"::text || ${suffix}`
+
+// Exact name equality, never `name like 'proposal-' || id || '-%'`: a pattern built per outer row is not a
+// planner-time constant, so it could not drive a prefix range scan (and a non-C collation would need
+// text_pattern_ops anyway), whereas equality against the outer row resolves through yjs_document's
+// name primary key.
+export const hasStep2CollabDocSql = sql<boolean>`exists (
+    select 1
+      from "yjs_document"
+     where "yjs_document"."name" = any(array[${sql.join(STEP2_DOC_SUFFIXES.map(docNameForStudyRow))}])
```
nathanstitt: The correlation to the outer row is implicit here — it rides entirely on the study id being embedded in the constructed document name, rather than on a `where "yjs_document"."study_id" = "study"."id"`. It is correct (the `does not read another study's documents` test pins it), and I can see why: `study_id` has only an FK and no index, while `name` is the primary key, so probing by name is the fast path.

Worth considering adding the `study_id` equality as a redundant conjunct anyway. It costs nothing on the PK probe, but it makes the subquery self-evidently correlated to a reader, and it fails closed rather than silently matching if the naming convention ever drifts from what `proposalFieldsDocName` builds. Right now the only thing keeping this scoped to one study is string construction, which is a slightly uncomfortable place for a cross-row isolation guarantee to live.

Either way, could we get a one-line comment stating the correlation is via the embedded id? That is the non-obvious part of this fragment.

src/server/db/step2-collab-doc.ts
```diff
@@ -0,0 +1,26 @@
...
+// `proposal-<studyId>-fields` document or to one per lexical field, and the columns are written only by
+// Previous / View as reviewer / Submit. A draft edited on Step 2 and left by any other exit (nav link,
+// closed tab, back button) therefore has empty Step 2 columns, which used to send the researcher back to
+// the Step 1 picker on reopen (OTTER-572).
+//
+// Every document name Step 2 can write for a study. The fields document alone would very nearly do, since
+// it is created the first time Step 2 mounts (use-yjs-form-map seeds title/datasets/piName/piUserId into a
+// document that does not exist yet), but listing the lexical documents too keeps the signal correct if that
+// seeding ever changes.
+const STEP2_DOC_SUFFIXES = [PROPOSAL_FIELDS_SUFFIX, ...PROPOSAL_TEXT_SLUGS.map((slug) => `-${slug}`)]
```
nathanstitt: Nice that this derives from `PROPOSAL_TEXT_SLUGS` rather than re-spelling the slugs, and that `PROPOSAL_FIELDS_SUFFIX` got extracted for the same reason — that is exactly the kind of coupling that rots quietly.

One gap: this builds the lexical names as `-${slug}` here, while `proposalTextFieldDocName` builds them as `${PROPOSAL_PREFIX}${studyId}-${FIELD_TO_SLUG[fieldKey]}`. The two agree today, but they are separate spellings of one convention, so a change to the separator in `collaboration-documents.ts` would leave this fragment quietly matching nothing — and the failure mode is a silent regression of the exact bug this PR fixes, not a test error, unless the `study-state-query` tests catch it (they would, since they build names via the real helpers — good).

Mostly flagging so it is a conscious tradeoff. A comment pointing at `proposalTextFieldDocName` as the source of truth would be enough.

src/lib/study-screen/state.ts
```diff
@@ -118,7 +118,7 @@ export function projectStudyState(raw: RawStudyState): StudyState {
     return {
         status: raw.status,
         isDraft: raw.status === 'DRAFT',
-        hasStep2Progress: draftHasStep2Progress(raw),
+        hasStep2Progress: draftHasStep2Progress(raw) || raw.hasStep2CollabDoc,
```
nathanstitt: it seems like `hasStep2CollabDoc` is unconditioned on `isDraft` here, so a submitted or approved study with proposal documents also projects `hasStep2Progress: true`.  Seems like we should add a `!isDraft` to this just so they can't?

nathanstitt (approved):
looks good, Marvin but might take a look at the isDraft vs hasStepTwo

src/server/db/step2-collab-doc.ts
therealmarv: Agreed, and done in a9ce3e79. `study_id` is now a conjunct alongside the name equality, and the comment above the fragment says why both are there: [step2-collab-doc.ts#L24-L39](https://github.com/safeinsights/management-app/blob/a9ce3e7934f34aad29a3f28efad4981d870849ab/src/server/db/step2-collab-doc.ts#L24-L39).

You read the plan right: `name` is the primary key and `study_id` carries only an FK with no index, so the name equality is still the access path and the `study_id` check is a filter on an already-fetched row. It costs nothing and it takes the cross-study isolation guarantee off string construction, which was the uncomfortable part.

It also fails closed as you describe, and there is now a test for exactly that: a row whose name embeds this study but whose `study_id` points at another is not counted ([study-state-query.test.ts#L79](https://github.com/safeinsights/management-app/blob/a9ce3e7934f34aad29a3f28efad4981d870849ab/src/server/db/study-state-query.test.ts#L79)). No writer can produce that row today (the editor service derives `study_id` from the same parsed name, `services/editor/auth.ts` `studyIdForDocument`), so the test stands in for the convention drifting rather than for a reachable state.

src/server/db/step2-collab-doc.ts
therealmarv: Good catch, and rather than document the tradeoff I removed it. `collaboration-documents.ts` now exports [`proposalTextFieldSuffix`](https://github.com/safeinsights/management-app/blob/a9ce3e7934f34aad29a3f28efad4981d870849ab/src/lib/collaboration-documents.ts#L36-L39), `proposalTextFieldDocName` builds from it, and the SQL fragment maps `PROPOSAL_TEXT_SLUGS` through the same function: [step2-collab-doc.ts#L20](https://github.com/safeinsights/management-app/blob/a9ce3e7934f34aad29a3f28efad4981d870849ab/src/server/db/step2-collab-doc.ts#L20). The separator is now spelled once, the same treatment `PROPOSAL_FIELDS_SUFFIX` already got, so there is no second spelling left to drift.

You were right that `study-state-query` would have caught it (it builds names through the real helpers), but a silent regression of the exact bug this PR fixes is not something I want relying on a test being present.

src/lib/study-screen/state.ts
therealmarv: Right on the observation, and gated in a9ce3e79: [state.ts#L118-L127](https://github.com/safeinsights/management-app/blob/a9ce3e7934f34aad29a3f28efad4981d870849ab/src/lib/study-screen/state.ts#L118-L127).

Two things worth separating, though. It is not only the new fact: `draftHasStep2Progress` was equally ungated, so a submitted study with `datasets` written already projected `hasStep2Progress: true` before this PR. And the gate changes no behavior today, because the single consumer (dashboard rule 1, `isDraft && hasStep2Progress`) already required `isDraft`. So I put the gate on the whole expression rather than just on `hasStep2CollabDoc`: the field's own doc comment says "a DRAFT that has reached Step 2", and now it is literally that from either layer, instead of a fact that happens to be safe because every caller remembers to `&&` it.

Worth doing because neither layer self-clears: `Submit` is what writes the columns, and a CHANGE-REQUESTED round writes the documents again, so the traces outlive the wizard.

Covered by [state.test.ts#L228](https://github.com/safeinsights/management-app/blob/a9ce3e7934f34aad29a3f28efad4981d870849ab/src/lib/study-screen/state.test.ts#L228), which asserts false across PENDING-REVIEW / APPROVED / REJECTED / CHANGE-REQUESTED from both layers. I also rendered the real `StudiesTableView` + `StudyActionLink` against fake rows carrying both signals on every status: the four non-DRAFT rows resolve to `/submitted` and only the DRAFT control resolves to `/proposal`. Docs table updated in `docs/study-screens-logic.md` too.

therealmarv:
Thanks @nathanstitt. All three comments are addressed in a9ce3e79, replies are in the threads:

- **isDraft vs hasStep2Progress** (the one from your review summary): `hasStep2Progress` is now gated on `isDraft` in the projection. No behavior change, since the only consumer already required `isDraft`, but the fact now matches its own definition instead of depending on every caller remembering the guard. Note the pre-existing column check was ungated too, so the gate covers both layers, not just the new one.
- **`study_id` correlation**: added as a conjunct next to the name equality, with a comment explaining that the name probe is the index path and `study_id` is the explicit scoping that fails closed.
- **Lexical name spelling**: removed rather than documented. `proposalTextFieldSuffix` is now the single source, used by both `proposalTextFieldDocName` and the SQL fragment.

Two new tests (non-DRAFT statuses report no Step 2 progress; a name/`study_id` mismatch is not counted). Full unit suite green at 266 files / 2615 tests, `pnpm run checks` clean. I also rendered the real dashboard table against fake rows for each case (collab doc only, columns only, both, neither, and every non-DRAFT status) and confirmed the resolved hrefs match. PR description updated.

PR #964, opened 2026-08-12
OTTER-695: route the feedback-only outputs decision to its own resear…

Rod0352:
## What this does

A clean run (`RUN-COMPLETE`, never `JOB-ERRORED`) that the Data Partner decided with **Share feedback only** used to fall through to the results-only Study Details screen. The researcher table now claims that state first and routes `/view` to a new `outputs-feedback` screen:

- **Rule** (`researcher-screen-rules.ts`): `['outputs-feedback', { when: (s) => s.resultsRejected && !s.resultsErrored }]`, inserted above `study-results`. Per `submitOutputsDecisionAction` (OTTER-675/676), "Share feedback only" is persisted as `FILES-REJECTED`, so this is the ticket's "results-approved + Share feedback only" condition expressed in projected state. An errored run's `FILES-REJECTED` still lands on `study-results` (the errored variant is its own card), as does a share-outputs decision (`resultsApproved`).
- **Screen** (`_screens/outputs-feedback-screen.tsx`): reuses `StudyPageHeader`, `ProposalStepHeader` (STEP 4 / Verify outputs, which owns the lg-spaced divider), the **action**-variant `StatusAlert` titled `Feedback on outputs available • {date}` (date = the `FILES-REJECTED` status change, undated degrade when missing), `FeedbackAndNotesSection`, and the `xxl` stack spacing used by the sibling outputs screens. Nav: **Previous step** (subtle) → `/view/code` (resolves to the code-approved screen for these studies), **Edit code** (outline, enabled) → `/resubmit` — already resubmit-eligible via `canResearcherResubmitCode` (`resultsRejected`), no route changes.
- **Data** (`getOutputsFeedbackAction`): the `RESULTS` counterpart of `getCodeReviewFeedbackAction`. Both now delegate to a shared `loadReviewFeedbackThread(db, studyId, reviewKind)` so entry shape, version labels, and newest-first ordering can't drift between the code and outputs surfaces. The CODE action's behavior is unchanged (its existing tests all pass untouched).
- **Docs**: researcher table + precedence note updated in `docs/study-screens-logic.md` per the screen-rules doc rule.
- `hasNextStepFromCode` delegation means the code-approved page's forward link starts working for this state with no change there (by design, see `next-step.ts`).

## Tests

- `resolve.test.ts`: redirection matrix — feedback-only → `outputs-feedback`; share-outputs, undecided `RUN-COMPLETE`, and errored+rejected all stay on `study-results`; walk-back (`/view/code`) resolves to `code-approved` for this state.
- `consistency.test.ts`: dashboard `View` link for the state resolves to a real screen.
- `outputs-feedback-screen.test.tsx` (12 tests): reused header/section-header/alert/feedback-section via their shared test ids, exact banner copy + DP-name interpolation, decision date (not approval date, not today), undated degrade, null-thread and wrong-state guards, study-scoped feedback isolation, CODE-entry exclusion, nav hrefs + `subtle`/`outline` variants + enabled-by-default.
- `page.test.tsx`: full `/view` dispatch lands on the new screen for `FILES-REJECTED` (moved out of the Study Details `it.each`, which documented the pre-refactor behavior).
- `study.actions.test.ts`: `getOutputsFeedbackAction` returns RESULTS + notes newest-first with round labels; kind isolation is symmetric with the CODE action.

### DIFF
--- a/src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx
+++ b/src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx
@@ -1,0 +1,98 @@
+import type { Route } from 'next'
+import { Box, Group, Stack } from '@mantine/core'
+import { CaretLeftIcon } from '@phosphor-icons/react/dist/ssr'
+import dayjs from 'dayjs'
+import { AlertNotFound } from '@/components/errors'
+import { ButtonLink } from '@/components/links'
+import { FeedbackAndNotesSection } from '@/components/study/feedback-and-notes'
+import { ProposalStepHeader } from '@/components/study/proposal-step-header'
+import { StatusAlert, STATUS_ALERT_SEPARATOR, STATUS_ALERT_VARIANT } from '@/components/study/status-alert'
+import { StudyPageHeader } from '@/components/study/study-page-header'
+import { Routes } from '@/lib/routes'
+import { displayOrgName } from '@/lib/string'
+import { latestStatusAt } from '@/lib/study-job-status'
+import { projectStudyState } from '@/lib/study-screen'
+import type { OutputsFeedbackEntry } from '@/server/actions/study.actions'
+import { getOrgNameFromId, latestSubmittedJobForStudy } from '@/server/db/queries'
+import { loadOutputsFeedback } from '../view/load-outputs-feedback'
+import type { ScreenComponentProps } from './types'
+
+const FeedbackOnlyBanner = ({ decidedAt, dataPartner }: { decidedAt: Date | string | null; dataPartner: string }) => {
+    // Display-only date: degrade to an undated banner rather than block a page routing already chose.
+    const decidedOn = decidedAt ? ` ${STATUS_ALERT_SEPARATOR} ${dayjs(decidedAt).format('MMM DD, YYYY')}` : ''
+    return (
+        <StatusAlert variant={STATUS_ALERT_VARIANT.action} title={`Feedback on outputs available${decidedOn}`}>
+            {dataPartner} has shared feedback on the latest code run. The outputs are not available for this study. When
+            you are ready, edit your code and resubmit.
+        </StatusAlert>
+    )
+}
+
+// Mirrors the code surface: a failed fetch swaps in the shared notice instead of hiding the section.
+const FeedbackSection = ({
+    feedbackLoadError,
+    entries,
+}: {
+    feedbackLoadError: boolean
+    entries: OutputsFeedbackEntry[]
+}) => {
+    if (feedbackLoadError) {
+        return <AlertNotFound title="Feedback could not be loaded" message="Please refresh and try again" />
+    }
+    return <FeedbackAndNotesSection entries={entries} alwaysExpandLatest />
+}
+
+// OTTER-695: clean run whose outputs the reviewer withheld with "Share feedback only"
+// (FILES-REJECTED without JOB-ERRORED); the researcher reads the feedback and resubmits.
+export async function OutputsFeedbackScreen({
+    study,
+    raw,
+    orgSlug,
+    returnTo,
+}: Pick<ScreenComponentProps, 'study' | 'raw' | 'orgSlug' | 'returnTo'>) {
+    const job = await latestSubmittedJobForStudy(study.id)
+    if (!job) {
+        return <AlertNotFound title="No submission found" message="This study has no submitted code yet." />
+    }
+
+    // Guards the same facts the researcher rule routes on, so routing and rendering cannot
+    // disagree; the query above supplies only the banner's date payload.
+    const state = projectStudyState(raw)
+    if (!state.resultsRejected || state.resultsErrored) {
+        return (
+            <AlertNotFound
+                title="Feedback not found"
+                message="This study does not have outputs feedback to display yet."
+            />
+        )
+    }
+
+    const { entries, feedbackLoadError } = await loadOutputsFeedback(study.id)
+    const dataPartner = displayOrgName(await getOrgNameFromId(study.orgId))
+    const decidedAt = latestStatusAt(job.statusChanges, 'FILES-REJECTED')
+    const previousHref = Routes.studyViewCode({ orgSlug, studyId: study.id, returnTo }) as Route
+    const editCodeHref = Routes.studyResubmit({ orgSlug, studyId: study.id }) as Route
+
+    return (
+        <Box bg="grey.10">
+            <Stack px="xl" gap="xxl" py="xl">
+                <StudyPageHeader>Secondary analysis study</StudyPageHeader>
+                <ProposalStepHeader
+                    stepLabel="STEP 4"
+                    heading="Verify outputs"
+                    studyTitle={study.title!}
+                    banner={<FeedbackOnlyBanner decidedAt={decidedAt} dataPartner={dataPartner} />}
+                />
+                <FeedbackSection feedbackLoadError={feedbackLoadError} entries={entries} />
+                <Group justify="space-between">
+                    <ButtonLink href={previousHref} variant="subtle" leftSection={<CaretLeftIcon />}>
+                        Previous step
+                    </ButtonLink>
+                    <ButtonLink href={editCodeHref} variant="outline" size="md">
+                        Edit code
+                    </ButtonLink>
+                </Group>
+            </Stack>
+        </Box>
+    )
+}
--- a/src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback.ts
+++ b/src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback.ts
@@ -1,0 +1,9 @@
+import { getOutputsFeedbackAction } from '@/server/actions/study.actions'
+import { isActionError } from '@/lib/errors'
+
+export async function loadOutputsFeedback(studyId: string) {
+    const entriesResult = await getOutputsFeedbackAction({ studyId })
+    const feedbackLoadError = isActionError(entriesResult)
+    const entries = feedbackLoadError ? [] : entriesResult
+    return { entries, feedbackLoadError }
+}
--- a/src/lib/study-screen/resolve.test.ts
+++ b/src/lib/study-screen/resolve.test.ts
@@ -29,6 +29,34 @@ describe('resolveScreen (researcher)', () => {
             ).screen,
         ).toBe('study-results')
     })
+    it('OTTER-695: feedback-only decision on a clean run → outputs-feedback (out-ranks study-results)', () => {
+        expect(
+            resolveScreen(
+                'researcher',
+                state({ hasResults: true, resultsRejected: true, codeDecision: 'CODE-APPROVED' }),
+            ).screen,
+        ).toBe('outputs-feedback')
+    })
+    it('OTTER-695: share-outputs decision (resultsApproved) still → study-results', () => {
+        expect(
+            resolveScreen(
+                'researcher',
+                state({ hasResults: true, resultsApproved: true, codeDecision: 'CODE-APPROVED' }),
+            ).screen,
+        ).toBe('study-results')
+    })
+    it('OTTER-695: clean run with no files decision yet (RUN-COMPLETE) still → study-results, not outputs-feedback', () => {
+        expect(
+            resolveScreen(
+                'researcher',
+                state({
+                    hasResults: true,
+                    resultsDisplayStatus: 'RUN-COMPLETE',
+                    codeDecision: 'CODE-APPROVED',
+                }),
+            ).screen,
+        ).toBe('study-results')
+    })
     it('errored job with a stale code decision (resubmission) → code-under-review, NOT study-results (OTTER-598)', () => {
         // Edge case raised in PR #837: resultsErrored excludes from study-results, the prior
         // CODE-APPROVED was dropped by dropStale (so codeDecision is null and codeAwaitingDecision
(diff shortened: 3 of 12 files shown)


src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const editCodeHref = Routes.studyResubmit({ orgSlug, studyId: study.id }) as Route
+
+    return (
+        <Box bg="grey.10">
+            <Stack px="xl" gap="xxl" py="xl">
+                <StudyPageHeader>Secondary analysis study</StudyPageHeader>
+                <ProposalStepHeader
+                    stepLabel="STEP 4"
+                    heading="Verify outputs"
+                    studyTitle={study.title!}
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx]: `study.title!` is an unchecked assertion here. `StudyShape` types `title` as `string | null`, and the sibling screens only reach for `!` *after* narrowing: `outputs-pending-screen` gets there via `guardExecutionStage`, which runs `isSubmittedStudy(study)` first, and `proposal-feedback-screen` calls it directly. This screen has no such guard, so the `!` is doing real work rather than restating a proven fact.

In practice routing should mean a FILES-REJECTED study is never a titleless draft, so this is likely unreachable today — but it's the kind of thing that only surfaces as a blank heading. Either add the `isSubmittedStudy` guard alongside the `!job` check, or follow `reviewer-outputs-available-screen` and use `study.title ?? ''`.

src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback.ts
```diff
@@ -0,0 +1,9 @@
+import { getOutputsFeedbackAction } from '@/server/actions/study.actions'
+import { isActionError } from '@/lib/errors'
+
+export async function loadOutputsFeedback(studyId: string) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback.ts]: This file is byte-for-byte identical to `load-code-review-feedback.ts` apart from which action it calls — same `isActionError` swallow, same `entries`/`feedbackLoadError` shape. Given the PR already went to the trouble of extracting `loadReviewFeedbackThread` server-side to stop the two surfaces drifting, it seems a shame to reintroduce the drift risk one layer up.

Could we collapse both into one parameterised helper? Something like `loadReviewFeedback(studyId, kind)` dispatching on `'CODE' | 'RESULTS'`, mirroring the shape the shared thread loader already takes. Then there's one place that decides what a failed feedback fetch looks like.

src/server/actions/study.actions.ts
```diff
@@ -866,117 +982,26 @@ export const getCodeReviewFeedbackAction = new Action('getCodeReviewFeedbackActi
...
-
-            return a.id.localeCompare(b.id)
-        })
+// OTTER-695: outputs counterpart of getCodeReviewFeedbackAction ('RESULTS' rows are written by
+// submitOutputsDecisionAction).
+export const getOutputsFeedbackAction = new Action('getOutputsFeedbackAction')
+    .params(z.object({ studyId: z.string().uuid() }))
+    .middleware(async ({ params: { studyId }, db }) => {
+        const study = await db
+            .selectFrom('study')
```
nathanstitt: The `.params()` + `.middleware()` + `.requireAbilityTo('view', 'Study')` prelude is now duplicated verbatim between `getCodeReviewFeedbackAction` and this one — same study lookup, same three selected columns, same `throwNotFound('study')`, same ability.

Since the handlers already delegate to a shared function, the remaining duplication is the authorization surface, which is the part worth keeping in one place: if the ability for viewing feedback ever changes, one of these two is going to get missed. Worth a small factory (e.g. `studyFeedbackAction(name, kind)`) that builds both, so the auth check exists once?

src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx
```diff
@@ -0,0 +1,98 @@
...
+}: Pick<ScreenComponentProps, 'study' | 'raw' | 'orgSlug' | 'returnTo'>) {
+    const job = await latestSubmittedJobForStudy(study.id)
+    if (!job) {
+        return <AlertNotFound title="No submission found" message="This study has no submitted code yet." />
+    }
+
+    // Guards the same facts the researcher rule routes on, so routing and rendering cannot
+    // disagree; the query above supplies only the banner's date payload.
+    const state = projectStudyState(raw)
+    if (!state.resultsRejected || state.resultsErrored) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx]: Nice that the guard restates the routing predicate so render and routing can't disagree. One thought: it restates it *by hand*, so the rule in `researcher-screen-rules.ts:12` (`s.resultsRejected && !s.resultsErrored`) and this negation are two copies of the same fact that have to be kept in sync manually.

The codebase already has form for this — `awaitingFilesDecisionOnError` was extracted precisely so the two screen-rule tables and the pill couldn't drift. Could we lift this one the same way (`isFeedbackOnlyOutcome(s)` or similar in `state.ts`) and have both the rule and this guard call it?

src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+// OTTER-695: clean run whose outputs the reviewer withheld with "Share feedback only"
+// (FILES-REJECTED without JOB-ERRORED); the researcher reads the feedback and resubmits.
+export async function OutputsFeedbackScreen({
+    study,
+    raw,
+    orgSlug,
+    returnTo,
+}: Pick<ScreenComponentProps, 'study' | 'raw' | 'orgSlug' | 'returnTo'>) {
+    const job = await latestSubmittedJobForStudy(study.id)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx]: Minor ordering nit: we hit the database for the job before the state guard on line 61 runs, but the job is only used for the banner's `decidedAt`. On the guard-fail path that's a wasted query.

Moving `projectStudyState(raw)` and its guard above this call would short-circuit before any I/O — `raw` is already in hand, so the check is free.

src/server/actions/study.actions.ts
```diff
@@ -855,6 +855,122 @@ export const submitCodeReviewDecisionAction = new Action('submitCodeReviewDecisi
         return { submitterFullName: submitter.fullName }
     })
 
+// Reviewer DECISION rows of one reviewKind merged with resubmission notes, newest first. Shared by
+// the code and outputs feedback actions so entry shape, version labels, and ordering can't drift.
+async function loadReviewFeedbackThread(
+    db: DBExecutor,
+    studyId: string,
+    reviewKind: Extract<StudyReviewCommentKind, 'CODE' | 'RESULTS'>,
```
nathanstitt, line 873: Good call typing this as `Extract<StudyReviewCommentKind, 'CODE' | 'RESULTS'>` rather than a bare string — it keeps `PROPOSAL` rows out by construction, and it'll fail to compile rather than silently widen if the kind union grows. Worth noting since the exclusion is the whole point of the kind-isolation tests.

nathanstitt (changes requested):
looks good but need to remove the non-null assertion, that's kinda bad practice, in fact we probably should add a eslint rule disallowing it.  other than that just some minor duplication

src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx]: swapped the ! for an isSubmittedStudy guard alongside the job check

src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback.ts
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback.ts]: Collapsed both into loadReviewFeedback(studyId, kind)

src/server/actions/study.actions.ts
Rod0352: Tried the factory, but validate-actions requires each export to be a literal new Action('<export name>') chain, so I extracted the shared lookup (feedbackStudyAuthz) and left a comment noting the constraint

src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx]: Lifted to isFeedbackOnlyOutcome in state.ts

src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx
Comment (Rod0352) [reply src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx]: Reordered

src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx
```diff
@@ -0,0 +1,101 @@
...
+                title="Feedback not found"
+                message="This study does not have outputs feedback to display yet."
+            />
+        )
+    }
+    if (!isSubmittedStudy(study)) {
+        return <AlertNotFound title="No submission found" message="This study has no submitted code yet." />
+    }
+
+    const job = await latestSubmittedJobForStudy(study.id)
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/outputs-feedback-screen.tsx]: Thanks for reordering this — the guard now short-circuits before any I/O, which is great.

something else to think about is the guard decides from `raw` (whose `latestJob` prefers the newest job carrying any non-`INITIATED` status), while `decidedAt` on line 75 comes from a separately-fetched `latestSubmittedJobForStudy`. Those two "latest job" definitions are close but not identical, and they're resolved by two different queries at two different moments. If they ever disagree, the symptom is the soft one the `FeedbackOnlyBanner` already handles — a missing date — so this is genuinely not a bug today. It just means the undated-banner degradation is loadbearing rather than purely defensive, which is worth knowing.

Since `raw` is already in hand, sourcing `decidedAt` from it would collapse the two sources into one and drop a query.

src/server/actions/study.actions.ts
```diff
@@ -865,122 +865,141 @@ export const submitCodeReviewDecisionAction = new Action('submitCodeReviewDecisi
...
+    .middleware(studyViewMiddleware)
+    .requireAbilityTo('view', 'Study')
+    .handler(async ({ params: { studyId }, db }) => loadReviewFeedbackThread(db, studyId, 'CODE'))
 
 export type CodeReviewFeedbackEntry = ActionSuccessType<typeof getCodeReviewFeedbackAction>[number]
 
+// OTTER-695: the researcher-facing outputs thread — RESULTS decisions plus resubmission notes.
+// Distinct from getOutputsDecisionFeedbackAction above, which returns the decisions alone for the
+// reviewer's post-review page.
+export const getOutputsFeedbackThreadAction = new Action('getOutputsFeedbackThreadAction')
```
nathanstitt, line 995: Fair enough on the factory — I checked and `validate-actions` really does require each export to be a literal `new Action('<export name>')` chain, so a builder genuinely isn't available here.

Worth noting the concern behind my original comment is mostly resolved anyway, just not where I expected: `studyViewMiddleware` was already shared, so the study lookup, the three selected columns and `throwNotFound('study')` all live in one place. What's left duplicated is four declarative chain lines, and the auth surface (`.requireAbilityTo('view', 'Study')`) is one grep away from both. That's an acceptable amount of repetition for a framework constraint.

(Minor: your reply mentions extracting `feedbackStudyAuthz`, but there's no such symbol at head — I think you landed on reusing the existing middleware instead, which is the better outcome. Just flagging so the thread doesn't mislead later.)

src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback-thread.ts
```diff
@@ -0,0 +1,5 @@
+import { getOutputsFeedbackThreadAction } from '@/server/actions/study.actions'
+import { loadFeedbackEntries } from './load-feedback-entries'
+
+export const loadOutputsFeedbackThread = (studyId: string) =>
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback-thread.ts:4]: This is a really clean resolution of the duplication thread — better than the parameterised `loadReviewFeedback(studyId, kind)` I originally suggested.

By making `loadFeedbackEntries` generic over the action rather than switching on a `'CODE' | 'RESULTS'` discriminant, there's exactly one place that decides what a failed feedback fetch looks like, and adding a third thread later is a two-line file with no edit to the shared helper. The `isActionError` swallow existing once is the part that matters.

src/lib/study-screen/resolve.test.ts
```diff
@@ -29,6 +29,34 @@ describe('resolveScreen (researcher)', () => {
...
+                'researcher',
+                state({ hasResults: true, resultsRejected: true, codeDecision: 'CODE-APPROVED' }),
+            ).screen,
+        ).toBe('outputs-feedback')
+    })
+    it('OTTER-695: share-outputs decision (resultsApproved) still → study-results', () => {
+        expect(
+            resolveScreen(
+                'researcher',
+                state({ hasResults: true, resultsApproved: true, codeDecision: 'CODE-APPROVED' }),
```
nathanstitt: Good instinct testing the two negative cases (`resultsApproved` still → `study-results`, and bare `RUN-COMPLETE` still → `study-results`) alongside the positive one. Precedence changes fail by over-claiming, and a rule promoted to #1 is precisely the kind that quietly swallows neighbouring states — these two pin the boundary rather than just the happy path.

The `consistency.test.ts` addition covering the Tier-1/Tier-2 pairing for this state is the right companion to it.

nathanstitt (approved):
looks good, take a look at the comments, it probably is good to use the `raw` and skip an extra query.

PR #965, opened 2026-08-13
OTTER-677: Add reviewer post-decision screen for outputs

stetradis:
## Summary

- Replaces the generic `reviewer-study-results screen` with a purpose-built `reviewer-outputs-decided` screen that shows the decision banner (approved/rejected x errored/completed), the reviewer's outputs feedback, and a re-decrypt section for viewing outputs again.
- Adds `OutputsDecidedBanner` with 4 variants covering every combination of errored/completed run and shared/feedback-only decision.
- Adds `DecryptAndViewOutputs` component that reuses SecurityKeyForm with customizable title/description, letting the reviewer re-enter their security key to view outputs post-decision.
- Adds getOutputsDecisionFeedbackAction to fetch RESULTS-kind review comments so the feedback panel shows outputs decision feedback instead of code review feedback.
- Updates reviewer screen rules to route hasResults to reviewer-outputs-decided and updates resolveReviewerCodeScreen comments accordingly.


src/app/[orgSlug]/study/[studyId]/_screens/registry.ts
```diff
@@ -15,7 +15,7 @@ import { OutputsPendingScreen } from './outputs-pending-screen'
 import { ReviewerOutputsPendingScreen } from './reviewer-outputs-pending-screen'
 import { ReviewerOutputsErroredScreen } from './reviewer-outputs-errored-screen'
 import { ReviewerOutputsAvailableScreen } from './reviewer-outputs-available-screen'
-import { ReviewerStudyResultsScreen } from './reviewer-study-results-screen'
+import { ReviewerOutputsDecidedScreen } from './reviewer-outputs-decided-screen'
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/_screens/registry.ts:18]: Now that this is the only consumer swap, `StudyDetailsReviewer` has no production callers left — `git grep StudyDetailsReviewer` only turns up its own test file plus the chain it drags along (`study-details-reviewer.tsx`, `study-results-redesign-with-review.tsx`, and `study-details-reviewer-view.tsx`, the last of which survives via `review-study-view.stories.tsx`).

Worth deleting `study-details-reviewer.tsx` + `study-details-reviewer.test.tsx` + `study-results-redesign-with-review.tsx` in this PR while the reason for their removal is fresh. As-is the test keeps a dead OTTER-538 screen green in CI, which reads to the next person like it's still routable.

src/app/[orgSlug]/study/[studyId]/review/reviewer-outputs-decided.tsx
```diff
@@ -0,0 +1,88 @@
...
+    feedbackLoadError,
+    entries,
+}) => {
+    if (feedbackLoadError) {
+        return <AlertNotFound title="Feedback could not be loaded" message="Please refresh and try again" />
+    }
+    return <FeedbackAndNotesSection entries={entries} alwaysExpandLatest />
+}
+
+export async function ReviewerOutputsDecided({ study, orgSlug }: ReviewerOutputsDecidedProps) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/reviewer-outputs-decided.tsx]: The two sibling screens deliberately guard on the *same* predicate the rule table routes on, rather than re-deriving it — `ReviewerOutputsErroredScreen` does `if (!awaitingFilesDecisionOnError(projectStudyState(raw)))` and the available screen does the equivalent, both with the comment "so routing and rendering cannot disagree (#922 review)".

This screen instead re-derives its own guard from `job.statusChanges` below, while rule 1c routes on `s.hasResults` from `projectStudyState`. Those come from different queries (`latestSubmittedJobForStudy` vs. the raw state projection), so they can drift — and the drift shows up as the "No decision found" alert on a study the router was sure belonged here.

Could we take `raw` off `ScreenComponentProps` and guard on the projected state like the siblings do? It'd also let the screen wrapper stay the thin pass-through the others are.

src/app/[orgSlug]/study/[studyId]/review/reviewer-outputs-decided.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const job = await latestSubmittedJobForStudy(study.id)
+    if (!job) {
+        return <AlertNotFound title="No submission found" message="This study has no submitted code to review." />
+    }
+
+    const statuses = new Set(job.statusChanges.map((c) => c.status))
+    const filesDecision = job.statusChanges.find((c) => c.status === 'FILES-APPROVED' || c.status === 'FILES-REJECTED')
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/reviewer-outputs-decided.tsx]: `@/lib/study-job-status` already has all three pieces this line hand-rolls:

- `ROUND_CLOSING_JOB_STATUSES` — literally `['FILES-APPROVED', 'FILES-REJECTED'] as const`, the exact pair inlined here
- `hasJobStatus(statusChanges, statuses)` — the existence test
- `latestStatusAt(statusChanges, status)` — max-by-`createdAt`, which both sibling screens use for their banner dates

So this could read:

```ts
if (!hasJobStatus(job.statusChanges, ROUND_CLOSING_JOB_STATUSES)) { /* No decision found */ }
const decidedAt =
    latestStatusAt(job.statusChanges, 'FILES-APPROVED') ?? latestStatusAt(job.statusChanges, 'FILES-REJECTED')
```

The reason I'd push on it rather than leave it as taste: `.find()` picks a row by *position* in a `desc`-ordered array, and the comment on `latestSubmittedJobHasLiveCodeDecision` documents that exact move as a bug we already shipped once — `jobStatusChange.createdAt` defaults to `now()` and is constant within a transaction, and v7 ids aren't reliably monotonic inside a millisecond, so statuses written together sort non-deterministically. That's why the existing helpers are all existence-or-reduce and never index-based.

It happens to be safe here, but only because `getOrCreateCurrentRoundJob` guarantees a round-closing status is never followed by another on the same job — exactly one FILES decision per job. That invariant is load-bearing for both this line and the set-based `resultsApproved` below, and it lives two files away with nothing here pointing at it.

src/components/study/decrypt-and-view-outputs.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+    return <DecryptedOutputs jobId={job.id} decryptedFiles={decryptedFiles} />
+}
+
+type DecryptedOutputsProps = {
+    jobId: string
+    decryptedFiles: JobFileInfo[]
+}
+
+const DecryptedOutputs: FC<DecryptedOutputsProps> = ({ jobId, decryptedFiles }) => {
```
nathanstitt: `DecryptedOutputs` is a near-exact copy of the files half of `ReviewBody` in `outputs-review-panel.tsx` — same `useOutputsFiles` call, same `previewFile` ternary, same `OutputsFilesSection` prop spread, same `FileOrImagePreviewModal` wiring. Only the decision section differs there.

Could we extract that block (say `OutputsFilesViewer({ jobId, decryptedFiles })`) and have both call it? Right now a change to the files table or the preview modal's props has to be made in two places, and the compiler won't tell you when you've only done one.

src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback.ts
```diff
@@ -0,0 +1,9 @@
+import { getOutputsDecisionFeedbackAction } from '@/server/actions/study.actions'
+import { isActionError } from '@/lib/errors'
+
+export async function loadOutputsFeedback(studyId: string) {
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/view/load-outputs-feedback.ts]: This is byte-for-byte `loadCodeReviewFeedback` in the sibling `load-code-review-feedback.ts` with a different action imported — same three lines, same `{ entries, feedbackLoadError }` shape.

A small generic would collapse both: `const loadFeedback = async (action, studyId) => { ... }`, or even just inline the three lines at each call site since the wrapper isn't buying much abstraction. Either beats two files that have to be kept in step.

src/server/actions/study.actions.ts
```diff
@@ -978,6 +978,51 @@ export const getCodeReviewFeedbackAction = new Action('getCodeReviewFeedbackActi
 
 export type CodeReviewFeedbackEntry = ActionSuccessType<typeof getCodeReviewFeedbackAction>[number]
 
+export const getOutputsDecisionFeedbackAction = new Action('getOutputsDecisionFeedbackAction')
```
nathanstitt, line 984: This action's `.params`/`.middleware`/`.requireAbilityTo` prelude is identical to `getCodeReviewFeedbackAction` directly above, and the row→entry mapper at the end differs only in that this one drops `criteria`. The handlers genuinely diverge (the code one does the lateral-join resubmission-note merge), so the whole thing isn't shared — but the study-lookup middleware is pure boilerplate repeated verbatim and would be worth hoisting into a named helper both `.middleware()` calls use.

Dropping `criteria` is right, by the way — `FeedbackAndNotesSection` never reads it and outputs decisions don't carry criteria. Might be worth a one-liner saying so, since the asymmetry with the neighbouring action otherwise looks like an oversight.

src/server/actions/study.actions.ts
```diff
@@ -978,6 +978,51 @@ export const getCodeReviewFeedbackAction = new Action('getCodeReviewFeedbackActi
...
+    })
+    .requireAbilityTo('view', 'Study')
+    .handler(async ({ params: { studyId }, db }) => {
+        const rows = await db
+            .selectFrom('studyReviewComment')
+            .innerJoin('user as author', 'author.id', 'studyReviewComment.authorId')
+            .select([
+                'studyReviewComment.id',
+                'studyReviewComment.authorId',
+                'studyReviewComment.entryType',
```
nathanstitt: `entryType` is selected here but then hardcoded to `'REVIEWER-FEEDBACK' as const` in the mapper, so the selected value is never read. (The `.where(entryType = 'DECISION')` filter is what makes the constant safe.) Same dead select exists in `getCodeReviewFeedbackAction`, so this is inherited rather than new — but since this action is being written fresh, dropping the column from the `.select([...])` would keep the query honest about what it uses.

src/app/[orgSlug]/study/[studyId]/review/reviewer-outputs-decided.tsx
```diff
@@ -0,0 +1,88 @@
...
+                        <OutputsDecidedBanner
+                            resultsErrored={resultsErrored}
+                            resultsApproved={resultsApproved}
+                            labName={labName}
+                            decidedAt={filesDecision.createdAt}
+                        />
+                    }
+                />
+                <FeedbackSection feedbackLoadError={feedbackLoadError} entries={feedbackEntries} />
+                <DecryptAndViewOutputs job={job} />
```
Comment (nathanstitt) [inline src/app/[orgSlug]/study/[studyId]/review/reviewer-outputs-decided.tsx:75]: The title/description passed down are fixed ("View outputs again" / "The outputs are encrypted…"), but this screen also serves both errored variants. On a `JOB-ERRORED` job what `fetchEncryptedJobFilesAction` returns is the encrypted *logs* — that's what the errored screen has the reviewer decrypt, and its own copy says "access the outputs and see what went wrong" rather than calling them outputs outright.

So on a `Code errored. Feedback shared` banner the section underneath still says "View outputs again", which reads as if a completed run's results are waiting. Since the banner already branches on `resultsErrored`, could the copy branch with it?

(Both the FILES-REJECTED and errored paths do decrypt fine — the reviewer is a manifest recipient regardless of the share decision — so this is wording, not broken behaviour.)

Comment (stetradis) [reply src/app/[orgSlug]/study/[studyId]/review/reviewer-outputs-decided.tsx:75]: Good catch — branching the copy here is possible and makes sense, since the banner already distinguishes the errored path.

I want to confirm the wording with UX before changing it. The pre-decision errored screen uses the same SecurityKeyForm copy (still “outputs”), and the post-decision Figma for the errored variant currently does too: https://www.figma.com/design/84pQzoYf7rKqOjo4qsDOeG/Card-59-69-71---Outputs-step?node-id=12287-25987&m=dev

All other concerns resolved.

PR #968, opened 2026-08-13
Don't report a note autosave that fails after navigating away

nathanstitt:
## Problem

Researchers saw `Unable to save resubmission note draft` toasts on QA with an
opaque reference, and the app kept working.

A Server Action posts to whatever route is current when the request goes out.
The debounced resubmission-note autosave can still be in flight when the
researcher navigates away, and it then resolves against the **new** route,
which has no matching action. Next returns a non-RSC 200, and the client throws:

```
An unexpected response was received from the server.
```

The Sentry event (MANAGEMENT-APP-3K1) shows the sequence plainly:

1. POST `/resubmit` → 200 (autosave, fine)
2. POST `/resubmit` → 200 (autosave, fine)
3. **navigation** `/resubmit` → `/view`
4. GET `/view?_rsc=…` → 200
5. POST **`/view`** → 200, `response_body_size: 2` ← throws

Every breadcrumb is **status 200**, which is what distinguishes this from a
rejected Server Action (those are 4xx/500). Next's throw needs both a non-RSC
content-type *and* no redirect header:

```js
if (!isRscResponse && !redirectLocation) throw new Error('An unexpected response was received from the server.')
```

## Change

Skip reporting when the provider has unmounted, in both the proposal and code
resubmit contexts (they had identical copies of the bug).

The queued-but-not-yet-fired case was already handled by the autosave effect's
`clearTimeout`; this covers the **in-flight** one, which wasn't.

Server Actions can't accept an `AbortSignal` (not serializable), so the request
itself can't be cancelled — only its now-meaningless failure suppressed. The
toast was reporting a save the researcher never asked for, on a page they'd
already left, that they have no way to retry.

## Tests

A regression test per context. Both fail without the fix, with the same error
text seen in Sentry:

```
expected "vi.fn()" to not be called at all, but actually been called 1 times
  [Error: An unexpected response was received from the server.]
```

## Verification

Typecheck and lint clean; the changed suites pass.

Note: `context.integration.test.tsx`, `use-resubmit-proposal.test.ts`, and
`use-submit-proposal.test.ts` fail locally with `missing ARN DB_SECRET_ARN` —
they need a database. They fail identically on a clean checkout of `main`, so
they're unaffected by this change, but I could not run them.

## Not the same bug as the encryption-key PRs

This is **not** a stale Server Action ID. Those return 4xx/500 and log
`Invalid Server Actions request.` server-side; this one is 200 throughout.
management-app #967 and iac #218 fix that separate cutover issue, which
produces a similar-looking toast.

### DIFF
--- a/src/contexts/edit-code-resubmit/context.test.tsx
+++ b/src/contexts/edit-code-resubmit/context.test.tsx
@@ -9,6 +9,13 @@ vi.mock('@/server/actions/study-request', () => ({
     saveCodeResubmissionNoteDraftAction: vi.fn(),
 }))
 
+// The provider passes reportMutationError's returned handler to useMutation's onError, so asserting
+// on the returned handler is what tells us whether a failure was surfaced to the researcher.
+const mutationErrorHandler = vi.fn()
+vi.mock('@/components/errors', () => ({
+    reportMutationError: vi.fn(() => mutationErrorHandler),
+}))
+
 const STUDY_ID = '11111111-1111-4111-8111-111111111111'
 
 function Harness({ onSaveResult }: { onSaveResult: (result: boolean) => void }) {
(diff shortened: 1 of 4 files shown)


src/contexts/edit-code-resubmit/context.tsx
```diff
@@ -56,9 +56,26 @@ export function EditCodeResubmitProvider({ children, studyId, initialNote }: Edi
     const savingValueRef = useRef<string | null>(null)
     const inFlightSaveRef = useRef<Promise<boolean> | null>(null)
 
+    // A Server Action posts to whatever route is current when the request goes out, so an autosave
```
stetradis, line 59: nit: The same long rationale is repeated in both providers and both tests. One short "why" in each place should be enough. Worth tightening our comment guidance to keep them brief

src/contexts/edit-code-resubmit/context.test.tsx
```diff
@@ -29,6 +36,44 @@ function Harness({ onSaveResult }: { onSaveResult: (result: boolean) => void })
...
+            </EditCodeResubmitProvider>,
+        )
+
+        const note = 'typed right before navigating away'
+        fireEvent.change(screen.getByLabelText('Resubmission note'), { target: { value: note } })
+        await waitFor(() => expect(saveDraftAction).toHaveBeenCalled())
+
+        unmount()
+        rejectSave(new Error('An unexpected response was received from the server.'))
+        await new Promise((resolve) => setTimeout(resolve, 50))
```
stetradis, line 72: nit: the 50ms sleep makes the negative assertion a bit timing-based. Not blocking (CI is green), but waiting for the rejected save promise to settle then asserting would drop the magic delay. Same in `edit-resubmit/context.test.tsx`.

